PR #2, opened 2021-04-12
Use public packages and add build action

mayank99:
- Added build.yml with build and test. Currently won't even work because packages are not available in npm
- Removed azure_pipelines.yml
- Updated package.json and lockfile using public dependencies, and removed npmrc
- Updated imports and some code for new itwinui packages


aruniverse:
@bentleyvk @veekeys lets talk tomorrow, but i think the icon imports should not be importing from the cjs module directly anymore. we had to do this in the old icons-react pkg because they werent exported at the root barrel file, but since the icons dir has been flattened it should be fine

PR #14, opened 2021-04-20
Add borderless styleType to DropdownButton

mayank99:
Previously, we were omitting the `styleType` prop, but after discussion with @FlyersPh9 and @raplemie, it is a valid case to have a borderless dropdown, so I added it.

Also fixed an issue with partial props syntax in unit tests.

### DIFF
--- a/src/core/Buttons/DropdownButton/DropdownButton.test.tsx
+++ b/src/core/Buttons/DropdownButton/DropdownButton.test.tsx
@@ -10,7 +10,7 @@ import { MenuItem } from '../../Menu';
 import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
 import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
 
-function renderComponent(...props: Partial<DropdownButtonProps>[]) {
+function renderComponent(props?: Partial<DropdownButtonProps>) {
   return render(
     <DropdownButton
       menuItems={(close) => [
@@ -82,4 +82,9 @@ it('should work with menu items', () => {
 
   tippy = document.querySelector('[data-tippy-root]') as HTMLElement;
   expect(tippy.style.visibility).toEqual('hidden');
+});
+
+it('should render borderless button correctly', () => {
+  const { container } = renderComponent({ styleType: 'borderless' });
+  expect(container.querySelector('.iui-button.iui-borderless')).toBeTruthy();
 });
--- a/src/core/Buttons/DropdownButton/DropdownButton.tsx
+++ b/src/core/Buttons/DropdownButton/DropdownButton.tsx
@@ -19,7 +19,13 @@ export type DropdownButtonProps = {
    * and returns a list of `MenuItem` components.
    */
   menuItems: (close: () => void) => JSX.Element[];
-} & Omit<ButtonProps, 'onClick' | 'styleType' | 'endIcon'>;
+  /**
+   * Style of the dropdown.
+   * Use 'borderless' to hide outline.
+   * @default 'default'
+   */
+  styleType?: 'default' | 'borderless';
+} & Omit<ButtonProps, 'onClick' | 'endIcon'>;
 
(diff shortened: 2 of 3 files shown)


src/core/Buttons/DropdownButton/DropdownButton.test.tsx
```diff
@@ -10,7 +10,7 @@ import { MenuItem } from '../../Menu';
 import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
 import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
 
-function renderComponent(...props: Partial<DropdownButtonProps>[]) {
+function renderComponent(props?: Partial<DropdownButtonProps>) {
```
mayank99, line 13: The previous syntax was not working, causing the test to fail with this error
image

mayank99: I went through all instances of "Partial" and found the same in SplitButton so I fixed there too

src/core/Buttons/DropdownButton/DropdownButton.tsx
```diff
@@ -19,7 +19,13 @@ export type DropdownButtonProps = {
...
    */
   menuItems: (close: () => void) => JSX.Element[];
-} & Omit<ButtonProps, 'onClick' | 'styleType' | 'endIcon'>;
+  /**
+   * Style of the dropdown button.
+   * Use 'borderless' to hide outline.
+   * @default 'default'
+   */
+  styleType?: 'default' | 'borderless';
+} & Omit<ButtonProps, 'onClick' | 'endIcon'>;
```
bentleyvk, line 28: Omit `styleType` too:
image

src/core/Buttons/DropdownButton/DropdownButton.tsx
```diff
@@ -19,7 +19,13 @@ export type DropdownButtonProps = {
    * and returns a list of `MenuItem` components.
    */
   menuItems: (close: () => void) => JSX.Element[];
-} & Omit<ButtonProps, 'onClick' | 'styleType' | 'endIcon'>;
+  /**
+   * Style of the dropdown button.
+   * Use 'borderless' to hide outline.
+   * @default 'default'
+   */
+  styleType?: 'default' | 'borderless';
```
bentleyvk, line 27: When switching between style types menu with stays the same:
image
It's a minor thing. Strange but when changing other props in Storybook, menu width changes accordingly.

bentleyvk: Oh, I found why 😀
image
We need to add `styleType` to deps list.

src/core/Buttons/DropdownButton/DropdownButton.test.tsx
veekeys: What, so it was not working all the time? How tests were passing then..

veekeys: Oh ok.. We were not passing anything, so it was not complaining... Interesting

src/core/Buttons/DropdownButton/DropdownButton.tsx
mayank99: Yeah I noticed that, but is it really needed? Who is gonna change styleType in practice?

src/core/Buttons/DropdownButton/DropdownButton.tsx
mayank99: Done

src/core/Buttons/DropdownButton/DropdownButton.tsx
bentleyvk: But it is an easy fix and since we know that it will bug me 😀

mayank99: Yeah that's why I already did it 😄

bentleyvk (approved):

PR #15, opened 2021-04-20
Cache node modules and run parallel jobs

mayank99:
- Added step for caching node_modules using yarn.lock as hash
- Split up tests+publish-tests and build+build-storybook into two jobs (now runs in parallel and takes half the time)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-    name: Build and run tests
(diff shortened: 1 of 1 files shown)


mayank99:
See https://github.com/iTwin/iTwinUI-react/actions/runs/768589516

bentleyvk (approved):
Nice. How you even found this? 😀

veekeys (approved):
Wow, amazing. Great finding!

PR #25, opened 2021-04-23
Fix tooltip blurred text by disabling gpu acceleration

mayank99:
3D transforms seem to cause issues with backdrop-filter.

Disabling [`gpuAcceleration`](https://popper.js.org/docs/v2/modifiers/compute-styles/#gpuacceleration) means the tooltip will use `inset` instead of `transform`.

### DIFF
--- a/src/core/Tooltip/Tooltip.tsx
+++ b/src/core/Tooltip/Tooltip.tsx
@@ -46,6 +46,14 @@ export const Tooltip = (props: TooltipProps) => {
       }
       offset={[0, 4]}
       ref={ref}
+      popperOptions={{
+        modifiers: [
+          {
+            name: 'computeStyles',
+            options: { gpuAcceleration: false },
+          },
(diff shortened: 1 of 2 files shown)


mayank99:
`gpuAcceleration: true` (default):

before

`gpuAccelration: false`:

after

mayank99:
Ugh this doesn't fully fix the issue. The children still get blurred on hover (more noticeable in dark theme and on high resolution screen).

blur

I could maybe apply `backface-visibility` same as https://github.com/iTwin/iTwinUI/pull/37 and have tested that it fixes it.

```jsx
      {React.cloneElement(children, {
        title: undefined,
        style: { backfaceVisibility: 'hidden', ...children.props.style },
      })}
```

bentleyvk:
For me it is the same. I tried adding `backface-visibility: hidden` but it just makes text blurry.

PR #33, opened 2021-04-28
Release 1.2.0

mayank99:
Updated @itwin/itwinui-css to 0.11.0 which has the tooltip fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.1.1]
+## [1.2.0]
 
 `2021-04-29`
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+`2021-04-29`
+
+### What's new
+
+- **`DropdownMenu` now accepts `popoverProps`**.
+
+### Fixes
+
+- **Fixed image sizing issues in `ErrorPage`**.
+- **Fixed tooltip text blurring issues** through base CSS package.
```
mayank99, line 14: Should I close #25 now that don't need it? It could still help with something idk

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@itwin/itwinui-react",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
mayank99, line 3: I can't decide if this should be 1.2.0 or 1.1.1

The `popoverProps` thing is technically an addition but it's not really a feature, so I went with 1.1.1

CHANGELOG.md
mayank99: Also @veekeys @bentleyvk can you test if the blurriness is actually gone? I think it looks good on my machine but it's hard to tell

package.json
bentleyvk: I would say it should be `1.2.0`.

CHANGELOG.md
bentleyvk: Seems like it is fixed for me.
And about #25 it is now unnecessary and may bring unknown bugs, I would say to close it for now.

veekeys: It is fixed, but controlled story does not work for me.
Also, before release, need to fix wizard story, which got broken for some reason.

veekeys (approved):

PR #40, opened 2021-04-29
Add FileUpload component with drag & drop

mayank99:
Added `FileUpload` component with drag & drop functionality using `file-upload.css` and [native drag events](https://developer.mozilla.org/en-US/docs/Web/API/DragEvent).

Can be used a wrapper around any component or standalone if no children provided.

## Checklist

## Screenshots

file-upload-1
file-upload-2

### DIFF
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -3,11 +3,11 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
 
 import { FileUpload } from './FileUpload';
 
-it('should render in its most basic state', () => {
+it('should render content and children', () => {
   const mockContent = 'Mock drag content';
   const mockChildren = 'Mock children to wrap';
   const { container } = render(
@@ -26,4 +26,31 @@ it('should render in its most basic state', () => {
   const children = component.lastChild as HTMLElement;
   expect(children).toBeTruthy();
   expect(children.textContent).toEqual(mockChildren);
+});
+
+it('should handle drag and drop events correctly', () => {
+  const mockFn = jest.fn();
+  const { container } = render(
+    <FileUpload content={'Mock content'} onFileDropped={mockFn} />,
+  );
+
+  const component = container.querySelector('.iui-file-upload') as HTMLElement;
+  expect(component).toBeTruthy();
+
+  expect(container.querySelector('.iui-content')).toBeTruthy();
+
+  const file = new File(['mock-file'], 'test.txt', { type: 'text/plain' });
+  const mockDataTransfer = {
+    dataTransfer: { files: [file], items: [{ kind: 'file', type: file.type }] },
+  };
+
+  fireEvent.dragOver(component, mockDataTransfer);
+  expect(container.querySelector('.iui-file-upload.iui-drag')).toBeTruthy();
+
+  fireEvent.dragLeave(component, mockDataTransfer);
+  expect(container.querySelector('.iui-file-upload.iui-drag')).toBeFalsy();
+
+  fireEvent.dragOver(component, mockDataTransfer);
+  fireEvent.drop(component, mockDataTransfer);
+  expect(mockFn).toHaveBeenCalledWith([file]);
 });
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -23,7 +23,7 @@ it('should render dragContent and children', () => {
   expect(content).toBeTruthy();
   expect(content.textContent).toEqual(mockContent);
 
-  const children = component.lastChild as HTMLElement;
+  const children = component.firstChild as HTMLElement;
   expect(children).toBeTruthy();
   expect(children.textContent).toEqual(mockChildren);
 });
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -8,13 +8,22 @@ import { render } from '@testing-library/react';
 import { FileUpload } from './FileUpload';
 
 it('should render in its most basic state', () => {
-  // TODO: Make sure all required props are passed in here
-  const { container } = render(<FileUpload />);
-  expect(container.querySelector('div')).toBeTruthy();
-});
+  const mockContent = 'Mock drag content';
+  const mockChildren = 'Mock children to wrap';
+  const { container } = render(
+    <FileUpload content={mockContent} onFileDropped={jest.fn}>
+      {mockChildren}
+    </FileUpload>,
+  );
 
-// TODO: Write tests here!
+  const component = container.querySelector('.iui-file-upload') as HTMLElement;
+  expect(component).toBeTruthy();
 
-it('should be improved', () => {
-  expect(false).toBe(true);
+  const content = container.querySelector('.iui-content') as HTMLElement;
+  expect(content).toBeTruthy();
+  expect(content.textContent).toEqual(mockContent);
+
+  const children = component.lastChild as HTMLElement;
+  expect(children).toBeTruthy();
+  expect(children.textContent).toEqual(mockChildren);
 });
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -7,11 +7,11 @@ import { fireEvent, render } from '@testing-library/react';
 
 import { FileUpload } from './FileUpload';
 
-it('should render content and children', () => {
+it('should render dragContent and children', () => {
   const mockContent = 'Mock drag content';
   const mockChildren = 'Mock children to wrap';
   const { container } = render(
-    <FileUpload content={mockContent} onFileDropped={jest.fn}>
+    <FileUpload dragContent={mockContent} onFileDropped={jest.fn}>
       {mockChildren}
     </FileUpload>,
   );
@@ -28,10 +28,22 @@ it('should render content and children', () => {
   expect(children.textContent).toEqual(mockChildren);
 });
 
+it('should apply content class to children if dragContent not passed', () => {
+  const mockChildren = 'Mock children to wrap';
+  const { container } = render(
+    <FileUpload onFileDropped={jest.fn}>{mockChildren}</FileUpload>,
+  );
+  expect(container.querySelector('.iui-file-upload')).toBeTruthy();
+
+  const content = container.querySelector('.iui-content') as HTMLElement;
+  expect(content).toBeTruthy();
+  expect(content.textContent).toEqual(mockChildren);
+});
+
 it('should handle drag and drop events correctly', () => {
   const mockFn = jest.fn();
   const { container } = render(
-    <FileUpload content={'Mock content'} onFileDropped={mockFn} />,
+    <FileUpload onFileDropped={mockFn}>Mock content</FileUpload>,
   );
 
   const component = container.querySelector('.iui-file-upload') as HTMLElement;
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { render } from '@testing-library/react';
+
+import { FileUpload } from './FileUpload';
+
+it('should render in its most basic state', () => {
+  // TODO: Make sure all required props are passed in here
+  const { container } = render(<FileUpload />);
+  expect(container.querySelector('div')).toBeTruthy();
+});
+
+// TODO: Write tests here!
+
+it('should be improved', () => {
+  expect(false).toBe(true);
+});
--- a/src/core/FileUpload/FileUpload.test.tsx
+++ b/src/core/FileUpload/FileUpload.test.tsx
@@ -44,13 +44,13 @@ it('should handle drag and drop events correctly', () => {
     dataTransfer: { files: [file], items: [{ kind: 'file', type: file.type }] },
   };
 
-  fireEvent.dragOver(component, mockDataTransfer);
+  fireEvent.dragEnter(component, mockDataTransfer);
   expect(container.querySelector('.iui-file-upload.iui-drag')).toBeTruthy();
 
   fireEvent.dragLeave(component, mockDataTransfer);
   expect(container.querySelector('.iui-file-upload.iui-drag')).toBeFalsy();
 
-  fireEvent.dragOver(component, mockDataTransfer);
+  fireEvent.dragEnter(component, mockDataTransfer);
   fireEvent.drop(component, mockDataTransfer);
   expect(mockFn).toHaveBeenCalledWith([file]);
 });
--- a/src/core/FileUpload/FileUpload.tsx
+++ b/src/core/FileUpload/FileUpload.tsx
@@ -44,52 +44,43 @@ export const FileUpload = React.forwardRef<HTMLDivElement, FileUploadProps>(
     const fileUploadRef = React.useRef<HTMLDivElement>(null);
     const refs = useMergedRefs(fileUploadRef, ref);
 
-    const onDragEnterHandler = React.useCallback((e: React.DragEvent) => {
+    const onDragOverHandler = (e: React.DragEvent) => {
       e.preventDefault();
       e.stopPropagation();
-    }, []);
+    };
 
-    const onDragOverHandler = React.useCallback(
-      (e: React.DragEvent) => {
-        e.preventDefault();
-        e.stopPropagation();
+    const onDragEnterHandler = (e: React.DragEvent) => {
+      e.preventDefault();
+      e.stopPropagation();
 
-        // only set active if a valid file is dragged over
-        if (!isDragActive && e.dataTransfer?.items?.[0]?.kind === 'file') {
-          setIsDragActive(true);
-        }
-      },
-      [isDragActive],
-    );
+      // only set active if a file is dragged over
+      if (!isDragActive && e.dataTransfer?.items?.[0]?.kind === 'file') {
+        setIsDragActive(true);
+      }
+    };
 
-    const onDragLeaveHandler = React.useCallback(
-      (e: React.DragEvent) => {
-        e.preventDefault();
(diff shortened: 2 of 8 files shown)


src/core/FileUpload/FileUpload.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+  expect(component).toBeTruthy();
+
+  const content = container.querySelector('.iui-content') as HTMLElement;
+  expect(content).toBeTruthy();
+  expect(content.textContent).toEqual(mockContent);
+
+  const children = component.lastChild as HTMLElement;
+  expect(children).toBeTruthy();
+  expect(children.textContent).toEqual(mockChildren);
+});
```
mayank99, line 29: @veekeys @bentleyvk Any idea how I would actually test the drag behavior? Just mock events?

stories/core/FileUpload.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const WithoutChildren: Story<FileUploadProps> = ({
+  children,
+  ...rest
+}) => {
+  const [file, setFile] = useState<File | undefined>(undefined);
+
+  return (
+    <FileUpload
+      content={file?.name ?? children ?? 'Drop file to upload'}
```
mayank99, line 51: I need to make the `children` include an input element
image

Not sure how the interaction would work with drop event (which does not need an input element)

mayank99: Actually I think they can stay unrelated, firing separate events 🤔
Will research accessibility approach tomorrow

src/core/FileUpload/FileUpload.test.tsx
bentleyvk: Looks like you need to simulate events yourself as jsdom does not support them:
https://react-dnd.github.io/react-dnd/docs/testing#simulate-the-dom
image

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,109 @@
...
+  onFileDropped: (file: File) => void;
+  /**
+   * Component to wrap `FileUpload` around.
+   * `content` will be shown if `children` is not provided.
+   */
+  children?: React.ReactNode;
+} & CommonProps;
+
+/**
+ * File upload component to be used as a wrapper or standlone.
```
veekeys, line 29: standalone

stories/core/FileUpload.stories.tsx
mayank99: @FlyersPh9 I think we need some focus styling on labels. If not, then we should make them non-focusable. Right now, you can go to `file-upload.html` demo and keep pressing **tab** (to focus without any visual) and **space** (to simulate keypress) to make the file explorer show up for each of those hidden input labels.

Maybe something like this: https://developer.mozilla.org/en-US/docs/Web/API/File/Using_files_from_web_applications#using_a_label_element_to_trigger_a_hidden_file_input_element

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+export const FileUpload = React.forwardRef<HTMLDivElement, FileUploadProps>(
+  (props, ref) => {
+    const { content, children, onFileDropped, className, ...rest } = props;
+    useTheme();
+
+    const [isDragActive, setIsDragActive] = React.useState(false);
+    const fileUploadRef = React.useRef<HTMLDivElement>(null);
+    const refs = useMergedRefs(fileUploadRef, ref);
+
+    const onDragEnterHandler = React.useCallback((e: React.DragEvent) => {
```
bentleyvk, line 47: If your are using callbacks on base HTML elements, you don't need to use `useCallback`. https://github.com/facebook/react/issues/17055#issuecomment-540339327
image

stories/core/FileUpload.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      />
+    </FileUpload>
+  );
+};
+
+Basic.args = {
+  content: 'Drop file to upload',
+};
+
+export const WithoutChildren: Story<FileUploadProps> = (args) => {
```
bentleyvk, line 48: When dragging files and border appears, the content moves. Need a fix in CSS.

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const onDropHandler = React.useCallback(
+      (e: React.DragEvent) => {
+        e.preventDefault();
+        e.stopPropagation();
+
+        if (isDragActive) {
+          setIsDragActive(false);
+        }
+        onFileDropped(e.dataTransfer?.files);
```
bentleyvk, line 89: Currently it lets you drag'n'drop selected text and replaces previously dropped files. Maybe we should call `onFIleDropped` only when there is files.

src/core/FileUpload/FileUpload.tsx
bentleyvk, line 47: What it does?

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const onDragLeaveHandler = React.useCallback(
+      (e: React.DragEvent) => {
+        e.preventDefault();
+        e.stopPropagation();
+
+        // only set inactive if secondary target is outside the component
+        if (
+          isDragActive &&
+          !fileUploadRef.current?.contains(e.relatedTarget as Node)
```
bentleyvk, line 73: Is this check necessary? Because event fires when drag leaves so `relatedTarget` should always be outside.

stories/core/FileUpload.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const [files, setFiles] = useState<Array<File>>([]);
+
+  return (
+    <FileUpload
+      content={
+        <>
```
bentleyvk, line 54: It would be nice to provide this from our side but I am not sure how...

stories/core/FileUpload.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      {...args}
+      onFileDropped={(files) => {
+        setFiles(Array.from(files));
+        action(`${files.length} files uploaded`)();
+      }}
+    >
+      <Textarea
+        placeholder={'Drag a file here'}
+        rows={1}
+        defaultValue={files.map((f) => f.name)}
```
veekeys, line 38: maybe add readonly here, not to edit the name manually

veekeys: btw, sure its defaultValue here? I think it should be just value, cause it changes.

stories/core/FileUpload.stories.tsx
veekeys: So users will need to copy paste this one themselves? I would agree we need to render this content ourselves...

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+export type FileUploadProps = {
+  /**
+   * Content shown over `children` when file is being dragged onto the component.
+   */
+  content: React.ReactNode;
+  /**
+   * Callback fired when files are dropped onto the component.
+   */
+  onFileDropped: (files: FileList) => void;
```
veekeys, line 20: Cant it be optional? Or you wanted to enforce it, cause it doesnt make sense not to listen to the event?

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+    return (
+      <div
+        className={cx(
+          'iui-file-upload',
+          { 'iui-drag': isDragActive },
+          className,
+        )}
+        onDragEnter={onDragEnterHandler}
```
veekeys, line 101: hmm... maybe destruct all these in props and add them in your handlers in case user provides those? Cause now they might be overriden and brake the component..

veekeys: Oh wait.. users cannot even provide them.. well mb ok for now.

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+      e.stopPropagation();
+    }, []);
+
+    const onDragOverHandler = React.useCallback(
+      (e: React.DragEvent) => {
+        e.preventDefault();
+        e.stopPropagation();
+
+        // only set active if a valid file is dragged over
+        if (!isDragActive && e.dataTransfer?.items?.[0]?.kind === 'file') {
```
veekeys, line 58: How the "invalid" file looks like?

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,115 @@
...
+  (props, ref) => {
+    const { content, children, onFileDropped, className, ...rest } = props;
+    useTheme();
+
+    const [isDragActive, setIsDragActive] = React.useState(false);
+    const fileUploadRef = React.useRef<HTMLDivElement>(null);
+    const refs = useMergedRefs(fileUploadRef, ref);
+
+    const onDragEnterHandler = React.useCallback((e: React.DragEvent) => {
+      e.preventDefault();
```
veekeys, line 48: Are you sure about this? Shouldnt this one have the logic of dragOver instead? It is called once when it is dragged over versus dragover is called all the time while you are dragging...If I understood correctly

src/core/FileUpload/FileUpload.tsx
mayank99: Yeah I wanted to enforce it because there's no point to this component without this callback

stories/core/FileUpload.stories.tsx
mayank99: `value` and `onChange` are for user inputs, which I don't want to control, so `defaultValue` made sense here

stories/core/FileUpload.stories.tsx
mayank99: But it's so basic. If we provide something, we should go with a better template and it also needs to be flexible

src/core/FileUpload/FileUpload.tsx
mayank99: It is very much necessary. I spent several hours trying to make the setState be called only once.

What happens is that `dragLeave` fires once for each child element inside the div, so the drag state flickers rapidly

src/core/FileUpload/FileUpload.tsx
mayank99: So I should be able to just move it in the if block above?

src/core/FileUpload/FileUpload.tsx
mayank99: Just overriding default behavior

stories/core/FileUpload.stories.tsx
mayank99: @FlyersPh9

src/core/FileUpload/FileUpload.tsx
mayank99: "invalid" just means it's not a file. Meaning if any draggle HTML element is dragged

src/core/FileUpload/FileUpload.tsx
mayank99: I thought I read somewhere that `dragEnter` is not called for files. But it seems to work ok so 🤷

src/core/FileUpload/FileUpload.tsx
mayank99: Done

src/core/FileUpload/FileUpload.tsx
veekeys: There is smth like that.. I read that in other example of this kinda component, but did not really go into details to understand...
https://medium.com/@650egor/simple-drag-and-drop-file-upload-in-react-2cb409d88929

stories/core/FileUpload.stories.tsx
veekeys: If there are classes user need to add, it is not basic anymore...

stories/core/FileUpload.stories.tsx
veekeys: User inputs? In storybook user can provide onChange?
I tohught in storybook we give actual user examples, how they would use stuff in their work.
Just by design, defaultValue is used wrong here, IMO.

src/core/FileUpload/FileUpload.tsx
veekeys: hmm.. I just saw in your tests, that there was a point (only for visual check, ofc, but still :D)

src/core/FileUpload/FileUpload.tsx
veekeys: hmm, interesting..

src/core/FileUpload/FileUpload.tsx
veekeys: Yeah, just found this one in the article mentioned in my comment above...

stories/core/FileUpload.stories.tsx
mayank99: We gave a proper example of `FileUpload`. The question then is about controlled input (`value`/`onChange`) vs uncontrolled (`defaultValue`).

I don't think it's used wrong because the value is not really being changed by the user directly.

stories/core/FileUpload.stories.tsx
mayank99: It is pretty basic. I think we decided that if we are going to provide something, we should go all the way. i.e. include upload tracker, final state, etc

For now, the main purpose of this component is to wrap user content/children

veekeys: I would agree, if this was our LabeledInput..Now its just plain html and most important thing, it needs our class to be applied. It is just a blocker in here, IMO.
We removed simple things (applying iui-icon class requirement), we cannot expect this to work and I imagine this workflow might be used more often than with wrapper element..

stories/core/FileUpload.stories.tsx
veekeys: Well, from what I tried, value changes when I drop the file in there...
Again, you are missusing here the defaultValue. https://reactjs.org/docs/uncontrolled-components.html
If you want to have it uncontrolled, you need to give defaultValue a static value: "no files were uploaded" or something like that.

stories/core/FileUpload.stories.tsx
FlyersPh9: I've got a fix ready.  Will push this fix along with the fix for [this](https://github.com/iTwin/iTwinUI-react/pull/40#discussion_r623978591).

stories/core/FileUpload.stories.tsx
mayank99: The value changes but that's not user input. I would even call it "initial state".

But I think the confusion here might actually come from using the Textarea wrong (we are using it for both user input and for showing filename).

So maybe I can make the textarea readonly, or I can put the filename outside the textarea.

veekeys: What do you mean it is not an user input? So if I just dragged the files in there, the names which appeared are not my input??
The only issue in here is, that you should use value instead of defaultValue, thats all...

veekeys: oh, and readonly would be useful too in this context

stories/core/FileUpload.stories.tsx
FlyersPh9: `:focus-within` has been added in [this PR](https://github.com/iTwin/iTwinUI/pull/54).

stories/core/FileUpload.stories.tsx
mayank99: Added `FileUploadTemplate` based on https://github.com/iTwin/iTwinUI/pull/54

stories/core/FileUpload.stories.tsx
mayank99: readonly textarea looks like it's disabled ([work item](https://github.com/iTwin/iTwinUI/projects/2#card-60322339))

so I've changed story to use `LabeledInput`

src/core/FileUpload/FileUpload.tsx
```diff
@@ -0,0 +1,105 @@
...
+        )}
+        onDragEnter={onDragEnterHandler}
+        onDragOver={onDragOverHandler}
+        onDragLeave={onDragLeaveHandler}
+        onDrop={onDropHandler}
+        ref={refs}
+        {...rest}
+      >
+        {dragContent && <div className='iui-content'>{dragContent}</div>}
+        {dragContent ? children : <div className='iui-content'>{children}</div>}
```
mayank99, line 99: I've slightly changed the api to make it user-friendly.

1. `content` is now called `dragContent` for clarity.
2. `dragContent` is optional now.
3. `children` is required and will be wrapped by `.iui-content` if `dragContent` is not provided. (This simulates the "Not Wrapping Anything" example in backstop html and the "Default" example in storybook).

So now `FileUploadTemplate` should be passed to `children` instead.

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -0,0 +1,60 @@
...
+        id='file-input'
+        onChange={onChange}
+        multiple
+      />
+      <SvgUpload className='iui-icon' />
+      <label htmlFor='file-input' className='iui-anchor'>
+        {label ?? (
+          <>
+            {'Choose a file.'}
+            <br />
```
mayank99, line 50: I would like to change this to `\n`. See https://github.com/iTwin/iTwinUI/pull/54#discussion_r625315645

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -0,0 +1,60 @@
...
+  useTheme();
+
+  return (
+    <>
+      <input
+        className='iui-browse-input'
+        type='file'
+        id='file-input'
+        onChange={onChange}
+        multiple
```
bentleyvk, line 43: Add a prop for this. I wonder if we can do the same for drag component.

mayank99: I don't think there's a way to control number of files for drag events, so I just left multiple.

mayank99:
image

Interesting...

bentleyvk (dismissed):

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <>
+      <input
+        className='iui-browse-input'
+        type='file'
+        id='file-input'
+        onChange={onChange}
+        multiple={acceptMultiple}
+      />
+      <SvgUpload className='iui-icon' />
```
veekeys, line 61: aria-hidden

veekeys:
When you click on upload and then cancel, this not nice style is there (its seenable on click too, active I guess)
image

veekeys:
Should it cover full height of the input?
image

I guess it should not, but our input is done in a weird way, so...

veekeys (dismissed):

veekeys (approved):

PR #42, opened 2021-04-30
Update createComponent.js with stories boilerplate

mayank99:
`yarn createComponent` now creates a `.stories.tsx` file under `/stories/core` with a basic story.

Also removed the prompt for collection because `utils` is broken and unneeded.

### DIFF
--- a/scripts/createComponent.js
+++ b/scripts/createComponent.js
@@ -75,6 +75,30 @@ export default ${name};
   };
 };
 
+const storiesFactory = (directory, name) => {
+  return {
+    path: `${directory}/${name}.stories.tsx`,
+    template: `${copyrightHeader}
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import { ${name}, ${name}Props } from '../../src/core';
+
+export default {
+  component: ${name},
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/${name}',
+} as Meta<${name}Props>;
+
+export const Basic: Story<${name}Props> = (args) => {
+  return <${name} {...args} />;
+};
+`,
+  };
+};
+
 const componentIndexFactory = (directory, name) => {
   return {
     path: `${directory}/index.ts`,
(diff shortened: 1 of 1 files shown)


scripts/createComponent.js
```diff
@@ -159,4 +183,8 @@ inquirer
     writeFile(componentOut);
     writeFile(componentTest);
     writeFile(componentIndex);
+
+    if (collection !== 'utils') {
```
bentleyvk, line 187: Do we need `utils` option? I don't think we ever used that.

scripts/createComponent.js
```diff
@@ -75,6 +75,30 @@ export default ${name};
...
+    path: `${directory}/${name}.stories.tsx`,
+    template: `${copyrightHeader}
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import { ${name}, ${name}Props } from '../../src/core';
+
+export default {
+  component: ${name},
+  argTypes: {
+    className: { control: { disable: true } },
```
veekeys, line 89: We want not to show them? I remember we discussed to maybe show them just to realize that you can use those props...?

scripts/createComponent.js
veekeys: Yeah, now it is broken.. lets remove it from options

scripts/createComponent.js
veekeys: A wait, its control... all good

scripts/createComponent.js
mayank99: It's broken?

veekeys: Well, I tried, it failed, so... Unless I missed something.

mayank99: utils is gone 💥🤯

veekeys: Sad doesnt work with subDirs, but.. it is ok for now, I guess
image

veekeys (approved):

scripts/createComponent.js
mayank99: 😄  Yeah we removed that in https://github.com/iTwin/iTwinUI-react/commit/2975e1a480afa07f141d0868dfcf691952a75578

PR #45, opened 2021-05-04
Add Header and HeaderButton

raplemie:
# Header components

~Based on iTwin/iTwinUI#33 for the header, which is not ready. (But would require that css branch to test this)~

Built with @iTwin/itwinui  ~0.13.1~ 0.14.0.

~Looking for discussions on various `TODO: Validate` points, I'll create review so we can discuss them. All of it is subject to change as the CSS might change, so no tests are written at the moment.~

## Checklist

## Screenshots

jPfIONmJ5a
**Don't worry about cursor position, it's related to the software I use to grab screenshot and Windows resizing for laptop**
### Header

Root component, split in 3 sections with an optional "more" menu and a slim mode.

### HeaderTitle

This component is intended to be used in the appTitle props of the `Header` component, allows focus if `onClick` is present. (Named it `HeaderTitle` instead of `HeaderLogo` as the title is the `children`, however, the `logo` prop is mandatory.

### HeaderBreadcrumbs

This component ~simply~ adds a `nav` component and then add `iui-chevron` between the items it's provided, to be used in the `breadcrumbs` prop of the `Header` component.

### HeaderButton

This component should be used within a `HeaderBreadcrumbs` component, supports `menuItems` which will make it switch between a `Button` and a `DropdownButton`. Handle `isActive` prop and the `iui-header-buttton-icon` which supports size changing icons depending on the `Header` `isSlim`.


src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+   */
+  centerContent?: React.ReactNode;
+  /**
+   * Expects `UserIcon`, will be given a drop down if `userMenuItems` is also provided;
+   */
+  userIcon?: React.ReactNode;
+  /**
+   * Handler on notification click, will show notification bell if provided.
+   */
+  onNotification?: () => void;
```
raplemie, line 53: This is only a handler instead of a `menuItems` list, because typically the bell button would open a side panel with the notifications (for which we might want to create a component eventually :) )

raplemie (commented):
Here are some discussion points, some we can decide if we are fine with or want to head in a different direction.

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+   */
+  appName?: React.ReactNode;
+  /**
+   * Handler on application click
+   */
+  onApplication?: () => void;
+  /**
+   * `HeaderButton` array, chevron will be added between each.
+   */
+  breadcrumbsButtons?: React.ReactNode[];
```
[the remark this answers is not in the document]
mayank99: I would just call it `breadcrumbs` because `breadcrumbsButtons` sounds redundant.

And you can add some more explanation in the JSDoc comment.

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+  /**
+   * Content to display in the center section of the header, should handle iui-slim class if used.
+   */
+  centerContent?: React.ReactNode;
```
mayank99, line 45: Hmm I'm not sure about asking the user to handle a css class.

Also would call this `children` because it's not part of the main content we provide.

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+  moreMenuItems?: (close: () => void) => JSX.Element[];
+  /**
+   * Reduce height of the header;
+   * @default false
+   */
+  slim?: boolean;
+  /**
+   * Provide localized strings.
+   */
+  translatedHeader?: HeaderTranslations;
```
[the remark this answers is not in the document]
mayank99: `translatedHeader` is not very intuitive, so I would call it `headerTranslations` or `translatedStrings` or anything plural

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+   * Items in the more dropdown menu.
+   * Pass a function that takes the `close` argument (to close the menu),
+   * and returns a list of `MenuItem` components.
+   */
+  moreMenuItems?: (close: () => void) => JSX.Element[];
+  /**
+   * Reduce height of the header;
+   * @default false
+   */
+  slim?: boolean;
```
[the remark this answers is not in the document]
mayank99: `visible` comes from tippy.js

It used to be called `isVisible`. The `is-` convention does seem more explanatory

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * Items in the dropdown menu.
+   * Pass a function that takes the `close` argument (to close the menu),
+   * and returns a list of `MenuItem` components.
+   */
+  menuItems: (close: () => void) => JSX.Element[];
+  /**
+   * Is the menu active.
+   * @default false
+   */
+  active?: boolean;
```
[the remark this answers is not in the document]
mayank99: `isActive` as in other places (e.g. `IconButton`)

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+import React from 'react';
+
+import { DropdownMenu } from '../DropdownMenu';
+import { useTheme } from '../utils/hooks/useTheme';
+
+export type HeaderButtonProps = {
+  /**
+   * Title to display.
+   */
+  title: React.ReactNode;
```
mayank99, line 19: I would go with `name` not to confuse with native html `title`

src/core/Header/Header.tsx
raplemie: I can switch to `children`, however it is put in a very specific place in the center, not like in the `Tile` where you can do whatever you want, but it needs to be aware of what happens when they select "slim"... otherwise it will do just like my example GIF where the searchbox overflow the header...

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Second line to display.
+   */
+  subTitle?: React.ReactNode;
```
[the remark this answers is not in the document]
mayank99: `name` and `subtitle` (not `subTitle`) would be fine, I think.

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+   */
+  appIcon: React.ReactNode;
+  /**
+   * Name of the application.
+   */
+  appName?: React.ReactNode;
+  /**
+   * Handler on application click
+   */
+  onApplication?: () => void;
```
mayank99, line 37: This could be `onAppClick`

src/core/Header/Header.tsx
raplemie: I can switch to `children`, however it is put in a very specific place in the center, not like in the `Tile` where you can do whatever you want, but it needs to be aware of what happens when they select "slim"... otherwise it will do just like my example GIF where the searchbox overflow the header...

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+  /**
+   * Handler on notification click, will show notification bell if provided.
+   */
+  onNotification?: () => void;
+  /**
+   * Items in the help dropdown menu.
+   * Pass a function that takes the `close` argument (to close the menu),
+   * and returns a list of `MenuItem` components.
+   */
+  helpMenuItems?: (close: () => void) => JSX.Element[];
```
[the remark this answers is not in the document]
mayank99: Depends on whether we want to provide more control or more simplicity

src/core/Header/Header.tsx
mayank99: Not sure we need a top level prop for that

raplemie: Well, we dont control the notification panel in the Header, how would the application know that the bell was clicked and that it should show that panel if we dont provide a click handler ?

src/core/Header/Header.tsx
mayank99: Well, the user of this component is providing the `isSlim` prop, so they should know to handle the height manually

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+   * Pass a function that takes the `close` argument (to close the menu),
+   * and returns a list of `MenuItem` components.
+   */
+  userMenuItems?: (close: () => void) => JSX.Element[];
+  /**
+   * Items in the more dropdown menu.
+   * Pass a function that takes the `close` argument (to close the menu),
+   * and returns a list of `MenuItem` components.
+   */
+  moreMenuItems?: (close: () => void) => JSX.Element[];
```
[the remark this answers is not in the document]
mayank99: I have mixed feelings about these props but I can't think of a better approach yet

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+        'aria-hidden': true,
+      })
+    : appIcon;
+  return (
+    <header className={cx('iui-header', { 'iui-slim': slim })} {...rest}>
+      <div className='iui-left'>
+        {/* TODO: Validate
+         * Button: Can't use as is because the 'iui-brand-logo' would also get 'iui-icon' and
+         *         associated behavior. (Not sure what the impact is)
+         * Is the "onApplication" mandatory ? If not, we might need a 'no click' version. */}
```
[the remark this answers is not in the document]
mayank99: I don't think it should be mandatory. There isn't always a reason to click the app logo/name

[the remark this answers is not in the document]
raplemie: That's also what I think, should this be raised to @FlyersPh9 in the css PR ?

src/core/Header/Header.tsx
mayank99: Oh you mean the actual bell icon click...

To me, "on notification click" implies that an actual notification item (from the panel) is clicked. Since that panel is not part of this component, I would actually remove this prop completely.

In #32 I suggested grouping the "right" part of header where the user can provide any arbitrary items. So you could do that, and leave it upto the user to provide both the bell icon and click handler

raplemie: Ok, I misunderstood @veekeys point in #32 , I can change that to only control User icon and More, but that means we'll have no way of smartly collapsing stuff when space is not available, at this point, not sure it's even worth trying to handle User and  More either, as the only thing that this is adding is the `iui-user-icon`, to override `iui-user-icon-medium` while in slim mode, which, if we follow the class names split, should probably be `iui-user-icon iui-medium` anyway, in which case we can decide to do that override without having to "add" the class at all...

raplemie: Actually, we dont even have to add that class, as we could simply override `iui-user-icon-medium` if it's in the header...

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Discussed with @FlyersPh9 , suggestion is to keep using the same css, but add `style={{pointerEvents: 'none'}}` if there is no click handler.

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+              <button className='iui-button iui-dropdown iui-borderless'>
+                {processedUserIcon}
+              </button>
+            </DropdownMenu>
+          ) : (
+            /* TODO: Validate
+             * if there is no menu items, I don't add the button (as it is not interactive)
+             * but then there is no margins, is it allowed to have no sub menu ?
+             */
+            processedUserIcon
```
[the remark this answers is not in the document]
raplemie: Like the application button, could use the same button but with `style={{pointerEvents: 'none'}}` to handle that.

src/core/Header/Header.tsx
raplemie: regarding the `iui-user-icon` that have already been changed in the CSS PR to use the new names, so technically, we wouldn't need to control any of the content of `iui-center` and `iui-right`. With some discussions, I'm starting to think that maybe we could reduce the complexity of this comopnent to `<Header leftContent centerContent rightContent />`, these contents being
```
<Header
  leftContent={
  <>
    <ApplicationButton />
    <Breadcrumb>
      <HeaderButton "project" />
      <HeaderButton "model" />
    </Breadcrumb>
  </>}
  centerContent={<Whatever />}
  rightContent={
  <>
    <IconButton><SvgNotification/></IconButton>
    <button className="iui-button iui-borderless" style={{pointerEvents: 'none'}}>
      <UserIcon />
    </button>
  </>}
/>
```

This way, they have full control over what is in their menu, we dont force anyone, and in a second time, we want to have a component that do control all this, along with collapsing to a full "..." menu if there is not enough space, we can do that, and they will be able to use it without breaking anyone.

mayank99: At that point, it doesn't even make sense to have props, and everything should be handled using `children` by providing granular components. Kinda like what material ui does: https://material-ui.com/components/cards/#api

But their reasoning is that their library needs to be very flexible. Our library is the opposite: enforce standards first, provide flexibility second.

We could continue this discussion in #32 or offline (over call) but I really think there is some middle ground where we don't provide 3 huge props and we don't provide 15 smaller props.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
mayank99: So we want to use button just for the extra spacing? That doesn't seem right to me. I would think the spacing should be applied separately and button should not be used if it's non-interactive

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Changed to translatedStrings, however, we only have one string now to translate, but I wasnt sure if I was to change it to a string instead of an object or not, string is ("More options")

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Changed

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Went with the function, simply because users will use them for the HeaderButton as they are now DropdownButtons, it seems cleaner to have all of these working in the same way. (However, we now only have 1 menuItems prop, for the "More options".)

src/core/Header/HeaderButton.tsx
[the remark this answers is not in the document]
raplemie: I went with `name` and `description`.

src/core/Header/HeaderButton.tsx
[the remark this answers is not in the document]
raplemie: Changed

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useTheme();
+  return (
+    <DropdownMenu
+      menuItems={menuItems}
+      onShow={() => setIsMenuOpen(true)}
+      onHide={() => setIsMenuOpen(false)}
+    >
+      {/** TODO: Validate
+       *   DropdownButton: Can't use as is because iui-thumnbail-area would be set to iui-icon.
+       */}
```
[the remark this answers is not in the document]
raplemie: Changed to use DropdownButton, CSS will be updated.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Now up to the component user to provide the full thing, so no longer relevant.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: This prop no longer exist, the question remains but we will want to adress that in `HeaderApplicationButton`

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,206 @@
...
+        className: cx('iui-icon', appIcon.props.className),
+        'aria-hidden': true,
+      })
+    : appIcon;
+  return (
+    <header className={cx('iui-header', { 'iui-slim': slim })} {...rest}>
+      <div className='iui-left'>
+        {/* TODO: Validate
+         * Button: Can't use as is because the 'iui-brand-logo' would also get 'iui-icon' and
+         *         associated behavior. (Not sure what the impact is)
```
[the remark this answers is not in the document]
raplemie: Changed to Button with a class, CSS will be updated.

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+   * <Tile
+   *  // ...
+   *  thumbnail='/url/to/image.jpg'
+   *  // or
+   *  thumbnail={<UserIcon image={<img src='icon.png' />} />}
+   *  // or
+   *  thumbnail={<SvgImodel2 />}
+   * />
+   */
+  thumbnail?: string | React.ReactNode;
```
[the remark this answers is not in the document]
raplemie: Removed all of these and we should use startIcon, but we must update so it allows variable size images in there.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: This is pretty much all gone now.

src/core/Header/Header.tsx
raplemie: Changed the API in may ways, will resolve this, we can create a different discussion if needed.

src/core/Header/Header.tsx
raplemie: Removed

src/core/Header/Header.tsx
raplemie: children it is.

src/core/Header/HeaderButton.tsx
raplemie: Changed

src/core/Header/HeaderButton.tsx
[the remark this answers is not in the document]
mayank99: Maybe leave this comment open until we have the css (https://github.com/iTwin/iTwinUI/pull/57)

Or I guess make a new comment when you update the code

src/core/Header/Header.tsx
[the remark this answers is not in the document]
bentleyvk: I would say to leave it as an object if we have some new strings in the future.

src/core/Header/HeaderButton.tsx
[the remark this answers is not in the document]
raplemie: Fixed with new css

raplemie:
There is currently only one story which have everything in it, however it is not built to show up properly in the "Story" panel, it need to be revisited. I'd take suggestions on what would be meaningful stories for this component.

src/core/Header/HeaderTitle.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const HeaderTitle = (props: HeaderTitleProps) => {
+  const { className, children, logo, ...rest } = props;
+
+  useTheme();
+  return (
+    <div
+      className={cx('iui-header-logo', className)}
+      role={rest.onClick ? 'button' : undefined}
+      {...rest}
```
mayank99, line 33: It needs `tabIndex={0}` and even a keydown handler if you're up for it

src/core/Header/HeaderTitle.tsx
```diff
@@ -0,0 +1,45 @@
...
+   * Logo shown before the main content.
+   */
+  logo: JSX.Element;
+} & React.ButtonHTMLAttributes<HTMLDivElement>;
+
+/**
+ * Header Title section
+ * @example
+ * <HeaderTitle logo="<SvgMicrostation />">Microstation</HeaderTitle>
+ * <HeaderTitle logo="<SvgItwin />" />
```
mayank99, line 23: Maybe use real examples (we dont have these svgs in the package) of svgs and possibly one for png too

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return (
+    <>
+      {items.reduce(
+        (previous: React.ReactNode[], current, index) => [
+          ...previous,
+          current,
+          index !== items.length - 1 && (
+            <SvgChevronRight
+              key={`chevron${index}`}
+              aria-hidden={true}
```
mayank99, line 39: No need for `={true}`

src/core/Header/HeaderTitle.tsx
```diff
@@ -0,0 +1,45 @@
...
+import React from 'react';
+
+import { useTheme } from '../utils/hooks/useTheme';
+
+export type HeaderTitleProps = {
+  /**
+   * Logo shown before the main content.
+   */
+  logo: JSX.Element;
+} & React.ButtonHTMLAttributes<HTMLDivElement>;
```
mayank99, line 17: `CommonProps` might be enough here, not sure we want all the native button props when we don't even use it

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react';
+
+import { useTheme } from '../utils/hooks/useTheme';
+
+export type HeaderBreadcrumbsProps = {
+  items: React.ReactNode[];
+};
+
+/**
+ * Header breadcrumbs, adds chevron between each provided items.
```
mayank99, line 17: Either here or in `HeaderButton`, I would add something about how the behavior changes in slim mode

mayank99 (commented):
Some initial comments to help get this finalized

src/core/Header/Header.tsx
```diff
@@ -16,74 +13,48 @@ import { useTheme } from '../utils/hooks/useTheme';
...
-   */
-  breadcrumbsButtons?: React.ReactNode[];
-  /**
-   * Content to display in the center section of the header, should handle iui-slim class if used.
-   */
-  centerContent?: React.ReactNode;
-  /**
-   * Expects `UserIcon`, will be given a drop down if `userMenuItems` is also provided;
+   * User icon
+   * (expects a `UserIcon`, potentially wrapped in a `DropdownMenu`)
```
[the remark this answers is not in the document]
bentleyvk: Maybe worth adding `@example` here?

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,107 @@
...
+    translatedStrings,
+    className,
+    children,
+    ...rest
+  } = props;
+  useTheme();
+  const strings = { ...headerTranslations, ...translatedStrings };
+  return (
+    <header
+      className={cx(className, 'iui-header', { 'iui-slim': isSlim })}
```
bentleyvk, line 83: Move `className` to the end

src/core/Header/Header.tsx
```diff
@@ -16,74 +13,48 @@ import { useTheme } from '../utils/hooks/useTheme';
...
    */
-  slim?: boolean;
+  isSlim?: boolean;
   /**
    * Provide localized strings.
    */
-  translatedHeader?: HeaderTranslations;
-};
+  translatedStrings?: HeaderTranslations;
+} & React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>;
```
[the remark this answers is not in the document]
raplemie: I thought iTwinUI-React components allowed using the core component default props, that's why it's there. I can remove them if we dont want to let the users do that. I guess we want to keep `className` ? Are there other props that should be kept ?

src/core/Header/Header.tsx
```diff
@@ -16,74 +13,48 @@ import { useTheme } from '../utils/hooks/useTheme';
...
 
 const headerTranslations: HeaderTranslations = {
```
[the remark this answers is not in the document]
raplemie: Had that because it was initialy in the HTML example, I'm not familiar with what aria-label really do. I can remove it if it is not something that is ususally there.

src/core/Header/HeaderTitle.tsx
raplemie: I can add `tabIndex`, but a keydown handler that does what ?

mayank99: keydown handler to call onClick when space and enter are pressed.

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#required_javascript_features

src/core/Header/HeaderTitle.tsx
raplemie: I can use <SvgPlaceholder> or <svg ... />, is there a preference ? (Not sure its worth having the full content of an svg here.)

src/core/Header/HeaderTitle.tsx
raplemie: Shouldnt be Button, it was with the previous implementation, for the same reason as the Header, but I'll change to `CommonProps`

src/core/Header/HeaderTitle.tsx
raplemie: Hmm, what does `tabIndex={0}` supposed to do ? Is it only if onClick is defined ?

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: used `CommonProps`

src/core/Header/Header.tsx
```diff
@@ -19,9 +19,9 @@ export type HeaderTranslations = {
 export type HeaderProps = {
   /**
    * Application title.
-   * (expects `HeaderApplicationButton`)
+   * (expects `HeaderTitle`)
    */
-  appButton: React.ReactNode;
+  appTitle: React.ReactNode;
```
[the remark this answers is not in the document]
raplemie: Renamed component to `HeaderLogo`

src/core/Header/HeaderTitle.tsx
mayank99: tabindex=0 makes the element focusable using tab key

src/core/Header/Header.tsx
raplemie: Updated

src/core/Header/HeaderTitle.tsx
mayank99: Any `Svg-` that's actually valid. So I guess use `SvgImodelHollow`

src/core/Header/Header.tsx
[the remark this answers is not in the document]
mayank99: Try Omit<CommonProps, 'title'> because title does not make sense here. This will give you className and style

I don't think we'd need anything else, as long as we are forwarding `...rest` (to allow data- and aria- attributes)

src/core/Header/Header.tsx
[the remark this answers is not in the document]
mayank99: After some research I'd say leave it in, but name can still be improved (i.e. "defaultTranslations").

And in the component, you can change "strings" to "headerTranslations" or anything more descriptive

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *     <HeaderButton key={...} ... />
+ *   ]}
+ * />
+ */
+export const HeaderBreadcrumbs = (props: HeaderBreadcrumbsProps) => {
+  const { items } = props;
+
+  useTheme();
+  return (
+    <>
```
mayank99, line 31: I did some research on accessible breadcrumbs and found that we should use `nav` with `aria-label='breadcrumbs'`. And the active breadcrumb should have `aria-current='page'`. https://www.aditus.io/patterns/breadcrumbs/#technical-considerationg

That's exactly how bootstrap does it too: https://getbootstrap.com/docs/4.0/components/breadcrumb/

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+import HeaderTitle from '../../src/core/Header/HeaderTitle';
+import HeaderBreadcrumbs from '../../src/core/Header/HeaderBreadcrumbs';
+import SvgHelpCircularHollow from '@itwin/itwinui-icons-react/cjs/icons/HelpCircularHollow';
+import SvgVersion from '@itwin/itwinui-icons-react/cjs/icons/Pin';
+import SvgNetwork from '@itwin/itwinui-icons-react/cjs/icons/Network';
+import SvgNotification from '@itwin/itwinui-icons-react/cjs/icons/Notification';
+
+export default {
+  title: 'Core/Header',
+  component: Header,
```
mayank99, line 29: You can add HeaderLogo and HeaderButton as subcomponents to show their props in Docs tab

src/core/Header/HeaderTitle.tsx
mayank99: If adding keydown is too much work, then I would skip tabindex too, because they both go together. I don't know if the app logo necessarily needs keyboard navigation.

@veekeys @bentleyvk Thoughts?

src/core/Header/HeaderBreadcrumbs.tsx
raplemie: Updated `HeaderButton`, breadcrumbs does not really do anything specific with slim.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
mayank99: image

I was talking about line 84 not line 80 😄

src/core/Header/HeaderBreadcrumbs.tsx
raplemie: If we want to add an element containing the breadcrumbs the css will need to be updated as the `.iui-chevron` currently require to be direct descendant of `.iui-left`.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: Added examples to every props and an overall to the component itself.

src/core/Header/Header.tsx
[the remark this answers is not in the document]
raplemie: I think I did what you meant, let me know if its right.

src/core/Header/HeaderTitle.tsx
raplemie: Added the keydown handler based on the example provided by the link, added related tests

src/core/Header/HeaderTitle.tsx
raplemie: Changed to `CommonProps`

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * <HeaderLogo logo={<SvgImodelHollow />} />
+ * <HeaderLogo logo={<img src='image.png' />} />
+ * <HeaderLogo logo={<img src='data:image/png;base64,...' />} >Downloaded Image</HeaderLogo>
+ */
+export const HeaderLogo = (props: HeaderLogoProps) => {
+  const { className, children, logo, onClick, ...rest } = props;
+  const keyDownHandler = React.useCallback<
+    React.KeyboardEventHandler<HTMLDivElement>
+  >(
+    (e) => {
```
mayank99, line 40: This doesn't need to be a `useCallback` since it's just being passed into a div (not a custom react component).

You can instead set the type on event as `React.KeyboardEvent<HTMLDivElement>`

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,68 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import '@itwin/itwinui-css/css/header.css';
+
+import cx from 'classnames';
+import React from 'react';
+
+import { useTheme } from '../utils/hooks/useTheme';
+import { CommonProps } from '../utils/props';
+
+export type HeaderLogoProps = React.PropsWithChildren<
```
mayank99, line 13: Would be better to define `children` explicitly rather than use `PropsWithChildren` so that we can add a JSDoc comment to it.

src/core/Header/HeaderLogo.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const logo = container.querySelector(
+    '.iui-header-button-icon:only-child',
+  ) as HTMLDivElement;
+  expect(logo).toBeTruthy();
+
+  const label = container.querySelector('.iui-label') as HTMLSpanElement;
+  expect(label).toBeNull();
+});
+
+it('trashes wrong logo type (JS only)', () => {
```
mayank99, line 81: This is interesting 😮

package.json
```diff
@@ -40,7 +40,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
-    "@itwin/itwinui-css": "^0.11.0",
+    "@itwin/itwinui-css": "^0.13.1",
```
mayank99, line 43: Can you resolve merge conflict?

mayank99 (commented):
Close to being done 👍

src/core/Header/HeaderLogo.tsx
raplemie: Updated

src/core/Header/HeaderLogo.tsx
raplemie: Added for `HeaderLogo` and `Header` with positionning information, couldnt find better to put in there.

stories/core/Header.stories.tsx
raplemie: Added

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,163 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import '@itwin/itwinui-css/css/header.css';
```
bentleyvk, line 5: @mayank99 @veekeys I don't remember now but shouldn't this be like the last import?

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,65 @@
...
+   * Logo shown before the main content.
+   */
+  logo: JSX.Element;
+  /**
+   * Click event handler.
+   * Will update the Logo to have mouse and keyboard interaction if provided.
+   */
+  onClick?: () => void;
+  /**
+   * Children is put on the right of the logo.
```
bentleyvk, line 24: Maybe write explicitly that it should be app name?

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * @example
+ * <HeaderLogo logo={<SvgImodelHollow />}>iTwin Application</HeaderLogo>
+ * <HeaderLogo logo={<SvgImodelHollow />} />
+ * <HeaderLogo logo={<img src='image.png' />} />
+ * <HeaderLogo logo={<img src='data:image/png;base64,...' />} >Downloaded Image</HeaderLogo>
+ */
+export const HeaderLogo = (props: HeaderLogoProps) => {
+  const { className, children, logo, onClick, ...rest } = props;
+  const keyDownHandler = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    if (onClick && (e.key === 'Enter' || e.key === ' ')) {
```
bentleyvk, line 40: Also check for `Spacebar` because some browsers return that instead of `' '`.

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import '@itwin/itwinui-css/css/header.css';
+import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
+
+import React from 'react';
+
+import { useTheme } from '../utils/hooks/useTheme';
+
+export type HeaderBreadcrumbsProps = {
+  items: React.ReactNode[];
```
bentleyvk, line 13: Add comment

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Header button that handles slim state of the `Header` it's in.
+ * When in slim mode, will only display the name and reduce icon size.
+ * Will display an arrow on the right if `menuItems` prop is provided.
+ * @example
+ * <HeaderButton name='Project A' description='0n00434' />
+ * <HeaderButton name='Project B' description='0n00434' startIcon={<SvgImodelHollow />}/>
+ * <HeaderButton name='Project C' startIcon={<img src='project.png' />}/>
+ * <HeaderButton name='Project D' isActive />
+ * <HeaderButton name='Project D' isActive menuItems={...}/>
```
bentleyvk, line 46: Minor but can you make this one as the first example as this one should be used the most.

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+  DropdownMenu,
+  getUserColor,
+  Header,
+  HeaderProps,
+  IconButton,
+  Input,
+  MenuItem,
+  UserIcon,
+} from '../../src/core';
+import HeaderButton from '../../src/core/Header/HeaderButton';
```
bentleyvk, line 19: I think these can be imported from `/src/core` too

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+  </MenuItem>,
+  <MenuItem key={2} value={2} onClick={buildClickHandler(menu, close)}>
+    {menu} item #2
+  </MenuItem>,
+  <MenuItem key={3} value={3} onClick={buildClickHandler(menu, close)}>
+    {menu} item #3
+  </MenuItem>,
+];
+
+export const full: Story<HeaderProps> = (args) => {
```
bentleyvk, line 64: Use PascalCase. I don't remember if we had some troubles with camelCase.

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+            <HeaderButton
+              key='project'
+              menuItems={buildMenu('Project')}
+              name={'Project A (Super Size Edition)'}
+              description={'YJC-2249'}
+              startIcon={<SvgNetwork />}
+            />,
+            <HeaderButton
+              key='iModel'
+              menuItems={buildMenu('IModel')}
```
bentleyvk, line 95: Should be `iModel` 😀

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+            margin: 12,
+            transition: 'all 0.2s ease',
+          }}
+        />
+      </div>
+    </Header>
+  );
+};
+
+export const basic: Story<HeaderProps> = (args) => {
```
bentleyvk, line 160: This one and another one story looks odd to me. Maybe we should have a story without breadcrumbs.

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+              }
+              title='Terry Rivers'
+            />
+          </IconButton>
+        </DropdownMenu>
+      }
+      menuItems={buildMenu('More')}
+      {...args}
+    >
+      <div>
```
bentleyvk, line 143: Is this div needed?

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+    HeaderLogo,
+    HeaderButton,
+  },
+  argTypes: {
+    appLogo: { control: { disable: true } },
+    breadcrumbs: { control: { disable: true } },
+    actions: { control: { disable: true } },
+    userIcon: { control: { disable: true } },
+    menuItems: { control: { disable: true } },
+    children: { control: { disable: true } },
```
bentleyvk, line 41: Do the same with `style` and `className`

src/core/Header/HeaderLogo.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+import HeaderLogo from './HeaderLogo';
+
+it('renders default correctly', () => {
+  const { container } = render(
+    <HeaderLogo logo={<SvgPlaceholder />}>Application</HeaderLogo>,
+  );
+
+  const root = container.querySelector('.iui-header-logo') as HTMLDivElement;
+  expect(root).toBeTruthy();
+  expect(root.getAttribute('role')).toBeNull();
```
bentleyvk, line 18: Maybe use `toBeFalsy` because `undefined` !== `null`

src/core/Header/Header.tsx
veekeys: Yeah, was something...
I think we needed to put it after useTheme import hook...?

src/core/Header/Header.tsx
```diff
@@ -0,0 +1,163 @@
...
+  useTheme();
+  const headerTranslations = { ...defaultTranslations, ...translatedStrings };
+  return (
+    <header
+      className={cx('iui-header', { 'iui-slim': isSlim }, className)}
+      {...rest}
+    >
+      <div className='iui-left'>
+        {appLogo}
+        {breadcrumbs && <div className='iui-divider' />}
```
veekeys, line 141: Shouldnt we have divider always after app logo even if there are no breadcrumbs??
@FlyersPh9

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,46 @@
...
+ *   ]}
+ * />
+ */
+export const HeaderBreadcrumbs = (props: HeaderBreadcrumbsProps) => {
+  const { items } = props;
+
+  useTheme();
+  return (
+    <>
+      {items.reduce(
```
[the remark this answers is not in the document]
veekeys: Raph is a fan of reduce from what I have seen.

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { useTheme } from '../utils/hooks/useTheme';
+
+export type HeaderButtonProps = {
+  /**
+   * Name.
+   */
+  name: React.ReactNode;
+  /**
+   * Description
```
veekeys, line 24: Looks really poor :D
Could you add maybe some more wording? description could have mentioned the font size change for example

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+    name,
+    description,
+    isActive,
+    menuItems,
+    className,
+    startIcon,
+    ...rest
+  } = props;
+  const buttonProps: ButtonProps & { styleType: 'borderless' } = {
+    startIcon: React.isValidElement<{ className: string }>(startIcon)
```
veekeys, line 59: is explicit className state needed here?

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * <HeaderButton name='Project B' description='0n00434' startIcon={<SvgImodelHollow />}/>
+ * <HeaderButton name='Project C' startIcon={<img src='project.png' />}/>
+ * <HeaderButton name='Project D' isActive />
+ * <HeaderButton name='Project D' isActive menuItems={...}/>
+ */
+export const HeaderButton = (props: HeaderButtonProps) => {
+  const {
+    name,
+    description,
+    isActive,
```
veekeys, line 52: assign default value

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,65 @@
...
+  children?: React.ReactNode;
+} & CommonProps;
+
+/**
+ * Header Title section
+ * @example
+ * <HeaderLogo logo={<SvgImodelHollow />}>iTwin Application</HeaderLogo>
+ * <HeaderLogo logo={<SvgImodelHollow />} />
+ * <HeaderLogo logo={<img src='image.png' />} />
+ * <HeaderLogo logo={<img src='data:image/png;base64,...' />} >Downloaded Image</HeaderLogo>
```
veekeys, line 35: no need space before right arrow..

src/core/Header/HeaderButton.tsx
veekeys: Also, if you could add space before closing tags, where missed... :D

src/core/Header/HeaderButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+    menuItems,
+    className,
+    startIcon,
+    ...rest
+  } = props;
+  const buttonProps: ButtonProps & { styleType: 'borderless' } = {
+    startIcon: React.isValidElement<{ className: string }>(startIcon)
+      ? React.cloneElement(startIcon, {
+          className: cx('iui-header-button-icon', startIcon.props.className),
```
veekeys, line 61: I dont see this class added in css html examples to the breadcrumbs...

veekeys: a ok, found it...

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onClick();
+    }
+  };
+  useTheme();
+  return (
+    <div
+      className={cx('iui-header-logo', className)}
+      role={onClick ? 'button' : undefined}
+      tabIndex={onClick ? 0 : undefined}
+      onKeyDown={onClick ? keyDownHandler : undefined}
```
veekeys, line 51: YOu check onClick double time in here and in function.. Remove one

src/core/Header/HeaderLogo.tsx
```diff
@@ -0,0 +1,65 @@
...
+      onClick={onClick}
+      {...rest}
+    >
+      {React.isValidElement<{ className: string }>(logo)
```
veekeys, line 55: I prefer && syntax versus undefined, but i guess no big diff... Just less undefined, looks better :D

src/core/index.ts
```diff
@@ -158,3 +158,6 @@ export { getUserColor } from './utils/common';
 
 export { useTheme } from './utils/hooks/useTheme';
 export type { ThemeType } from './utils/hooks/useTheme';
+
+export { Header } from './Header';
```
veekeys, line 162: I think in barrel we need to export all the header compnents as they will be used to build Header...
Also, could you put it alphabeticaly ordered? :D

stories/core/Header.stories.tsx
veekeys: Not right now, I think.. need to add them in barrel export first.

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+          Microstation
+        </HeaderLogo>
+      }
+      breadcrumbs={
+        <HeaderBreadcrumbs
+          items={[
+            <HeaderButton
+              key='project'
+              menuItems={buildMenu('Project')}
+              name={'Project A (Super Size Edition)'}
```
veekeys, line 89: Minor, but no need curlies if it is simple string value..

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,209 @@
...
+          onClick={() => action('Clicked on the title')()}
+        >
+          Microstation
+        </HeaderLogo>
+      }
+      breadcrumbs={
+        <HeaderBreadcrumbs
+          items={[
+            <HeaderButton
+              key='project'
```
veekeys, line 87: Oh, I think we should be adding key prop as we are taking the array and itterating through it.. Users might not add it and I dont think they will see any warning before the build...

stories/core/Header.stories.tsx
veekeys: Yeah..Do not look really useful.. I guess just have a story with simple buttons (not dropdowns) in breadcrumbs and you can have on the right side only user-icon lets say..
SOmething from more used examples and that should be enough.

veekeys:
I have left few minor comments, but in general looks great!
The only thought is about tests, cause currently they are written in a harder way to read (nth-child selectors, for example). Just not sure, if we want to spend time changing them a little, or will be fine for now.
Thoughts @bentleyvk @mayank99 ?

stories/core/Header.stories.tsx
mayank99: I think users should see a warning. Key is their responsibility

veekeys: Disagree. If we are iterating through the array, key is our respnsibility. It always was and we handle that in other places. I dont see this different here.

src/core/Header/HeaderLogo.tsx
raplemie: Added

src/core/Header/HeaderLogo.tsx
raplemie: That's one of the reason I did not want to handle that at all... but documentation does not mention "spacebar" as a valid key: https://www.w3.org/TR/uievents-key/#keys-whitespace

src/core/Header/HeaderLogo.tsx
raplemie: I did not want to have a keyDownHandler if there was no reason for it to be there, but I can leave it always there... And if I dont check it in the function, there is a branch that cannot be reached in the tests.

src/core/Header/HeaderLogo.tsx
raplemie: Changed

src/core/Header/Header.tsx
raplemie: Moved for all the files

src/core/Header/Header.tsx
raplemie: Yep, it's how it was, but it was a change requested yesterday by @mayank99

src/core/Header/HeaderLogo.tsx
mayank99: Spacebar is used in older browsers 
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values#:~:text=older%20browsers%20may%20return%20%22spacebar%22%20instead%20of%20%22%20%22

src/core/Header/HeaderBreadcrumbs.tsx
raplemie: Added

src/core/Header/HeaderBreadcrumbs.tsx
mayank99: Fixed in https://github.com/iTwin/iTwinUI/pull/81 and will be available in next release

src/core/Header/HeaderButton.tsx
raplemie: I'm not aware of a font-size change, but it does disappear in slim mode.

src/core/Header/HeaderButton.tsx
raplemie: Moved the menuItems part to the Project A example

src/core/Header/HeaderButton.tsx
raplemie: added

src/core/Header/HeaderButton.tsx
raplemie: if I don't put it, it complains that I'm using a prop that cant be validated in the cloneElement. And it cannot "guess" that its there because the element is a JSX.Element, I'll change it to `ClassNameProps` though

src/core/Header/HeaderLogo.test.tsx
raplemie: Well, getAttribute() returns either a string | null: 
image

src/core/index.ts
raplemie: Pretty sure the script created that, I can move it and add other stuff.

stories/core/Header.stories.tsx
raplemie: Added to barrel, moved all imports to /src/core

stories/core/Header.stories.tsx
raplemie: Added

stories/core/Header.stories.tsx
raplemie: Should probably be handled by prettier or linter or something...

stories/core/Header.stories.tsx
raplemie: Well, I dont really see the use of those 2 different stories either, I dont have example of "more used", the full one is the recommended approach. The Full example already shows a button with no dropdown with the "Version" menu. I can addit to the "SimpleUserIcon" and name this other one the "Basic" ?

stories/core/Header.stories.tsx
raplemie: Nice, I based my example on the only that was wrong, found it odd too... (ErrorPage.stories)

stories/core/Header.stories.tsx
raplemie: @veekeys I dont see any example where we clone an element to add the key, all other places we create elements based on object array (Table, TableRow, HorizontalTab), we are not provided with an array of ReactNode in these cases. I can add that but it seems weird.

src/core/Header/HeaderLogo.tsx
raplemie: Added

stories/core/Header.stories.tsx
raplemie: Kinda was when I was using the `Full.args = { children: ....}` format, probably not needed anymore.

raplemie:
> The only thought is about tests, cause currently they are written in a harder way to read (nth-child selectors, for example). Just not sure, if we want to spend time changing them a little, or will be fine for now.
> Thoughts @bentleyvk @mayank99 ?

The only thing that the header is really doing is ordering it's components, if we dont test the order, there isnt much to test, hence the positional selectors. (Otherwise, I can remove all of those checks, but then it means that if someone for some reason changes the order because of a bad copy/paste, nothing will ring.)

stories/core/Header.stories.tsx
veekeys: Ok, then my bad.. I checked and it gives error if there is no key, so should be fine...

src/core/index.ts
veekeys: Yeah, script does not handle alphabetical ordering, sadly.

stories/core/Header.stories.tsx
```diff
@@ -0,0 +1,215 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { Meta, Story } from '@storybook/react/types-6-0';
```
veekeys, line 5: Can import from '@storybook/react'

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## [1.5.0]
+
+`2021-05-13`
+
+### What's new
+
+- **Added new `Header` component.** Use `HeaderLogo`, `HeaderBreadcrumbs` and `HeaderButton` to build application header, support `isSlim` for 3D content.
+- **Added new `HeaderLogo` component.** Use within `Header` component, support size adjusting logo for `isSlim` header mode.
+- **Added new `HeaderBreadcrumbs` component.** Use within `Header` component, adds chevrons between items.
+- **Added new `HeaderButton` component.** Use within `HeaderBreadcrumbs` component, support size adjusting `startIcon` and `description` display depending on `Header` `isSlim` prop.
```
mayank99, line 12: I meant to say this earlier but these subcomponent points are kinda unnecessary, or at least they don't warrant so much extra text.

Try this:
```suggestion
- **Added new `Header` component,** with a condensed version available through `isSlim` prop.
  - **Added new `HeaderLogo` component** for use with the Header `appLogo` prop.
  - **Added new `HeaderBreadcrumbs` and `HeaderButton` component** for use with the Header `breadcrumbs` prop.
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## [1.5.0]
+
+`2021-05-13`
```
mayank99, line 5: Placeholder would be better here.

https://github.com/iTwin/iTwinUI-react/blob/4e5c2f174e8bb2f661d04c979ffbd814e6268a13/CONTRIBUTING.md#L128

raplemie: Then contributing should be modified to just mention the placeholder... Everyone would waste less time updating this.

src/core/Header/HeaderBreadcrumbs.tsx
mayank99: use css v0.14.0

https://github.com/iTwin/iTwinUI/blob/fbe27813a05cd53cefdfd97a2c45ce2a90a893e1/backstop/tests/header.html#L117

CHANGELOG.md
raplemie: What I mean is that as an external contributor, there is no way I can predict when and in which version this will actually be published, meaning that you guys will always have to validate whatever is in the CHANGELOG, so, the Contrib could be reworded to take that into account. The less doubt there is, the simpler it is to contribute, something like:
``` The `CHANGELOG.md` file must be updated for any new components or changes that you add. Add a placeholder version and date (on the top) and iTwinUI-React team will handle the actual date and version when publishing the package.```
Which tells me that I dont have to worry at all about these 2 informations, but I must put it there, and it will be handled correctly. :)

On top of that, it could be interresting to ask a format of the changelog, that you guys will be able to adapt for consistency and verbosity (just like you did, but instead of me trying to figure what to write right, I could have added only that I added `Header`, `HeaderButton`... components and let you guys figure how you want to surface it.

File: src/core/Header/HeaderBreadcrumbs.tsx (hunk 27 above)
Comment (raplemie) [reply src/core/Header/HeaderBreadcrumbs.tsx]: Updated the `HeaderBreadcrumbs` and `HeaderButton` components, the strings are localized so I added the `translatedStrings` prop to both components along tests for it (added such a test for `Header` component that was missing for `translatedStrings`) I also added JSDoc to explain where these strings were used to help fill them correctly. (So there is on only this line to review.)

File: stories/core/Header.stories.tsx (hunk 57 above)
Comment (raplemie) [reply stories/core/Header.stories.tsx]: Updated (HorizontalTab have it this way too, with the `types-6-0` part)

File: src/core/Header/HeaderBreadcrumbs.tsx (hunk 27 above)
Comment (mayank99) [reply src/core/Header/HeaderBreadcrumbs.tsx]: Now it's just missing `aria-current=location`

File: stories/core/Header.stories.tsx (hunk 41 above)
Comment (raplemie) [reply stories/core/Header.stories.tsx]: I dont have a story with the translatedStrings, but, since this is only changing the aria-label, I'm not sure its worth a story.

File: src/core/Header/HeaderButton.tsx [hunk 60]
```diff
@@ -69,6 +86,7 @@ export const HeaderButton = (props: HeaderButtonProps) => {
       },
       className,
     ),
+    ['aria-label']: isActive ? buttonTranslations.page : undefined,
```
mayank99, line 89: This should really be `aria-current` and it does not need translations. Also, use `location` not `page`. If the user wants `page` for any reason, they can still override.

See https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent

src/core/Header/HeaderBreadcrumbs.tsx
raplemie: I mixed the `page` with `aria-label` instead of `aria-current`, which does not need to be localized at all.

raplemie: > Now it's just missing `aria-current=location`
This is referring to the other comment in `HeaderButton` or it should have that here too ?

mayank99: The other comment. I will close this.

src/core/Header/HeaderButton.tsx
raplemie: Fixed

src/core/Header/HeaderBreadcrumbs.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const HeaderBreadcrumbs = (props: HeaderBreadcrumbsProps) => {
+  const { items, translatedStrings } = props;
+  const breadcrumbsTranslations = {
+    ...defaultTranslations,
+    ...translatedStrings,
+  };
+
+  useTheme();
+  return (
+    <nav aria-label={breadcrumbsTranslations.breadcrumbs}>
```
mayank99, line 59: If you pass `...rest` here, I think this won't need translations, right?

raplemie: Yes, but it doesnt really makes it obvious that we are passing in english values in there that should be translated. What would be the type of the `HeaderBreadcrumbsProps` ?

mayank99: I don't think it needs a specific type. Just passing rest is enough. All JSX components can accept `data-` and `aria-` attributes without a special type.

As for the "makes it obvious" debate, you can say the same thing for the prop itself. It's not obvious that the translation is used for aria-label.

src/core/Header/HeaderButton.tsx
raplemie: Because I confused computed attribute names, and linter does not care about this particular thing apparently. Fixed.

src/core/Header/HeaderBreadcrumbs.tsx
raplemie: I dont want to debate I just want to know what to put in there according to iTwinUI standards, so we can finalize this...

raplemie: Changed to use `{...rest}` on the `nav` component.

mayank99 (approved):

PR #46, opened 2021-05-04
Update to itwinui-css 0.12.0

mayank99:
Updated @itwin packages:
- itwinui-css: 0.11.0 → 0.12.0
- itwinui-icons-react: 0.1.0 → 1.1.0
- itwinui-illustrations-react: 0.1.0 → 1.0.0

Changes made to reflect the new packages:
- Updated `UserIcon` classes and dom structure.
- Added `aria-hidden` and `aria-label` attributes where applicable.
- Updated `Modal` and `Tile` to use correct icons.
- Added `iui-icon` and `aria-hidden` to `Table` sort icon (will need to do the same for filter).
- Removed size from `HorizontalTabs` story (pill).

### DIFF
--- a/src/core/UserIcon/UserIcon.tsx
+++ b/src/core/UserIcon/UserIcon.tsx
@@ -89,9 +89,9 @@ export const UserIcon = (props: UserIconProps) => {
   const statusTitles = { ...defaultStatusTitles, ...translatedStatusTitles };
 
   const iconMap: { [key in UserIconStatus]: JSX.Element } = {
-    away: <SvgAway className='iui-status-symbol' />,
-    offline: <SvgCloseSmall className='iui-status-symbol' />,
-    online: <SvgCheckmark className='iui-status-symbol' />,
+    away: <SvgAway className='iui-status-symbol' aria-hidden />,
+    offline: <SvgCloseSmall className='iui-status-symbol' aria-hidden />,
+    online: <SvgCheckmark className='iui-status-symbol' aria-hidden />,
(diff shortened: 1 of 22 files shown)


src/core/UserIcon/UserIcon.tsx
```diff
@@ -89,35 +89,35 @@ export const UserIcon = (props: UserIconProps) => {
...
+    away: <SvgAway className='iui-status-symbol' />,
+    offline: <SvgCloseSmall className='iui-status-symbol' />,
+    online: <SvgCheckmark className='iui-status-symbol' />,
     busy: <></>,
   };
 
   return (
     <span
-      className={cx(`iui-user-icons-${size}`, className)}
+      className={cx(`iui-user-icon iui-${size}`, className)}
```
veekeys, line 100: Medium size is not existing in css anymore...
But I think it makes sense to have, just to understand the sizing.
Or we could have an item to change this (to default? does not sound better...) with the next major update, but that would be only to reflect better the css. Thoughts?
Id personally be for leaving the medium in this case, to be honest.

veekeys (dismissed):

src/core/UserIcon/UserIcon.tsx
mayank99: We can leave the prop but not apply the class?

veekeys: I guess we could, for cleanliness..

veekeys (approved):

PR #50, opened 2021-05-07
Table expander implementation

Juliakas:
EDIT: In addition to subComponent, I added support for subrows. Subrows are pretty simple to use, just use a `subRows` field in `data` prop object and make a tree-like data structure. For example `const data = [{ ...someData1, subRows: [{..someSubrowsData1}]},{ ...someData2, subRows: [{..someSubrowsData2}]}]`. Padding for nesting has to be provided manually for now.

## Checklist

### Subcomponent
https://user-images.githubusercontent.com/36437897/117477921-3a53e100-af67-11eb-8e9a-2f04f718c430.mp4
### Subrows demo
https://user-images.githubusercontent.com/36437897/122430183-09ba8a80-cf9c-11eb-95a7-3e96cd7bcf9a.mp4

## Additional Comments
~~This PR is mainly for initiating discussion and evaluating design solutions to match everyone's expectation on how table expanding should work.~~

For me at least, its important to have a flexible way to provide expanding behavior to iTwinUI tables. Solution is inspired by react-table example: https://react-table.tanstack.com/docs/examples/sub-components

- Provided component can be in free, unconstrained format: can be a nested table that fits well, or just some additional information container like in the video example.
- ~~Custom styles can be provided to 'iui-tables-rowgroup' class that wraps both expanded row and expanded content.~~
- Expanding can be invoked freely from Table parent, for example it can be done by providing a column with a button - clicking it will expand a row that is associated with it.

~~Note: there probably should be small stylesheet adjustments, as this selector will break as I have inserted iui-tables-rowgroup div in between~~


src/core/Table/Table.tsx
```diff
@@ -86,6 +87,20 @@ export type TableProps<
    * @default 300
    */
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  expandedSubComponent?: (row: Row<T>) => ReactNode;
+  /**
```
veekeys, line 95: Not sure about the prop name.. But we can discuss this.

src/core/Table/Table.tsx
```diff
@@ -230,6 +249,7 @@ export const Table = <
     },
     useFlexLayout,
     useSortBy,
+    useExpanded,
```
veekeys, line 252: Just curious..Why decided to put this in the middle? :D

src/core/Table/TableRowMemoized.tsx
```diff
@@ -42,18 +46,21 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
+            style: getCellStyle(cell.column),
+          });
+          return (
+            <div {...cellProps} key={cellProps.key}>
+              {cell.render('Cell')}
+            </div>
+          );
+        })}
+      </div>
+      {isExpanded && expandedSubComponent?.(row)}
```
veekeys, line 63: I think we might want to arrange spacing ourselves.
image

And somehow I feel that content should start with the 2nd column (I mean not to be aligned with expander column)

@FlyersPh9 Could you check up this PR from visual aspect?

stories/core/Table.stories.tsx
```diff
@@ -307,6 +308,98 @@ Sortable.args = {
...
+    [],
+  );
+
+  const tableColumns = useMemo(
+    () => [
+      {
+        Header: 'Table',
+        columns: [
+          {
+            id: 'expander',
```
veekeys, line 325: Oh, I think we want to provide this from us. Something like selectable.
Cause this is defined by standards, so we could control that from us, if expanded is enabled (need new prop?). And then check, if we can give users ability to override this if they want.

stories/core/Table.stories.tsx
```diff
@@ -307,6 +308,98 @@ Sortable.args = {
...
+};
+
+Expandable.args = {
+  data: [
+    { name: 'Name1', description: 'Description1' },
+    { name: 'Name3', description: 'Description3' },
+    { name: 'Name2', description: 'Description2' },
+  ],
+};
+
```
veekeys, line 402: Also, I think we will need a story with full features on.
THis way we could test more real life scenarios and also see, how features work together.
For this case, we have a standard, how selected expanded table row looks like, so. need to check that:
image

Not necessarily for this PR, we could create a separate one.
@bentleyvk @mayank99

src/core/Table/Table.tsx
Juliakas: I initially put it as last parameter. But this is the reason, I had to put it in the middle :D
image

stories/core/Table.stories.tsx
Juliakas: I moved the expander column to table with latest commit. The behavior can be disabled, if "provideDefaultExpander" prop (true by default) is set to false.

@veekeys What do you think about this?

src/core/Table/Table.tsx
```diff
@@ -86,6 +87,20 @@ export type TableProps<
    * @default 300
    */
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  expandedSubComponent?: (row: Row<T>) => ReactNode;
```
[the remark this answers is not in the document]
Juliakas: Edited to use React.ReactNode

stories/core/Table.stories.tsx
Juliakas: image
Added another storybook entry for that.

src/core/Table/Table.tsx
mayank99: Maybe just `subcomponent` (which is what react-table v6 called it) or `subRow`

src/core/Table/TableRowMemoized.tsx
FlyersPh9: This looks really good, I'm happy you used the borderless button for the expand / collapse arrow.

Question about the row height: is the borderless button causing the rows height to be taller than it would be if it were just text?  If yes, can we adjust the padding so that the row height is close to / the same height as a table row without a button in it?

Aside from that, gets my 👍.

veekeys: @FlyersPh9  And what about padding inside the expanded content? Shouldnt it have some or we leave that freely for the user?

FlyersPh9: Can you place 1 line of text within the expanded section and try to get the height of that to match the row above?

Juliakas: @FlyersPh9  Yes, it looks like the button is stretching the row a bit since small icon button has fixed height of 27px and the text in other table cells takes up only 22px (due to line-height: 22px). Just by adding cellClassName: 'iui-tables-slot' seems to solve that problem since it removes the paddings from that particular cell. I see the same strategy used for populating checkbox columns.

src/core/Table/Table.tsx
```diff
@@ -149,6 +170,9 @@ export const Table = <
     onBottomReached,
     onRowInViewport,
     intersectionMargin = 300,
+    expandedSubComponent,
+    onExpand,
+    provideDefaultExpander: provideExpanderColumn = true,
```
veekeys, line 175: New variable looks weird here.. Why you needed it?

src/core/Table/Table.tsx
```diff
@@ -333,6 +392,8 @@ export const Table = <
                 intersectionMargin={intersectionMargin}
                 state={state}
                 key={rowProps.key}
+                expandedSubComponent={expandedSubComponent}
+                isExpanded={row.isExpanded}
```
veekeys, line 396: Sure need separate prop `isExpanded`? you can read the value from the row itself inside `TableRowMemoized`.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -66,5 +73,7 @@ export const TableRowMemoized = React.memo(
     prevProp.onBottomReached === nextProp.onBottomReached &&
     prevProp.row.original === nextProp.row.original &&
     prevProp.state.selectedRowIds?.[prevProp.row.id] ===
-      nextProp.state.selectedRowIds?.[nextProp.row.id],
+      nextProp.state.selectedRowIds?.[nextProp.row.id] &&
+    prevProp.expandedSubComponent === nextProp.expandedSubComponent &&
+    prevProp.isExpanded === nextProp.isExpanded,
```
veekeys, line 78: Wouldnt this be covered with the check of row.original ?

stories/core/Table.stories.tsx
veekeys: I do not think expand row should be true by default.. Users should turn it on, when they need it.
I like the hook (same as select). Now we need to think, how we could allow this for the user to customize.. I guess same is applicable to selection column.

veekeys (dismissed):

src/core/Table/Table.tsx
Juliakas: Oh.. this is some weird F2 renaming thing in Vscode. When I renamed prop name, it preserved older name and put it after colon... Will fix it

stories/core/Table.stories.tsx
Juliakas: After further experimenting with this, I noticed a couple of issues when combining multiple features. For instance, upon sorting, bottom row reached gets triggered which results in new data being generated, after that sorting simply gets reset. I think this needs to go on a separate PR where fixes could be applied as well.

src/core/Table/Table.tsx
Juliakas: Fixed with new commits

stories/core/Table.stories.tsx
veekeys: Lets leave this full-on story for separate PR and fixes gonna be there. In this just have a story for expander.

src/core/Table/Table.tsx
Juliakas: I previously tried to just use row and would prefer that option, however React.memo prop check for some reason was skipping updates since `prevProp.row.isExpanded` was always equal to `nextProp.row.isExpanded` . I added `isExpanded` separately to fix this issue and now it seems `nextProp.isExpanded` and `prevProp.isExpanded` are actually different.

I'd like more insight on why it is behaving like that, so for now I passed additional prop like that...

src/core/Table/TableRowMemoized.tsx
Juliakas: Same as my above comment, it wasn't covered by just those prop checks

stories/core/Table.stories.tsx
Juliakas: For me at least, customizing expander from the user side is pretty easy, I only need to add additional Cell at the beggining which invokes `props.row.toggleRowExpanded()` on button click.

I made `provideDefaultExpander` to be `false` by default.

src/core/Table/Table.tsx
Juliakas: Renamed to `subComponent`

src/core/Table/Table.tsx
veekeys: @bentleyvk thats how it should work?

src/core/Table/Table.tsx
veekeys: I think expanded prefix I liked better.. That was concrete.. Cause subComponent could be also not necessarily the ex expanded content..Not sure.

Juliakas:
@FlyersPh9 @veekeys

I am currently tweaking some css styling locally and seeing how it behaves with row selection. This is what I came up with so far:
1. I'm thinking about having slightly different backround for subcompoent to distingush from row. Maybe should add even more transparency than it currently has in the video.
2. I adjusted some border css rules - now if row is selected, border will wrap around entire rowgroup. Also added some transparent borders to prevent table from stretching and shrinking on selecting.

### Light
https://user-images.githubusercontent.com/36437897/118250726-6dd6c400-b4af-11eb-9786-7c4fd5186c31.mp4

### Dark
https://user-images.githubusercontent.com/36437897/118250760-76c79580-b4af-11eb-815c-406136202d08.mp4

If all is good, I can make a PR afterwards at iTwinUI repository.

veekeys:
> @FlyersPh9 @veekeys
> 
> I am currently tweaking some css styling locally and seeing how it behaves with row selection. This is what I came up with so far:
> 
> 1. I'm thinking about having slightly different backround for subcompoent to distingush from row. Maybe should add even more transparency than it currently has in the video.
> 2. I adjusted some border css rules - now if row is selected, border will wrap around entire rowgroup. Also added some transparent borders to prevent table from stretching and shrinking on selecting.
> 
> Anything else that should be added as well? How about a padding rule for subcomponent? I guess it can be overriden in case user really needs to.
> 
> ### Light
>  Light.mp4 
> ### Dark
>  Dark.mp4 
> If all is good, I can make a PR afterwards at iTwinUI repository.

I like the background!
Border should be around whole section according to standards too. Just not sure about the row border in that case..
Anyway, I think this is already a good base to work on and @FlyersPh9 can improve minor details, where he thinks its needed. You can create PR in iTwinUI repo with this one, if want.

src/core/Table/Table.tsx
bentleyvk: For me `subCompenent` is still not clear without description. Maybe `expansionComponent`, `expandedRow` or `expandedContent.`

src/core/Table/Table.tsx
bentleyvk: We had the same issue with select, you can check like this but with `expanded`:
image

src/core/Table/Table.tsx
```diff
@@ -373,6 +432,7 @@ export const Table = <
             const rowProps = row.getRowProps({
               className: cx('iui-tables-row', {
                 'iui-tables-row-active': row.isSelected,
+                'iui-tables-expanded': row.isExpanded,
```
bentleyvk, line 435: Does it do anything?

stories/core/Table.stories.tsx
bentleyvk: I think we could provide object/map of predefined columns. It could have expander and maybe even selection. I now think we shouldn't have provided that `isSelectable` flag...

src/core/Table/Table.tsx
```diff
@@ -88,6 +91,24 @@ export type TableProps<
...
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  subComponent?: (row: Row<T>) => React.ReactNode;
+  /**
+   * Handler for row expand events. Will trigger when expanding and condensing rows.
+   */
+  onExpand?: (
```
bentleyvk, line 102: Maybe return row data and `newState` only like in other callbacks.

Juliakas:
> 
> 
> > @FlyersPh9 @veekeys
> > I am currently tweaking some css styling locally and seeing how it behaves with row selection. This is what I came up with so far:
> > 
> > 1. I'm thinking about having slightly different backround for subcompoent to distingush from row. Maybe should add even more transparency than it currently has in the video.
> > 2. I adjusted some border css rules - now if row is selected, border will wrap around entire rowgroup. Also added some transparent borders to prevent table from stretching and shrinking on selecting.
> > 
> > Anything else that should be added as well? How about a padding rule for subcomponent? I guess it can be overriden in case user really needs to.
> > ### Light
> > Light.mp4
> > ### Dark
> > Dark.mp4
> > If all is good, I can make a PR afterwards at iTwinUI repository.
> 
> I like the background!
> Border should be around whole section according to standards too. Just not sure about the row border in that case..
> Anyway, I think this is already a good base to work on and @FlyersPh9 can improve minor details, where he thinks its needed. You can create PR in iTwinUI repo with this one, if want.

@veekeys 
I'm not sure why I can't directly reply to you on this comment :/ . But anyway, which part of row border exactly? I tried to mimic expanded + selected row from UX standards.

I made a very initial PR for css changes: https://github.com/iTwin/iTwinUI/pull/82
Will fix some failing scenarios now since table structure changes slightly.

src/core/Table/Table.tsx
Juliakas: Yes, just made a PR for styling changes, will be useful for combination of active and expanded rows. And probably for a different background as well. https://github.com/iTwin/iTwinUI/pull/82

src/core/Table/Table.tsx
```diff
@@ -88,6 +91,24 @@ export type TableProps<
    * @default 300
    */
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  subComponent?: (row: Row<T>) => React.ReactNode;
```
bentleyvk, line 98: Not sure whether pass row or row data (`row.original`)

Juliakas: I think `row` could be quite useful since it gives much more information about the row state. I can even do `row.toggleRowExpanded()` inside subcomponent for example.

src/core/Table/Table.tsx
Juliakas: Changed to what you suggested and it does work!

Juliakas:
In this PR I didn't include transition animations - I still need css classes that support trasitions (similar to Toast).

Juliakas:
In addition to subComponent, I added support for subrows. Subrows are pretty simple to use, just use a `subRows` field in `data` prop object and make a tree-like data structure. For example `const data = [{ ...someData1, subRows: [{..someSubrowsData1}]},{ ...someData2, subRows: [{..someSubrowsData2}]}]`. Padding for nesting has to be provided manually for now.

src/core/Table/Table.tsx
bentleyvk: I guess it is not needed anymore.

src/core/Table/Table.tsx
```diff
@@ -21,6 +21,7 @@ import {
   useTable,
   ActionType,
   TableInstance,
+  useExpanded,
```
bentleyvk, line 24: Can you uncomment expand related stuff inside `src\types\react-table-config.ts`

src/core/Table/Table.tsx
```diff
@@ -31,6 +32,8 @@ import SvgSortDown from '@itwin/itwinui-icons-react/cjs/icons/SortDown';
 import SvgSortUp from '@itwin/itwinui-icons-react/cjs/icons/SortUp';
 import { getCellStyle } from './utils';
 import { TableRowMemoized } from './TableRowMemoized';
+import { IconButton } from '../Buttons';
+import { SvgChevronDown, SvgChevronRight } from '@itwin/itwinui-icons-react';
```
bentleyvk, line 36: Import SVGs like we are importing other SVGs (lines 31, 32). We import like that because other projects does not use tree-shaking and their build will end up with all posible icons.

src/core/Table/Table.tsx
bentleyvk: @mayank99 @veekeys can we help Julius with the naming here?

src/core/Table/Table.tsx
```diff
@@ -274,6 +323,9 @@ export const Table = <
       case TableActions.setFilter:
         onFilterHandler(newState, action, previousState, instance);
         break;
+      case TableActions.toggleRowExpanded:
```
bentleyvk, line 326: Add `toggleAllRowsExpanded` too if someone decides to manually make expand all header column.

src/core/Table/Table.tsx
```diff
@@ -88,6 +91,19 @@ export type TableProps<
...
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  subComponent?: (row: Row<T>) => React.ReactNode;
+  /**
+   * Handler for row expand events. Will trigger when expanding and condensing rows.
+   */
+  onExpand?: (state: TableState<T>) => void;
```
bentleyvk, line 102: Add expanded rows data like we do with `onSelect`.

stories/core/Table.stories.tsx
bentleyvk: Maybe instead of `provideDefaultExpander`, we should have a prop for function like `canExpand` or `isExpandable` which would return a boolean whether row should have expansion button.

stories/core/Table.stories.tsx
```diff
@@ -469,6 +472,262 @@ Filters.args = {
...
+    [tableColumns],
+  );
+
+  return (
+    <Table
+      {...rest}
+      columns={tableColumnsWithHeaders}
+      data={data}
+      emptyTableContent='No data.'
+      onExpand={(newState) => {
```
bentleyvk, line 674: You can write jsut `onExpand={onExpand}`.

src/core/Table/Table.tsx
Juliakas: Will do. Just a question, why don't we re-export types from react-table directly?

src/core/Table/Table.tsx
Juliakas: So which data do we want to represent? Should we return row ids for that? For me it would make sense at least.

bentleyvk: I would say expanded rows data (`row.original`) array.

src/core/Table/Table.tsx
bentleyvk: Too hard question 😀

src/core/Table/Table.tsx
mayank99: I'm fine with `expandedContent` but I'm also fine with `subcomponent` because that's the term react-table uses

Juliakas: It is currently renamed to subComponent

src/core/Table/Table.tsx
```diff
@@ -88,6 +92,22 @@ export type TableProps<
...
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
+   */
+  subComponent?: (row: Row<T>) => React.ReactNode;
+  /**
+   * Handler for row expand events. Will trigger when expanding and condensing rows.
+   */
+  onExpand?: (
+    selectedData: T[] | undefined,
```
bentleyvk, line 104: Maybe not `selected` but `expandedData`.

Juliakas: Changed

Juliakas:
I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots

veekeys:
> I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots

@Juliakas, have you tried using delay? It was enough in Header tests.

Juliakas:
> 
> 
> > I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots
> 
> @Juliakas, have you tried using delay? It was enough in Header tests.

I initially thought this would still be flaky. Now I copied delay parameter from Header tests, seems to be working.

veekeys:
> > > I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots
> > 
> > 
> > @Juliakas, have you tried using delay? It was enough in Header tests.
> 
> I initially thought this would still be flaky. Now I copied delay parameter from Header tests, seems to be working. I still adjusted Radio button group, the expected image had radio button focused.

It is not something you can trust much, but as for now it seems to work on all the cases...
Radio buttons story is supposed to have focused one. Story shows that you can setFocus on init :D And visual tests test the focus style, so.. How did you make it to be without focus?
https://itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group

Juliakas:
> 
> 
> > > > I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots
> > > 
> > > 
> > > @Juliakas, have you tried using delay? It was enough in Header tests.
> > 
> > 
> > I initially thought this would still be flaky. Now I copied delay parameter from Header tests, seems to be working. I still adjusted Radio button group, the expected image had radio button focused.
> 
> It is not something you can trust much, but as for now it seems to work on all the cases...
> Radio buttons story is supposed to have focused one. Story shows that you can setFocus on init :D And visual tests test the focus style, so.. How did you make it to be without focus?
> https://itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group

In my latest commit I reverted expected image. When running tests it doesn't succeed on the first attempt, I guess this made me think its supposed to be unfocused.

src/core/Table/Table.tsx
Juliakas: Removed it

veekeys:
> > > > > I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots
> > > > 
> > > > 
> > > > @Juliakas, have you tried using delay? It was enough in Header tests.
> > > 
> > > 
> > > I initially thought this would still be flaky. Now I copied delay parameter from Header tests, seems to be working. I still adjusted Radio button group, the expected image had radio button focused.
> > 
> > 
> > It is not something you can trust much, but as for now it seems to work on all the cases...
> > Radio buttons story is supposed to have focused one. Story shows that you can setFocus on init :D And visual tests test the focus style, so.. How did you make it to be without focus?
> > [itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group](https://itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group)
> 
> In my latest commit I reverted expected image. When running tests it doesn't succeed on the first attempt, I guess this made me think its supposed to be unfocused.

Ok, now it is weird.. In your PR somehow tests decided to fail, even though on previous PRs they were not failing...
Interesting...

mayank99:
I re-ran the workflow and now it's passing 😆

Juliakas:
> 
> 
> > > > > > I have tweaked Creevey tests a bit (at least until the flakiness can be fixed). Most of the tests will ignore images now since it looks like they are not being loaded fast enough for snapshots
> > > > > 
> > > > > 
> > > > > @Juliakas, have you tried using delay? It was enough in Header tests.
> > > > 
> > > > 
> > > > I initially thought this would still be flaky. Now I copied delay parameter from Header tests, seems to be working. I still adjusted Radio button group, the expected image had radio button focused.
> > > 
> > > 
> > > It is not something you can trust much, but as for now it seems to work on all the cases...
> > > Radio buttons story is supposed to have focused one. Story shows that you can setFocus on init :D And visual tests test the focus style, so.. How did you make it to be without focus?
> > > [itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group](https://itwin.github.io/iTwinUI-react/?path=/story/input-inputgroup--radio-group)
> > 
> > 
> > In my latest commit I reverted expected image. When running tests it doesn't succeed on the first attempt, I guess this made me think its supposed to be unfocused.
> 
> Ok, now it is weird.. In your PR somehow tests decided to fail, even though on previous PRs they were not failing...
> Interesting...

It is very strange... Because those components that are failing all of the sudden were never touched. Except now when I added 1 second delay for tests only, but they were failing anyway. This time it did succeed though...

mayank99:
> This time it did succeed though...

It didn't at first. So I re-ran the workflow and now it's passing 😆 

I don't think the delay is doing anything.

Juliakas:
> 
> 
> > This time it did succeed though...
> 
> It didn't at first. So I re-ran the workflow and now it's passing 😆
> 
> I don't think the delay is doing anything.

Yeah, I meant it like that. Delay for image component seems to work. For Radio group delay probably not..

src/core/Table/Table.tsx
```diff
@@ -88,6 +92,22 @@ export type TableProps<
    * @default 300
    */
   intersectionMargin?: number;
+  /**
+   * A function that will be used for rendering a component for each row if that row is expanded.
+   * Component will be placed right after the row, in the same row group div.
```
mayank99, line 97: Update this comment. The row group part is not true anymore.

Also add the part about returning `false` or `null` if you add that functionality.

src/core/Table/Table.tsx
```diff
@@ -190,6 +213,36 @@ export const Table = <
...
+    hooks.allColumns.push((columns: ColumnInstance<T>[]) => [
+      {
+        id: 'iui-table-expander',
+        disableResizing: true,
+        disableGroupBy: true,
+        minWidth: 48,
+        width: 48,
+        maxWidth: 48,
+        columnClassName: 'iui-tables-slot',
+        cellClassName: 'iui-tables-slot',
```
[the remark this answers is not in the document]
Juliakas: Changing to iui-slot, removing columnClassName field

src/core/Table/Table.tsx
```diff
@@ -190,6 +213,36 @@ export const Table = <
...
+        Cell: ({ row }: CellProps<T>) => (
+          <IconButton
+            styleType='borderless'
+            size='small'
+            onClick={() => {
+              row.toggleRowExpanded();
```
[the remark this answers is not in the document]
Juliakas: Will look like this: 
image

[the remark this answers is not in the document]
mayank99: i think checkbox column should go before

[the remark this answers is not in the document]
Juliakas: My bad, I originally intended it to be before, I think ordering of hooks should fix it

src/core/Table/Table.tsx
```diff
@@ -88,6 +92,22 @@ export type TableProps<
...
+  /**
+   * Handler for row expand events. Will trigger when expanding and condensing rows.
+   */
+  onExpand?: (
+    expandedData: T[] | undefined,
+    tableState?: TableState<T>,
+  ) => void;
+  /**
+   * Flag whether to provide default expander column
+   */
```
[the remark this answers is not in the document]
Juliakas: Decided to get rid of this prop and use `expanderCell` instead for overriding default

.creevey/config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   browsers: {
     chrome: {
       browserName: 'chrome',
+      version: '88.0',
```
mayank99, line 20: Will need to do something about this in a different PR

mayank99 (approved):
Looks good!

PR #51, opened 2021-05-10
Release 1.3.0

mayank99:


CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Changelog
 
-## 1.X.X
+## [1.3.0]
 
-`2021-05-XX`
+`2021-05-10`
 
 ### What's new
```
bentleyvk, line 8: Maybe worth mentioning that random build/test failure should be fixed because of the icons package update?

CHANGELOG.md
```diff
@@ -1,8 +1,8 @@
 # Changelog
 
-## 1.X.X
+## [1.3.0]
 
-`2021-05-XX`
+`2021-05-10`
```
veekeys, line 5: need update day too now :D

bentleyvk (approved):

package.json
```diff
@@ -105,7 +105,8 @@
   },
   "resolutions": {
     "braces": "^2.3.1",
-    "hosted-git-info": ">=3.0.8"
+    "hosted-git-info": ">=3.0.8",
+    "trim": "^0.0.3"
```
bentleyvk, line 109: Do we need resolution for this? Can you try removing this from resolution? I think it should work as `yarn.lock` is changed.

mayank99: If I remove resolution and do `yarn install`, it goes back to 0.0.1 because there is no `^`

bentleyvk: Yeah, why libs do this...

Issue #57, opened 2021-05-13
Toasts: `toaster.closeAll()` does not work just before adding a new toast.


Juliakas:
#### Environment
- __Package version(s)__: 1.4.0
- __Operating System__: Microsoft Windows 10 Enterprise
- __Browser name and version__: Mozilla Firefox 88.0.1 (64-bit)

#### Code Sandbox
Link to a minimal repro (fork [this code sandbox](https://codesandbox.io/s/itwinui-react-example-son74)): https://codesandbox.io/s/itwinui-react-example-forked-hcymp?file=/src/App.tsx

#### Steps to reproduce
1. Inspect code, notice `toaster.closeAll()` in "Add toast" click handler.
1. Click button "Add toast" several times.

#### Actual behavior
Multiple toasts appear.

#### Expected behavior
Last shown toast should be closed before generating next toast.

#### Additional information
For some reason `toaster.closeAll()` function does not close toasts if it is used just before adding a new toast.
On the other hand closing all toasts on a separate occasion (by clicking "Close all" button) does work as expected.

veekeys:
If the bug you are describing is that Toasts are stacking, that is intended.
If you want to achieve that only 1 toast is visible, you need to call closeAll public method yourself, before adding new one.

Juliakas:
> 
> 
> If the bug you are describing is that Toasts are stacking, that is intended.
> If you want to achieve that only 1 toast is visible, you need to call closeAll public method yourself, before adding new one.
> 
> EDIT: ok, now I understood. Need to read the repro code before...

Yes - so I edited repro steps now to be more clear.

veekeys:
> > If the bug you are describing is that Toasts are stacking, that is intended.
> > If you want to achieve that only 1 toast is visible, you need to call closeAll public method yourself, before adding new one.
> > EDIT: ok, now I understood. Need to read the repro code before...
> 
> Yes - so I edited repro steps now to be more clear.

Yeah thanks. I am working on toasts right now and tested same flow, it works correct. So this should be fixed together with this:
#54

PR #58, opened 2021-05-13
Toaster not using document in constructor

veekeys:

## Checklist

- [ ] Add screenshots of the key elements of the component below
- [ ] Add any additional comments below describing the features you've implemented

## Additional Comments

Refactored toaster a little.  Main reason is to support SSR. It was not working, because we used document in the constructor (not available in server side). So I have moved that to the runtime. 
Ended up having pretty empty ToastMaster (maybe we should rename it) and toaster handling all the logic.
It is impportant to test the behavior and animation (had some issues with it) that those are same as they were.

### DIFF
--- a/src/core/Toast/ToastMaster.tsx
+++ b/src/core/Toast/ToastMaster.tsx
@@ -6,70 +6,17 @@ import '@itwin/itwinui-css/css/toast-notification.css';
 import React from 'react';
 import Toast, { ToastProps } from './Toast';
 
-export type ToastSettings = Omit<ToastProps, 'isVisible' | 'id'>;
-
 type ToastMasterProps = {
-  addToastHandler: (addToast: (settings: ToastSettings) => void) => void;
-  closeAllHandler: (closeAll: () => void) => void;
+  toasts: ToastProps[];
 };
 
 export const ToastMaster = (props: ToastMasterProps) => {
-  const { addToastHandler, closeAllHandler } = props;
-
-  const idCounter = React.useRef<number>(0);
-  const [toasts, setToasts] = React.useState<ToastProps[]>([]);
-
-  const addToast = (settings: ToastSettings) => {
-    const instance = createToastInstance(settings);
-
-    setToasts([instance, ...toasts]);
-    return instance;
-  };
-
-  const createToastInstance = (settings: ToastSettings) => {
-    const uid = ++idCounter.current;
-
-    return {
-      ...settings,
-      id: uid,
-      isVisible: true,
-    };
-  };
-
-  const removeToast = (id: number) => {
-    setToasts(toasts.filter((toast) => toast.id !== id));
-  };
-
-  const closeAll = () => {
-    setToasts(
-      toasts.map((toast) => {
-        return {
-          ...toast,
-          isVisible: false,
-        };
-      }),
-    );
-  };
-
-  React.useEffect(() => {
-    addToastHandler(addToast);
-    closeAllHandler(closeAll);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [toasts]);
+  const { toasts } = props;
 
   return (
     <span className='iui-toast-wrapper'>
       {toasts.map((toastProps) => {
-        return (
-          <Toast
-            key={toastProps.id}
-            onRemove={() => {
-              removeToast(toastProps.id);
-              toastProps.onRemove && toastProps.onRemove();
-            }}
-            {...toastProps}
-          />
-        );
+        return <Toast key={toastProps.id} {...toastProps} />;
       })}
     </span>
   );
--- a/src/core/Toast/Toaster.tsx
+++ b/src/core/Toast/Toaster.tsx
@@ -6,7 +6,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { getContainer } from '../utils/common';
 import { ToastCategory, ToastProps } from './Toast';
-import { ToastMaster } from './ToastMaster';
+import { ToastWrapper } from './ToastWrapper';
 
 const TOASTS_CONTAINER_ID = 'iui-toasts-container';
 
@@ -65,7 +65,7 @@ export default class Toaster {
 
   private updateView() {
     ReactDOM.render(
-      <ToastMaster toasts={this.toasts} />,
+      <ToastWrapper toasts={this.toasts} />,
       getContainer(TOASTS_CONTAINER_ID),
     );
   }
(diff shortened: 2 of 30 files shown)


src/core/Toast/Toaster.tsx
```diff
@@ -51,22 +38,45 @@ export default class Toaster {
     category: ToastCategory,
     settings?: ToastOptions,
   ) {
-    this.addToast({
-      ...settings,
-      content,
-      category,
-    });
+    const lastId = this.toasts[0]?.id ?? 0;
```
bentleyvk, line 41: New toast won't always have new id. Idk if it is bad but maybe to avoid random errors e.g. with keys, we should keep static count as it was done previously.

src/core/Toast/Toaster.tsx
```diff
@@ -51,22 +38,45 @@ export default class Toaster {
...
+    const currentId = lastId + 1;
+    this.toasts = [
+      {
+        ...settings,
+        content,
+        category,
+        onRemove: () => {
+          this.removeToast(currentId);
+          settings?.onRemove;
```
bentleyvk, line 50: I think you should call it 😀

src/core/utils/common.tsx
```diff
@@ -43,3 +43,16 @@ export const getUserColor = (emailOrName: string) => {
   }
   return USER_COLORS[hash];
 };
+
+/**
+ * Get the portal container, or create one if it doesn't exist.
+ */
+export const getContainer = (containerId: string) => {
+  let container = document.getElementById(containerId);
+  if (container === null) {
```
bentleyvk, line 52: Better use `==` as it will handle `undefined` and `null`.

src/core/Toast/ToastMaster.tsx
```diff
@@ -6,70 +6,17 @@ import '@itwin/itwinui-css/css/toast-notification.css';
...
 
-export type ToastSettings = Omit<ToastProps, 'isVisible' | 'id'>;
-
 type ToastMasterProps = {
-  addToastHandler: (addToast: (settings: ToastSettings) => void) => void;
-  closeAllHandler: (closeAll: () => void) => void;
+  toasts: ToastProps[];
 };
 
 export const ToastMaster = (props: ToastMasterProps) => {
```
mayank99, line 13: Since this component is very basic now, I would rename it to just `ToastWrapper`, aligning it with the class name.

src/core/utils/common.tsx
```diff
@@ -43,3 +43,16 @@ export const getUserColor = (emailOrName: string) => {
   }
   return USER_COLORS[hash];
 };
+
+/**
+ * Get the portal container, or create one if it doesn't exist.
+ */
+export const getContainer = (containerId: string) => {
```
mayank99, line 50: So this is not just for portal anymore. Maybe need to rename and update JSDoc comment

src/core/Toast/Toaster.tsx
```diff
@@ -51,22 +38,45 @@ export default class Toaster {
...
+    this.toasts = this.toasts.filter((toast) => toast.id !== id);
+    this.updateView();
   }
 
-  private _addHandler(addHandler: (settings: ToastSettings) => void) {
-    this.addToast = addHandler;
+  private updateView() {
+    ReactDOM.render(
+      <ToastMaster toasts={this.toasts} />,
+      getContainer(TOASTS_CONTAINER_ID),
```
mayank99, line 69: So instead of constructor, `document` is being used `updateView`. Have you checked that this actually solves the SSR problem?

src/core/Toast/Toaster.tsx
```diff
@@ -51,22 +38,45 @@ export default class Toaster {
...
-    this.closeAllHandler();
+  private removeToast(id: number) {
+    this.toasts = this.toasts.filter((toast) => toast.id !== id);
+    this.updateView();
   }
 
-  private _addHandler(addHandler: (settings: ToastSettings) => void) {
-    this.addToast = addHandler;
+  private updateView() {
+    ReactDOM.render(
```
mayank99, line 67: so this will create a whole new renderer separate from the main app? Sure this is the best way? 🤔 

I'm surprised this is how we were doing it before too 😮

mayank99: One key difference is that render() will be called multiple times now that it's not in the constructor. ~~Might be bad for performance~~ Should be fine 😄 https://twitter.com/dan_abramov/status/691301224541503488

src/core/Toast/Toaster.tsx
```diff
@@ -4,32 +4,19 @@
...
-    document.body.appendChild(container);
-    ReactDOM.render(
-      <ToastMaster
-        addToastHandler={this._addHandler}
-        closeAllHandler={this._closeAllHandler}
-      />,
-      container,
-    );
-  }
+  private toasts: ToastProps[] = [];
```
mayank99, line 19: This doesn't need state?

src/core/Toast/Toaster.tsx
mayank99: > New toast won't always have new id

Isn't that what the next line does? Or are you saying the user could provide any random id?

mayank99: Also, wasn't static count the cause of #57? Or was it something else?

src/core/Toast/Toaster.tsx
veekeys: I havent. I guess we will release a dev version and test in real project. The idea is to run it when it is needed, not before..

src/core/Toast/Toaster.tsx
veekeys: What do you mean? `useState` ?

src/core/Toast/Toaster.tsx
veekeys: yea, agree

src/core/Toast/Toaster.tsx
veekeys: Not sure what to do with this one...

src/core/utils/common.tsx
veekeys: fixed

src/core/Toast/ToastMaster.tsx
veekeys: renamed

src/core/utils/common.tsx
veekeys: updated

src/core/Toast/Toaster.tsx
mayank99: Oh it's not a react component, nevermind

bentleyvk (approved):

src/core/Toast/Toaster.tsx
veekeys: Added class variable

bentleyvk (approved):

PR #61, opened 2021-05-13
Improve overview story logic for hiding addons panel

mayank99:
Changed approach for hiding addons panel after noticing that `'button[title="Hide addons [A]"]'` selector was returning null, and `isPanelOpen` was also broken because `style` was not being set at the time of checking.

The new way involves opening the shortcuts menu (because the options inside are not part of the default DOM structure anymore) and checking if this svg is present (and clicking the button to hide if present):
image

Also was able to disable Storysource addon, so opening panel will now show this:
image

Also fixed logo theming.

### DIFF
--- a/stories/intro/Overview.stories.tsx
+++ b/stories/intro/Overview.stories.tsx
@@ -21,6 +21,7 @@ export default {
     options: { showPanel: false },
     controls: { disable: true },
     actions: { disable: true },
+    storysource: { disable: true },
   },
 } as Meta;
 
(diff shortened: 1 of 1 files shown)


stories/intro/Overview.stories.tsx
```diff
@@ -30,18 +31,23 @@ export const Overview: Story = () => {
...
-      'button[title="Hide addons [A]"]',
+    const shortcutsButton = parent.document.querySelector(
+      'button[title="Shortcuts"]',
     ) as HTMLButtonElement;
+    shortcutsButton?.click();
 
-    const isPanelOpen = !!parent.document.querySelector(
-      '#storybook-preview-wrapper',
-    )?.parentElement?.parentElement?.style.height;
+    setTimeout(() => {
```
veekeys, line 39: Sure it always works with timeout? Usually timeout solutions are not stable...

veekeys (approved):

stories/intro/Overview.stories.tsx
mayank99: I tested in Chrome and Firefox in incognito mode and in both light and dark themes.

Will deploy storybook after merging this so others can test.

PR #63, opened 2021-05-14
Add date filter

bentleyvk:

## Checklist

## Screenshots
image
image

## Additional Comments
I was not sure how to handle user input. Currently I only parse it when it is in correct format (I use regex for that) so user won't see strange behaviors when entering half of the date and it is being parsed e.g. entering "May 1" and it becomes "May 1, 2000".
I couldn't wrap icon button with `Popover` as it has `position: absolute`.
I fixed the issue when clicking inside filter triggers sorting.


src/core/Table/filters/BaseFilter.tsx
```diff
@@ -35,7 +35,11 @@ export const BaseFilter = (props: BaseFilterProps) => {
   useTheme();
 
   return (
-    <div className={cx('iui-column-filter', className)} style={style}>
+    <div
+      className={cx('iui-column-filter', className)}
+      style={style}
+      onClick={(event) => event.stopPropagation()}
```
bentleyvk, line 41: This fixes https://github.com/iTwin/iTwinUI-react/issues/69

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -49,12 +49,16 @@ export const FilterToggle = <T extends Record<string, unknown>>(
...
           onClickOutside={close}
+          appendTo={document.body}
         >
           <div
             className={cx('iui-filter', {
               'iui-active': isVisible || column.filterValue,
             })}
-            onClick={() => setIsVisible((v) => !v)}
+            onClick={(e) => {
+              e.stopPropagation();
```
bentleyvk, line 59: Fix for https://github.com/iTwin/iTwinUI-react/issues/69

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -49,12 +49,16 @@ export const FilterToggle = <T extends Record<string, unknown>>(
           placement='bottom'
           visible={isVisible}
           onClickOutside={close}
+          appendTo={document.body}
```
bentleyvk, line 52: Fix for https://github.com/iTwin/iTwinUI-react/issues/69. Otherwise filter would be as a child of a header column which handles clicks and pointer cursor for sorting.

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+      const parsedDate = parseInput(value);
+      if (parsedDate && !isNaN(parsedDate.valueOf())) {
+        onChange(parsedDate);
+      }
+    },
+    [onChange, parseInput],
+  );
+
+  return (
+    <Popover
```
bentleyvk, line 56: I can't place it next to the icon button as it is positioned absolutely and datepicker goes under other elements.

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+      content={<DatePicker date={date} onChange={onDateSelected} setFocus />}
+      placement='bottom'
+      visible={isVisible}
+      onClickOutside={(_, e) => {
```
bentleyvk, line 60: @mayank99 maybe you know how else I could handle that clicking on icon button won't close and open instantly the datepicker?

mayank99: Why does it do that? 🤔 

The DropdownMenu example does not need that. I wonder if you need to put the Popover inside the `svgIcon` prop of LabeledInput

mayank99: Oh I see the other comment now 😄 

So then idk. Would have to debug to see the order of events, etc. One thing you could try looking at is `e.relatedTarget`. But it won't be much different from the current approach

src/core/Table/filters/FilterToggle.tsx
mayank99: Is that necessary?

I would think the `stopPropagation` should be enough. Do we need both?

bentleyvk: Without this line:
filter-sort

basanta078: Basefilter takes care of stopping the event from going to the header
The only issue that remains is the pointer icon when hovering on the side. 
I ignored it on my PR because it is just a UI Issue and not affecting the sorting 
Correct fix would be to fix the css for the header

mayank99: @basanta078  It's not just the pointer. Notice the table header getting highlighted. That's because the filter dropdown is located inside the header div.

CSS will not help here, so it needs to be in your PR or in this one.

CHANGELOG.md
```diff
@@ -266,6 +267,7 @@
 - **New ThemeProvider component added.** It allows to switch between light and dark themes.
 - **iTwinUI-React is generated to `CommonJS` and `ES` modules.** Usage of `ES` modules allows bundlers to tree-shake unused code resulting in smaller bundle sizes.
 
+[1.5.0]: https://github.com/iTwin/iTwinUI-react/compare/v1.5.0...v1.4.1
```
mayank99, line 270: This is reversed 😄

src/core/Table/Table.test.tsx
```diff
@@ -563,3 +562,42 @@ it('should show message when there is no data after filtering', () => {
   expect(rows.length).toBe(0);
   screen.getByText('No results. Clear filter.');
 });
+
+it('should not trigger sorting when filter is clicked', () => {
```
mayank99, line 566: If #70 gets merged, will this test be necessary?

src/core/Table/Table.tsx
```diff
@@ -163,6 +164,7 @@ export const Table = <
     intersectionMargin = 300,
     onFilter,
     emptyFilteredTableContent,
+    filterTypes: userFilterFunctions,
```
mayank99, line 167: First I read this as `useFilterFunctions` which made me think it's a hook. 😄 

Maybe rename to something else?

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+    >
+      <LabeledInput
+        displayStyle='inline'
+        value={inputValue}
+        onChange={onInputChange}
+        onClick={close}
+        svgIcon={
+          <IconButton
+            styleType='borderless'
+            onClick={() => setIsVisible((v) => !v)}
```
mayank99, line 74: This would need `useCallback`, right?

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -0,0 +1,109 @@
...
+  to: string;
+};
+
+const defaultStrings: DateRangeTranslation = {
+  from: 'From',
+  to: 'To',
+};
+
+const defaultFormatDate = (date: Date) => {
+  const formatter = new Intl.DateTimeFormat('en-us', {
```
mayank99, line 27: I think `en-us` is not necessary and it can just be `undefined` to get the browser default

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -0,0 +1,109 @@
...
+    month: 'short',
+    day: 'numeric',
+    year: 'numeric',
+  });
+  return formatter.format(date);
+};
+
+const defaultParseInput = (text: string) => {
+  if (/^[a-z]{3}\s\d{1,2},\s\d{4}$/gi.test(text)) {
```
mayank99, line 36: can you assign this regex to a const with a comment next to it to make it clear what it does?

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -0,0 +1,109 @@
...
+  props: DateRangeFilterProps<T>,
+) => {
+  const {
+    column,
+    translatedLabels,
+    setFilter,
+    clearFilter,
+    formatDate = defaultFormatDate,
+    parseInput = defaultParseInput,
+    placeholder = 'MMM DD, YYYY',
```
mayank99, line 61: I'm pretty sure this should say "Mon" not "MMM". Need to research standard date format placeholders

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
mayank99: Well, maybe it's no big deal if it re-renders

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
mayank99: Oh and it would need to be localized I guess

bentleyvk: Changed. So user can provide his own placeholder. I don't think we can handle every language.

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: Added the comment.

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: If it takes user browser language it may break stuff - parser and placeholder may be wrong. As we use English in all places, I left for the developer to handle this.

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
bentleyvk: I thought the same 😀

src/core/Table/Table.tsx
bentleyvk: Renamed.

CHANGELOG.md
bentleyvk: Upsy 😀 Fixed.

src/core/Table/filters/FilterToggle.tsx
bentleyvk: Yeah, CSS won't help because without this line filter is rendered as a header child.

src/core/Table/Table.test.tsx
basanta078: there is no test for this file, though it would be nice to have to cover more usecase

src/core/Table/filters/FilterToggle.tsx
basanta078: sounds good, I added that fix to the other PR

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -0,0 +1,110 @@
...
+    day: 'numeric',
+    year: 'numeric',
+  });
+  return formatter.format(date);
+};
+
+const defaultParseInput = (text: string) => {
+  // Checks whether date is in correct format e.g. Feb 11, 2021
+  if (/^[a-z]{3}\s\d{1,2},\s\d{4}$/gi.test(text)) {
+    return new Date(text);
```
mayank99, line 38: I was trying to research how best to parse dates and I came across this (notice emphasis on "strongly" 😄)

image

I guess it's fine here but we should make it extra clear that we expect users to provide their own parser

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: Everywhere in our UX standards this short format is used, also Jon said it would be better to use it.

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: Other option would be to use some date parsing lib like `date-fns` but native parse with our date format worked perfectly to my surprise.

src/core/Table/Table.test.tsx
bentleyvk: Merged.

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
mayank99: Yeah no, I think it's fine how we are doing it but we should also explicitly state that we expect users to handle this if they want reliable results.

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
mayank99: ok so short format is fine but re: the original placeholder comment, I did some research and came across this Java convention: https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html

According to that, y and d should be lowercase, and also:
>  If the number of pattern letters is 3 or more, the month is interpreted as text; otherwise, it is interpreted as a number.

So I'd say go with: **MMM dd, yyyy**

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: Added this comment to `tableFilters`:
image

src/core/Table/filters/DateRangeFilter/DateRangeFilter.tsx
bentleyvk: Changed.

src/core/Table/filters/tableFilters.tsx
```diff
@@ -44,6 +44,8 @@ export const tableFilters = {
   ) => <TextFilter {...props} translatedLabels={translatedLabels} />,
   /**
    * Date range filter.
+   * By default it handles `en-us` date format if you want to use other format
```
mayank99, line 47: Punctuation
```suggestion
   * By default it handles `en-us` date format. If you want to use other format
```

mayank99 (dismissed):

src/core/Table/filters/customFilterFunctions.ts
```diff
@@ -0,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { IdType, Row } from 'react-table';
+
+const isValidDate = (date: Date | undefined) =>
+  !!date && !isNaN(date.valueOf());
+
+const betweenDate = <T extends Record<string, unknown>>(
```
veekeys, line 10: Name sounds weird... Does it return values between min and max?

bentleyvk: Default one from react-table is `between` which is for numbers, so I went with the same naming. And yes it return values between min and max dates.

veekeys (approved):

PR #66, opened 2021-05-17
Update to itwinui-css 0.15.0

mayank99:
Bumped css package version.

Added `required` prop and `.iui-required` classname to:
- InputGroup
- LabeledSelect
- LabeledTextarea
- LabeledInput

Updated stories and tests for these 4 components.

### DIFF
--- a/src/core/InputGroup/InputGroup.tsx
+++ b/src/core/InputGroup/InputGroup.tsx
@@ -33,6 +33,11 @@ export type InputGroupProps = {
    */
   disabled?: boolean;
   /**
+   * Whether the whole input group is required.
+   * @default false
+   */
+  required?: boolean;
+  /**
    * Custom icon. If group has status, default status icon is used instead.
    */
   svgIcon?: JSX.Element;
@@ -68,6 +73,7 @@ export const InputGroup = (props: InputGroupProps) => {
     svgIcon,
     className,
     style,
+    required = false,
     ...rest
   } = props;
   useTheme();
@@ -87,7 +93,15 @@ export const InputGroup = (props: InputGroupProps) => {
       style={style}
       {...rest}
     >
-      {label && <div className='iui-label'>{label}</div>}
+      {label && (
+        <div
+          className={cx('iui-label', {
+            'iui-required': required,
+          })}
+        >
+          {label}
+        </div>
+      )}
       {children}
       {(message || icon) && (
         <div className='iui-message'>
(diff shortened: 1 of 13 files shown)


src/core/InputGroup/InputGroup.tsx
```diff
@@ -32,6 +32,11 @@ export type InputGroupProps = {
    * @default false
    */
   disabled?: boolean;
+  /**
+   * Whether the whole input group is required.
+   * @default false
+   */
+  required?: boolean;
```
mayank99, line 39: This prop does nothing (other than the visual change) so the user still needs to set `required` manually to the actual inputs inside the group.

But it's the same as how `disabled` works so I didn't add any details in the JSDoc comment.

src/core/LabeledSelect/LabeledSelect.tsx
```diff
@@ -42,6 +42,11 @@ export type LabeledSelectProps<T> = {
    * @default 'default'
    */
   displayStyle?: 'default' | 'inline';
+  /**
+   * If true, shows a red asterisk but does not prevent form submission.
+   * @default false
+   */
+  required?: boolean;
```
mayank99, line 49: Since this is different from how `LabeledInput` and `LabeledTextarea` work, I explicitly mentioned that it does not prevent form submission.

mayank99: Created a discussion #67 so we can address the root cause in the future

src/core/LabeledTextarea/LabeledTextarea.tsx
```diff
@@ -98,7 +99,15 @@ export const LabeledTextarea = React.forwardRef<
...
+          className={cx('iui-label', {
+            'iui-required': required,
+          })}
+        >
+          {label}
+        </div>
+      )}
       <Textarea
         disabled={disabled}
         className={textareaClassName}
```
veekeys, line 113: Dont you need to add required on textarea???

mayank99: Done

src/core/InputGroup/InputGroup.tsx
veekeys, line 39: Tests?
everywhere...

mayank99: Added

veekeys (approved):

PR #68, opened 2021-05-17
Add SideNavigation component

mayank99:
Added `SideNavigation` and `SidenavButton` based on side-navigation.css from itwinui-css 0.15.0

### Checklist

### Screenshots

sidebar (1)


src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,85 @@
...
+      {...rest}
+    >
+      {expanderVisibility === 'top' && ExpandButton}
+      <div className='iui-sidenav-content'>
+        <div className='iui-top'>
+          {mainItems.map((sidebarButton) =>
+            React.cloneElement(sidebarButton as JSX.Element, {
+              showTooltip: !isExpanded,
+            }),
+          )}
```
mayank99, line 76: This is the best I could come up with after some brainstorming.

It still needs some work to validate that it's a `SidebarButton` element.

src/core/Sidebar/SidebarButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+      disabled={disabled}
+      {...rest}
+    >
+      {children}
+    </Button>
+  );
+
+  return showTooltip ? (
+    <Tooltip content={children} placement='right'>
+      {button}
```
mayank99, line 58: I wanted to do this:
```suggestion
      {!disabled ? button : <span>{button}</span>}
```

But of course the css is not designed for that. 😕

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,85 @@
...
+    mainItems,
+    secondaryItems,
+    expanderVisibility = 'top',
+    className,
+    ...rest
+  } = props;
+
+  useTheme();
+
+  const [isExpanded, setIsExpanded] = React.useState(false);
```
mayank99, line 46: Should we provide this as a controlled prop too?

src/core/Sidebar/SidebarButton.tsx
mayank99: This sorta works:

```jsx
{!disabled ? button : <span style={{ display: 'contents' }}>{button}</span>}
```

but it messses up tooltip positioning

src/core/Sidebar/Sidebar.tsx
veekeys: Yeah, I think we better do.. Maybe there is a reason they want to show it expanded straight away...

src/core/Sidebar/SidebarButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Tooltip } from '../Tooltip';
+
+export type SidebarButtonProps = {
+  /**
+   * Whether the sidebar button is active.
+   */
+  isActive?: boolean;
+  /**
+   * If true, button label will also be shown as a tooltip.
+   * By default, tooltip is shown when sidebar is collapsed, and hidden when expanded.
```
veekeys, line 21: This comment is weird...Default value is true or false? Or is it some 3rd option.

src/core/Sidebar/SidebarButton.tsx
veekeys: What is wrong with current one..?

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useTheme } from '../utils/hooks/useTheme';
+import '@itwin/itwinui-css/css/side-navigation.css';
+import { CommonProps } from '../utils/props';
+import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
+import { IconButton } from '../Buttons';
+
+export type SidebarProps = {
+  /**
+   * Buttons shown in the top portion sidebar.
+   * Recommended to use `SidenavButton`s.
```
veekeys, line 16: Apostrophe looks weird.. Maybe just 'SidenavButton' components.

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,85 @@
...
+   */
+  mainItems: React.ReactNode[];
```
veekeys, line 18: Need to think about naming...

src/core/Sidebar/SidebarButton.tsx
mayank99: I guess technically the default is undefined. But practically, it is set to true in `Sidebar` when collapsed and false when expanded

src/core/Sidebar/SidebarButton.tsx
mayank99: It won't show tooltip on disabled button

src/core/Sidebar/Sidebar.tsx
mayank99: What apostrophe? 😄 
image

veekeys: well I mean.. it is written in there, but anyway.. point being, looks bad :D

mayank99: I pushed a change earlier

src/core/Sidebar/Sidebar.tsx
bentleyvk: What about wrapping with a `Tooltip` here and switching between just a `button` and then wrapped with `Tooltip`?

src/core/Sidebar/SidebarButton.tsx
bentleyvk: What about using `title` prop instead of this flag?

src/core/Sidebar/SidebarButton.tsx
veekeys: Like it should... :D

src/core/Sidebar/Sidebar.tsx
mayank99: I tried this but the tooltip never shows up. And it's not a problem with `sidebarButton.props.children` because I'm able to get the tooltip to show if I wrap just a static span instead of `sidebarButton`.

```jsx
{mainItems.map((sidebarButton: JSX.Element) =>
  !isExpanded ? (
    <Tooltip content={sidebarButton.props.children} placement='right'>
      {sidebarButton}
      // also tried these two:
      // React.cloneElement(sidebarButton)
      // (() => sidebarButton())()
    </Tooltip>
  ) : (
    sidebarButton
  ),
)}
```

src/core/Sidebar/SidebarButton.tsx
mayank99: Removed `showTooltip` and added `title`

mayank99: Removed this prop and moved tooltip logic to `Sidebar`

src/core/Sidebar/Sidebar.tsx
mayank99: Update: it works now. Just had to forward the ref in `SidebarButton` 😆 

Reason: https://github.com/atomiks/tippyjs-react#component-children

src/core/Sidebar/Sidebar.tsx
mayank99: Added `isExpanded` and `onExpanderClick` props

mayank99: `onExpanderClick` may be unnecessary idk

src/core/Sidebar/Sidebar.tsx
bentleyvk: You had the same issue that Basanta told us about 😀 Maybe add this link to out `Popover` or even to all components that are using `Popover`.

src/core/Sidebar/Sidebar.tsx
bentleyvk: Any ideas? 😀 Cause I couldn't think other name, maybe just `items` but when I look ar other component libs they are using very simple naming for these kind of stuff like `start/top/leftSomething`.

src/core/Sidebar/Sidebar.tsx
bentleyvk: In other components we provide these callbacks, so maybe it is ok to have it here too.

src/core/Sidebar/SidebarButton.tsx
bentleyvk: > This sorta works:
> 
> ```js
> {!disabled ? button : <span style={{ display: 'contents' }}>{button}</span>}
> ```
> 
> but it messses up tooltip positioning

Does this now works?

stories/core/Sidebar.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+import React from 'react';
+import { Sidebar, SidebarButton, SidebarProps } from '../../src/core';
+
+export default {
+  component: Sidebar,
+  subcomponents: { SidebarButton },
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
```
bentleyvk, line 16: Disable main and secondary items.

stories/core/Sidebar.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  title: 'Core/Sidebar',
+} as Meta<SidebarProps>;
+
+export const Basic: Story<SidebarProps> = (args) => {
+  return (
+    <Sidebar
+      {...args}
+      mainItems={[
+        <SidebarButton startIcon={<SvgPlaceholder />} key={0}>
```
bentleyvk, line 25: Maybe handle click to make it active.

src/core/Sidebar/SidebarButton.tsx
bentleyvk: I still think we should use `title` for tooltip content as it allows user to customize tooltip.

veekeys: What do you mean to customize? There we should show tooltip, not title by design.

src/core/Sidebar/Sidebar.tsx
veekeys: Id think about topItems and bottomItems to be honest.. Directly say, where they are.. Cause not sure if those are main and secondary in workflorw...

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Buttons shown at the bottom of sidebar.
+   */
+  secondaryItems?: React.ReactNode[];
+  /**
+   * Control the visibility of "expander" icon button.
+   * @default 'top'
+   */
+  expanderVisibility?: 'top' | 'bottom' | 'hidden';
```
veekeys, line 28: Visibility or Position...

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * Sidebar component.
+ * @example
+ * <Sidebar
+ *  mainItems={[
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 1</SidebarButton>,
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 2</SidebarButton>,
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 3</SidebarButton>,
+ *  ]}
+ * secondaryItems={[
```
veekeys, line 48: alignment

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * ]}
+ * />
+ */
+export const Sidebar = (props: SidebarProps) => {
+  const {
+    mainItems,
+    secondaryItems,
+    expanderVisibility = 'top',
+    className,
+    isExpanded,
```
veekeys, line 59: or just give default value and no need to check for undefined

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <SvgChevronRight />
+    </IconButton>
+  );
+
+  return (
+    <div
+      className={cx(
+        'iui-side-navigation',
+        {
+          'iui-expanded': _isExpanded,
```
veekeys, line 88: weird implementation to have classes for both states...

src/core/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,133 @@
...
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 1</SidebarButton>,
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 2</SidebarButton>,
+ *    <SidebarButton startIcon={<SvgPlaceholder />}>App 3</SidebarButton>,
+ *  ]}
+ * secondaryItems={[
+ *  <SidebarButton
+ * ]}
+ * />
+ */
+export const Sidebar = (props: SidebarProps) => {
```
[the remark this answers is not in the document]
mayank99: The name? I thought `SideNavigation` (and `SideNavigationButton`) would be too long and "navigation" would not be fully accurate

src/core/Sidebar/Sidebar.tsx
mayank99: "position: hidden" would not make sense

veekeys: True, but visibility: "top" also is weird..

src/core/Sidebar/Sidebar.tsx
mayank99: I like `items` and `secondaryItems`.

"secondary" makes sense because it will only include settings icon or something similar

src/core/Sidebar/SidebarButton.tsx
mayank99: nah it doesn't show up

src/core/Sidebar/Sidebar.tsx
mayank99: The problem is that it's not fully controlled. We change the value of `_isExpanded` (state) even when `isExpanded` (prop) is provided

src/core/Sidebar/Sidebar.tsx
mayank99: it made sense when I thought to myself: it is "visible" on the top 😄

src/core/Sidebar/Sidebar.tsx
mayank99: Initially I wanted to differentiate between controlled prop and internal state so I left it undefined

src/core/Sidebar/Sidebar.tsx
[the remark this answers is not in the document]
mayank99: I think that's fine.

As an example: our docs say "flextable" and so did our css, but react always said Table

src/core/Sidebar/Sidebar.tsx
mayank99: Popover already points to tippyjs though 🤔

src/core/Sidebar/Sidebar.tsx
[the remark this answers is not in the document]
veekeys: flextable was removed already, from what I rememebr..
I dont think here is same example, I just want us to be consistent as much as possible.
@FlyersPh9, SideNavigation or SideBar or both or ... ?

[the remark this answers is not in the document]
FlyersPh9: Where does this sidebar text show up?  If they devs have to place something like `<sidebar>` in their code, I'd prefer if the name matches what's in the docs so devs could easily search for the standards.  
Also yes we did change flextable to table.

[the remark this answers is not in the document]
mayank99: Changed to `SideNavigation` and `SidenavButton`

stories/core/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+    mainItems: { control: { disable: true } },
+    secondaryItems: { control: { disable: true } },
+  },
+  args: { style: { height: 'calc(100vh - 24px) ' } },
+  title: 'Core/SideNavigation',
+} as Meta<SideNavigationProps>;
+
+export const Basic: Story<SideNavigationProps> = (args) => {
```
bentleyvk, line 28: Is it still needed when `ActiveItem` story shows everything?

src/core/Sidebar/Sidebar.tsx
bentleyvk: I see that we are doing the same in other components.

src/core/Sidebar/SidebarButton.tsx
bentleyvk: I mean to use `title` value for `Tooltip` content - not to use `children` but `title`.

src/core/Sidebar/Sidebar.tsx
bentleyvk: Yeah I missed this one but it really looks odd, position would be better here. And maybe it would sound better if `hidden` would be changed with `none`?

stories/core/SideNavigation.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <SideNavigation
+      {...args}
+      mainItems={[
+        <SidenavButton startIcon={<SvgPlaceholder />} key={0}>
```
bentleyvk, line 33: For me placeholder svgs look ugly here, can we have some different svg like home? But maybe it is only for me 😀

src/core/Sidebar/SidebarButton.tsx
mayank99: That could be confusing and unintuitive. Plus we'd have to hide the native tooltip by unassigning `title` again 🤔

src/core/Sidebar/Sidebar.tsx
mayank99: How about `expanderPlacement`? to be consistent with `placement` from popover

stories/core/SideNavigation.stories.tsx
mayank99: Could say the same about 70% of the stories in all components

stories/core/SideNavigation.stories.tsx
bentleyvk: Home, Flag and maybe something from Folder?

mayank99: Done
image

veekeys (approved):

PR #71, opened 2021-05-18
New StylingProps added to ButtonGroup

maxime4000:
Self explaining title.

### DIFF
--- a/src/core/ButtonGroup/ButtonGroup.tsx
+++ b/src/core/ButtonGroup/ButtonGroup.tsx
@@ -3,10 +3,12 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
+import cx from 'classnames';
+import { CommonProps } from '../utils/props';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/button.css';
 
-export type ButtonGroupProps = {
+export type ButtonGroupProps = CommonProps & {
   /**
    * Buttons in the ButtonGroup.
    */
@@ -26,11 +28,11 @@ export type ButtonGroupProps = {
  * </ButtonGroup>
  */
 export const ButtonGroup = (props: ButtonGroupProps) => {
-  const { children, ...rest } = props;
+  const { children, className, ...rest } = props;
 
   useTheme();
   return (
-    <div className='iui-button-group' {...rest}>
+    <div className={cx('iui-button-group', className)} {...rest}>
       {children}
     </div>
   );
--- a/src/core/ButtonGroup/ButtonGroup.tsx
+++ b/src/core/ButtonGroup/ButtonGroup.tsx
@@ -4,16 +4,16 @@
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
 import cx from 'classnames';
-import { CommonProps } from '../utils/props';
+import { ClassNameProps } from '../utils/props';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/button.css';
 
-export type ButtonGroupProps = CommonProps & {
+export type ButtonGroupProps = {
   /**
    * Buttons in the ButtonGroup.
    */
   children: React.ReactNode;
-};
+} & ClassNameProps;
 
 /**
  * Group buttons together for common actions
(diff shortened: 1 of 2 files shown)


src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -3,10 +3,12 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
+import cx from 'classnames';
+import { CommonProps } from '../utils/props';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/button.css';
 
-export type ButtonGroupProps = {
+export type ButtonGroupProps = CommonProps & {
```
mayank99, line 11: This should be placed at the end, not the beginning.

Also, `title` should be excluded.

mayank99 (changes requested):

src/core/ButtonGroup/ButtonGroup.tsx
maxime4000: I'm not sure why title is in CommonProps in the first place if it shouldn't be apply to any component... At this point, we should be moving style to ClassnameProps so I could use ClassnameProps instead of CommonProps.

mayank99: If you're not sure how to use our props, then you should open an issue instead of creating a PR directly.

https://github.com/iTwin/iTwinUI-react/blob/4e5c2f174e8bb2f661d04c979ffbd814e6268a13/CONTRIBUTING.md#L140

maxime4000: I'm sure it doesn't make sense to have styling and classname in a separate props and having title with style in CommonProps. And still being ask to omit the title props in CommonProps when Common mean that any component could use those props... So if it's common, why should I `Omit<CommonProps, "title">` ? Why isn't style with className when both are styling related and style and title are not closely related ?

mayank99: We can look into making the `title` situation nicer but it cannot be part of this PR. Moving `style` into `ClassNameProps` is not something we can just do without checking and updating all references of `CommonProps`.

This PR is supposed to tackle a very _specific_ problem. The solution to that problem is to use `Omit<CommonProps, 'title'>`.

maxime4000: CommonProps is still the same definition and there is an additive feature to ClassNamesProps which cannot cause problems, so there is no issue here... `Omit<CommonProps, "title">` is not a decent fix, it's the bandage you put on a broken bone when you should be installing a metal plate instead.

veekeys: It is kinda a breaking change... I would think about having a new name for both className and style props, if needed.
CommonProps were created for a reason. They are called common, cause the idea was to have there all the native HTML most often used attributes, which we loose support for with our custom components props (not extending the native ones). In most cases, title is used, but components as this one (some grouping ones or big ones like table), just does not make sense to have a title in there, so we need to omit it.
I would suggest just creating a new type, which could have className and style props in there.
Just not sure about the naming, but it is something connected with styling/design/etc (both of those props do the same to the element).

maxime4000: Does StylingProps is reasonable ?

src/core/utils/props.ts
```diff
@@ -9,6 +9,17 @@ export type ClassNameProps = {
...
 
+export type StylingProps = {
+  /**
+   * Custom CSS style properties.
+   */
+  style?: React.CSSProperties;
+  /**
+   * Custom CSS class name.
+   */
+  className?: string;
```
veekeys, line 20: Just extend `ClassNameProps` here instead of copying same definition..

veekeys: You would need to change the places where ClassNameProps is used too, if you want to just remove that :)

veekeys: @maxime4000 are you planning to finish up this PR or you just want us to add the props you need for this component?

src/core/utils/props.ts
```diff
@@ -9,6 +9,13 @@ export type ClassNameProps = {
...
 };
 
+export type StylingProps = {
+  /**
+   * Custom CSS style properties.
+   */
+  style?: React.CSSProperties;
+} & ClassNameProps;
+
 export type CommonProps = {
```
veekeys, line 19: Now CommonProps should take StylingProps, I guess...

maxime4000: ...

mayank99: Can you also merge main? Then I think we can complete this.

mayank99 (approved):
Now I guess we'll have to go update all instances of `Omit<CommonProps, 'title'>` in another PR.

PR #75, opened 2021-05-21
Add useWindowSize hook to DropdownButton

mayank99:
Hook listens to window `resize` event and updates size. Used in `DropdownButton` to update menu width when window gets resized.

## Checklist

- [ ] Add screenshots of the key elements of the component below

## Additional Comments

After #73, it became obvious that the header dropdowns were not being updated on resizing window so this PR is mainly intended to fix that.

### DIFF
--- a/src/core/Buttons/DropdownButton/DropdownButton.tsx
+++ b/src/core/Buttons/DropdownButton/DropdownButton.tsx
@@ -11,6 +11,7 @@ import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
 
 import { useTheme } from '../../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/button.css';
+import { useWindowSize } from '../../utils/hooks/useWindowSize';
 
 export type DropdownButtonProps = {
   /**
@@ -43,6 +44,8 @@ export const DropdownButton: React.FC<DropdownButtonProps> = (props) => {
 
   const [isMenuOpen, setIsMenuOpen] = React.useState(false);
 
+  const windowSize = useWindowSize(500);
+
   const [menuWidth, setMenuWidth] = React.useState(0);
   const ref = React.useRef<HTMLButtonElement>(null);
 
@@ -50,7 +53,7 @@ export const DropdownButton: React.FC<DropdownButtonProps> = (props) => {
     if (ref.current) {
       setMenuWidth(ref.current.offsetWidth);
     }
-  }, [children, size, styleType]);
+  }, [children, size, styleType, windowSize]);
 
   return (
     <DropdownMenu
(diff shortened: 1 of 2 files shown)


src/core/Buttons/DropdownButton/DropdownButton.tsx
```diff
@@ -43,14 +44,16 @@ export const DropdownButton: React.FC<DropdownButtonProps> = (props) => {
...
+
   const [menuWidth, setMenuWidth] = React.useState(0);
   const ref = React.useRef<HTMLButtonElement>(null);
 
   React.useEffect(() => {
     if (ref.current) {
       setMenuWidth(ref.current.offsetWidth);
     }
-  }, [children, size, styleType]);
+  }, [children, size, styleType, windowSize]);
```
bentleyvk, line 56: Is this really needed? Cause I looked at other libs and no one is doing this. Also, it just adds additional complexity.

mayank99: Hmm idk, it's nice. Which libs did you check? Most of them are meant to be as customizable as possible, rather than enforcing standards.

I guess we could also just provide a util.

https://material-ui.com/components/use-media-query/
https://material-ui.com/components/textarea-autosize/
https://blueprintjs.com/docs/#popover2-package/resize-sensor2
https://material.angular.io/cdk/layout/overview

bentleyvk: antd and material-ui changes the select element and menu stays the same size when resizing. react-bootstrap and blueprint does not change anything on resize so hard to tell.

mayank99: Material likes to keep the dropdown width same as select, but it does not have any example where select is taking full width.
image

I would think if you were to code a dynamic (resizable) dropdown using Material, you'd want the width to respond to the change.

But AntD does have some examples.
antd-select

mayank99: Oh this is interesting: [https://popper.js.org/docs/v2/modifiers/community-modifiers/#:~:text=samewidth%20modifier%3A%20make%20your%20popper%20the%20same%20width%20as%20the%20reference.](https://popper.js.org/docs/v2/modifiers/community-modifiers/)

bentleyvk: Your AntD gif does the same as we do right now 😀 and if understand correctly your hook will work with opened menu and resized window.
That modifier would solve a lot of duplication. But does it handle live resizes?

mayank99: This is our current behavior and really the main thing I'm trying to solve.
resize

And I haven't gotten that modifier to work tbh.

bentleyvk: I don't think this is a bad behavior:
image
Maybe @FlyersPh9 can comment on this one.

PR #77, opened 2021-05-21
Added range filter for numbers

bentleyvk:

## Checklist

- [ ] Add any additional comments below describing the features you've implemented

## Screenshots
image

### DIFF
--- a/src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
+++ b/src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
@@ -126,7 +126,7 @@ it('should set filter when both values entered and Enter is pressed', () => {
   expect(setFilter).toHaveBeenCalledWith([1, 3]);
 });
 
-it('should set filter with empty values when invalid date is entered', () => {
+it('should set filter with empty values when invalid number is entered', () => {
   const setFilter = jest.fn();
   const { container } = renderComponent({ setFilter });
 
--- a/src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
+++ b/src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
@@ -1,0 +1,143 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+import { HeaderGroup } from 'react-table';
+import { NumberRangeFilter, NumberRangeFilterProps } from './NumberRangeFilter';
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const renderComponent = (
+  initialProps?: Partial<NumberRangeFilterProps<any>>,
+) => {
+  const props = {
+    column: {} as HeaderGroup,
+    setFilter: jest.fn(),
+    clearFilter: jest.fn(),
+    close: jest.fn(),
+    ...initialProps,
+  } as NumberRangeFilterProps<any>;
+  return render(<NumberRangeFilter {...props} />);
+};
+/* eslint-enable @typescript-eslint/no-explicit-any */
+
+it('should render correctly', () => {
+  const { container } = renderComponent();
+
+  const labeledInputs = container.querySelectorAll(
+    '.iui-input-container.iui-inline',
+  );
+  expect(labeledInputs.length).toBe(2);
+
+  expect(labeledInputs[0].querySelector('.iui-label')?.textContent).toEqual(
+    'From',
+  );
+  expect(labeledInputs[1].querySelector('.iui-label')?.textContent).toEqual(
+    'To',
+  );
+
+  screen.getByText('Filter');
+  screen.getByText('Clear');
+});
+
+it('should render correctly with set filter value', () => {
+  const { container } = renderComponent({
+    column: {
+      filterValue: [1, 3],
+    } as HeaderGroup,
+  });
+
+  const labeledInputs = container.querySelectorAll(
+    '.iui-input-container.iui-inline input',
+  ) as NodeListOf<HTMLInputElement>;
+  expect(labeledInputs.length).toBe(2);
+
+  expect(labeledInputs[0].value).toEqual('1');
+  expect(labeledInputs[1].value).toEqual('3');
+});
+
+it('should set filter when both values entered', () => {
+  const setFilter = jest.fn();
+  const { container } = renderComponent({ setFilter });
+
+  const labeledInputs = container.querySelectorAll(
+    '.iui-input-container.iui-inline input',
+  ) as NodeListOf<HTMLInputElement>;
+  expect(labeledInputs.length).toBe(2);
+
+  fireEvent.change(labeledInputs[0], { target: { value: '1' } });
+  fireEvent.change(labeledInputs[1], { target: { value: '3' } });
+
+  screen.getByText('Filter').click();
+
+  expect(setFilter).toHaveBeenCalledWith([1, 3]);
+});
+
+it('should set filter when only From is entered', () => {
+  const setFilter = jest.fn();
+  const { container } = renderComponent({ setFilter });
+
+  const labeledInputs = container.querySelectorAll(
+    '.iui-input-container.iui-inline input',
+  ) as NodeListOf<HTMLInputElement>;
+  expect(labeledInputs.length).toBe(2);
+
+  fireEvent.change(labeledInputs[0], { target: { value: '1' } });
+
+  screen.getByText('Filter').click();
+
+  expect(setFilter).toHaveBeenCalledWith([1, undefined]);
+});
+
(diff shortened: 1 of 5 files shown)


src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const translatedStrings = { ...defaultStrings, ...translatedLabels };
+
+  const [from, setFrom] = React.useState<string | undefined>(
+    column.filterValue?.[0] ?? '',
+  );
+  const [to, setTo] = React.useState<string | undefined>(
+    column.filterValue?.[1] ?? '',
+  );
+
+  const parseInputValue = React.useCallback((value: string | undefined) => {
```
veekeys, line 48: Does it make sense to have it in useCallback? Cause you are using it in FilterButtonBar still as anonymous function...

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  expect(labeledInputs.length).toBe(2);
+
+  expect(labeledInputs[0].querySelector('.iui-label')?.textContent).toEqual(
+    'From',
+  );
+  expect(labeledInputs[1].querySelector('.iui-label')?.textContent).toEqual(
+    'To',
+  );
+
+  screen.getByText('Filter');
```
veekeys, line 40: I know in more cases we took getByText from render than from screen.. Not sure, whats the diff, but you would not need to import it :D

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  fireEvent.keyDown(labeledInputs[1], {
+    key: 'Enter',
+    charCode: 13,
+  });
+
+  expect(setFilter).toHaveBeenCalledWith([1, 3]);
+});
+
+it('should set filter with empty values when invalid date is entered', () => {
```
veekeys, line 129: date you say?

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const labeledInputs = container.querySelectorAll(
+    '.iui-input-container.iui-inline input',
+  ) as NodeListOf<HTMLInputElement>;
+  expect(labeledInputs.length).toBe(2);
+
+  fireEvent.change(labeledInputs[0], { target: { value: 'abc' } });
+
+  screen.getByText('Filter').click();
+
+  expect(setFilter).toHaveBeenCalledWith([undefined, undefined]);
```
veekeys, line 142: I dont think we should allow to call set with wrong values..

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onChange={(e) => setFrom(e.target.value)}
+        onKeyDown={onKeyDown}
+        type='number'
+        displayStyle='inline'
+        setFocus
+      />
+      <LabeledInput
+        label={translatedStrings.to}
+        value={to}
+        onChange={(e) => setTo(e.target.value)}
```
veekeys, line 72: My guess would be that useCallback for this one might help...

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
bentleyvk: I saw recommendations to use it from `screen` instead destructing from `render` result.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
bentleyvk: With `type='number'` it doesn't allow you to enter anything except number. It is more just to be sure if some browsers (I guess IE) don't handle number inputs.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
bentleyvk: Removed `useCallback` because it really doesn't make sense when every input change still rerenders everything and makes memorized functions to change too.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.test.tsx
bentleyvk: Upsy. Changed.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
bentleyvk: Nah, change of `to` will still trigger rerender.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+        type='number'
```
mayank99, line 73: This looks extremely ugly in dark theme 
image

Wonder if we should use `inputmode` instead.
https://css-tricks.com/what-to-use-instead-of-number-inputs/

Also maybe we should set `min` and `max` so that someone doesn't try to do this:
image

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const [from, setFrom] = React.useState<string | undefined>(
+    column.filterValue?.[0] ?? '',
+  );
+  const [to, setTo] = React.useState<string | undefined>(
+    column.filterValue?.[1] ?? '',
+  );
+
+  const parseInputValue = (value: string | undefined) => {
+    return !value || isNaN(Number(value)) ? undefined : Number(value);
```
mayank99, line 49: Weird logic. Maybe we should allow user to also provide their own parser, similar to date.

Also got me thinking it could be useful for float type values 🤔

stories/core/Table.stories.tsx
```diff
@@ -314,14 +314,17 @@ Sortable.args = {
...
-      date: Date;
-    }>,
-  ) => action(props.row.original.name)();
+  type TableStoryDataType = {
+    index: number;
+    name: string;
+    description: string;
+    ids: number[];
+    date: Date;
+  };
```
mayank99, line 323: This story code is getting too complicated. Can't we infer this type using `typeof something`?

stories/core/Table.stories.tsx
```diff
@@ -314,14 +314,17 @@ Sortable.args = {
...
+  const onClickHandler = useCallback(
+    (props: CellProps<TableStoryDataType>) => action(props.row.original.name)(),
+    [],
+  );
```
mayank99, line 327: Stuff like this adds up. Maybe we should remove this column

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
bentleyvk: I need to convert string to number `Number(undefined) === 0` so I have to check whether `value` is defined - it won't allow `''`, `undefined`.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
bentleyvk: Don't use dark theme then 😀 I don't know from this article and the one it referenced it seems that it is a problem with number input fields that isn't just a regular numbers like credit card number.
By `min`, `max` you mean to use it inside with current input values or/and have it as props?

stories/core/Table.stories.tsx
bentleyvk: I can't use `typeof` because data is array, maybe I could move it outside.

stories/core/Table.stories.tsx
bentleyvk: Removed.

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
mayank99: min max with current input values, although that could get complicated 🤔

stories/core/Table.stories.tsx
mayank99: leave it I guess. Moving outside would not make sense as every story differs

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
mayank99: and what about float values?

mayank99 (approved):
Looks good I guess 😄

src/core/Table/filters/NumberRangeFilter/NumberRangeFilter.tsx
bentleyvk: It support float.

PR #83, opened 2021-05-24
Time picker component

veekeys:

## Checklist

## Additional Comments

image

image

Currently waiting for css changes, to make separate stories for TimePicker

### DIFF
--- a/setupTests.js
+++ b/setupTests.js
@@ -1,0 +1,6 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+window.HTMLElement.prototype.scrollIntoView = () => {};
+Element.prototype.scrollIntoView = () => {};
--- a/setupTests.js
+++ b/setupTests.js
@@ -3,4 +3,3 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 window.HTMLElement.prototype.scrollIntoView = () => {};
-Element.prototype.scrollIntoView = () => {};
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -227,7 +227,7 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
 
   const getNewFocusedDate = (newYear: number, newMonth: number) => {
     const newDate = new Date(selectedDay ?? new Date());
-    newDate?.setFullYear(newYear);
+    newDate.setFullYear(newYear);
     newDate?.setMonth(newMonth);
     return newDate;
   };
@@ -253,6 +253,8 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
       setMonthAndYear(day.getMonth(), day.getFullYear());
     }
     const newDate = new Date(selectedDay ?? new Date());
+    newDate.setFullYear(displayedYear);
+    newDate.setMonth(displayedMonthIndex);
     newDate.setDate(day.getDate());
     setSelectedDay(newDate);
     setFocusedDay(newDate);
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -6,7 +6,6 @@ import SvgChevronLeft from '@itwin/itwinui-icons-react/cjs/icons/ChevronLeft';
 import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
 import cx from 'classnames';
 import React from 'react';
-import { CommonProps } from '../utils/props';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/date-picker.css';
 import { IconButton } from '../Buttons/IconButton';
@@ -130,8 +129,7 @@ export type DatePickerProps = {
    * @default false
    */
   showTime?: boolean;
-} & Omit<CommonProps, 'title'> &
-  Omit<TimePickerProps, 'date' | 'onChange' | 'setFocusHour'>;
+} & Omit<TimePickerProps, 'date' | 'onChange' | 'setFocusHour'>;
 
 /**
  * Date picker component
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -165,6 +165,8 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
   const [displayedYear, setDisplayedYear] = React.useState(
     selectedDay ? selectedDay.getFullYear() : new Date().getFullYear(),
   );
+  // Used to focus days only when days are changed
+  // e.g. without this, when changing months day would be focused
   const needFocus = React.useRef(setFocus);
 
   React.useEffect(() => {
@@ -224,10 +226,10 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
   }, [days]);
 
   const getNewFocusedDate = (newYear: number, newMonth: number) => {
-    return selectedDay?.getMonth() === newMonth &&
-      selectedDay?.getFullYear() === newYear
-      ? selectedDay
-      : new Date(newYear, newMonth, 1);
+    const newDate = new Date(selectedDay ?? new Date());
+    newDate?.setFullYear(newYear);
+    newDate?.setMonth(newMonth);
+    return newDate;
   };
 
   const handleMoveToPreviousMonth = () => {
@@ -250,9 +252,11 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     if (day.getMonth() !== selectedDay?.getMonth()) {
       setMonthAndYear(day.getMonth(), day.getFullYear());
     }
-    setSelectedDay(day);
-    setFocusedDay(day);
-    onChange?.(day);
+    const newDate = new Date(selectedDay ?? new Date());
+    newDate.setDate(day.getDate());
+    setSelectedDay(newDate);
+    setFocusedDay(newDate);
+    onChange?.(newDate);
   };
 
   const handleCalendarKeyDown = (
@@ -393,7 +397,7 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
           hourStep={hourStep}
           minuteStep={minuteStep}
           secondStep={secondStep}
-          onChange={(date) => onDayClick(date)}
+          onChange={(date) => onChange?.(date)}
         />
       )}
     </div>
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -10,6 +10,7 @@ import { CommonProps } from '../utils/props';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/date-picker.css';
 import { IconButton } from '../Buttons/IconButton';
+import { TimePicker, TimePickerProps } from '../TimePicker';
 
 const isSameDay = (a: Date | undefined, b: Date | undefined) => {
   return (
@@ -124,7 +125,9 @@ export type DatePickerProps = {
    * @default false
    */
   setFocus?: boolean;
-} & Omit<CommonProps, 'title'>;
+  showTime?: boolean;
+} & Omit<CommonProps, 'title'> &
+  Omit<TimePickerProps, 'date' | 'onChange'>;
 
 /**
  * Date picker component
@@ -139,6 +142,12 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     className,
     style,
     setFocus = false,
+    showTime = false,
+    use12Hours = false,
+    precision,
+    hourStep,
+    minuteStep,
+    secondStep,
     ...rest
   } = props;
 
@@ -376,6 +385,17 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
           })}
         </div>
       </div>
+      {showTime && (
+        <TimePicker
+          date={selectedDay}
+          use12Hours={use12Hours}
+          precision={precision}
+          hourStep={hourStep}
+          minuteStep={minuteStep}
+          secondStep={secondStep}
+          onChange={(date) => onDayClick(date)}
+        />
+      )}
     </div>
   );
 };
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -152,6 +152,7 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     hourStep,
     minuteStep,
     secondStep,
+    setFocusHour,
     ...rest
   } = props;
 
@@ -415,6 +416,7 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
           minuteStep={minuteStep}
           secondStep={secondStep}
           onChange={(date) => onChange?.(date)}
+          setFocusHour={setFocusHour}
         />
       )}
     </div>
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -125,6 +125,10 @@ export type DatePickerProps = {
    * @default false
    */
   setFocus?: boolean;
+  /**
+   * Show time picker near the calendar.
+   * @default false
+   */
   showTime?: boolean;
 } & Omit<CommonProps, 'title'> &
   Omit<TimePickerProps, 'date' | 'onChange'>;
@@ -226,9 +230,15 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
   }, [days]);
 
   const getNewFocusedDate = (newYear: number, newMonth: number) => {
-    const newDate = new Date(selectedDay ?? new Date());
-    newDate.setFullYear(newYear);
-    newDate?.setMonth(newMonth);
+    const currentDate = selectedDay ?? new Date();
+    const newDate = new Date(
+      newYear,
+      newMonth,
+      currentDate.getDate(),
+      currentDate.getHours(),
+      currentDate.getMinutes(),
+      currentDate.getSeconds(),
+    );
     return newDate;
   };
 
@@ -252,10 +262,15 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     if (day.getMonth() !== selectedDay?.getMonth()) {
       setMonthAndYear(day.getMonth(), day.getFullYear());
     }
-    const newDate = new Date(selectedDay ?? new Date());
-    newDate.setFullYear(displayedYear);
-    newDate.setMonth(displayedMonthIndex);
-    newDate.setDate(day.getDate());
+    const currentDate = selectedDay ?? new Date();
+    const newDate = new Date(
+      day.getFullYear(),
+      day.getMonth(),
+      day.getDate(),
+      currentDate.getHours(),
+      currentDate.getMinutes(),
+      currentDate.getSeconds(),
+    );
     setSelectedDay(newDate);
     setFocusedDay(newDate);
     onChange?.(newDate);
--- a/src/core/DatePicker/DatePicker.tsx
+++ b/src/core/DatePicker/DatePicker.tsx
@@ -131,7 +131,7 @@ export type DatePickerProps = {
    */
   showTime?: boolean;
 } & Omit<CommonProps, 'title'> &
-  Omit<TimePickerProps, 'date' | 'onChange'>;
+  Omit<TimePickerProps, 'date' | 'onChange' | 'setFocusHour'>;
 
 /**
  * Date picker component
@@ -152,7 +152,6 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     hourStep,
     minuteStep,
     secondStep,
-    setFocusHour,
     ...rest
   } = props;
 
@@ -416,7 +415,6 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
           minuteStep={minuteStep}
           secondStep={secondStep}
           onChange={(date) => onChange?.(date)}
-          setFocusHour={setFocusHour}
         />
       )}
     </div>
--- a/src/core/Select/Select.test.tsx
+++ b/src/core/Select/Select.test.tsx
@@ -62,16 +62,8 @@ function renderComponent(props?: Partial<SelectProps<number>>) {
   );
 }
 
-const originalScrollIntoView = window.HTMLElement.prototype.scrollIntoView;
-const scrollIntoViewMock = jest.fn();
-
 beforeEach(() => {
-  window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock;
   jest.clearAllMocks();
-});
-
-afterAll(() => {
-  window.HTMLElement.prototype.scrollIntoView = originalScrollIntoView;
 });
 
 it('should render empty select', () => {
@@ -252,6 +244,8 @@ it('should show menu with disabled item', () => {
 });
 
 it('should show selected item in menu', () => {
+  const mockedScroll = jest.fn();
+  window.HTMLElement.prototype.scrollIntoView = mockedScroll;
   const { container } = renderComponent({
     value: 1,
     options: [...new Array(3)].map((_, index) => ({
@@ -266,7 +260,7 @@ it('should show selected item in menu', () => {
   fireEvent.click(select.querySelector('.iui-select-button') as HTMLElement);
   const menu = document.querySelector('.iui-menu') as HTMLUListElement;
   assertMenu(menu, { selectedIndex: 1 });
-  expect(scrollIntoViewMock).toHaveBeenCalledTimes(1);
+  expect(mockedScroll).toHaveBeenCalledTimes(1);
 });
 
 it('should call onChange on item click', () => {
--- a/src/core/Select/Select.test.tsx
+++ b/src/core/Select/Select.test.tsx
@@ -244,7 +244,7 @@ it('should show menu with disabled item', () => {
 });
 
 it('should show selected item in menu', () => {
-  spyOn(window.HTMLElement.prototype, 'scrollIntoView');
+  const scrollSpy = spyOn(window.HTMLElement.prototype, 'scrollIntoView');
   const { container } = renderComponent({
     value: 1,
     options: [...new Array(3)].map((_, index) => ({
@@ -259,9 +259,7 @@ it('should show selected item in menu', () => {
   fireEvent.click(select.querySelector('.iui-select-button') as HTMLElement);
   const menu = document.querySelector('.iui-menu') as HTMLUListElement;
   assertMenu(menu, { selectedIndex: 1 });
-  expect(
-    container.querySelector('.iui-active')?.scrollIntoView,
-  ).toHaveBeenCalledTimes(1);
+  expect(scrollSpy).toHaveBeenCalledTimes(1);
 });
 
 it('should call onChange on item click', () => {
--- a/src/core/Select/Select.test.tsx
+++ b/src/core/Select/Select.test.tsx
@@ -244,8 +244,7 @@ it('should show menu with disabled item', () => {
 });
 
 it('should show selected item in menu', () => {
-  const mockedScroll = jest.fn();
-  window.HTMLElement.prototype.scrollIntoView = mockedScroll;
+  spyOn(window.HTMLElement.prototype, 'scrollIntoView');
   const { container } = renderComponent({
     value: 1,
     options: [...new Array(3)].map((_, index) => ({
@@ -260,7 +259,9 @@ it('should show selected item in menu', () => {
   fireEvent.click(select.querySelector('.iui-select-button') as HTMLElement);
   const menu = document.querySelector('.iui-menu') as HTMLUListElement;
   assertMenu(menu, { selectedIndex: 1 });
-  expect(mockedScroll).toHaveBeenCalledTimes(1);
+  expect(
+    container.querySelector('.iui-active')?.scrollIntoView,
+  ).toHaveBeenCalledTimes(1);
 });
 
 it('should call onChange on item click', () => {
--- a/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -7,9 +7,11 @@ import { render } from '@testing-library/react';
 
 import { ThemeProvider } from './ThemeProvider';
 
-const expectLightTheme = () => {
-  expect(document.documentElement.classList).toContain('iui-theme-light');
-  expect(document.documentElement.classList).not.toContain('iui-theme-dark');
+const expectLightTheme = (ownerDocument = document) => {
+  expect(ownerDocument.documentElement.classList).toContain('iui-theme-light');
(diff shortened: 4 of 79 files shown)


src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+  const scrollIntoView = (ref: HTMLLIElement | null, isSame: boolean) => {
+    isSame && ref?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
+  };
+
+  const getTimeTemplate = (
+    data: number[],
+    step: number,
+    isSame: (number: number, date: Date | undefined) => boolean,
+    onClick: (number: number) => void,
+  ) => {
```
mayank99, line 156: So this already has 4 params. What other ones do you think would be needed here to customize? If it's like 2-3 more props, I would say that's not too bad if they still reuse some code. If not, maybe we need 3 different template generators.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+          {data
+            .filter((_, index) => index % step === 0)
+            .map((value) => {
+              return (
+                <li
+                  className={cx({
+                    'iui-selected': isSame(value, selectedTime),
+                  })}
+                  key={value}
+                  tabIndex={isSame(value, selectedTime) ? 0 : undefined}
```
mayank99, line 169: Yeah, this won't be nice for keyboard navigation, better remove, even if you're not gonna add arrow key support initially.

When it comes to accessibility, "none" is better than "bad".

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  const onSecondsClick = (seconds: number) => {
+    const adjustedSelectedDay = selectedTime ?? new Date();
+    adjustedSelectedDay.setSeconds(seconds);
+    updateCurrentTime(adjustedSelectedDay);
+  };
+
+  const updateCurrentTime = (time: Date) => {
+    setSelectedTime(time);
+    onChange?.(time);
```
mayank99, line 108: Single callback exposed with single object 🤔 I'm thinking users might want more granular feedback so they know exactly what changed.

mayank99: Well I guess they could compare with old object themselves.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+                    'iui-selected': isSame(value, selectedTime),
+                  })}
+                  key={value}
+                  tabIndex={isSame(value, selectedTime) ? 0 : undefined}
+                  ref={(ref) => {
+                    scrollIntoView(ref, isSame(value, selectedTime));
+                  }}
+                  onClick={() => onClick(value)}
+                >
+                  {`0${value}`.slice(-2)}
```
mayank99, line 175: This is interesting 🤔 Could it cause problems anywhere?

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+const areSameMinutes = (minutes: number, date: Date | undefined) => {
+  return !!date && minutes === date.getMinutes();
+};
+
+const areSameSeconds = (seconds: number, date: Date | undefined) => {
+  return !!date && seconds === date.getSeconds();
+};
+
+const isSamePeriod = (period: string, date: Date | undefined) => {
```
mayank99, line 22: This is called "meridiem" btw

src/core/TimePicker/TimePicker.tsx
mayank99: @veekeys What other feedback do you need regarding this? I think this approach is pretty okay-ish

veekeys: I guess I will just try to implement something working from keyboard navigation and see, how it goes..

src/core/TimePicker/TimePicker.tsx
veekeys: I know the latin word in there, but can I just call them like that? I think I found this called period in some lib of many...

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+          focusedTime={focusedTime}
+          selectedTime={selectedTime}
+          onFocusChange={onSecondFocus}
+          onSelectChange={onSecondsClick}
+          isSame={areSameSeconds}
+          step={secondStep}
+        />
+      )}
+      {use12Hours && (
+        <div className='iui-time'>
```
bentleyvk, line 218: It should be `iui-period`

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+                    (index) => onFocusChange?.(data[index]),
+                    (index) => onSelectChange?.(data[index]),
+                    index,
+                  );
+                }}
+                className={cx({
+                  'iui-selected': isSame(value, selectedTime),
+                })}
+                key={value}
+                tabIndex={isSame(value, focusedTime) ? 0 : undefined}
```
bentleyvk, line 353: Need to handle precision because now it is not selected.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+    focusedTime,
+    selectedTime,
+    onFocusChange,
+    onSelectChange,
+    isSame,
+    step = 0,
+  } = props;
+  const needFocus = React.useRef(false);
+
+  React.useEffect(() => {
```
bentleyvk, line 285: Add comment becaue we will forget again what it is for 😀

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+                  handleTimeKeyDown(
+                    event,
+                    data.length - 1,
+                    (index) => onFocusChange?.(data[index]),
+                    (index) => onSelectChange?.(data[index]),
+                    index,
+                  );
+                }}
+                className={cx({
+                  'iui-selected': isSame(value, selectedTime),
```
bentleyvk, line 350: You can move this call result up because now it is called 4 times.

src/core/DatePicker/DatePicker.tsx
```diff
@@ -241,9 +252,11 @@ export const DatePicker = (props: DatePickerProps): JSX.Element => {
     if (day.getMonth() !== selectedDay?.getMonth()) {
       setMonthAndYear(day.getMonth(), day.getFullYear());
     }
-    setSelectedDay(day);
-    setFocusedDay(day);
-    onChange?.(day);
+    const newDate = new Date(selectedDay ?? new Date());
+    newDate.setDate(day.getDate());
+    setSelectedDay(newDate);
```
bentleyvk, line 257: Something is fucked up because when selecting a day in next month, it selects the same day but in the current month.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+      adjustedSelectedTime.setHours(currentHours - 12);
+    }
+    if (value === 'PM' && currentHours <= 12) {
+      setPeriod(value);
+      adjustedSelectedTime.setHours(currentHours + 12);
+    }
+    updateCurrentTime(adjustedSelectedTime);
+  };
+
+  const updateCurrentTime = (time: Date) => {
```
bentleyvk, line 143: We might need to look at the precision because currently if precision is minutes and I select 10:40, it can set like 10:40:34. If I am going to use that in filter in from field I would expect that 10:40:05 would be shown for me but with current implementation it would be not.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+  const onSecondsClick = (seconds: number) => {
+    const adjustedSelectedTime = selectedTime ?? new Date();
+    adjustedSelectedTime.setSeconds(seconds);
+    updateCurrentTime(adjustedSelectedTime);
+  };
+
+  const onPeriodClick = (value: string) => {
+    const adjustedSelectedTime = selectedTime ?? new Date();
+    const currentHours = adjustedSelectedTime.getHours();
+    if (value === 'AM' && currentHours > 11) {
```
bentleyvk, line 132: Won't work with localization 😕

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+        {data
+          .filter((_, index) => index % step === 0)
+          .map((value, index) => {
+            return (
+              <li
+                onKeyDown={(event) => {
+                  handleTimeKeyDown(
+                    event,
+                    data.length - 1,
+                    (index) => onFocusChange?.(data[index]),
```
bentleyvk, line 344: Are you going to export this to root index? If not then these functions should be required.

setupTests.js
```diff
@@ -0,0 +1,6 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+window.HTMLElement.prototype.scrollIntoView = () => {};
```
bentleyvk, line 5: Why you need these?

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+                key={value}
+                tabIndex={isSame(value, focusedTime) ? 0 : undefined}
+                ref={(ref) => {
+                  scrollIntoView(ref, isSame(value, focusedTime));
+                  needFocus.current &&
+                    isSame(value, focusedTime) &&
+                    ref?.focus();
+                }}
+                onClick={() => onSelectChange?.(value)}
+                onBlur={() => {
```
bentleyvk, line 361: I guess this one is not needed anymore.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+    use12Hours = false,
+    precision = 'minutes',
+    hourStep = 1,
+    minuteStep = 1,
+    secondStep = 1,
+  } = props;
+
+  useTheme();
+
+  const [selectedTime, setSelectedTime] = React.useState(date);
```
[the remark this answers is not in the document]
bentleyvk: So it would always have the default value of current time. I don't think this should be a problem because you currently use current time everywhere.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+  const applyFocus = (
+    value: number | undefined,
+    callback: (value: number, date: Date) => void,
+  ) => {
+    if (!!value) {
+      const date = new Date(focusedTime);
+      callback(value, date);
+      setFocusedTime(date);
+    } else {
```
bentleyvk, line 171: When else is called?

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+  React.useEffect(() => {
+    if (needFocus.current) {
+      needFocus.current = false;
+      // onFocusChange?.(undefined);
+    }
+  });
+
+  const scrollIntoView = (ref: HTMLLIElement | null, isSame: boolean) => {
+    isSame && ref?.scrollIntoView({ block: 'nearest', inline: 'nearest' });
```
bentleyvk, line 293: Can you use here `needFocus` instead of passing options `scrollIntoView`? Because they are not fully supported by all browsers.

src/core/TimePicker/TimePicker.tsx
bentleyvk: Ok I found it. It is triggered when focusing item with 0 and it causes focus to jump to currently seelected value or if nothing is selected to the current time value.

bentleyvk: Yup, it is not needed as you always get a value. So you can change `value` type to just `number` also the same `on...Focus`

src/core/TimePicker/TimePicker.tsx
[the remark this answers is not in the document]
veekeys: I have selectedDate in DatePicker and there it can be undefined.. So I either leave it like that or change in both places then..

src/core/TimePicker/TimePicker.tsx
veekeys: I guess in future we should provide ability to format the shown value.. Just dont want to add all the features at once here, it is really endless.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+export type TimePickerProps = {
+  /**
+   * Selected date.
+   */
+  date?: Date;
+  /**
+   * Callback when date is changed.
+   */
+  onChange?: (date: Date) => void;
```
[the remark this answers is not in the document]
veekeys: Added

src/core/TimePicker/TimePicker.tsx
veekeys: good catch

src/core/TimePicker/TimePicker.tsx
veekeys: Moved

setupTests.js
veekeys: For tests.. It is undefined in jest.

src/core/TimePicker/TimePicker.tsx
veekeys: Not always.. onBlur I send undefined.. It is this hack to get out from the focus..
Ill try to check again if there is a way without it..

src/core/TimePicker/TimePicker.tsx
veekeys: how needFocus would help me to scroll whole thing?
i tested in other browsers (like IE), it is not that bad. Scroll goes a little (to the nearest border), but its ok.
Much worse is the visual look (css issue):
image

src/core/DatePicker/DatePicker.tsx
veekeys: yeah, I thought I fixed it.. Might have undone it for other issues...
Will double check..

veekeys: yeah, fixed...

src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -54,7 +54,7 @@ it('should set default theme', () => {
   expectDarkTheme();
 
   render(<ThemeProvider />);
-  expectDarkTheme();
+  expectLightTheme();
```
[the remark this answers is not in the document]
veekeys: I have no idea how it was passing before. Read the test.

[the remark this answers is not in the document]
veekeys: or wait, maybe it should be dark...
hmmm..
Now I am confused.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,401 @@
...
+  const updateCurrentTime = (time: Date) => {
+    setFocusedTime(time);
+    setSelectedTime(time);
+    onChange?.(time);
+  };
+
+  const onHourFocus = (hour: number) => {
+    const adjustedHour =
+      use12Hours && hour ? formatHourFrom12(hour, period) : hour;
+    applyFocus(adjustedHour, (value, date) => setHours(value, date));
```
bentleyvk, line 191: You can just pass `setHours` no need for arrow fn.

bentleyvk: Ok, `applyFocus` doesn't do anything, you just can call `setFocusTime` here.

src/core/TimePicker/TimePicker.tsx
veekeys: added

src/core/TimePicker/TimePicker.tsx
veekeys: changed

src/core/TimePicker/TimePicker.tsx
veekeys: removed

src/core/TimePicker/TimePicker.tsx
veekeys: removed

src/core/TimePicker/TimePicker.tsx
veekeys: yup, simplified

src/core/TimePicker/TimePicker.tsx
veekeys: Something is working!

src/core/TimePicker/TimePicker.tsx
veekeys: Check up current state, if its "bad".

src/core/TimePicker/TimePicker.tsx
veekeys: Now I have valueRenderer prop, which we could expose to the users and they could then show what ever they want.

src/core/TimePicker/TimePicker.tsx
veekeys: Fixed

src/core/TimePicker/TimePicker.tsx
veekeys: Not sure, what was this about now..

src/core/DatePicker/DatePicker.tsx
```diff
@@ -124,7 +125,13 @@ export type DatePickerProps = {
...
    */
   setFocus?: boolean;
-} & Omit<CommonProps, 'title'>;
+  /**
+   * Show time picker near the calendar.
+   * @default false
+   */
+  showTime?: boolean;
+} & Omit<CommonProps, 'title'> &
+  Omit<TimePickerProps, 'date' | 'onChange'>;
```
bentleyvk, line 134: Omit `setFocusHour`

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,425 @@
...
+      ? formatHourFrom12(date.getHours(), period)
+      : date.getHours();
+    setFocusedTime(setHours(adjustedHour, focusedTime));
+  };
+
+  const onPeriodFocus = (value: string) => {
+    let adjustedSelectedTime = selectedTime ?? new Date();
+    const currentHours = adjustedSelectedTime.getHours();
+    if (value === 'AM' && currentHours > 11) {
+      setPeriod(value);
```
bentleyvk, line 169: This can be moved before `if`

setupTests.js
bentleyvk: Can you check other tests for this?

mayank99: Yeah I thought we had one in Select

src/core/TimePicker/TimePicker.tsx
mayank99: How is the valueRenderer (part of `TimePickerColumn`) exposed to the user? I'm confused

veekeys: It is not yet, but it might be.. Just need to think of API.. Multiple props (hour, second, etc..) or something else..

src/core/TimePicker/TimePicker.tsx
bentleyvk: I also don't remember 😀

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,425 @@
...
+  const adjustedHours = period
+    ? formatHourFrom12(date1.getHours(), period)
+    : date1.getHours();
+  if (!!period) {
+    return !!date2 && adjustedHours % 12 === date2.getHours() % 12;
+  }
+  return !!date2 && adjustedHours === date2.getHours();
+};
+
+const areSameMinutes = (date1: Date, date2: Date | undefined) => {
```
bentleyvk, line 24: Maybe rename to `isSameMinute`, `isSameSecond` to be consistent also `are*` looks odd.

src/core/TimePicker/TimePicker.tsx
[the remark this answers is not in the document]
bentleyvk: I think here might be a different scenario because if no date is provided and you select hours then minutes and seconds gets auto selected. Maybe we should initialize with `00:00:00` also in datepicker if date is not provided then set time with zeros.

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,425 @@
...
+
+type TimePickerColumnProps<T = Date> = {
+  /**
+   * Data to render in column.
+   */
+  data: T[];
+  /**
+   * Compare function for focus.
+   */
+  isSameFocused: (date1: T) => boolean;
```
bentleyvk, line 302: Rename here to `value`

src/core/DatePicker/DatePicker.tsx
veekeys: YOu think there wont be a need to focus hour before day?

src/core/TimePicker/TimePicker.tsx
veekeys: renamed

src/core/TimePicker/TimePicker.tsx
veekeys: renamed

src/core/TimePicker/TimePicker.tsx
veekeys: moved

src/core/DatePicker/DatePicker.tsx
veekeys: omited..

src/core/TimePicker/TimePicker.tsx
veekeys: Now it should work, I think ...

src/core/TimePicker/TimePicker.tsx
veekeys: Added new props for every column cell value. So now it can be customized by the user, what we gonna show in there (including AM and PM for localization)

setupTests.js
veekeys: changed in select a little

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,453 @@
...
+    ...rest
+  } = props;
+
+  useTheme();
+
+  const [selectedTime, setSelectedTime] = React.useState(date);
+  const [focusedTime, setFocusedTime] = React.useState(
+    selectedTime ?? new Date(),
+  );
+  const [period, setPeriod] = React.useState<'PM' | 'AM' | undefined>(
```
mayank99, line 147: Can use `PeriodType` here?

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,453 @@
...
+    onChange,
+    use12Hours = false,
+    precision = 'minutes',
+    hourStep = 1,
+    minuteStep = 1,
+    secondStep = 1,
+    setFocusHour = false,
+    hourRenderer = (date: Date) => <>{date.getHours()}</>,
+    minuteRenderer = (date: Date) => <>{date.getMinutes()}</>,
+    secondRenderer = (date: Date) => <>{date.getSeconds()}</>,
```
mayank99, line 135: All these renderers, I wonder if they should default to two digits?

Maybe using [`toLocaleString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)
```tsx
date.getHours().toLocaleString(undefined, { minimumIntegerDigits: 2 })
```

Just weird seeing 24 hour format without preceding zeroes. 08:23 makes more sense than 8:23 (without AM).

src/core/TimePicker/TimePicker.tsx
veekeys: yeah, good catch..

src/core/TimePicker/TimePicker.tsx
veekeys: Adjusting date now according to the precision on select.

src/core/TimePicker/TimePicker.tsx
veekeys: using now for hours, mins and secs...

stories/core/TimePicker.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+import SvgCalendar from '@itwin/itwinui-icons-react/cjs/icons/Calendar';
+
+export default {
+  title: 'Core/TimePicker',
+  component: TimePicker,
+  argTypes: {
+    onChange: { control: { disable: true } },
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+    hourRenderer: {
```
veekeys, line 25: Dont like these, but somehow storybook stringified default values and I was getting errors...
Might be connected:
https://github.com/storybookjs/storybook/issues/12455

mayank99: Found the problem: https://github.com/storybookjs/storybook/issues/12098#issuecomment-758153653

I don't think there is a better workaround than what you did

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,491 @@
...
+   */
+  onSelectChange: (value: T) => void;
+  /**
+   * Set initial focus.
+   */
+  setFocus?: boolean;
+  /**
+   * What value to display in every cell.
+   */
+  valueRenderer: (value: T) => JSX.Element;
```
mayank99, line 388: Why does it need to be JSX.Element and not just ReactNode?

src/core/DatePicker/DatePicker.tsx
```diff
@@ -124,7 +125,13 @@ export type DatePickerProps = {
    * @default false
    */
   setFocus?: boolean;
-} & Omit<CommonProps, 'title'>;
+  /**
+   * Show time picker near the calendar.
+   * @default false
+   */
+  showTime?: boolean;
+} & Omit<CommonProps, 'title'> &
```
mayank99, line 133: Think `Omit<CommonProps, 'title'` is not needed anymore as `StylingProps` are coming from `TimePicker`

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,491 @@
...
+  /**
+   * Class to apply on root.
+   */
+  className?: string;
```
mayank99, line 392: Could use classNameProps? Or even StylingProps

stories/core/TimePicker.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+    action(`New Time value: ${date}`, { clearOnStoryChange: false })();
+  };
+
+  useEffect(() => {
+    setCurrentDate(new Date(date));
+    return () => action('', { clearOnStoryChange: true })();
+  }, [date]);
+  return (
+    <>
+      <LabeledInput
```
mayank99, line 73: I'd say make the input `readOnly` since it doesn't do anything.

Or even remove it and go with an icon button (similar to DatePicker)

stories/core/TimePicker.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+        displayStyle='inline'
+        value={`${currentDate.getHours()}:${currentDate.getMinutes()}:${currentDate.getSeconds()}`}
```
mayank99, line 75: Looks weird 
image

Try just `currentDate.toLocaleTimeString()`

stories/core/TimePicker.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+  IconButton,
+  LabeledInput,
+  MeridiemType,
+} from '../../src/core';
+import SvgCalendar from '@itwin/itwinui-icons-react/cjs/icons/Calendar';
+
+export default {
+  title: 'Core/TimePicker',
+  component: TimePicker,
+  argTypes: {
```
mayank99, line 21: Could use `date: { control: { type: 'date' } }` for adding date picker control 😄

src/core/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,491 @@
...
+  /**
+   * Change step of the seconds displayed.
+   * @default 1
+   */
+  secondStep?: number;
+  /**
+   * Set focus on hour.
+   * @default false
+   */
+  setFocusHour?: boolean;
```
mayank99, line 95: How do I demo this? It seems to do nothing in storybook currently

mayank99 (commented):
Looks pretty good. Few minor comments

src/core/TimePicker/TimePicker.tsx
mayank99: Ah nevermind, had to reload because I was testing smooth scrolling (which is broken). Arrow keys work

stories/core/TimePicker.stories.tsx
mayank99: Oh btw can you use `args` instead?

`argTypes.defaultValue` will be deprecated soon: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-argtypedefaultvalue

src/core/TimePicker/TimePicker.tsx
veekeys: Had something in my mind, but cannot remember now..

src/core/TimePicker/TimePicker.tsx
veekeys: Wrote it that way as inner component and its mostly for the iui-period reason.. but I can take classNameProps I guess..

stories/core/TimePicker.stories.tsx
veekeys: Kinda wanted to show its achievable (what Maxime requested in some issue)

src/core/DatePicker/DatePicker.tsx
veekeys: oh, that got a little messy...

stories/core/TimePicker.stories.tsx
veekeys: Added

stories/core/TimePicker.stories.tsx
veekeys: Tried, fixed

stories/core/TimePicker.stories.tsx
veekeys: made readonly, looks great

src/core/TimePicker/TimePicker.tsx
veekeys: using classNameProps

src/core/DatePicker/DatePicker.tsx
veekeys: removed..

src/core/TimePicker/TimePicker.tsx
veekeys: using ReactNode

stories/core/TimePicker.stories.tsx
veekeys: moved to args..

mayank99 (approved):

src/core/Select/Select.test.tsx
```diff
@@ -252,6 +244,8 @@ it('should show menu with disabled item', () => {
 });
 
 it('should show selected item in menu', () => {
+  const mockedScroll = jest.fn();
+  window.HTMLElement.prototype.scrollIntoView = mockedScroll;
```
bentleyvk, line 248: Better use `spyOn` here. Previously I couldn't because `scrollIntoView` was missing.

veekeys: used..

bentleyvk (approved):

PR #93, opened 2021-05-26
Update issue templates to include project

mayank99:
- Changed issue templates to use `config.yml` with the `projects=` url param
- Removed most frontmatter from template yaml to avoid them being listed multiple times

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,10 +1,5 @@
 ---
-name: Bug report 🐞
-about: Create a report to help us improve
-title: ''
-labels: bug
-assignees: ''
-
+about: Create a bug report to help us improve
 ---
 
 #### Environment
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,11 @@
 blank_issues_enabled: false
 contact_links:
+  - name: Bug report 🐞
+    url: https://github.com/iTwin/iTwinUI-react/issues/new?labels=bug&template=bug_report.md&projects=iTwin/iTwinUI-react/1
+    about: Create a report to help us improve
+  - name: Feature request 💄
+    url: https://github.com/iTwin/iTwinUI-react/issues/new?labels=enhancement&template=feature_request.md&projects=iTwin/iTwinUI-react/1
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,10 +1,5 @@
 ---
-name: Bug report 🐞
-about: Create a report to help us improve
-title: ''
-labels: bug
-assignees: ''
-
+about: Create a bug report to help us improve
```
mayank99, line 2: I kept this to preserve this text:
image

Not sure it will work

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,11 @@
 blank_issues_enabled: false
 contact_links:
+  - name: Bug report 🐞
+    url: https://github.com/iTwin/iTwinUI-react/issues/new?labels=bug&template=bug_report.md&projects=iTwin/iTwinUI-react/1
```
bentleyvk, line 4: Have you tried accessing these links with your private account? Because I am wondering whether they work for us as we are owners of the repo.

.github/ISSUE_TEMPLATE/bug_report.md
veekeys: Now these props (about) look duplicated.. We really need that on both files now?

mayank99: I kept the text slightly different to check which one is being used. There is no other way to really test this 🙁

If we find it's not being used, we can remove it

.github/ISSUE_TEMPLATE/config.yml
mayank99: Not sure what you mean.

If you try to create an issue from a template right now, it goes to the same url. I just added the `projects=` part

.github/ISSUE_TEMPLATE/bug_report.md
veekeys: But you did remove the name..and it is shown, I guess.
So should be same.

mayank99: it is shown? no that is not a real screenshot. My guess is the name will just say "Issue".

Again, there is no way to test this without actually merging it.

veekeys (approved):

PR #94, opened 2021-05-26
Update to itwinui-css 0.17.1

mayank99:
- Updated to latest itwinui-css
- Added inline width to wizard step, calculated based on total steps (needed after https://github.com/iTwin/iTwinUI/pull/104)
- Removed before/after tracks from non-default wizard steps (needed after https://github.com/iTwin/iTwinUI/pull/119)
- Updated icons packages (paths fixed in https://github.com/iTwin/iTwinUI-icons/pull/13)
- Added additional progress indicator sizes (from https://github.com/iTwin/iTwinUI/issues/79)

### DIFF
--- a/src/core/Wizard/Step.test.tsx
+++ b/src/core/Wizard/Step.test.tsx
@@ -318,3 +318,33 @@ describe('Wizard step (workflow)', () => {
     expect(container.querySelector('.iui-wizards-step-track-main')).toBeNull();
   });
 });
+
+it('should set dynamic percentage width based on total steps', () => {
+  const { container: container1 } = render(
+    <Step
+      title='Mock step'
+      index={1}
+      currentStepNumber={0}
+      totalSteps={4}
+      type={'default'}
+    />,
+  );
+
+  const step1 = container1.querySelector('.iui-wizards-step') as HTMLElement;
+  expect(step1).toBeTruthy();
+  expect(step1.style.width).toEqual('25%');
+
+  const { container: container2 } = render(
+    <Step
+      title='Mock step'
+      index={1}
+      currentStepNumber={1}
+      totalSteps={8}
+      type={'default'}
+    />,
+  );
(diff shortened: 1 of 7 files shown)


src/core/Wizard/Step.tsx
```diff
@@ -80,7 +80,12 @@ export const Step = (props: StepProps) => {
       )}
 
       {type !== 'workflow' && (
-        <span className='iui-wizards-step-title'>{title}</span>
+        <span
+          className='iui-wizards-step-title'
+          style={{ width: `${100 / totalSteps}vw` }}
```
bentleyvk, line 85: I think it can cause some bugs because most of the cases wizard won't take the whole screen e.g. page having an expanded sidenav and maybe some container with padding then 100% wizard width != 100vw

veekeys: Yeah, that was the reason I suggested to have % in css (which broke things a little...).

mayank99: Then why didn't we try to fix those broken things? 😄

% is obviously better, but maybe there's a better solution?

mayank99: CSS fixed in https://github.com/iTwin/iTwinUI/pull/119. Using percentage in Step now

src/core/Wizard/Step.test.tsx
```diff
@@ -318,3 +318,84 @@ describe('Wizard step (workflow)', () => {
...
 });
+
+it('should set dynamic inline width based on total steps', () => {
+  const { container: container1 } = render(
+    <Step
+      title='Mock step'
+      index={1}
+      currentStepNumber={0}
+      totalSteps={4}
+      type={'default'}
```
veekeys, line 329: habit with curlies...

veekeys (approved):

Issue #96, opened 2021-05-27
Allow ThemeProvider to support themes in Popup Windows


bsteinbk:
#### Feature
The ThemeProvider should allow the specification of an owningDocument so classes are added to the proper document when the component is used in a Popout window.

#### Examples
    const childWindow = window.open(url, "", `width=800,height=600,menubar=no,resizable=yes,scrollbars=no,status=no,location=no`);
    const reactConnectionDiv = childWindow.document.getElementById("root");
    ReactDOM.render( <ThemeProvider theme="light" owningDocument={childWindow.document}>*** component that contains Slider ***</ThemeProvider>, reactConnectionDiv);

mayank99:
This was closed through #97 and should be released early next week.

PR #97, opened 2021-05-27
Add ability to have themes work properly in popup window.

bsteinbk:
Add ability to have themes work properly in popup window by:

- adding `ThemeOptions` type with an optional `ownerDocument` property
- adding optional `themeOptions` argument to both `useTheme` and `ThemeProvider`

This PR addresses issue https://github.com/iTwin/iTwinUI-react/issues/96.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,8 @@
 
 ### What's new
 
-- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.\*\*
-- **Update `ThemeProvider` component to allow optional owningDocument prop** to allow use in popup browser windows.
-- **Update `useTheme` hook to allow specification of owningDocument** to allow use in popup browser windows.
+- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.
+- **Updated `ThemeProvider` component and `useTheme` hook to allow specification of ownerDocument**. This provides support for theme in popup browser windows.
 
 ## [1.5.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ### What's new
 
-- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.
+- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.\*\*
+- **Update `ThemeProvider` component to allow optional owningDocument prop** to allow use in popup browser windows.
+- **Update `useTheme` hook to allow specification of owningDocument** to allow use in popup browser windows.
 
 ## [1.5.0]
 
--- a/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -48,7 +48,12 @@ it('should set light theme specifying ownerDocument', () => {
     `<!DOCTYPE html><body><p>Test</p></body>`,
     'text/html',
   );
-  render(<ThemeProvider theme='light' ownerDocument={testDocument} />);
+  render(
+    <ThemeProvider
+      theme='light'
+      themeOptions={{ ownerDocument: testDocument }}
+    />,
+  );
   expectLightTheme(testDocument);
 });
 
--- a/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -44,9 +44,12 @@ it('should set light theme', () => {
 });
 
 it('should set light theme specifying ownerDocument', () => {
-  const ownerDocument = new Document();
-  render(<ThemeProvider theme='light' ownerDocument={ownerDocument} />);
-  expectLightTheme(ownerDocument);
+  const testDocument = new DOMParser().parseFromString(
+    `<!DOCTYPE html><body><p>Test</p></body>`,
+    'text/html',
+  );
+  render(<ThemeProvider theme='light' ownerDocument={testDocument} />);
+  expectLightTheme(testDocument);
 });
 
 it('should set dark theme', () => {
--- a/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -7,9 +7,11 @@ import { render } from '@testing-library/react';
 
 import { ThemeProvider } from './ThemeProvider';
 
-const expectLightTheme = () => {
-  expect(document.documentElement.classList).toContain('iui-theme-light');
-  expect(document.documentElement.classList).not.toContain('iui-theme-dark');
+const expectLightTheme = (ownerDocument = document) => {
+  expect(ownerDocument.documentElement.classList).toContain('iui-theme-light');
+  expect(ownerDocument.documentElement.classList).not.toContain(
+    'iui-theme-dark',
+  );
 };
 
 const expectDarkTheme = () => {
@@ -41,9 +43,10 @@ it('should set light theme', () => {
   expectLightTheme();
 });
 
-it('should set light theme specifying owningDocument', () => {
-  render(<ThemeProvider theme='light' owningDocument={document} />);
-  expectLightTheme();
+it('should set light theme specifying ownerDocument', () => {
+  const ownerDocument = new Document();
+  render(<ThemeProvider theme='light' ownerDocument={ownerDocument} />);
+  expectLightTheme(ownerDocument);
 });
 
 it('should set dark theme', () => {
--- a/src/core/ThemeProvider/ThemeProvider.test.tsx
+++ b/src/core/ThemeProvider/ThemeProvider.test.tsx
@@ -41,6 +41,11 @@ it('should set light theme', () => {
   expectLightTheme();
 });
 
+it('should set light theme specifying owningDocument', () => {
+  render(<ThemeProvider theme='light' owningDocument={document} />);
+  expectLightTheme();
+});
+
 it('should set dark theme', () => {
   render(<ThemeProvider theme='dark' />);
   expectDarkTheme();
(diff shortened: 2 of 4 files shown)


src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -14,15 +14,19 @@ export type ThemeProviderProps = {
    * Optional children.
    */
   children?: React.ReactNode;
+  /**
+   * Optional document specification to support popup windows.
+   */
+  owningDocument?: Document;
```
mayank99, line 20: I think this prop should be called `ownerDocument` to match the API (https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument)

src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -14,15 +14,19 @@ export type ThemeProviderProps = {
    * Optional children.
    */
   children?: React.ReactNode;
+  /**
+   * Optional document specification to support popup windows.
+   */
```
mayank99, line 19: The comment could clarify the purpose and specify the default.

```suggestion
   * Document to which the theme will be applied.
   * Can be specified to handle popup windows.
   * @default document
   */
```

src/core/utils/hooks/useTheme.ts
```diff
@@ -12,42 +12,47 @@ export type ThemeType = 'light' | 'dark' | 'os';
  * Defaults to light theme if none provided or set elsewhere.
  * @param theme Light, dark, or based on OS setting.
  */
-export const useTheme = (theme?: ThemeType): void => {
+export const useTheme = (
+  theme?: ThemeType,
+  owningDocument = document,
```
mayank99, line 17: Same deal here, prop can be renamed and JSDoc `@param` can be added after line 13.

src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -41,6 +41,11 @@ it('should set light theme', () => {
   expectLightTheme();
 });
 
+it('should set light theme specifying owningDocument', () => {
+  render(<ThemeProvider theme='light' owningDocument={document} />);
```
mayank99, line 45: I think this test can be improved to actually use a mock document that is not the same as `document`.

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 ### What's new
 
-- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.
+- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.\*\*
+- **Update `ThemeProvider` component to allow optional owningDocument prop** to allow use in popup browser windows.
+- **Update `useTheme` hook to allow specification of owningDocument** to allow use in popup browser windows.
```
mayank99, line 11: These two can be combined.

CHANGELOG.md
```diff
@@ -6,7 +6,9 @@
 
 ### What's new
 
-- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.
+- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.\*\*
```
mayank99, line 9: Typo here.
```suggestion
- **Added number range filter to `Table`.** Use `tableFilters.NumberRangeFilter`.
```

src/core/ThemeProvider/ThemeProvider.tsx
mayank99: Unless there was a reason for this name that I'm not aware of?

src/core/utils/hooks/useTheme.ts
```diff
@@ -12,42 +12,47 @@ export type ThemeType = 'light' | 'dark' | 'os';
...
         } else {
-          addLightTheme();
+          addLightTheme(owningDocument);
         }
         break;
       default:
-        if (!document.documentElement.classList.value.includes('iui-theme')) {
-          addLightTheme();
+        if (
+          !owningDocument.documentElement.classList.value.includes('iui-theme')
```
mayank99, line 42: @veekeys Should we change `value.includes` to `contains` here to avoid conflicts with your PR?

mayank99 (commented):
Thanks for the PR!

Looks mostly ok but I wonder how it would work with our components. We call `useTheme()` (without any params) at the top of every single component. Might need to change something to support popout windows.

CHANGELOG.md
mayank99: Also they should use past tense to be consistent.

```suggestion
- **Added `ownerDocument` prop to `ThemeProvider` and `useTheme`** to allow use in popup browser windows.

File: src/core/ThemeProvider/ThemeProvider.tsx (hunk 1 above)
Comment (bsteinbk) [reply src/core/ThemeProvider/ThemeProvider.tsx]: renamed

File: src/core/ThemeProvider/ThemeProvider.tsx (hunk 2 above)
Comment (bsteinbk) [reply src/core/ThemeProvider/ThemeProvider.tsx]: Made suggested change.

File: src/core/utils/hooks/useTheme.ts (hunk 3 above)
Comment (bsteinbk) [reply src/core/utils/hooks/useTheme.ts]: param added

File: src/core/ThemeProvider/ThemeProvider.test.tsx (hunk 4 above)
Comment (bsteinbk) [reply src/core/ThemeProvider/ThemeProvider.test.tsx]: updated

File: CHANGELOG.md (hunk 5 above)
Comment (bsteinbk) [reply CHANGELOG.md]: updated

File: CHANGELOG.md (hunk 6 above)
Comment (bsteinbk) [reply CHANGELOG.md]: oops... updated

**Comment by bsteinbk** [comment]
@mayank99 Why does every component need the useTheme hook, can't you just require use of ThemeProvider for theme support?

File: src/core/ThemeProvider/ThemeProvider.test.tsx (hunk 4 above)
Comment (mayank99) [reply src/core/ThemeProvider/ThemeProvider.test.tsx]: oh the test is failing now because the mock document doesn't have `body`.

Might need to do something like: 
```ts
const { document } = (new JSDOM(`...`)).window;
```

bsteinbk: yep, just saw that, will fix.

bsteinbk: I could not get JSDOM to work, so I opted for an approach that did not require a new import.

src/core/utils/hooks/useTheme.ts
veekeys: Its fine, Ill merge if needed.

veekeys:
> Thanks for the PR!
> 
> Looks mostly ok but I wonder how it would work with our components. We call `useTheme()` (without any params) at the top of every single component. Might need to change something to support popout windows.

That is a very good point.. 
I think we should pass refs of every component to the useTheme hook to decide, where that component is and apply theming there..
Thats a big change, unfortunately.

I just wonder if we could make a single param for both of the cases in useTheme. We could pass either document or maybe a ref of the component. Remember, I also mentioned, that people might want to apply theming not on document.body, but just on some random div in their application.
Id like to explore here the options, not to make breaking changes later on.

bsteinbk:
@veekeys  - I don't understand everything that is going on in useTheme but based on your comments I could change it to something like the following?

```
export const useTheme = (
  theme?: ThemeType,
  themeHost?: Document | React.RefObject<HTMLElement>,
): void => {
  if (undefined === themeHost) {
    themeHost = document;
  }

  let ownerDocument: Document;
  let hostElement: HTMLElement;

  if (themeHost instanceof Document) {
    ownerDocument = themeHost;
    hostElement = ownerDocument.documentElement;
  } else {
    ownerDocument = themeHost.current?.ownerDocument ?? document;
    hostElement = themeHost.current ?? document.documentElement;
  }

  React.useLayoutEffect(() => {
    if (!ownerDocument.body.classList.contains('iui-body')) {
      ownerDocument.body.classList.add('iui-body');
    }
  }, [ownerDocument]);

  React.useLayoutEffect(() => {
    switch (theme) {
      case 'light':
        addLightTheme(hostElement);
        break;
      case 'dark':
        addDarkTheme(hostElement);
        break;
      case 'os':
        if (window.matchMedia?.('(prefers-color-scheme: dark)').matches) {
          addDarkTheme(hostElement);
        } else {
          addLightTheme(hostElement);
        }
        break;
      default:
        if (!hostElement.classList.value.includes('iui-theme')) {
          addLightTheme(hostElement);
        }
    }
  }, [hostElement, theme]);
};

const addLightTheme = (hostElement: HTMLElement) => {
  hostElement.classList.add('iui-theme-light');
  hostElement.classList.remove('iui-theme-dark');
};

const addDarkTheme = (hostElement: HTMLElement) => {
  hostElement.classList.add('iui-theme-dark');
  hostElement.classList.remove('iui-theme-light');
};

```

src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -14,15 +14,21 @@ export type ThemeProviderProps = {
    * Optional children.
    */
   children?: React.ReactNode;
+  /**
+   * Document to which the theme will be applied.
+   * Can be specified to handle popup windows.
+   * @default document
+   */
+  ownerDocument?: Document;
```
veekeys, line 22: We were thinking about it, safer option might be to have a prop called `themeOptions`, which could be an object with `ownerDocument` property for now. So we have an easy way to add more stuff in future, when needed.
For now that might be enough and we would need to think how we gonna handle this in our components, when needed.

bsteinbk: ThemeOptions implemented,

mayank99: Shouldn't it be in `useTheme` (where `ThemeType` is also defined)?

bsteinbk: Moved

mayank99 (approved):

PR #99, opened 2021-05-27
Replace document calls for popup window support

mayank99:
- [x] `getContainer`: Added `ownerDocument` arg and test
- [x] `Modal`: Added `ownerDocument` prop (passed to getContainer)
- [x] `Popover`: Updated `hideOnEsc` plugin to use popper element instead of `document`
- [x] `Table`: added `ownerDocument` (from table ref) to `FilterToggle`
- [ ] `useTheme`: Hook fixed in #97 but none of the components use that arg currently

<details>
<summary>All references to "document" in project</summary>

image
</details>

### DIFF
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -170,6 +170,8 @@ export const Table = <
 
   useTheme();
 
+  const tableRef = React.useRef<HTMLDivElement>(null);
+
   const defaultColumn = React.useMemo(
     () => ({
       maxWidth: 0,
@@ -318,6 +320,7 @@ export const Table = <
 
   return (
     <div
+      ref={tableRef}
       {...getTableProps({
         className: cx('iui-tables-table', className),
         style,
@@ -349,7 +352,10 @@ export const Table = <
                   <div {...columnProps} key={columnProps.key}>
                     {column.render('Header')}
                     {!isLoading && data.length != 0 && (
-                      <FilterToggle column={column} />
+                      <FilterToggle
+                        column={column}
+                        ownerDocument={tableRef.current?.ownerDocument}
+                      />
                     )}
                     {!isLoading && data.length != 0 && column.canSort && (
                       <div className='iui-sort'>
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -170,7 +170,7 @@ export const Table = <
 
   useTheme();
 
-  const tableRef = React.useRef<HTMLDivElement>(null);
+  const [ownerDocument, setOwnerDocument] = React.useState<Document>();
 
   const defaultColumn = React.useMemo(
     () => ({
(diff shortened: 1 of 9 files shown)


src/core/utils/Popover.tsx
```diff
@@ -85,6 +85,9 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
 
 export const hideOnEsc = {
   fn(instance: Instance) {
+    // will always refer to the correct document instance (e.g. in a popout window)
+    const _document = instance.reference?.ownerDocument ?? document;
```
mayank99, line 89: Maybe we need to add a util like this: https://github.com/mui-org/material-ui/blob/next/packages/material-ui-utils/src/ownerDocument.ts

mayank99:
This is ready for review now. Check PR description.

I don't think we need to update `useTheme` call in all components yet (it could get complicated with refs and stuff, might need a util).

mayank99:
> About table filter, can we get `ownerDocument` of table ref and is that?

Done. But react-table still has some `document` references, which we can't do anything about.

https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useResizeColumns.js#L105
https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useGridLayout.js#L82

src/core/Table/Table.tsx
```diff
@@ -349,7 +352,10 @@ export const Table = <
                   <div {...columnProps} key={columnProps.key}>
                     {column.render('Header')}
                     {!isLoading && data.length != 0 && (
-                      <FilterToggle column={column} />
+                      <FilterToggle
+                        column={column}
+                        ownerDocument={tableRef.current?.ownerDocument}
```
mayank99, line 357: Actually, this won't work because it's passed at the same time as the ref as initialized so it's null

mayank99: Fixed and added test

bentleyvk:
> > About table filter, can we get `ownerDocument` of table ref and is that?
> 
> Done. But react-table still has some `document` references, which we can't do anything about.
> 
> https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useResizeColumns.js#L105
> https://github.com/tannerlinsley/react-table/blob/master/src/plugin-hooks/useGridLayout.js#L82

Yeah but we are not using them right now.

bentleyvk:
Have you tested all these changes with a popup? 😀

mayank99:
> Have you tested all these changes with a popup? 😀

Did now, and it all works! (Took me a while to get linked cra working locally and I had to copy styles from main window into popup)
<details>
<summary>copyStyles</summary>

```ts
function copyStyles(src: Document, dest: Document) {
  Array.from(src.styleSheets).forEach(styleSheet => {
    styleSheet && dest.head.appendChild(styleSheet!.ownerNode!.cloneNode(true))
  })
}
```
</details>

I tested linking a build from main branch as well, to compare!
- [x] `Table` filter dropdown is broken on main (and works here)
- [x] `DropdownMenu` escape key doesn't work on main (and works here)
- [x] `Modal`: Passing `ownerDocument` works
- [x] `ThemeProvider`: Passing `ownerDocument` works

veekeys (approved):

PR #105, opened 2021-06-01
Update issue and PR templates

mayank99:
Reverted #93 because apparently it only works for users with write access.

Also:
- made the code sandbox link easier to copy
- simplified PR template and made it harder to _not_ link an issue

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,5 +1,9 @@
 ---
-about: Create a bug report to help us improve
+name: Bug report 🐞
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
 ---
 
 #### Environment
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -1,14 +1,15 @@
-Thank you for your contribution to the iTwinUi-react project! Please make sure to complete all of the items below before submitting the PR.
+<!--
+Thank you for your contribution to the iTwinUI-react project!
+Please describe your PR here and make sure to complete all of the items below before submitting.
+-->
+
+Closes issue #
 
 ## Checklist
```
bentleyvk, line 8: Maybe add that issue is linked. Just to be more explicit.

mayank99: I don't think it's possible for anyone without write access to link issues using the UI. That's why I added this "Closes" keyword.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,5 +1,9 @@
 ---
+name: Bug report 🐞
```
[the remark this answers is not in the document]
mayank99: I think it only sees the main branch, not sure

veekeys (approved):

PR #107, opened 2021-06-01
Close dropdown menu when tabbing out

mayank99:
<details>
<summary>❌ Approach 1: Add `focusout` event listener in Popover.</summary>

Doesn't work because focusout is called before click, so it closes the menu and immediately reopens it.
  - [x] Tried `stopPropagation` and `stopImmediatePropagation` on both click event and focusout
  - [x] Tried `capture: true` and `capture: false`
  - [x] Tried `setTimeout` (without delay) in focusout. Only works with arbitrary delay (e.g. 100) but it's hacky.
</details>

<details>
<summary>❌ Approach 2: Use `onBlur` on Menu and maintain state for click vs blur.</summary>

Not reliable, and is dependent on content and adds extra state management (which needs to be overridden by `visible` prop too). Got complicated real fast and was still not working perfectly between keyboard and mouse clicks when repeatedly opening/closing.
</details>

✔ **Approach 3**: Listen for "Tab" `keydown` event and hide tippy if none of the children are tabbable.

## Checklist

- [x] ~~Add screenshots of the key elements of the component below~~

### DIFF
--- a/src/core/utils/Popover.tsx
+++ b/src/core/utils/Popover.tsx
@@ -112,18 +112,14 @@ export const hideOnBlur = {
   fn(instance: Instance) {
     return {
       onCreate() {
-        instance.popper.addEventListener(
-          'focusout',
-          (event) => {
-            if (
-              event.relatedTarget && // only hide when focus is on a valid target
-              !instance.popper.contains(event.relatedTarget as Node)
-            ) {
-              instance.hide();
-            }
-          },
-          { capture: true },
-        );
+        instance.popper.addEventListener('focusout', (event) => {
+          if (
+            event.relatedTarget && // only hide when focus is on a valid target
+            !instance.popper.contains(event.relatedTarget as Node)
+          ) {
+            setTimeout(() => instance.hide(), 100);
+          }
+        });
       },
     };
   },
--- a/src/core/utils/Popover.tsx
+++ b/src/core/utils/Popover.tsx
@@ -70,7 +70,12 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
     role: undefined,
     offset: [0, 0],
     ...props,
-    plugins: [lazyLoad, removeTabIndex, hideOnEsc, ...(props.plugins || [])],
+    plugins: [
+      lazyLoad,
+      removeTabIndex,
+      hideOnEscOrTab,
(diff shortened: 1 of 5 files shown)


src/core/utils/Popover.tsx
```diff
@@ -83,11 +88,26 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
...
-        instance.hide();
+      switch (event.key) {
+        case 'Escape':
+          instance.hide();
+          break;
+        case 'Tab':
+          const descendents = Array.from<HTMLElement>(
+            instance.popper.querySelectorAll('*'),
+          );
+          if (!descendents.some((el) => el?.tabIndex >= 0)) {
```
mayank99, line 106: This approach is working pretty well on normal `DropdownMenu`, but not so much on `Select`.

mayank99: Why are we setting tabindex=0 on selected menuitem?

https://github.com/iTwin/iTwinUI-react/blob/df6f9fac901ffa771dca4db642e8cdb611551aac/src/core/Menu/MenuItem.tsx#L96

veekeys: Weird behavior I find is that doesnt matter if you Tab or Shift+Tab you always go to the menu focus...

veekeys: Btw, have you noticed that clicking again on the dropdown does not close it in w3? :D
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

bentleyvk: Doesn't work with table filters. I tabbed out and filter is still visible.

mayank99: Tab or Shift+Tab should both go back to button, I made that intentional (using `preventDefault`).

Table filter needs extra logic to keep track of the last focusable element. So I specifically excluded it.

For Select, try selecting an item, reopening the menu, and tabbing out. It won't work because the selected item gets tabindex=0

mayank99: Fixed Tab / Shift+Tab behavior.

For selected item, I will remove tabindex=0 in next PR

bentleyvk (approved):

PR #112, opened 2021-06-04
Combine classnames in Footer

mayank99:
Noticed `classnames` was completely missing when I was investigating #111. So if a user were to pass a custom class, it would completely replace the `iui-legal-footer` class.

Added a test as well.

### DIFF
--- a/stories/core/Footer.stories.tsx
+++ b/stories/core/Footer.stories.tsx
@@ -9,6 +9,10 @@ import { Footer, FooterProps } from '../../src/core';
 export default {
   title: 'Core/Footer',
   component: Footer,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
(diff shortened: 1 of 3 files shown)


stories/core/Footer.stories.tsx
```diff
@@ -9,6 +9,10 @@ import { Footer, FooterProps } from '../../src/core';
 export default {
   title: 'Core/Footer',
   component: Footer,
+  argTypes: {
```
veekeys, line 12: Didnt you tell me in other PR not to use argTypes...?

mayank99: I said not to use `argTypes.defaultValue` 😄

veekeys: yeah right, double standards..

veekeys (approved):

PR #113, opened 2021-06-07
Moved ownerDocument get to function

veekeys:
After #97 SSR got broken as this line was causing issues (reported by @irenajur):
image

And giving this error on build in server side:
image

Document is not defined in server, so I guess it has some auto checks on code (not in hooks or functions, like variables) for document and it just fails...

Solution: moved our ownerDocument assignment to function so server would not complain, but the outcome should be same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [1.7.0]
+
+`2021-06-07`
+
+### What's new
+
+- **Added `TimePicker` component.**
+
+### Fixes
+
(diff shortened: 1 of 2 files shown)


irenajur:
Converted PR changes to JS and added them to local version of itwinui/react (1.6.1) package. 

image

Portal is now building w/o errors & exported version served locally works.

bentleyvk (approved):

bentleyvk:
Add changelog entry.

mayank99 (approved):

PR #118, opened 2021-06-08
Replace document and window with function getters

veekeys:
After #99 SSR build was broken again. This time with document in FilterToggle.
Decided to replace all document/window with functions, where we check if those are defined (this is [recommended ](https://nextjs.org/docs/migrating/from-create-react-app#safely-accessing-web-apis) to do in SSR applications).

### DIFF
--- a/src/core/Table/filters/FilterToggle.tsx
+++ b/src/core/Table/filters/FilterToggle.tsx
@@ -10,6 +10,7 @@ import { HeaderGroup } from 'react-table';
 import '@itwin/itwinui-css/css/table.css';
 import { useTheme } from '../../utils/hooks/useTheme';
 import { Popover } from '../../utils/Popover';
+import { getDocument } from '../../utils/common';
 
 export type FilterToggleProps<T extends Record<string, unknown>> = {
   column: HeaderGroup<T>;
@@ -22,7 +23,7 @@ export type FilterToggleProps<T extends Record<string, unknown>> = {
 export const FilterToggle = <T extends Record<string, unknown>>(
   props: FilterToggleProps<T>,
 ) => {
-  const { column, ownerDocument = document } = props;
+  const { column, ownerDocument = getDocument() } = props;
 
   useTheme();
 
@@ -50,7 +51,7 @@ export const FilterToggle = <T extends Record<string, unknown>>(
           placement='bottom'
           visible={isVisible}
           onClickOutside={close}
-          appendTo={ownerDocument.body}
+          appendTo={ownerDocument?.body}
         >
           <div
             className={cx('iui-filter', {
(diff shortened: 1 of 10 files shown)


src/core/utils/common.tsx
```diff
@@ -60,3 +60,19 @@ export const getContainer = (containerId: string, ownerDocument = document) => {
   }
   return container;
 };
+
```
bentleyvk, line 63: Maybe change to `getDocument`
image

src/core/utils/hooks/useIntersection.tsx
```diff
@@ -24,7 +25,7 @@ export const useIntersection = (
 
   const setRef = React.useCallback(
     (node: HTMLElement | null) => {
-      if (!window.IntersectionObserver) {
+      if (!getWindow()?.IntersectionObserver) {
```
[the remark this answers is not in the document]
veekeys: Where? 
Would it run with that if above?

src/core/utils/common.tsx
veekeys: added as default parameter value

src/core/utils/common.tsx
```diff
@@ -51,12 +51,31 @@ export const getUserColor = (emailOrName: string) => {
...
     ownerDocument.body.appendChild(container);
   }
   return container;
 };
+
+/**
+ * Get document if it is defined.
+ * Used to support SSR/SSG applications.
+ */
+export const getDocument = () => {
```
mayank99, line 71: Wonder if this should take an optional Element parameter and use that to return `element.ownerDocument`.

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -50,7 +51,7 @@ export const FilterToggle = <T extends Record<string, unknown>>(
           placement='bottom'
           visible={isVisible}
           onClickOutside={close}
-          appendTo={ownerDocument.body}
+          appendTo={ownerDocument?.body}
```
mayank99, line 54: I would have just gone with `() => ownerDocument.body` 😄

src/core/utils/common.tsx
veekeys: I would like not to include any other logic in here..ownerDocument stuff is different.

src/core/Table/filters/FilterToggle.tsx
veekeys: huh? How that would have helped

mayank99: Huh.
image

Well ok 🤷‍♀️

veekeys: Yes, it was in FilterToggle... Initially, wanted to fix only that, but...
Decided to just replace all places not to bother about stuff anymore.

mayank99 (approved):

mayank99 (approved):

mayank99 (approved):

PR #123, opened 2021-06-08
Add `acceptType` prop to FileUploadTemplate

### LINKED ISSUE #122: FileUploadTemplate: Expose "accept" input attribute
#### Feature
`FileUploadTemplate` exposes `multiple` attribute through `acceptMultiple`, adding the `accept` attribute of file would be helpful to guide the user. (Although it is not supported in IE, Chrome and Firefox supports it)

#### Examples
Could probably be a pass through (however it might be confusing with the `acceptMultiple` already there:
`<FileUploadTemplate accept='.dgn' acceptMultiple={false} />`

or

`<FileUploadTemplate acceptType='.dgn,.rvt' acceptMultiple={false} />`

mayank99:
Closes #122 

## Checklist

- [x] ~~Add any additional comments describing the features you've implemented~~


src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -57,6 +63,7 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
         id='file-input'
         onChange={onChange}
         multiple={acceptMultiple}
+        {...inputProps}
```
mayank99, line 66: @veekeys @bentleyvk Need thoughts on this. Mostly done to allow any arbitrary input attributes (e.g. [`capture`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture)) but it does make the `acceptMultiple` prop redundant.

The alternative is:
```suggestion
        accept={acceptType}
```

bentleyvk: Ugh, very hard decision... There is only 4 props related to file upload:
image
Maybe we can just manually have those because I don't know if other props are needed in this case.

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -57,6 +63,7 @@ export const FileUploadTemplate = (props: FileUploadTemplateProps) => {
         id='file-input'
```
bentleyvk, line 63: I just saw that we are setting `id`, this can cause errors if 2 `FileUploadTemplate` components are present. We need to have a prop for this.

mayank99: `inputProps` would fix this too 😄

mayank99: I will remove `id` and `for` after https://github.com/iTwin/iTwinUI/issues/146 is available.

But I think we would still need to support adding `id`. The user might have a use case for it 🤔

src/core/FileUpload/FileUploadTemplate.tsx
veekeys: If its really only 4 props, maybe no need to go complicated way...
And at the end, i guess its not a bad deal to end up with a lot of props as long as they make sense and guarantee quality (implementation from our side).

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -42,25 +47,26 @@ export type FileUploadTemplateProps = {
...
-      <input
-        className='iui-browse-input'
-        type='file'
-        id='file-input'
-        onChange={onChange}
-        multiple={acceptMultiple}
-      />
       <SvgUpload className='iui-icon' aria-hidden />
       <div className='iui-template-text'>
-        <label htmlFor='file-input' className='iui-anchor'>
```
veekeys, line 63: hmm... Wasnt there a reason that input was used separately from label and connected with htmlFor?
Somehow I feel there was something, that we needed to do it only that way...

src/core/FileUpload/FileUploadTemplate.tsx
```diff
@@ -15,9 +15,14 @@ export type FileUploadTemplateProps = {
   /**
    * Whether the file input accepts multiple files.
    * Passed to the `multiple` attribute of native file input.
-   * @deprecated Use `inputProps` instead
+   * @default true
```
veekeys, line 18: maybe leave it as false..

veekeys: oh wait, it was true by default.. I just checked the last commit, so thought its added as new one...

mayank99: "leave"? It was `true` before this PR.

I had removed the default value because of the deprecation so I'm just reverting that

src/core/FileUpload/FileUploadTemplate.tsx
veekeys: well ok, seems like its still working, so.. i guess its ok...

veekeys (approved):

bentleyvk (approved):

Issue #126, opened 2021-06-09
Table: Highlighting data to complement search features


MichaelBelousov:
#### Feature
Add the ability in tables to specify a "highlighted subsection" string/regex, which can complement searches that filter the table rows, by drawing user attention to the part of the filtered columns that matches on their search.

This is especially useful when it runs as you type.
When a user only remembers keywords from a row they are search for, getting visual feedback of what projects match their keyword help them remember additional keywords they can use to narrow down their search query quickly. 

#### Examples
[https://review.itwin.bentley.com]([https://review.itwin.bentley.com) already does this:

![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/16315257/121377305-3a515300-c910-11eb-9e57-14b28a7f0b76.png)

You can see in their use case, they are passing a case-ignoring regex, so the `Ke` matches rows that have a lowercase `k`, and those are also highlighted.

MichaelBelousov:
I plan on (eventually) submitting a PR for this myself

mayank99:
Maybe we'll see this in 2022 👀 https://www.w3.org/TR/css-highlight-api-1/

mayank99:
Looks like the CSS Highlight API is available in Chrome/Safari now, so this can be implemented in userland. Here's a demo: https://microsoftedge.github.io/Demos/custom-highlight-api/

PR #129, opened 2021-06-11
Table highlight specifier

MichaelBelousov:
Closes issue #126 

Please advise on how to add styles (particularly the original implementation used a color known as bwc-yellow, which I placed the hexcode here but I couldn't see how was best do that here).

## Screenshot of new table story:

image

The story includes a function for compiling "simple search query syntax" to regular expressions, where a "simple search query"
is any text with `*` being a wild card that matches any amount of non-space characters, and all matches are caseless. See the screenshot for an example

## Checklist

  - I'm not sure I covered all lines, I'll go back and add some more
  - [ ] I forgot to change the text saying "enter a regular expression" to say "enter a *-wildcard search string"
  - if I need to add it to the repo, then I haven't yet


package.json
```diff
@@ -47,6 +47,7 @@
     "@tippyjs/react": "^4.2.5",
     "@types/react-table": "^7.0.18",
     "classnames": "^2.2.6",
+    "react-string-replace": "^0.4.4",
```
mayank99, line 50: We try to avoid external dependencies unless they are:
1. reliable, stable, and popular
2. necessary

react-string-replace is pre-1.0 ("unstable") and has less than 400 stars (not very "popular"). More importantly, it requires lodash, which is a huge dependency.

stories/core/Table.stories.tsx
```diff
@@ -718,3 +722,158 @@ export const NoData: Story<TableProps> = ({ columns, data, ...rest }) => {
...
+      // convert wildcard character to valid regex repeated non-space text definition
+      const queryWithWildcards = escapedQuery.replace(/\*/g, '\\S*');
+      return RegExp(queryWithWildcards, 'gi');
+    };
+    return compileSimpleSearchQuery(input);
+  }, [input]);
+
+  const defaultColumn: Partial<Column> = React.useMemo(
+    () => ({
+      Cell: HighlightingCell,
```
mayank99, line 782: If the user has to use a custom cell, then I wonder what is the point of adding a new prop to `Table`.

It should be fairly trivial for the user to handle the highlighting on their end (since `HighlightingCell` just adds `<mark>` with a color), without needing any changes to the Table.

mayank99 (commented):
Thanks for the PR.

I was gonna leave some more comments but I'm afraid the external dependency will be a stopper.

package.json
MichaelBelousov: That's totally fine, I did not see the transitive lodash dependency and was mostly porting this implementation from another app that uses React table.
I will draft the PR until I remove the dependency.

stories/core/Table.stories.tsx
MichaelBelousov: perhaps the "feature" I'm adding is just prop clutter and maybe downgrading this PR to just a story showing how to make such a column is better?

mayank99: Yeah I think that would make more sense. But can you confirm that it works without needing any changes to Table? Re-rendering memoized rows is the only thing I can think of that might not work.

MichaelBelousov: Maybe still providing/exporting the `HighlightingCell` is good. I'm new to react-table@7 but I'll look into how else to pass the regex state to the cell without a prop on the table

MichaelBelousov: yeah this would only work if I made the cell render a stateful component that re-rendered itself through some hacks. Perhaps you there should be some `cellDependencies` prop?

bentleyvk: I am still investigating what can we do about re-renders. From the first sight it seems that it isn't working for you without any changes because you are using `defaultColumn` prop and we are not checking it for row re-renders. I will write a comment later what I was able to find.

MichaelBelousov: defaultColumn probably should not be changing and should never cause re-renders, it is memoized here outside of the `Table` component to prevent that. But the Cell render method does need access to the changing `highlightRegex` probably as a prop. I think it would be difficult to memoize anything if we let defaultColumn change every time `highlightRegex` changes.

bentleyvk: Hey @MichaelBelousov, sorry for such a late response. We currently don't know what is the best way to handle table changes that are causing re-rendering. Here is a codesandbox with a working cell highlight and it doesn't require any changes to the Table: https://codesandbox.io/s/itwinui-react-minimal-example-forked-5qv8k

MichaelBelousov: the `tableColumns` not being memoized make me worried a bit about how it would perform compared to the prop-aware table implementation. When I have some time I think I will just turn this PR into removing any changes I made to the table and making a story using the changes you made in that sandbox. If people have actual performance issues using it then we can look into it again

MichaelBelousov:
closing, will eventually hopefully open up a separate PR with a more general solution for custom cell props or just a story using the technique in the attached code sandbox

PR #132, opened 2021-06-11
Add `size` prop to Input and Select

### LINKED ISSUE #131: Add additional sizes that came with CSS changes.

mayank99:
Added `size` prop to:
- Input (overriding native [size](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefsize) attribute)
- Select (doesn't change menu item size for now)

Added new story and test image for:
- Input

Enabled prop (but didn't add separate stories) for:
- LabeledInput
- Select and LabeledSelect

Closes #131

## Screenshots
image
image

image
image

## Checklist


src/core/Select/Select.test.tsx
```diff
@@ -336,3 +336,12 @@ it('should use custom renderer for menu items', () => {
   expect(menuItems[1].style.color).toEqual('green');
   expect(menuItems[2].style.color).toEqual('red');
 });
+
+it('should render small and large sizes', () => {
```
bentleyvk, line 340: Jest supports iterating trough test params with `it.each`, you can look at line 191 in this file.

stories/core/LabeledInput.stories.tsx
```diff
@@ -20,6 +20,9 @@ export default {
     disabled: { type: 'boolean' },
     required: { type: 'boolean' },
   },
+  args: {
+    size: undefined,
```
bentleyvk, line 24: Why it is needed?

src/core/Select/Select.test.tsx
mayank99: Oh that's cool

stories/core/LabeledInput.stories.tsx
mayank99: To explicitly enable the prop. Otherwise it doesn't show up (since it's inherited from `Input`).

src/core/Input/Input.test.tsx
```diff
@@ -44,3 +44,12 @@ it('should take class and style', () => {
   expect(input).toBeTruthy();
   expect(input.style.width).toBe('50px');
 });
+
+it('should render small and large sizes', () => {
```
bentleyvk, line 48: Can you use `it.each` here too?

stories/core/LabeledInput.stories.tsx
mayank99: Restarted storybook and it works without args. So removed

src/core/Input/Input.test.tsx
mayank99: Done

bentleyvk (approved):

PR #135, opened 2021-06-11
Update table classes from itwinui-css 0.20

mayank99:
- Bumped itwinui-css to 0.20.2
- Updated Table class names (from iTwin/iTwinUI#80)
- Wrapped FilterToggle in borderless IconButton (and added StylingProps)
- Updated tests to pass

## Checklist


src/core/Table/filters/FilterToggle.tsx
```diff
@@ -53,21 +55,18 @@ export const FilterToggle = <T extends Record<string, unknown>>(
           onClickOutside={close}
           appendTo={ownerDocument?.body}
         >
-          <div
-            className={cx('iui-filter', {
-              'iui-active': isVisible || column.filterValue,
-            })}
+          <IconButton
```
mayank99, line 58: After going through all stories, this is the only problem I've noticed so far:
image

src/core/Table/Table.tsx
```diff
@@ -322,31 +322,27 @@ export const Table = <
...
           return (
             <div {...headerGroupProps} key={headerGroupProps.key}>
               {headerGroup.headers.map((column) => {
                 const columnProps = column.getHeaderProps({
                   ...column.getSortByToggleProps(),
                   className: cx(
-                    'iui-tables-cell',
-                    'iui-tables-head',
-                    { 'iui-active-sort': column.isSorted },
+                    'iui-cell',
```
bentleyvk, line 341: I see that header cell always has `cursor: pointer` if it was always sortable but that is not always true.
image
Created an issue: https://github.com/iTwin/iTwinUI/issues/151

src/core/Table/Table.tsx
```diff
@@ -322,31 +322,27 @@ export const Table = <
...
+        className: cx('iui-table', className),
         style,
       })}
       {...ariaDataAttributes}
     >
       <div>
         {headerGroups.slice(1).map((headerGroup: HeaderGroup<T>) => {
           const headerGroupProps = headerGroup.getHeaderGroupProps({
-            className: 'iui-tables-row',
+            className: cx('iui-table-header', 'iui-row'),
```
bentleyvk, line 333: In HTML example this is the structure:
image
Row should be a new element.

src/core/Table/filters/FilterToggle.tsx
bentleyvk: So `padding-right` might actually be needed 😀

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -11,19 +11,21 @@ import '@itwin/itwinui-css/css/table.css';
...
 import { Popover } from '../../utils/Popover';
 import { getDocument } from '../../utils/common';
+import { IconButton } from '../../Buttons';
+import { StylingProps } from '../../utils/props';
 
 export type FilterToggleProps<T extends Record<string, unknown>> = {
   column: HeaderGroup<T>;
   ownerDocument?: Document;
-};
+} & StylingProps;
```
bentleyvk, line 20: Is this needed as we don't expose `FilterToggle` in `index.ts`?

src/core/Table/Table.tsx
```diff
@@ -375,13 +375,13 @@ export const Table = <
           );
         })}
       </div>
-      <div {...getTableBodyProps({ className: 'iui-tables-body' })}>
+      <div {...getTableBodyProps({ className: 'iui-table-body' })}>
         {data.length !== 0 &&
```
bentleyvk, line 379: Can you set `min-width: undefined` on line 178 so it would take `min-width` from CSS?
image

src/core/Table/filters/FilterToggle.tsx
mayank99: I added it just in case there is some need for it (internally or externally).

My first use case was to add padding-right, but I think I will do that in css

src/core/Table/Table.tsx
mayank99: and what about line 205?

mayank99: and what about maxwidth and width?

src/core/Table/Table.tsx
mayank99: Wow, I'm surprised it even worked before

mayank99: Oh it does nothing 😄 
image

because react-table sets flex automatically
image

src/core/Table/filters/FilterToggle.tsx
mayank99: Fixed but now I notice another problem :/

image

src/core/Table/Table.tsx
mayank99: Fixed

src/core/Table/Table.tsx
```diff
@@ -174,9 +174,9 @@ export const Table = <
 
   const defaultColumn = React.useMemo(
     () => ({
-      maxWidth: 0,
-      minWidth: 0,
-      width: 0,
+      maxWidth: undefined,
```
bentleyvk, line 177: Maybe add a comment why we did this because we will forget 😀

src/core/Table/filters/FilterToggle.tsx
bentleyvk: Can you add `title={undefined}` to line 347? Because `react-table` does this:
image
About sort going outside, I don't know what to do 😕 I see we had this problem previously, we either can get width using JS and apply it to all that column's cells or we can leave this to handle by the user if he wants to set fixed width. I would go with the second one.

src/core/Table/Table.tsx
mayank99: Done

bentleyvk (approved):

Issue #141, opened 2021-06-16
Table filter triggers onSelect


klopez213:
#### Environment
- __Package version(s)__: 1.8.0
- __Operating System__: Windows
- __Browser name and version__: Chrome 91.0.4472.101

#### Code Sandbox
Link to a minimal repro: https://codesandbox.io/s/table-filter-triggers-onselect-zjwk5

#### Steps to reproduce
1. Filter any of the columns.

#### Actual behavior
onSelect is triggered when filtering.  Please see Code Sandbox sample.

#### Expected behavior
onSelect shouldn't be triggered when filtering.

mayank99:
Fixed in 1.8.1 🎉

PR #143, opened 2021-06-17
Refactored table select

bentleyvk:
Closes issue #141 

## Checklist

Manually handling selected rows as previously sorting or filtering would trigger `onSelect` callback because I think `selectedFlatRows` reference changes every time and triggers `onSelect`.

### DIFF
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -16,7 +16,6 @@ import {
   TableState,
   useFlexLayout,
   useFilters,
-  useMountedLayoutEffect,
   useRowSelect,
   useSortBy,
   useTable,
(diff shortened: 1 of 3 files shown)


src/core/Table/Table.tsx
```diff
@@ -259,6 +274,12 @@ export const Table = <
       case TableActions.setFilter:
         onFilterHandler(newState, action, previousState, instance);
         break;
+      case TableActions.toggleRowSelected:
```
veekeys, line 277: tests would be great..

bentleyvk: Added to check that this bug is not reproducible.

veekeys (approved):

src/core/Table/Table.tsx
```diff
@@ -259,6 +274,12 @@ export const Table = <
...
+      case TableActions.toggleAllRowsSelected:
+      case TableActions.toggleAllPageRowsSelected: {
+        onSelectHandler(newState, instance);
```
mayank99, line 280: Have you thought about how this would work for controlled selections?

bentleyvk: This is just a callback when user selects something. 
I have started investigating `useControledState` prop and it still works - it allows to control what rows should be selected. Also, initial state also works.

bentleyvk: table-controlled-state

mayank99: Ok great! Just wanted to make sure it doesn't need additional refactoring when we add those in the future 😄

mayank99 (approved):

PR #149, opened 2021-06-18
Add nested menus

### LINKED ISSUE #29: Dropdown: multi-level support

mayank99:
Added new `subMenuItems` prop to `MenuItem` to allow nesting menus.

Submenu is shown when parent menu item is hovered or using arrow-right.
Submenu hides when focus returns to parent menu item (accessible from context) using arrow-left or mouseleave.

image

## Checklist


src/core/Menu/Menu.tsx
```diff
@@ -91,7 +109,35 @@ export const Menu = (props: MenuProps) => {
...
     >
-      {children}
+      {React.Children.map(children, (item) => {
+        if (React.isValidElement(item)) {
+          const subItems = (item.props as MenuItemProps)?.subMenuItems ?? [];
+          if (subItems.length === 0) {
+            return item;
+          } else {
+            return (
+              <Popover
```
bentleyvk, line 119: Why handling submenu here and not in `MenuItem`?

src/core/Menu/Menu.tsx
```diff
@@ -91,7 +109,35 @@ export const Menu = (props: MenuProps) => {
       ref={menuRef}
       {...rest}
     >
-      {children}
+      {React.Children.map(children, (item) => {
+        if (React.isValidElement(item)) {
```
bentleyvk, line 113: It is recommended to fail fast as it allows to avoid deep if conditions. Here better write `if (!React.isValidElement(item)) return item;`

src/core/Menu/Menu.tsx
mayank99: I've tried many different things, but I'm still struggling with focus

src/core/Menu/MenuItem.tsx
```diff
@@ -40,10 +44,18 @@ export type MenuItemProps = {
...
+   */
+  subMenuItems?: JSX.Element[];
   /**
    * Content of the menu item.
    */
   children?: React.ReactNode;
+  /**
+   * Parent `MenuItem` reference. Used in sub-menus.
+   */
+  parentMenuItemRef?: React.RefObject<HTMLLIElement>;
```
bentleyvk, line 58: I don't like this prop myself as it is exposed to the users. I have seen other libs are using `Context` for these kind of stuff so it is not visible for the user.

veekeys: After seeing Context in action in Slider PR (there it was not really needed), I am thinking if we could check more and maybe apply here? Exactly for the same reason to not to show internal stuff to the user...
@mayank99  thoughts?

mayank99: I was actually looking at context myself yesterday. Should have something working by the end of the day

mayank99: I've been playing around for a few hours and can't seem to do it properly yet.

I think I will need to move the submenu logic back into `Menu` and share the state there in the context.

veekeys: I ran context implementation and it seemed to work same as before.. There is something I havent noticed?
But thinking from logical perspective, having a context of the menu might help us solve issues when there is  multiple sub menus opened...

mayank99: What I have pushed works the same as before, so it also has the same problems: opening multiple submenus, and the hack for closing.

I was trying to improve it by moving the logic and context from MenuItem to Menu, but it got really messy and was not working properly. ☹️

veekeys: hmmm..
I think this looks better than the prop.
I would have guessed keeping context on whole menu should work better, but if you say its not.. idk. we could either spend some time on it together to see, or we could go forward with this approach and check up later (should not be breaking change as its inner handling)

mayank99: > What I have pushed works the same as before, so it also has the same problems: opening multiple submenus, and the hack for closing.

I was able to fix both of these by adding event listeners to the a new wrapper div around sub menu.

src/core/Menu/MenuItem.tsx
```diff
@@ -119,12 +171,42 @@ export const MenuItem = React.forwardRef<HTMLLIElement, MenuItemProps>(
...
+      listItem
+    ) : (
+      <MenuItemContext.Provider value={{ ref: menuItemRef }}>
+        <Popover
+          placement='right-start'
+          visible={isSubmenuVisible}
+          content={
+            <div
+              onMouseLeave={() => setIsSubmenuVisible(false)}
+              onKeyDown={(e) => {
```
veekeys, line 194: What for this was added? Now this one is being called together with the left key down handler above (where we focus and that should close menu anyway...).
Also, while playing around, still managed to get this one somehow
image

veekeys: I have fixed double opened sub menus with the onBlur event addition on a new div..
Now the only issue left is that the Menu component losses the correct `focusedIndex` when you start playing around with mouse + keyboard combo...
But honestly, I am a little lost, how that should even work...

veekeys: I was trying to look for any examples how keyboard nav should work here, but.. Not much I could find.
There was this:
https://codepen.io/vj/pen/rOgPEY
Which works using tabbing (might be easier, but I am not sure if it is better...).
Also w3 has some article about it, but not much helpful in our case though.
https://www.w3.org/WAI/tutorials/menus/flyout/
And noticed SmashingMagazine article and this attribute `aria-haspopup` which we do not use for menu for some reason...
https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

mayank99: Removed onKeyDown as it's not needed now that you're closing submenu in onBlur.

> Now the only issue left is that the Menu component losses the correct `focusedIndex` when you start playing around with mouse + keyboard combo... But honestly, I am a little lost, how that should even work...

That's fine, I don't think we need to make it perfect for v1. Mouse + keyboard is a weird use case anyway.

> attribute `aria-haspopup` which we do not use for menu for some reason...

Added.

veekeys (approved):

PR #158, opened 2021-06-21
Add `onRowClick` handler to improve `Table` selection

### LINKED ISSUE #155: Table: Single Select or onClick event handler for rows
#### Feature
A single select table or an onClick handler for rows.

It may be technically possible to implement a single select table using onSelect (I am still in the process of figuring it out), but it's very hacky.

#### Examples
I am currently working on a demo showcase where on selecting/clicking a row, it will zoom into that particular part of an iModel. I would need an event handler for clicking on rows to do this. A single select table would be ok but not ideal in this situation, but either would work for my use case. There is a key value that is stored within the row clicked that is used to zoom in on the element within the viewer, so at the very least I need the data from the clicked row.

mayank99:
Closes #155 

## Checklist

### DIFF
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -63,6 +63,10 @@ export type TableProps<
     tableState?: TableState<T>,
   ) => void;
   /**
+   * Handler for when a row is clicked.
+   */
+  onRowClick?: (row: Row<T>) => void;
+  /**
    * Flag whether table columns can be sortable.
    * @default false
    */
@@ -156,6 +160,7 @@ export const Table = <
     id,
     isSelectable = false,
     onSelect,
+    onRowClick,
     isSortable = false,
     onSort,
     stateReducer,
@@ -261,6 +266,13 @@ export const Table = <
     });
     onSelect?.(selectedData, newState);
   };
+
+  const onRowClickHandler = React.useCallback(
+    (row) => {
+      onRowClick?.(row);
+    },
+    [onRowClick],
+  );
 
   const tableStateReducer = (
     newState: TableState<T>,
@@ -407,6 +419,7 @@ export const Table = <
                 intersectionMargin={intersectionMargin}
                 state={state}
                 key={rowProps.key}
+                onClick={onRowClickHandler}
               />
             );
           })}
--- a/src/core/Table/TableRowMemoized.tsx
+++ b/src/core/Table/TableRowMemoized.tsx
@@ -9,7 +9,7 @@ import { useIntersection } from '../utils/hooks/useIntersection';
 import { getCellStyle } from './utils';
 
 /**
- * Memorization is needed to avoid unnecessary re-renders of all rows when additional data is added when lazy-loading.
+ * Memoization is needed to avoid unnecessary re-renders of all rows when additional data is added when lazy-loading.
  * Using `isLast` here instead of passing data length to avoid re-renders of all rows when more data is added. Now only the last row re-renders.
  * Although state is not used it is needed for `React.memo` to check state that changes row state e.g. selection.
  * When adding new features check whether it changes state that affects row. If it does then add equality check to `React.memo`.
@@ -22,6 +22,7 @@ const TableRow = <T extends Record<string, unknown>>(props: {
   onBottomReached: React.MutableRefObject<(() => void) | undefined>;
   intersectionMargin: number;
   state: TableState<T>; // Needed for explicitly checking selection changes
+  onClick?: (row: Row<T>) => void;
 }) => {
   const {
     row,
@@ -30,6 +31,7 @@ const TableRow = <T extends Record<string, unknown>>(props: {
     onRowInViewport,
     onBottomReached,
     intersectionMargin,
+    onClick,
   } = props;
 
   const onIntersect = React.useCallback(() => {
@@ -42,7 +44,14 @@ const TableRow = <T extends Record<string, unknown>>(props: {
   });
 
   return (
-    <div {...rowProps} key={rowProps.key} ref={rowRef}>
+    <div
+      {...rowProps}
+      key={rowProps.key}
+      ref={rowRef}
+      onClick={() => {
+        onClick?.(row);
+      }}
+    >
       {row.cells.map((cell) => {
         const cellProps = cell.getCellProps({
           className: cx('iui-cell', cell.column.cellClassName),
--- a/stories/core/Table.stories.tsx
+++ b/stories/core/Table.stories.tsx
@@ -235,6 +235,89 @@ export const Selectable: Story<TableProps> = (args) => {
 
 Selectable.args = { isSelectable: true };
 
+export const SelectOnRowClick: Story<TableProps> = (args) => {
+  const onClickHandler = (
+    event: React.MouseEvent,
+    props: CellProps<{ name: string; description: string }>,
+  ) => {
+    event.stopPropagation();
+    action(props.row.original.name)();
+  };
+
+  const tableColumns = useMemo(
+    () => [
+      {
+        Header: 'Table',
+        columns: [
+          {
+            id: 'name',
+            Header: 'Name',
+            accessor: 'name',
+          },
+          {
+            id: 'description',
+            Header: 'Description',
+            accessor: 'description',
+            maxWidth: 200,
+          },
+          {
+            id: 'click-me',
+            Header: 'Action',
+            width: 100,
+            Cell: (props: CellProps<{ name: string; description: string }>) => {
+              const onClick = (e: React.MouseEvent) => onClickHandler(e, props);
+              return (
+                <span
+                  className='iui-anchor'
+                  onClick={onClick}
+                  style={{
+                    width: '100%',
+                    height: '100%',
+                    display: 'flex',
+                    alignItems: 'center',
+                  }}
+                >
+                  Click me!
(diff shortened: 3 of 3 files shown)


src/core/Table/TableRowMemoized.tsx
```diff
@@ -30,6 +31,7 @@ const TableRow = <T extends Record<string, unknown>>(props: {
     onRowInViewport,
     onBottomReached,
     intersectionMargin,
+    onClick,
```
bentleyvk, line 34: Need to add to `memo` deps check.

src/core/Table/Table.tsx
```diff
@@ -262,6 +267,13 @@ export const Table = <
     onSelect?.(selectedData, newState);
   };
 
+  const onRowClickHandler = React.useCallback(
```
bentleyvk, line 270: It doesn't give you anything. If original `onRowClick` is not memoized then this one will also return new function reference every time.

veekeys: Maybe add comment above onRowClick with a recommendation to memo it?

mayank99: I was thinking of adding selection stuff here, instead of story. Because it is part of standards.

src/core/Table/TableRowMemoized.tsx
mayank99: Added

src/core/Table/Table.tsx
mayank99: Ahhh I can't get that to work without maintaining my own state, so I've removed it.

mayank99: Added back with some basic row selection

src/core/Table/Table.test.tsx
```diff
@@ -205,6 +205,37 @@ it('should handle checkbox clicks', () => {
...
+    onRowClick,
+  });
+
+  expect(screen.queryByText('Header name')).toBeFalsy();
+  const rows = container.querySelectorAll('.iui-table-body .iui-row');
+  expect(rows.length).toBe(3);
+
+  fireEvent.click(getByText(mockedData()[1].name));
+  expect(rows[1].classList).toContain('iui-selected');
+  // expect(onSelect).toHaveBeenCalledTimes(1);
```
mayank99, line 223: This does not currently work as intended. It's called 2 times because I'm also using `instance.toggleAllRowsSelected()` to deselect all rows first. I don't think the user would expect this to be called twice when a row is only clicked once ☹️ 

I wonder if there's a way to "collect" all the events in the reducer, before firing `onSelect`.

bentleyvk: You could dispatch custom action:
image
image
Need to think a better name and how to handle if conditions.

`dispatch` taken from here:
image

stories/core/Table.stories.tsx
```diff
@@ -210,11 +207,23 @@ export const Selectable: Story<TableProps> = (args) => {
...
+                    display: 'flex',
+                    width: '100%',
+                    height: '100%',
+                    alignItems: 'center',
+                  }}
+                  onClick={(e) => e.stopPropagation()} // prevent row selection when this cell is clicked
+                >
+                  <a
+                    className='iui-anchor'
+                    onClick={() => action(props.row.original.name)()}
```
bentleyvk, line 222: You can write like this:
```suggestion
                    onClick={action(props.row.original.name)}
```

stories/core/Table.stories.tsx
```diff
@@ -210,11 +207,23 @@ export const Selectable: Story<TableProps> = (args) => {
             Header: 'Click',
             width: 100,
             Cell: (props: CellProps<{ name: string; description: string }>) => {
-              const onClick = () => onClickHandler(props);
               return (
-                <a className='iui-anchor' onClick={onClick}>
-                  Click me!
-                </a>
+                <span
```
bentleyvk, line 211: No need for the wrapper. You can add `e.stopPropagation()` in anchor `onClick` handler.

stories/core/Table.stories.tsx
mayank99: Is it meant to be written like that? 

https://jakearchibald.com/2021/function-callback-risks/

stories/core/Table.stories.tsx
mayank99: The wrapper is to show that it takes up full space of the cell

stories/core/Table.stories.tsx
bentleyvk: `action(...)` returns a function that accepts any args so I guess it is not bad doing that 🤷‍♂️

stories/core/Table.stories.tsx
bentleyvk: But it seems that you need to wrap with some HTML element that has `e.stopPropagation()` ( at least I did until I tried to remove it).

mayank99: I could change it, but then clicking anywhere outside the link in the cell will trigger row selection

bentleyvk: I guess it is ok that you don't click on the link and it triggers select.

stories/core/Table.stories.tsx
```diff
@@ -213,9 +210,14 @@ export const Selectable: Story<TableProps> = (args) => {
             Header: 'Click',
             width: 100,
             Cell: (props: CellProps<{ name: string; description: string }>) => {
-              const onClick = () => onClickHandler(props);
               return (
-                <a className='iui-anchor' onClick={onClick}>
+                <a
+                  className='iui-anchor'
+                  onClick={(e) => {
+                    e.stopPropagation();
```
bentleyvk, line 217: Maybe add a comment why it is needed.

mayank99: Added

src/core/Table/Table.test.tsx
mayank99: Added, but not sure about putting it in `TableActions` record.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -48,7 +50,14 @@ const TableRow = <T extends Record<string, unknown>>(props: {
 
   return (
     <>
-      <div {...rowProps} ref={rowRef}>
+      <div
+        {...rowProps}
+        key={rowProps.key}
```
bentleyvk, line 55: Is it needed here as key is added in Table?

src/core/Table/Table.tsx
```diff
@@ -416,6 +439,24 @@ export const Table = <
...
   const areFiltersSet = allColumns.some((column) => !!column.filterValue);
 
+  const onRowClickHandler = React.useCallback(
+    (event: React.MouseEvent, row: Row<T>) => {
+      if (isSelectable) {
+        if (!row.isSelected && !event.ctrlKey) {
+          dispatch({
+            type: TableActions.singleRowSelected,
+            id: row.id,
+            value: true,
```
bentleyvk, line 449: Started to think that `value` might not be needed as it is always true.

src/core/Table/Table.tsx
```diff
@@ -37,6 +37,8 @@ import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
 import { FilterToggle, TableFilterValue } from './filters';
 import { customFilterFunctions } from './filters/customFilterFunctions';
 
+TableActions.singleRowSelected = 'singleRowSelected';
```
bentleyvk, line 40: Better don't mutate `react-table` object. Maybe have constant value instead.

stories/core/Table.stories.tsx
```diff
@@ -139,9 +139,6 @@ export const Basic: Story<TableProps> = (args) => {
 
 export const Selectable: Story<TableProps> = (args) => {
   const { columns, data, ...rest } = args;
-  const onClickHandler = (
-    props: CellProps<{ name: string; description: string }>,
-  ) => action(props.row.original.name)();
```
bentleyvk, line 142: Maybe add `onRowClick` instead. Also, write above the table that with `Ctrl` you can select multiple rows.

src/core/Table/Table.tsx
mayank99: Changed but I don't know whether it should be declared up here.

src/core/Table/Table.tsx
mayank99: Removed

src/core/Table/TableRowMemoized.tsx
mayank99: `rowProps.key` was explicitly destructed before it was removed in #50, so I guess it was just a merge (This PR was created before #50 was merged).

Removed.

stories/core/Table.stories.tsx
mayank99: Is `onRowClick` really necessary? This is the main story for `Selectable` Table so `onSelect` makes more sense here.

As for the `Ctrl` part, I am hesitant to advertise that explicitly because we don't support `Shift` yet.

bentleyvk: It is a new prop/functionality so we need to show it somewhere.

mayank99: What if I leave `onSelect` and also add `onRowClick`?

bentleyvk (approved):

PR #162, opened 2021-06-24
Add controlled `visible` prop to `Select`

mayank99:
Updated internal `isOpen` state to use `visible` (from `popoverProps`) when available.

## Checklist


bentleyvk:
Why this one is needed now?

src/core/Select/Select.tsx
```diff
@@ -175,7 +172,11 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
 
   useTheme();
 
-  const [isOpen, setIsOpen] = React.useState(false);
+  const [isOpen, setIsOpen] = React.useState(popoverProps?.visible ?? false);
+  React.useEffect(() => {
+    setIsOpen((open) => popoverProps?.visible ?? open);
+  }, [popoverProps]);
```
veekeys, line 178: Maybe `popoverProps?.visible` would work in dep list?

mayank99: That's what I had initially, but the unit test was failing 😕 I don't think it was updating

bentleyvk (approved):

PR #163, opened 2021-06-24
Update storybook to 6.3

mayank99:
- Ran `npx sb@latest upgrade` to bump sb packages from 6.1.21 to 6.3.4.
- Added `@babel/plugin-proposal-class-properties` to main.js with `loose: true` to fix build error.
  - https://github.com/storybookjs/storybook/issues/14197
  - https://github.com/storybookjs/storybook/issues/14805
- Updated generated arrays syntax to use `Array(N).fill()` because stories were not being rendered.
- Removed all references to `argType.defaultValue` (deprecated).
- Cleaned up `UserIcon` stories.
- Updated images.


bentleyvk:
As I understand it shows undefined props with these buttons:
image

veekeys:
What I noticed (but I guess not critical..).
-- DropdownMenu: default value for role is not shown (it was shown before)
image
For tabs also:
image
Or in Modal its taken as undefined, but actually they have values set...
image
image

stories/core/Select.stories.tsx
```diff
@@ -195,23 +195,20 @@ DisabledWithSelectedValue.args = {
...
-          [...new Array(20)].map((_, index) => ({
-            label: `Item #${index}`,
-            value: index,
-          }))
-        }
         value={value}
         onChange={setValue}
         placeholder={placeholder}
         {...rest}
+        options={[...Array(20).fill(null)].map((_, index) => ({
```
veekeys, line 208: Why needed to move and set Partial props too?

stories/core/Table.stories.tsx
```diff
@@ -735,7 +738,9 @@ export const RowInViewport: Story<TableProps> = (args) => {
         <a
           className='iui-anchor'
           onClick={() =>
-            parent.document.getElementById('tabbutton-actions')?.click()
+            (parent.document.querySelector(
+              '[id^="tabbutton-actions"]',
```
veekeys, line 742: Why this was needed to change?

.storybook/main.js
```diff
@@ -35,4 +35,8 @@ module.exports = {
     }
     return config;
   },
+  babel: async (options) => ({
+    ...options,
+    plugins: [['@babel/plugin-proposal-class-properties', { loose: true }]],
```
veekeys, line 40: It the issue you linked, they said in 6.3 it was fixed.. Sure you still had problems?

mayank99:
> What I noticed (but I guess not critical..).
> -- DropdownMenu: default value for role is not shown (it was shown before)

> Or in Modal its taken as undefined, but actually they have values set...

This is because of the new way they are handling default values in sb 6.3. Many open issues that might be related:
- https://github.com/storybookjs/storybook/issues/15446
- https://github.com/storybookjs/storybook/issues/15420
- https://github.com/storybookjs/storybook/issues/15378

sb devs seem to be aware of the problem, so I guess lets wait for a patch.

stories/core/Table.stories.tsx
mayank99: the id 'tabbutton-actions' was changed to 'tabbutton-actions-{count}' so I changed the selector to use `^` (target prefix)

stories/core/Select.stories.tsx
mayank99: I think that might not be needed anymore. I was playing around with the array (before I switched to `Array(N).fill()` syntax).

I'll revert next time I run this branch.

.storybook/main.js
mayank99: I was still having problems but I'll double check

mayank99: Tried removing and deleted node modules and cache, and ran `yarn install --force`. Storybook failed to start with this error:
```
SyntaxError: \iTwinUI-react\src\core\Toast\Toaster.tsx: Missing class properties transform.
  17 |
  18 | export default class Toaster {
> 19 |   private toasts: ToastProps[] = [];
```

veekeys: hmm...well ok.
If we gonna wait for patch, lets try this again after updating.
Im gonna leave this open, unless we decide to merge.

stories/core/Select.stories.tsx
veekeys: yeah..I guess what I dont like with `Partial` it kinda changes whole original interface.. Not sure it has huge influence for Storybook, though..

veekeys (approved):

PR #169, opened 2021-06-25
Add css transition to expandable table rows

mayank99:
- Wrapped table subcomponent in `CSSTransition` from transition group.
  - Adds new classes with a timeout of 200ms:
    - `iui-enter`, `iui-enter-active`, `iui-enter-done` when entering
    - `iui-exit` and `iui-exit-active` when exiting (and then unmounts)
  - Updates inline height for animation purposes.
- Bumped itwinui-css to 0.22.0 which has transitions for these classes.
- Fixed expander icon button class name and svg

transitionr

## Checklist


src/core/Table/TableRowMemoized.tsx
```diff
@@ -58,8 +61,25 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
-      {row.isExpanded && subComponent && (
-        <div className='iui-row iui-expanded-content'>{subComponent(row)}</div>
+      {subComponent && (
+        <CSSTransition
+          in={row.isExpanded}
+          timeout={200}
+          unmountOnExit={true}
+          onEnter={(node) => {
+            if (expandedHeight === 0) {
+              setExpandedHeight(node.getBoundingClientRect().height);
```
mayank99, line 71: @bentleyvk Do you know a better way to "save" the original height before setting inline height to 0?

mayank99: Maybe from the ref?

bentleyvk: I see Material-UI is using wrapper ref for this:
https://github.com/mui-org/material-ui/blob/e724d98eba018e55e1a684236a2037e24bcf050c/packages/material-ui/src/Collapse/Collapse.js#L167

mayank99: They have like 5 different wrappers 😄 

But yeah, I'll try with ref

bentleyvk (approved):

src/core/Table/TableRowMemoized.tsx
```diff
@@ -43,6 +44,8 @@ const TableRow = <T extends Record<string, unknown>>(props: {
     rootMargin: `${intersectionMargin}px`,
   });
 
+  const expandedHeight = React.useRef(0);
```
veekeys, line 47: I guess I would have had here just `expandedContentRef `and then take its height.. In case we need that ref for anythin else, would not need to change it.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -58,8 +61,33 @@ const TableRow = <T extends Record<string, unknown>>(props: {
           );
         })}
       </div>
-      {row.isExpanded && subComponent && (
```
veekeys, line 61: This prop is a leftover and not needed?

src/core/Table/TableRowMemoized.tsx
mayank99: I think element ref's height would not stay static, would it? Since we are changing inline height on the same element

src/core/Table/TableRowMemoized.tsx
mayank99: image

Because I set `unmountOnExit=true`, it will leave the DOM when `row.isExpanded` is false

PR #176, opened 2021-06-29
Add `useResizeObserver` hook to update tab stripe width

mayank99:
Added hook to wrap `ResizeObserver` API. Takes a ref to the element that needs to be observed, and returns a new `size` object every time that element is resized.

Used this hook in `HorizontalTabs` to update active stripe width for pill tabs.

tabs

## Checklist

- [x] ~~Add an entry to the changelog for any new components and changes~~

### DIFF
--- a/src/core/HorizontalTabs/HorizontalTabs.tsx
+++ b/src/core/HorizontalTabs/HorizontalTabs.tsx
@@ -6,6 +6,7 @@ import cx from 'classnames';
 import React from 'react';
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/tabs.css';
+import { useResizeObserver } from '../utils/hooks/useResizeObserver';
 
 export type HorizontalTabsProps = {
   /**
@@ -81,6 +82,8 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
 
   const tablistRef = React.useRef<HTMLUListElement>(null);
 
+  const tablistSize = useResizeObserver(tablistRef);
+
   const [currentIndex, setCurrentIndex] = React.useState(getValidIndex());
   const [stripeStyle, setStripeStyle] = React.useState<React.CSSProperties>({});
 
@@ -88,14 +91,17 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
     if (activeIndex != null && currentIndex !== activeIndex) {
       setCurrentIndex(getValidIndex());
     }
+  }, [activeIndex, currentIndex, getValidIndex]);
+
+  React.useLayoutEffect(() => {
     const activeTab = tablistRef.current?.children[currentIndex] as HTMLElement;
     if (type !== 'default') {
       setStripeStyle({
-        width: activeTab?.getBoundingClientRect().width,
+        width: activeTab?.getBoundingClientRect()?.width,
         left: activeTab?.offsetLeft,
       });
     }
-  }, [activeIndex, currentIndex, getValidIndex, type]);
+  }, [currentIndex, type, tablistSize]);
 
   const onTabClick = (index: number) => {
     if (onTabSelected) {
--- a/src/core/HorizontalTabs/HorizontalTabs.tsx
+++ b/src/core/HorizontalTabs/HorizontalTabs.tsx
@@ -94,11 +94,11 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
   }, [activeIndex, currentIndex, getValidIndex]);
 
   React.useLayoutEffect(() => {
-    const activeTab = tablistRef.current?.children[currentIndex] as HTMLElement;
     if (type !== 'default') {
+      const activeTab = tablistRef.current?.children[currentIndex];
       setStripeStyle({
-        width: activeTab?.getBoundingClientRect()?.width,
-        left: activeTab?.offsetLeft,
+        width: activeTab?.getBoundingClientRect().width,
+        left: (activeTab as HTMLElement)?.offsetLeft,
       });
     }
   }, [currentIndex, type, tablistSize]);
--- a/src/core/utils/hooks/useResizeObserver.tsx
+++ b/src/core/utils/hooks/useResizeObserver.tsx
@@ -5,13 +5,14 @@
 import React from 'react';
 
 /**
- * Hook that returns window dimensions updated through the `resize` event.
- * @param delay maximum ms to debounce the update by
- * @returns state object containing height and width of the window
+ * Hook that uses `ResizeObserver` to return an element's size every time it updates.
+ * @param elementRef ref of the element to observe resizes on.
+ * @param delay maximum timeout in ms to debounce the update by.
+ * @returns stateful object containing height and width of the element.
  */
 export const useResizeObserver = <T extends HTMLElement>(
   elementRef: React.RefObject<T>,
-  { delay = 0 },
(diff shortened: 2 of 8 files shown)


src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * Hook that uses `ResizeObserver` to return an element's size every time it updates.
+ * @param elementRef ref of the element to observe resizes on.
+ * @param delay maximum timeout in ms to debounce the update by.
+ * @returns stateful object containing height and width of the element.
+ */
+export const useResizeObserver = <T extends HTMLElement>(
+  elementRef: React.RefObject<T>,
+  { delay = 0 } = {},
+) => {
+  const [size, setSize] = React.useState({
```
bentleyvk, line 17: This init code will be called always, so you can change it to function. That way it will be initialized only once.

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,50 @@
...
+    );
+    resizeObserver.current?.observe(elementRef.current as T);
+    return () => resizeObserver.current?.disconnect();
+  }, [elementRef, delay]);
+
+  return size;
+};
+
+/** Returns a debounced fn that waits delayed by the provided ms */
+const debounce = <F extends (...args: never[]) => unknown>(
```
bentleyvk, line 40: Do we need that right now? It just adds additional complexity which also has to be tested.

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  React.useLayoutEffect(() => {
+    const updateSize = (entry: DOMRectReadOnly) => {
+      setSize({ height: entry.height, width: entry.width });
+    };
+    const debouncedUpdate = delay ? debounce(updateSize, delay) : updateSize;
+    resizeObserver.current = new ResizeObserver(([{ contentRect }]) =>
+      debouncedUpdate(contentRect),
+    );
+    resizeObserver.current?.observe(elementRef.current as T);
```
bentleyvk, line 32: Isn't `elementRef.current` `null` at first? You could do nothing then.

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -81,21 +82,26 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
   const [stripeStyle, setStripeStyle] = React.useState<React.CSSProperties>({});
 
   React.useLayoutEffect(() => {
     if (activeIndex != null && currentIndex !== activeIndex) {
       setCurrentIndex(getValidIndex());
     }
+  }, [activeIndex, currentIndex, getValidIndex]);
+
+  React.useLayoutEffect(() => {
     const activeTab = tablistRef.current?.children[currentIndex] as HTMLElement;
```
bentleyvk, line 97: Move this under `if`.

src/core/utils/hooks/useResizeObserver.tsx
mayank99: I think it should almost always be debounced for performance. It's just in this case, it looks particularly bad, because only the active stripe lags behind.

It's a pretty standard function, I would leave it.

src/core/utils/hooks/useResizeObserver.tsx
mayank99: Changed

src/core/utils/hooks/useResizeObserver.tsx
mayank99: Moved to if

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: Moved

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,53 @@
...
+  React.useLayoutEffect(() => {
+    const updateSize = (entry: DOMRectReadOnly) => {
+      setSize({ height: entry.height, width: entry.width });
+    };
+    const debouncedUpdate = delay ? debounce(updateSize, delay) : updateSize;
+    resizeObserver.current = new ResizeObserver(([{ contentRect }]) =>
+      debouncedUpdate(contentRect),
+    );
+
+    if (elementRef.current) {
```
veekeys, line 33: Do you need to create ResizeObserver if there is no element which you want to observe?

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,53 @@
...
+import React from 'react';
+
+/**
+ * Hook that uses `ResizeObserver` to return an element's size every time it updates.
+ * @param elementRef ref of the element to observe resizes on.
+ * @param delay maximum timeout in ms to debounce the update by.
+ * @returns stateful object containing height and width of the element.
+ */
+export const useResizeObserver = <T extends HTMLElement>(
+  elementRef: React.RefObject<T>,
```
veekeys, line 14: How would you extend this if there was a need to observe multiple elements?

mayank99: I would think there should multiple instances of this hook

src/core/utils/hooks/useResizeObserver.tsx
mayank99: Moved inside `if`

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const resizeObserver = React.useRef<ResizeObserver | null>(null);
+
+  const observe = React.useCallback(() => {
+    const updateSize = (entry: DOMRectReadOnly) => {
+      setSize({ height: entry.height, width: entry.width });
+    };
+
+    if (elementRef.current) {
+      resizeObserver.current = new ResizeObserver(([{ contentRect }]) =>
+        updateSize(contentRect),
```
veekeys, line 30: you could just call setSize here, is there a need of const?

mayank99: True. There was a need when there was debounce, but not anymore

veekeys (approved):

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -81,21 +82,26 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
-    const activeTab = tablistRef.current?.children[currentIndex] as HTMLElement;
+  }, [activeIndex, currentIndex, getValidIndex]);
+
+  React.useLayoutEffect(() => {
     if (type !== 'default') {
+      const activeTab = tablistRef.current?.children[currentIndex];
       setStripeStyle({
         width: activeTab?.getBoundingClientRect().width,
-        left: activeTab?.offsetLeft,
+        left: (activeTab as HTMLElement)?.offsetLeft,
```
bentleyvk, line 101: I see `ResizeObserver` also return `left`/`x` prop but I don't know if there is a big difference using that or `offsetLeft`.

bentleyvk:
Liquid behavior 😁 but I think it is fine.
mayank-resize-tabs

bentleyvk (approved):

PR #185, opened 2021-07-01
Disable extra Table props in storybook

mayank99:
Noticed a bunch of extra controls in Table in the newly released Storybook. https://itwin.github.io/iTwinUI-react/?path=/story/core-table
image

Disabled them and formatted them to take up less space.

### DIFF
--- a/stories/core/Table.stories.tsx
+++ b/stories/core/Table.stories.tsx
@@ -65,6 +65,8 @@ export default {
     autoResetSortBy: { table: { disable: true } },
     autoResetHiddenColumns: { table: { disable: true } },
     autoResetFilters: { table: { disable: true } },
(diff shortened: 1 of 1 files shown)


stories/core/Table.stories.tsx
```diff
@@ -39,78 +39,39 @@ export default {
     emptyTableContent: 'No data.',
   },
   argTypes: {
```
bentleyvk, line 41: Also add `disableFilters`

mayank99: I didn't add that because I thought it would be nice to have (as it does allow you to disable filtering in storybook)

bentleyvk (approved):

PR #186, opened 2021-07-02
Adding User Icon Group component

gretanausedaite:
Adding User Icon group component

image

image

image

## Checklist


src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+          {childrenArray.map((child, index) =>
+            index < length
+              ? React.cloneElement(child as JSX.Element, {
+                  status: '',
+                  key: index,
+                })
+              : null,
+          )}
+
+          <UserIcon
```
gretanausedaite, line 117: Extra members icon color: 
orange in https://itwin.github.io/iTwinUI/backstop/minified/user-icon.html
image

gray in https://ux.bentley.com/itwin/components/user-icon.html#team-displays
image

white when no "background-color" prop passed
image

green when empty string as "background-color" prop passed
image

gretanausedaite:
Ideas for responsiveness:
1. Should we stack more User Icons in small areas? Currently it is recommended in our docs to show 5 UserIcons and stack everything above. 
Eg. show 5 icons when area allows and stack more icons when area shrinks
image
image

2. Animated version: should UserIcons be unstacked on hover? 
Not hover:
image
Hover:
image
vs
image

gretanausedaite:
Tittle vs Tooltip:
CSS version of this group uses Tooltip visual:
image
If I add tooltip in React, the last user icon is visually ungrouped:
image

Should I change CSS to accommodate additional depth in code or use Title (code currently does this)
image

gretanausedaite:
User Icon takes only 2 symbols for abbreviation. Cant make it show 10+
image
vs
image

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: @FlyersPh9

veekeys:
> Ideas for responsiveness:
> 
> 1. Should we stack more User Icons in small areas? Currently it is recommended in our docs to show 5 UserIcons and stack everything above.
>    Eg. show 5 icons when area allows and stack more icons when area shrinks
>    image
>    image
> 2. Animated version: should UserIcons be unstacked on hover?
>    Not hover:
>    image
>    Hover:
>    image
>    vs
>    image

1. I think it would be nice if we could react to the paren width change. There is already resize observer which could be helpful here. In other words, show all (according to the params) if they fit, if not show maximum possible.

2. I would think thats the point of that animation and hover.. So all icons could be visible? But the issue is, it might brake layouts, cause theoretically they gonna expand and change the width.. and if it was overflown more issues there.. so I am not sure. @FlyersPh9, whats the expected pattern here?

veekeys:
> Tittle vs Tooltip:
> CSS version of this group uses Tooltip visual:
> image
> If I add tooltip in React, the last user icon is visually ungrouped:
> image
> 
> Should I change CSS to accommodate additional depth in code or use Title (code currently does this)
> image

I would say show title here... Tooltip does not look good and we already show title in UserIcon, so that should be consistent..
But I also think css should be changed, so rules still work if userIcon is even wrapped with some element (need to fill work item for this in future, in case...).

mayank99:
> 2. Animated version: should UserIcons be unstacked on hover?
>    Not hover:
>    image

I think that case is not even valid. If there is space to show "+1" icon, might as well show the actual icon, even when stacked. And I don't know if this animation pattern should even be used when there's more than a few icons. @FlyersPh9 

> > If I add tooltip in React, the last user icon is visually ungrouped:
> > image
> > Should I change CSS to accommodate additional depth in code or use Title (code currently does this)
> > image
> 
> I would say show title here... Tooltip does not look good and we already show title in UserIcon, so that should be consistent..
> But I also think css should be changed, so rules still work if userIcon is even wrapped with some element (need to fill work item for this in future, in case...).

I would agree, css should be changed to accommodate tooltips. But I don't think we should show title. Instead, we should allow the user to pass click/hover handlers so that they can do whatever they want (e.g. clicking on "+9" could open a modal or different page)

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const defaultLength = 5;
+const maxLength = 9;
+
+export type UserIconGroupProps = {
+  /**
+   * UserIconGroup group length
+   * @default 5
+   */
+  length?: number;
```
mayank99, line 19: This prop name could be confusing. 🤔

Maybe `maxLength` or `maxIcons`?

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+export const UserIconGroup = (props: UserIconGroupProps) => {
+  const { children, type = 'default', length = defaultLength } = props;
+
+  const childrenArray = React.Children.toArray(children);
+  const childrenLength = React.Children.count(children);
+  let additionalChildren = '';
+  if (childrenLength > length) {
+    React.Children.forEach(children, (child: JSX.Element, index) => {
+      if (index >= length) {
+        additionalChildren += child.props.title + '\n';
```
mayank99, line 91: Hmm... Weird logic

mayank99 (commented):
I started reviewing, but I think the API will need to be redesigned a little bit, as current API is not very flexible or smart.

Imagine: There are 200 user icons. Do we really need all of them to be passed in here? We should just ask for the first few, and a callback to handle what happens when clicking "10+".

Also, if it's called a `UserIconGroup` (a generic name), I think maybe it will need to support a "non-stacked" version like this:
image

bentleyvk:
> Tittle vs Tooltip:
> CSS version of this group uses Tooltip visual:
> image
> If I add tooltip in React, the last user icon is visually ungrouped:
> image
> 
> Should I change CSS to accommodate additional depth in code or use Title (code currently does this)
> image

Tooltip looks better for me. Also, I agree with Mayank that we need to allow users to pass `onClick`, `onHover` handlers but need to think what is the best approach here.

bentleyvk:
> User Icon takes only 2 symbols for abbreviation. Cant make it show 10+
> image
> vs
> image

Antd has nice handling for this:
https://ant.design/components/avatar/#components-avatar-demo-group
Their code:
https://github.com/ant-design/ant-design/blob/13a829d20c8a1a83c3bfa1f9d7c30d95659f5b57/components/avatar/avatar.tsx#L51

But I don't know if we really need that. Maybe we could just show `+99` as a max value because I have seen this in some places. But we still need to make it fit.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+      })}
+    >
+      {childrenLength <= length ? (
+        children
+      ) : (
+        <>
+          {childrenArray.map((child, index) =>
+            index < length
+              ? React.cloneElement(child as JSX.Element, {
+                  status: '',
```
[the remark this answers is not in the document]
gretanausedaite: Added undefined

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+import cx from 'classnames';
+import { UserIcon } from '../UserIcon';
+export type UserIconGroupType = 'default' | 'animated';
+
+const defaultLength = 5;
+const maxLength = 9;
```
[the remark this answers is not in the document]
gretanausedaite: Moved into UserIconGroup

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+          <UserIcon
+            size={(childrenArray[0] as JSX.Element).props.size}
+            title={additionalChildren}
+            abbreviation={
+              childrenLength <= maxLength
+                ? `+${childrenLength - length}`
+                : `${maxLength}+`
+            }
+            backgroundColor=''
```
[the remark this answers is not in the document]
gretanausedaite: Changed color to grey

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: I chose maxIcons :)

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: Could you elaborate?

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * UserIconGroup group length
+   * @default 5
+   */
+  length?: number;
+  /**
+   * UserIconGroup type
+   * @default 'default'
+   */
+  type?: UserIconGroupType;
```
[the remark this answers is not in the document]
gretanausedaite: Changed to animated?: boolean.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * UserIconGroup max number of icons unstacked
+   * @default 5
+   */
+  maxIcons?: number;
+  /**
+   * UserIconGroup is group needs to be stacked
+   * @default 'true'
+   */
+  stacked?: boolean;
```
gretanausedaite, line 21: @mayank99 Added stacked?: boolean for unstacked version

src/core/UserIconGroup/UserIconGroup.tsx
mayank99: Well, I was going to, but then I thought it would be better to scrap this whole block and rethink the API.

If we allow passing `onClick`/`onMouseOver` handlers, then we won't need any of this tooltip logic.

src/core/UserIconGroup/UserIconGroup.tsx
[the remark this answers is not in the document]
mayank99: @FlyersPh9 can comment on what the exact color should be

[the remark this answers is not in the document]
mayank99: Resolving this as duplicate of https://github.com/iTwin/iTwinUI-react/pull/186#discussion_r662961797

src/core/UserIconGroup/UserIconGroup.tsx
FlyersPh9: Created a new PR to give the user icon with the count inside it slightly different visual styling: https://github.com/iTwin/iTwinUI/pull/177

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <UserIcon
+            size={(childrenArray[0] as JSX.Element).props.size}
+            title={additionalChildren}
+            abbreviation={
+              childrenLength <= maxLength
+                ? `+${childrenLength - maxIcons}`
+                : `${maxLength}+`
+            }
+            backgroundColor='#dce0e3'
+          />
```
mayank99, line 114: > User Icon takes only 2 symbols for abbreviation. Cant make it show 10+

This is by design.
https://github.com/iTwin/iTwinUI-react/blob/1b3488d3cae8bb22263947c780ba8e2e6ad5152b/src/core/UserIcon/UserIcon.tsx#L110-L112

Since we need to add a new class (https://github.com/iTwin/iTwinUI/pull/177) and support custom events anyway, I'd say don't use `UserIcon` and instead go with a custom div.

```suggestion
          <div
            {...excessIconProps} // this should be a new prop
            className={cx(
              'iui-user-icon',
              { [`iui-${size}`]: (childrenArray[0] as JSX.Element).props.size !== 'medium' },
              'iui-user-icon-count',
              excessIconProps.className
            )}
          >
            <abbr className='iui-initials'>
              {childrenLength <= maxLength ? `+${childrenLength - maxIcons}` : `${maxLength}+`}
            </abbr>
            <span className='iui-stroke' />
          </div>
```

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,121 @@
...
+            index < maxIcons
+              ? React.cloneElement(child as JSX.Element, {
+                  status: undefined,
+                  key: index,
+                })
+              : null,
+          )}
+
+          <UserIcon
+            size={(childrenArray[0] as JSX.Element).props.size}
```
mayank99, line 106: What happens if the user icons are different sizes? 😮

src/core/UserIconGroup/UserIconGroup.tsx
mayank99: `excessIconProps` will add support for custom events if it's extending native div attributes, but I think we need one more prop to add support for tooltips/dropdowns 🤔 And then it starts getting complicated. Maybe we just need to allow passing a custom icon (which user can wrap in tooltip/dropdown) and show a storybook example.

---

AntD just forces a Popover, and has a `maxPopoverPlacement` prop. https://ant.design/components/avatar/#Avatar.Group-(4.5.0+)

FlyersPh9:
> User Icon takes only 2 symbols for abbreviation. Cant make it show 10+
> image
> vs
> image

Spoke with Mayank & Virginie about this.  We're going to adjust the standard.  We can display 1-99 without any + symbol at all.
Screen Shot 2021-07-06 at 1 28 54 PM

Once the count reaches 100 or higher, we will display 99+.
Screen Shot 2021-07-06 at 1 30 00 PM

We understand that it is hard to fit 3 characters inside of the user icon, especially at smaller sizes.  It seems likely that a user icon stack will hit 10+, but seems less likely that it will hit 100+.  This change will keep all user stacks less than 100 at 2 characters which should display well in all sizes.

src/core/UserIconGroup/UserIconGroup.tsx
mayank99: Another problem is that `UserIcon` does not forward ref, so popover won't show at all. We could add a wrapper but that breaks styling.

src/core/UserIconGroup/UserIconGroup.tsx
mayank99: I updated and simplified the stories to focus on the props of `UserIconGroup` (instead of repeating `UserIcon` sizes).

Seems that `stacked: false` doesn't do anything:
image

We need a new css class for it.

stories/core/UserIconGroup.stories.tsx
```diff
@@ -200,15 +201,25 @@ export const WithTooltip: Story<UserIconGroupProps> = (args) => {
     'Justice Harrington',
   ];
 
-  const [tooltipVisible, setTooltipVisible] = React.useState<boolean>(false);
-
+  const [, setTooltipVisible] = React.useState<boolean>(false);
```
mayank99, line 204: Needed?

stories/core/UserIconGroup.stories.tsx
```diff
@@ -200,15 +201,25 @@ export const WithTooltip: Story<UserIconGroupProps> = (args) => {
...
+  const userIconRef = React.useRef();
   const excessIconProps = {
+    id: 'someid',
```
mayank99, line 207: Not needed

stories/core/UserIconGroup.stories.tsx
```diff
@@ -200,15 +201,25 @@ export const WithTooltip: Story<UserIconGroupProps> = (args) => {
...
     onMouseOut: () => setTooltipVisible(false),
   };
 
+  const arrayLength = 5;
+  const usersSubArray = userNames.slice(arrayLength);
+  const tooltipContent = usersSubArray.join(`\n`) as string;
+
   return (
-    <div className='iui-tooltip-container'>
+    <Tooltip
```
mayank99, line 218: You can place tooltip as sibling. #198 was merged

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export type UserIconGroupProps = {
+  /**
+   * UserIconGroup max number of icons unstacked
+   * @default 5
+   */
+  maxIcons?: number;
+  /**
+   * UserIconGroup is group needs to be stacked
+   * @default 'true'
```
veekeys, line 18: its boolean, so `@default true`, not `@default 'true'`

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+import cx from 'classnames';
+import { CommonProps } from '../utils/props';
+
+export type UserIconGroupProps = {
+  /**
+   * UserIconGroup max number of icons unstacked
```
veekeys, line 12: No need componennt name in front. Already clear, which components props are these

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * @default 5
+   */
+  maxIcons?: number;
+  /**
+   * UserIconGroup is group needs to be stacked
```
veekeys, line 17: amm, weird english... Just maybe `Is group stacked` or `If true, group will be stacked to take less space.`

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * @default 'true'
+   */
+  stacked?: boolean;
+  /**
+   * UserIconGroup type
```
veekeys, line 22: Its the hover effect? So then can say that `When group is stacked, on hover it will be expanded.` or something like that..

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+   */
+  iconSize: 'small' | 'medium' | 'large' | 'x-large';
+  /**
+   * User Icons in the UserIconGroup.
+   */
+  children: React.ReactNode;
+
+  /**
+   * User count icon props, eg. onClick
```
veekeys, line 36: It is `e.g.`. 
No need empty line above too.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+   * @default 'true'
+   */
+  stacked?: boolean;
+  /**
+   * UserIconGroup type
+   * @default 'false'
+   */
+  animated?: boolean;
+  /**
+   * UserIconGroup type
```
veekeys, line 27: Its not type.. Size more, I guess..

veekeys: interesting about the size too.. Cause user needs to pass size prop to the group component and also hopefully same in every UserIcon component...

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * User count icon props, eg. onClick
+   */
+  countIconProps?: React.HTMLAttributes<HTMLDivElement>;
+} & CommonProps;
+
+/**
+ * Group User Icons together.
+ *
+ * User Icons stacking is based on maxIcons count. If you provide 8 UserIcons and keep default 5 maxIcons count,
```
veekeys, line 44: Wrapping props in your comment with the ` symbol, might make it look nicer to the user too.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+ *    abbreviation="TR"
+ *    backgroundColor="#6AB9EC"
+ *    size="medium"
+ *    title="Terry Rivers"
+ *    type="default"
+ *  />
+ * </UserIconGroup>
+ */
+export const UserIconGroup = (props: UserIconGroupProps) => {
+  const defaultLength = 5;
```
veekeys, line 75: No need this const.. You are not using it anywhere else, just assign value to the `maxIcons` directly.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+    children,
+    animated = false,
+    stacked = true,
+    maxIcons = defaultLength,
+    iconSize,
+    countIconProps,
+  } = props;
+
+  const childrenArray = React.Children.toArray(children);
+  const childrenLength = React.Children.count(children);
```
veekeys, line 87: Cant you use just `childrenArray.length` instead?
Also, now I wonder, if we better expect array in Children (`ReactNodeArray`).. Not sure, whats more comfortable to the user.

src/core/UserIconGroup/UserIconGroup.tsx
veekeys: oh, I see you just override all user icons with the container size.. Might mention that in comment too.

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,146 @@
...
+                })
+              : null,
+          )}
+
+          <div
+            {...countIconProps}
+            className={cx(
+              'iui-user-icon',
+              `iui-${iconSize}`,
+              'iui-user-icon-count',
```
veekeys, line 130: dont forget to take the `className` from `countIconProps`.

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+              .split(' ')
+              .map((token) => token[0])
+              .join('')}
+            backgroundColor={getUserColor(name)}
+            title={name}
+          />
+        ))}
+      </UserIconGroup>
+      <br />
+      <UserIconGroup animated={false} iconSize='large'>
```
veekeys, line 127: maybe take here args too and just override `iconSize`..

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+            .split(' ')
+            .map((token) => token[0])
+            .join('')}
+          backgroundColor={getUserColor(name)}
+          title={name}
+        />
+      ))}
+    </UserIconGroup>
+  );
+};
```
veekeys, line 85: Either, new line after story everywhere, or nowhere. Would make it look consistently.

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const [, setDoSomething] = React.useState<boolean>(false);
+
+  /**
+   * User Icon ref is needed for Tooltip positioning.
+   */
+  const userIconRef = React.useRef();
+
+  const countIconProps = {
+    ref: userIconRef,
+    onMouseOver: () => setDoSomething(true),
```
veekeys, line 206: Not needed anymore, I guess? Together with that `doSomething`

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,241 @@
...
+              .map((token) => token[0])
+              .join('')}
+            backgroundColor={getUserColor(name)}
+            title={name}
+          />
+        ))}
+      </UserIconGroup>
+      <Tooltip
+        reference={userIconRef.current}
+        content={tooltipContent as string}
```
veekeys, line 232: still need casting?

stories/core/UserIconGroup.stories.tsx
gretanausedaite: Removed

stories/core/UserIconGroup.stories.tsx
gretanausedaite: Placed as sibling

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: Changed into not weird English

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+`Date`
+
+- **Added `UserIconGroup` component** - stacked/non-stacked, animated lists with custom `.iui-user-icon-count` icon behavior
```
mayank99, line 7: Don't need to mention classname here. Users only care about the component and props.

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+    <UserIconGroup iconSize='medium'>
+      <UserIcon
+        abbreviation='RM'
+        backgroundColor='#C8C2B4'
+        size='medium'
+        title='Robin Mercer'
+      />
+    </UserIconGroup>,
+  );
+  expect(container.querySelector('div')).toBeTruthy();
```
mayank99, line 22: This test needs to be fixed

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+  );
+  const userGroupContainer = container.querySelector(
+    '.iui-user-icon-list',
+  ) as HTMLElement;
+  expect(userGroupContainer.classList).not.toContain(`iui-animated`);
+});
+
+it('should render stacked', () => {
+  const { container } = render(
+    <UserIconGroup iconSize='medium'>
```
mayank99, line 73: I would suggest creating a helper function to render the components

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+        title='Mercer Robin'
+      />
+    </UserIconGroup>,
+  );
+
+  const userGroupContainer = container.querySelector(
+    '.iui-user-icon-list',
+  ) as HTMLElement;
+  expect(userGroupContainer.classList).not.toContain(`iui-stacked`);
+});
```
mayank99, line 190: Need more test cases, for things like 99+ length, last icon props, etc.

`yarn test` should show you which lines were not covered

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+    </UserIconGroup>
+  );
+};
+
+Animated.args = {
+  animated: true,
+  iconSize: 'medium',
+};
+
+export const ManyIcons: Story<UserIconGroupProps> = (args) => {
```
mayank99, line 93: The original purpose of this story was to show the "9+" variant. Since that has changed to "99+", I think we might need to show that instead.

src/core/UserIconGroup/UserIconGroup.test.tsx
veekeys: You need to check if the element with class `iui-user-icon-list` exists.

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <UserIcon
+        abbreviation='RM'
+        backgroundColor='#C8C2B4'
+        size='medium'
+        title='Robin Mercer'
+      />
+    </UserIconGroup>,
+  );
+  expect(container.querySelector('div')).toBeTruthy();
+});
```
veekeys, line 23: ALso could check the amount of user icons rendered and that the correct size class is applied.

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+      />
+      <UserIcon
+        abbreviation='MR'
+        backgroundColor='#C8C2B4'
+        size='medium'
+        title='Mercer Robin'
+      />
+    </UserIconGroup>,
+  );
+  const userGroupContainer = container.querySelector(
```
veekeys, line 42: or just could expect to be truthy element with `.iui-user-icon-list.iui-animated`

src/core/UserIconGroup/UserIconGroup.test.tsx
veekeys: and iui-stacked class should be applied here too

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,238 @@
...
+    'Tanner Fraser',
+    'Ashley Miles',
+    'Jean Mullins',
+  ];
+
+  return (
+    <UserIconGroup {...args}>
+      {userNames.map((name) => (
+        <UserIcon
+          size='medium'
```
veekeys, line 42: now when I look at it, no need to pass size prop to the UserIcon. Enough to the group, cause it overrides it.
So suggest changing all examples (stories and docs)

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,148 @@
...
+  return (
+    <div
+      className={cx(
+        'iui-user-icon-list',
+        {
+          'iui-animated': animated,
+        },
+        {
+          'iui-stacked': stacked,
+        },
```
mayank99, line 99: These can be combined under the same object

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,148 @@
...
+  stacked?: boolean;
+  /**
+   * If true, group will be animated on hover.
+   * @default false
+   */
+  animated?: boolean;
+  /**
+   *  Size of user and count icons.
+   */
+  iconSize: 'small' | 'medium' | 'large' | 'x-large';
```
mayank99, line 29: Should have a default size, no?

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,148 @@
...
+ *    title="Robin Mercer"
+ *    type="default"
+ *  />
+ *  <UserIcon
+ *    abbreviation="TR"
+ *    backgroundColor="#6AB9EC"
+ *    size="medium"
+ *    title="Terry Rivers"
+ *    type="default"
+ *  />
```
mayank99, line 69: Update this example.

1. Names should probably not be repeated.
2. Background color should use `getUserColor` function.
3. Size does not need be set on on user icons.
4. There is no `type` prop.

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,322 @@
...
+    'Jean Mullins',
+    'Nico Triplett',
+    'Drew Abel',
+    'Kendall Simons',
+    'Kennedy Gray',
+    'Charlie Mayfield',
+    'Peyton Pennington',
+    'Justice Harrington',
+    'Jessie Dodd',
+  ];
```
mayank99, line 198: If you're just repeating the names, you don't really need an array that's so long.

You can just define the unique ones, and iterate over them using `%`.

stories/core/UserIconGroup.stories.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+  return (
+    <>
+      <UserIconGroup {...args}>
+        {userNames.map((name) => (
+          <UserIcon
+            key={name}
```
mayank99, line 205: using `name` as key is bad here, because the names are not unique

stories/core/UserIconGroup.stories.tsx
gretanausedaite: Changed story to show 99+

stories/core/UserIconGroup.stories.tsx
gretanausedaite: removed user icon size props

src/core/UserIconGroup/UserIconGroup.test.tsx
gretanausedaite: changed this this suggestion

src/core/UserIconGroup/UserIconGroup.test.tsx
gretanausedaite: updated basic test

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+    >
+      {childrenLength <= length ? (
+        children
+      ) : (
+        <>
+          {childrenArray.map((child, index) =>
+            index < length
+              ? React.cloneElement(child as JSX.Element, {
+                  status: '',
+                  key: index,
```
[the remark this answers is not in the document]
mayank99: Also, this will override the user-provided key.

Maybe do something like:
```suggestion
                  key: child.props.key ?? index,
```

It's still bad to use index as key but at least user can override it.

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: removed empty line

src/core/UserIconGroup/UserIconGroup.test.tsx
mayank99: Still need a test for `countIconProps`.

stories/core/UserIconGroup.stories.tsx
mayank99: I guess no need for showing two sizes there anymore.

The original intent was that small size would not have the "+", but now they are the same.

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const userGroupIconCount = container.querySelectorAll(
+    '.iui-user-icon-list > .iui-user-icon',
+  );
+  expect(userGroupIconCount.length).toBe(6);
+  const countIcon = container.querySelector(
+    '.iui-user-icon-count',
+  ) as HTMLElement;
+  expect(countIcon.textContent).toBe('2');
+  expect(userGroup.querySelectorAll('.iui-stroke').length).toBeGreaterThan(0);
+  expect(userGroup.querySelectorAll('.iui-initials').length).toBeGreaterThan(0);
```
veekeys, line 173: I think you knkow the exact amount of these, you could check if it equals

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+});
+
+it('should render animated', () => {
+  const { container } = render(
+    <UserIconGroup iconSize='medium' animated={true}>
+      {generateUserIcons(7)}
+    </UserIconGroup>,
+  );
+  expect(
+    container.querySelector('.iui-user-icon-list.iui-animated') as HTMLElement,
```
veekeys, line 183: no need cast here

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+  ).toBeTruthy();
+});
+
+it('should render many icons', () => {
+  const { container } = render(
+    <UserIconGroup iconSize='medium'>{generateUserIcons(105)}</UserIconGroup>,
+  );
+  const userGroup = container.querySelector(
```
veekeys, line 191: no need to set it as separate constant, as you ar enot using it anywhere else.. can straight away check if truthy

veekeys: same in other places

src/core/UserIconGroup/UserIconGroup.test.tsx
```diff
@@ -0,0 +1,264 @@
...
+    </UserIconGroup>,
+  );
+
+  const userGroupContainer = container.querySelector(
+    '.iui-user-icon-list.iui-stacked',
+  ) as HTMLElement;
+  expect(userGroupContainer).toBeFalsy();
+});
+
+it('should render user icons x-large', () => {
```
veekeys, line 218: instead you could iterate and check all the sizes.. there is example in UserIcon tests..

src/core/UserIconGroup/UserIconGroup.tsx
```diff
@@ -0,0 +1,143 @@
...
+        className,
+      )}
+      {...rest}
+    >
+      {childrenLength <= maxIcons + 1 ? (
+        <>
+          {childrenArray.map((child, index) =>
+            React.cloneElement(child as JSX.Element, {
+              status: undefined,
+              key: index,
```
mayank99, line 102: This whole logic can be improved to avoid duplicate code and bugs.

You could change the structure of the component to something like:
```tsx
<div className='iui-user-icon-list' ...>
  {​​​​​​​​​childrenArray.slice(0, maxIcons + 1).map(...)}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
  {​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​childrenArray.length > maxIcons && <div {​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​...countIconProps}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​ ... >}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
</div>
```

This way you're not repeating code, and you're not mapping over items when not needed.

src/core/UserIconGroup/UserIconGroup.tsx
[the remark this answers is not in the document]
mayank99: Oh wait, now I'm getting this error:
`Warning: UserIcon: `key` is not a prop. Trying to access it will result in `undefined` being returned.`

Do you even need to set key here?

src/core/UserIconGroup/UserIconGroup.tsx
gretanausedaite: I still need one more condition for 6 icons use case to not show +1 icon.
```
<div className='iui-user-icon-list' ...>
  {​​​​​​​​​childrenArray.length <= maxIcons + 1 && childrenArray.slice(0, maxIcons + 1).map(...)}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
{​​​​​​​​​childrenArray.length > maxIcons + 1  && childrenArray.slice(0, maxIcons).map(...)}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
  {​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​childrenArray.length > maxIcons + 1 && <div {​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​...countIconProps}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​ ... >}​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
</div>
```

veekeys (approved):
Congrats with the contribution, good job!

mayank99 (approved):

PR #187, opened 2021-07-02
Allowing to disable specific rows in Table

### LINKED ISSUE #125: Table: Disabled rows
#### Feature
Table currently allows all rows to be selectable when passed isSelectable prop
However there is no api to disable selection for some rows that might not be selectable

Similar feature that exists is of making column sortable by passing isSortable for the whole columns
And you can manually disable/enable sorting for individual column

#### Examples
Provide an api to say whether a row is disabled or not
Disabled row can not be selectable
Disabled row should have a different darker visual background to indicate that the row is disabled

bentleyvk:
Closes #125 

- Disabling selection checkboxes and expansion controls.
- Select All selects only enabled rows.
- Copied how header select checkbox should work from https://material-ui.com/components/data-grid/selection/#disable-selection-on-certain-rows
- There was a bug in `Checkbox` that after click indeterminate state disappears although it is set in props.

disabled-select

## Checklist


src/core/Checkbox/Checkbox.tsx
```diff
@@ -86,7 +86,7 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
           ? false
           : inputElementRef.current.checked;
       }
-    }, [indeterminate]);
+    });
```
bentleyvk, line 89: Without this change:
checkbox-with-intermediate

With change:
checkbox-with-intermediate2

mayank99: Should it really be `indeterminate` though? I would think that it should be `checked` when all selectable rows are selected. ("selectable rows" = non-disabled rows)

Maybe @FlyersPh9 can comment here.

src/core/Table/Table.tsx
```diff
@@ -118,6 +118,11 @@ export type TableProps<
    * Content shown when there is no data after filtering.
    */
   emptyFilteredTableContent?: React.ReactNode;
+  /**
+   * Function that returns whether row is disabled or not.
+   * It also disables selection and expansion cells.
+   */
+  isRowDisabled?: (rowData: T) => boolean;
```
mayank99, line 125: This prop is a bit unintuitive. Wouldn't it be better to have `isDisabled` property for each row in `data`? You can access it from `row.original`.

src/core/Table/Table.tsx
```diff
@@ -267,11 +274,29 @@ export const Table = <
...
+        Header: ({ getToggleAllRowsSelectedProps }: HeaderProps<T>) => {
+          const onChange = () => {
+            const isAllEnabledRowsSelected =
+              instance.isAllRowsSelected ||
+              instance.rows.every(
+                (row) =>
+                  instance.state.selectedRowIds[row.id] ||
+                  isRowDisabled(row.original),
+              );
+            instance.toggleAllRowsSelected(!isAllEnabledRowsSelected);
```
mayank99, line 286: Interesting logic. From what I understand, this is a binary toggle. Where does the `indeterminate` state come from?

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Version]
```
mayank99, line 3: Can you change this to `Unreleased` to prevent merge conflicts?

https://github.com/iTwin/iTwinUI-react/blob/0aaa0c31dfbc10b88888cf52156d326d2a78da29/CONTRIBUTING.md#L175-L181

src/core/Checkbox/Checkbox.tsx
veekeys: that is interesting :O
I would agree, that clicking the all checkbox should go to checked state. Staying indeterminate does not look ok.

src/core/Table/Table.tsx
```diff
@@ -267,11 +274,29 @@ export const Table = <
...
+          };
+          return (
+            <Checkbox
+              {...getToggleAllRowsSelectedProps()}
```
veekeys, line 290: Wouldnt from here also come onChange callback, which you need to merge?

stories/core/Table.stories.tsx
```diff
@@ -788,6 +788,100 @@ RowInViewport.argTypes = {
...
+        width: 100,
+        // Manually handling disabled state in custom cells
+        Cell: (props: CellProps<{ name: string; description: string }>) => (
+          <>
+            {isRowDisabled(props.row.original) ? (
+              <>Click me!</>
+            ) : (
+              <a
+                className='iui-anchor'
+                onClick={action(props.row.original.name)}
```
veekeys, line 831: we should not try to disable everything ourselves in table? like eat all the events, etc...

src/core/Table/Table.tsx
bentleyvk: I want to override it and did it here.

src/core/Table/Table.tsx
```diff
@@ -317,11 +342,17 @@ export const Table = <
     }
 
     const selectedData: T[] = [];
-    instance.rows.forEach((row) => {
-      if (newState.selectedRowIds[row.id]) {
-        selectedData.push(row.original);
+    const newSelectedRowIds = {} as Record<string, boolean>;
+    Object.keys(newState.selectedRowIds).forEach((id) => {
```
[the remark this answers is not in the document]
bentleyvk: Because I thought that `newState.selectedRowIds` would be smaller array in most cases 😃

stories/core/Table.stories.tsx
bentleyvk: Do we really want? What if there is a dropdown menu that has item to enable that row. Or even this example, idk how I could override it.

src/core/Table/Table.tsx
bentleyvk: From `...getToggleAllRowsSelectedProps()`

src/core/Checkbox/Checkbox.tsx
bentleyvk: I have no idea what is the proper way here as antd shows selected state, material-ui indeterminate. Also, all possible rows are selected so it should be checked but on the other hand not all rows are selected, so idk 😃

stories/core/Table.stories.tsx
veekeys: Well, I think by design, if row is disabled, it should not be interactive in any case? but maybe I agree, lets leave it... Just then need to be sure and inform users to handle it (I bet they wont!, that is my concern.. and all those menus will be still openable in disabled row).

src/core/Table/Table.tsx
veekeys: What if user has something added for his custom stuff? Or its not possible..

src/core/Table/Table.tsx
bentleyvk: But it is easier to use this way because I really doubt that there is a project that gets data from the server with the field `isDisabled` 😃 also if there is you can just pass `rowData => rowData.isDisabled` or `rowData => rowData.status !== 'success`. Another thing that users won't need to remap their data to have extra field for disabled state.

CHANGELOG.md
bentleyvk: Changed.

src/core/Table/Table.tsx
mayank99: > But it is easier to use this way

"easier" is subjective. I just thought it would make sense to have `isDisabled`, similar to `isExpanded` and `isSelected` properties of row. But those are not passed by the user, so maybe it's fine this way 😕 🤷‍♀️ 👍

CHANGELOG.md
mayank99: There will still be merge conflicts. Should be `Unreleased` not `[Unreleased]` 😆

src/core/Table/Table.tsx
mayank99: So we need to override if we want it to be `checked` when all non-disabled rows are selected.

```tsx
checked: instance.rows.every((row) => row.isSelected || isRowDisabled(row.original)),
```

src/core/Checkbox/Checkbox.tsx
FlyersPh9: I like that clicking on the Check All checkbox selects first, then deselects on the second click.

The indeterminate VS regular checkmark when a disabled row is displayed… it does seem weird to click & not get the regular checkmark because there is a disabled row present.  I think often times the disabled row may only be temporarily disabled as the file is uploading & processing.  Once that row becomes enabled, do we auto select it?  Or would it remain unselected? 🤔  I would think it would remain unselected, and if that is the case, then the indeterminate checkbox would make sense.

I would like to check how the other big players are handling this.

mayank99: > I like that clicking on the Check All checkbox selects first, then deselects on the second click.

That part is just standard behavior (try on a normal table without any disabled rows).

> I would like to check how the other big players are handling this.

As Vilius said, Material-UI and AntD (the two most popular React UI libraries) have opposite behavior for both `indeterminate` state as well as the "select first, then deselect" part.
https://material-ui.com/components/data-grid/selection/#disable-selection-on-certain-rows
https://ant.design/components/table/#components-table-demo-row-selection

[BlueprintJS](https://blueprintjs.com/docs/#table/api.table), [Semantic UI](https://react.semantic-ui.com/collections/table/#variations-selectable-row), [React-Bootstrap](https://react-bootstrap.github.io/components/table/) and [Fluent UI](https://developer.microsoft.com/en-us/fluentui#/controls/web/detailslist) don't support this use case.

src/core/Table/Table.tsx
mayank99: To alleviate the unintuitiveness, we could improve the JSDoc comment. Instead of _`Function that returns whether row is disabled or not. It also disables selection and expansion cells.`_, maybe we should be more clear about what is expected. Something like: _`Should return true if a row is disabled (i.e. cannot be selected or expanded). If not specified, all rows are enabled.`_.

Another source of confusion can be the param type (someone might ask "What is `rowData: T`?"). We could give an example to clarify that.

src/core/Table/Table.tsx
bentleyvk: What you mean by saying custom? It is controlled by us and comes from react-table.

src/core/Checkbox/Checkbox.tsx
bentleyvk: Wouldn't that confuse users when checked becomes indeterminate without user interaction?

src/core/Table/Table.tsx
veekeys: I meant, if user cannot pass same prop on row himself?

src/core/Checkbox/Checkbox.tsx
veekeys: No.. I dont get it.
Its either checked or none if user clicks on select all checkbox... Disabled is disabled and it cannot be a part of calculations for selected all.

FlyersPh9: Let's go with this approach:
> > Once that row becomes enabled, do we auto select it? Or would it remain unselected?
> 
> it should remain unselected, and the header cell should become `indeterminate` if it was previously `checked`.

Like Vilius said, it may confuse users, but I think it's a good logical starting point and we can test with users later on and change if needed.

src/core/Table/Table.tsx
bentleyvk: Changed description. About `rowData`, we have it in several places and IDE should show your data type instead of `T` so I don't think it is worth to put example on every prop, we have Storybook for that.

src/core/Checkbox/Checkbox.tsx
bentleyvk: Changed to show checked when selected all non-disabled rows.

src/core/Table/Table.tsx
bentleyvk: Changed.

CHANGELOG.md
bentleyvk: Changed.

stories/core/Table.stories.tsx
bentleyvk: Resolved?

src/core/Table/Table.tsx
bentleyvk: It is controlled completely by us, so user can't do anything here.

mayank99 (approved):

PR #188, opened 2021-07-02
Change `useResizeObserver` to return ref instead of accepting it

mayank99:
Changed hook implementation to return a callback ref instead of accepting an object ref as param. This makes it more reliable to respond to ref value changes.

Updated reference in `HorizontalTabs`.

Also updated tests.


src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -6,29 +6,33 @@ import React from 'react';
...
+ * @example
+ * const elementRef = React.useRef<HTMLDivElement>(null);
+ * const { height, width } = useResizeObserver(elementRef.current);
+ * ...
+ * return <div ref={elementRef}>...</div>;
  */
-export const useResizeObserver = <T extends HTMLElement>(
-  elementRef: React.RefObject<T>,
-) => {
+export const useResizeObserver = <T extends HTMLElement>(element: T | null) => {
```
bentleyvk, line 18: Do you have an example with the issue of using ref?

veekeys: Yeah, it is the weird one which imodeljs users reported..
Basically it is connected with dynamic components, which rendered conditionally in some parent component, which has the ref. Firstly it can be null and then appear, which does not change the ref, but ref.current is not caught in our observer..

veekeys (dismissed):

src/core/utils/hooks/useResizeObserver.tsx
mayank99: > Lets change implementation to the same as useIntersection

Why? Current implementation makes more sense.

src/core/utils/hooks/useResizeObserver.tsx
```diff
@@ -6,29 +6,33 @@ import React from 'react';
...
+    width: element?.getBoundingClientRect().width,
   }));
 
   const resizeObserver = React.useRef<ResizeObserver | null>(null);
 
   React.useLayoutEffect(() => {
     resizeObserver.current?.disconnect();
-    if (elementRef.current) {
+    if (element) {
       resizeObserver.current = new ResizeObserver(([{ contentRect }]) =>
```
veekeys, line 29: Also, need to check if ResizeObserver API is available.. Not only for IE, but also for client apps tests (so they dont start failing randomly). Check in `useIntersection` hook..Its handled there.
image

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -82,7 +83,17 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
   const tablistRef = React.useRef<HTMLUListElement>(null);
 
-  const tablistSize = useResizeObserver(tablistRef);
+  const [tablistSize, setTablistSize] = React.useState(() => ({
+    height: tablistRef.current?.getBoundingClientRect().height,
+    width: tablistRef.current?.getBoundingClientRect().width,
+  }));
+  const updateTablistSize = React.useCallback((s) => {
+    setTablistSize(s);
+    console.log(s);
```
veekeys, line 92: ups

mayank99: 🤫

bentleyvk (approved):

PR #204, opened 2021-07-14
Update to itwinui-css 0.24.1

mayank99:
- Bumped itwinui-css from 0.23.0 to 0.24.1
- Added class to menu item
- Updated unit tests
- Updated images


src/core/Select/Select.tsx
```diff
@@ -270,11 +270,10 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
       <DropdownMenu
         menuItems={menuItems}
         placement='bottom-start'
-        className={menuClassName}
+        className={cx('iui-scroll', menuClassName)}
```
veekeys, line 273: Why we needed separate class for this? Wouldnt it make sense that its always scrollable?

mayank99: There must be some reason that previously we didn't even have it in the css.

Maybe Select makes sense to always be scrollable, but not other Menus.

veekeys: which ones not?

mayank99: idk, just a normal menu.

I'm not seeing the problem here. All I did was move `overflow` from inline style into a css class.
image

We could make it default but that will be a separate css release.

veekeys (approved):

PR #207, opened 2021-07-19
Add keyboard navigation to tabs

### LINKED ISSUE #177: HorizontalTabs: add arrow-key keyboard navigation
#### Feature
For keyboard accessibility, the user should be able to use the left & right arrow keys to activate the tabs. The tabs also need some type of focus indication so the user knows when the left/right arrow key will activate the adjacent tab.

mayank99:
- Closes #177 
- Added `focusActivationMode` prop to toggle whether tabs should automatically be [selected when focused](https://www.w3.org/TR/wai-aria-practices-1.2/#examples-12).
- Renamed `currentIndex` to `currentActiveIndex` to prevent confusion with `focusedIndex`.
- Changed references of `typeof` and `JSX.Element` to use `React.isValidElement` instead.
- Bumped itwinui-css to 0.24.2
- Misc cleanup.

## Checklist


src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -112,24 +108,35 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
   React.useLayoutEffect(() => {
-    if (activeIndex != null && currentIndex !== activeIndex) {
-      setCurrentIndex(getValidIndex());
+    if (activeIndex != null && currentActiveIndex !== activeIndex) {
+      setCurrentActiveIndex(getBoundedValue(activeIndex, 0, labels.length - 1));
     }
-  }, [activeIndex, currentIndex, getValidIndex]);
+  }, [activeIndex, currentActiveIndex, labels.length]);
 
+  const [stripeStyle, setStripeStyle] = React.useState<React.CSSProperties>({});
```
veekeys, line 122: move this up, close to other vars of useState, diff wuld also be nicer shown

mayank99: I moved it down closer to the useEffect which uses this state

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -143,7 +150,55 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
     if (onTabSelected) {
       onTabSelected(index);
     }
-    setCurrentIndex(index);
+    setCurrentActiveIndex(index);
+  };
+
+  const onKeyDown = (event: React.KeyboardEvent<HTMLUListElement>) => {
+    // alt + left/right is used by browser / assistive technologies
+    if (event.altKey) {
```
veekeys, line 158: oh, this we might need to all the components then? need to research...

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -143,7 +150,55 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
+      return;
+    }
+
+    const isTabDisabled = (index: number) => {
```
veekeys, line 162: I would prefer this out, than function in function in function...

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -143,7 +150,55 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
+      const tab = labels[index];
```
veekeys, line 163: `labels` could be passed as strings, no? This would not work then

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -143,7 +150,55 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
+    const focusTabAt = (index: number) => {
+      setFocusedIndex(index);
+      focusActivationMode === 'auto' && onTabClick(index);
+    };
+
+    let newIndex = focusedIndex ?? currentActiveIndex;
+    switch (event.key) {
+      case 'ArrowRight': {
+        do {
+          newIndex = (newIndex + 1 + labels.length) % labels.length;
```
veekeys, line 176: when you go right from the latest ,it should go to first? not sure...

mayank99: That is the W3 recommendation

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -161,30 +216,32 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
         ref={refs}
+        onKeyDown={onKeyDown}
         {...rest}
       >
         {labels.map((label, index) => {
           const onClick = () => onTabClick(index);
           return (
             <li key={index}>
-              {typeof label === 'string' ? (
+              {!React.isValidElement(label) ? (
```
veekeys, line 226: how this check changes/helps than before?

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: That's why I'm checking `isValidElement` in next line

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: It is a typeguard (implemented using `is` keyword), so there is no need to cast to `JSX.Element` anymore.

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: I think DatePicker/TimePicker is the only one where we are handling left/right keys.

Other than that, Menu is the only one where we are handling up/down keys.

mayank99: This thread may be relevant: https://github.com/GoogleChromeLabs/howto-components/issues/19

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: Honestly, I think we should always check for this before cloning (in all places)

src/core/HorizontalTabs/HorizontalTabs.tsx
mayank99: I like keeping small helper functions confined to the scope they are used in.

We are doing it the same way in Menu:
https://github.com/iTwin/iTwinUI-react/blob/8f3e071c95016dd7722c5b3767460d647c2ccf04/src/core/Menu/Menu.tsx#L58-L59

src/core/HorizontalTabs/HorizontalTabs.tsx
veekeys: stripeStyle looks still randomly placed to me...

veekeys: ah, you put there because of setter..well..ok

src/core/HorizontalTabs/HorizontalTabs.tsx
```diff
@@ -143,7 +150,55 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
     if (onTabSelected) {
       onTabSelected(index);
     }
-    setCurrentIndex(index);
+    setCurrentActiveIndex(index);
```
veekeys, line 153: I think clicked tab should also be focused. I do that in date picker and also thats how w3 works. Now when you click with mouse and use arrow on keyboard, active tab firstly gets focused (you do not move...).

veekeys: And another thing, the styling... When you select tab with keyboard, outline should not be visible (act same as mouse click).
image

mayank99: clicked tab also updates focused index now, good catch.

Styling looks fine to me, but if we want to update, it would go in css.

veekeys (approved):

PR #210, opened 2021-07-20
Improve storybook UX

mayank99:
Added storybook addon-essentials, replacing individual addons for docs, controls and actions.

Includes new viewport and background switcher.
toolbar

Renamed Story tab to Code.
image

Changed Docs tab to respect OS theme, and updated background color for non-inline stories.


.storybook/itwinTheme.js
```diff
@@ -51,7 +51,7 @@ export const darkTheme = create({
 
   // UI
   appBg: '#2D373C',
-  appContentBg: '#4F5D65',
+  appContentBg: 'rgb(90, 105, 115)', // background-1
```
veekeys, line 54: hex would look better....

veekeys: Didnt need to change for both themes?

mayank99: This is for Docs tab. Previous background looked really bad in Dark theme, but was ok in Light theme.

veekeys (approved):

.storybook/itwinTheme.js
veekeys: Its not only for docs..its for main too, from what I remember

mayank99: Yeah but it doesn't look that bad in main tab.

Light theme:
image

Dark theme Docs (current):
image

Dark theme Docs (previous) notice lack of contrast:
image

Dark theme Canvas (current):
image

Dark theme Canvas (previous):
image

PR #212, opened 2021-07-21
Add new `Breadcrumbs` component

### LINKED ISSUE #79: Breadcrumb component
image

mayank99:
Added new `Breadcrumbs` component. Closes #79 

Supports styling breadcrumb-items as `Button` or `a` or any arbitrary element/component (e.g. a react-router `Link`).
image
image

Supports custom separator through prop.
image

Handles overflow based on available width (and responds to resizes through `useResizeObserver`).
image

Also added story example for folder navigation.
folders

## Checklist


mayank99:
@veekeys Can you do an initial review before I implement overflow logic?

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import { useTheme } from '../utils/hooks/useTheme';
+import { CommonProps } from '../utils/props';
+import { SvgChevronRight } from '@itwin/itwinui-icons-react';
```
veekeys, line 9: Full path please

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
...
+   * Index of the currently active breadcrumb.
+   * Defaults to last index in the items array.
+   */
+  currentIndex?: number;
+  /**
+   * Breadcrumb items.
+   */
+  items: JSX.Element[];
+  /**
+   * Specifiy a custom separator element to show between breadcrumb items.
```
veekeys, line 23: mistype

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
...
+   */
```
veekeys, line 24: Can mention that default arrow is provided

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
...
+    currentIndex = items.length - 1,
+    separator,
+    className,
+    ...rest
+  } = props;
+
+  useTheme();
+
+  const Separator = () => (
+    <li className='iui-breadcrumbs-separator' aria-hidden>
```
veekeys, line 57: Not sure if `aria-hidden` should go on li element... Usually its applied on svg, will need to read more..

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,90 @@
...
+      {separator ?? <SvgChevronRight />}
+    </li>
+  );
+
+  return (
+    <nav className='iui-breadcrumbs' {...rest}>
+      <ol className={cx('iui-breadcrumbs-list', className)}>
```
veekeys, line 64: Creating `Modal` component situation, where `style` and `className` are applied on different parts of the component.

stories/core/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,121 @@
...
+            Item 2
+          </BreadcrumbItem>,
+        ]
+      }
+      {...rest}
+    />
+  );
+};
+
+export const Links: Story<BreadcrumbsProps> = ({ items, ...rest }) => {
```
veekeys, line 57: hm.. This height change between using Button and anything else could be a concern....

src/core/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+type BreadcrumbItemOwnProps<E extends 'a' | 'button' | 'span'> = {
+  /**
+   * Control which element the breadcrumb item should use.
+   *
+   * Use 'button' if you specify an `onClick` handler.
+   * Use 'a' if you also specify a meaningful `href`.
+   * Use 'span' if the item should not be clickable (e.g. for the current item).
+   * @default 'button'
+   */
+  element?: E;
```
veekeys, line 21: I also wonder if we leave option to set the element to the user, wont we get hybrid approaches which might not look good? like button + anchor... Maybe we should ask element/type on the `Breadcrumbs` component instead...Not sure yet

src/core/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+  props: BreadcrumbItemProps<E>,
+) => {
+  const { className, children, element = 'button', ...rest } = props;
+
+  const Element = element === 'button' ? Button : element;
+
+  useTheme();
+
+  return (
+    <Element
```
veekeys, line 58: Other libs advertise a lot integration with react-router... Would it work here too?

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Oh wow. Fixed 😅

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Added

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: 🤫

src/core/Breadcrumbs/BreadcrumbItem.tsx
mayank99: Added React.ComponentType so it will work now

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: The item needs to be hidden from the list, otherwise it will mess with screen readers.

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Moved classname to root

stories/core/Breadcrumbs.stories.tsx
mayank99: Yeah I'm not sure why the height is smaller for non-anchor elements. But maybe that's ok if it gets vertically centered when placed to a taller element.

@FlyersPh9 can comment here

src/core/Breadcrumbs/BreadcrumbItem.tsx
veekeys: Does that mean that simple 'div' would not work?

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+   * Defaults to the `SvgChevronRight` icon.
+   */
+  separator?: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * A breadcrumb trail is used as a navigational aid to help users keep track
+ * of their place in the application. It is often placed before a page's main content.
+ *
+ * For accessibility, make sure to provide an aria-label describing the type of navigation.
```
veekeys, line 35: Stories code does not have aria-label. Maybe we could add some default value? I see 'Breadcrumb' in w3 example.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+export type BreadcrumbsProps = {
+  /**
+   * Index of the currently active breadcrumb.
+   * Defaults to last index in the items array.
+   */
+  currentIndex?: number;
+  /**
+   * Breadcrumb items.
+   */
+  items: JSX.Element[];
```
bentleyvk, line 23: Why not `children`?

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+    const refs = useMergedRefs(breadcrumbsRef, resizeRef, ref);
+
+    React.useLayoutEffect(() => {
+      if (!breadcrumbsRef.current) {
+        return;
+      }
+
+      // hide items when there's no space available
+      if (
+        breadcrumbsRef.current.offsetWidth < breadcrumbsRef.current.scrollWidth
```
bentleyvk, line 81: Invert this if condition to avoid deep nesting.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+      [],
+    );
+    const [resizeRef] = useResizeObserver(updateBreadcrumbsWidth);
+    const refs = useMergedRefs(breadcrumbsRef, resizeRef, ref);
+
+    React.useLayoutEffect(() => {
+      if (!breadcrumbsRef.current) {
+        return;
+      }
+
```
bentleyvk, line 78: If there is not enough space, maybe show just a current breadcrumb.
image

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+      ) {
+        if (postItems.length > 0) {
+          setHiddenItems([...hiddenItems, postItems.shift()!]); // eslint-disable-line @typescript-eslint/no-non-null-assertion
+          setPostItems(postItems);
+          overflowBreakpoints.current.push(breadcrumbsWidth);
+        }
+      }
+      // restore items when there's enough space again
+      else if (
+        overflowBreakpoints.current.length > 0 &&
```
bentleyvk, line 91: Huh? Why you need this variable?

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+      if (!breadcrumbsRef.current) {
+        return;
+      }
+
+      // hide items when there's no space available
+      if (
+        breadcrumbsRef.current.offsetWidth < breadcrumbsRef.current.scrollWidth
+      ) {
+        if (postItems.length > 0) {
+          setHiddenItems([...hiddenItems, postItems.shift()!]); // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
bentleyvk, line 84: `shift`, `pop` mutate existing array and it is considered a bad practice. Also, you might need to use callback setter here to avoid random bugs when resizing and it is set many times.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+          overflowBreakpoints.current.push(breadcrumbsWidth);
+        }
+      }
+      // restore items when there's enough space again
+      else if (
+        overflowBreakpoints.current.length > 0 &&
+        breadcrumbsWidth >
+          overflowBreakpoints.current[overflowBreakpoints.current.length - 1]
+      ) {
+        if (hiddenItems.length > 0) {
```
bentleyvk, line 95: This if condition can be moved to the parent `if`.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+              : undefined,
+        })}
+      </li>
+    );
+
+    return (
+      <nav className={cx('iui-breadcrumbs', className)} ref={refs} {...rest}>
+        <ol className='iui-breadcrumbs-list'>
+          <ListItem item={items[0]} index={0} />
+          <Separator />
```
bentleyvk, line 134: If there is just one item, then separator will be still shown.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+        {React.cloneElement(item, {
+          'aria-current':
+            item.props['aria-current'] ?? currentIndex === index
+              ? 'location'
+              : undefined,
+        })}
+      </li>
+    );
+
+    return (
```
bentleyvk, line 130: Who should be responsible for this menu? We or user?
image

src/core/Breadcrumbs/Breadcrumbs.tsx
bentleyvk: Found in UX page:
image

src/core/Breadcrumbs/BreadcrumbItem.tsx
```diff
@@ -0,0 +1,75 @@
...
+export type BreadcrumbItemProps<
+  E extends 'a' | 'button' | 'span' | React.ComponentType
+> = BreadcrumbItemOwnProps<E> &
+  Omit<React.ComponentPropsWithoutRef<E>, keyof BreadcrumbItemOwnProps<E>>;
+
+/**
+ * Individual breadcrumb component to be used in the `Breadcrumbs` items array.
+ * @example
+ * const items = [
+ *   <BreadcrumbItem>Root</BreadcrumbItem>,
```
bentleyvk, line 42: Maybe worth adding `onClick`.

src/core/Breadcrumbs/BreadcrumbItem.tsx
bentleyvk: But current breakcrumb is inactive and is of different type e.g. span so we can't control that. User can also pass completely different component, not `BreadcrumbItem`, we can't control everything.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,158 @@
...
+      separator,
+      className,
+      ...rest
+    } = props;
+
+    useTheme();
+
+    const breadcrumbsRef = React.useRef<HTMLElement>(null);
+
+    const [hiddenItems, setHiddenItems] = React.useState<JSX.Element[]>([]);
```
bentleyvk, line 62: I am just thinking, do you really need `hiddenItems`? You are not showing them anythere. You kind of only need to know how many items to show and can `slice` from original array.

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Maybe 🤔

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Not sure what you mean

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: It's to keep track of the breakpoints where we hide one or more items, and use those to make them visible again (when resized and there is enough space again).

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: That menu is not really part of the component, so I would say user.

src/core/Breadcrumbs/BreadcrumbItem.tsx
mayank99: Added

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Added

src/core/Breadcrumbs/Breadcrumbs.tsx
bentleyvk: Instead of nesting ifs, you could write
```tsx
if (breadcrumbsRef.current.offsetWidth >= breadcrumbsRef.current.scrollWidth) {
    return;
}
```

mayank99: Where's nesting?

the `postItems.length > 0` and `hiddenItems.length > 0` have nothing to do with this

src/core/Breadcrumbs/Breadcrumbs.tsx
veekeys: Could you remove that 'Primary' aria-label from examples? Maybe dont need to mention anything about it.. or rephrase it somehow, that there is the default one, but you can override.

mayank99: I think it should always be overridden by the user. "Breadcrumb" is not descriptive enough.

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Changed to use `visibleCount`

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Changed whole code so this is gone 💥

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Changed whole code so this is gone 💥

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Fixed

src/core/Breadcrumbs/BreadcrumbItem.tsx
mayank99: We only want users to be passing these 3 elements:
image

For anything else, it should be custom component which needs to handle styles.

If we allow passing any arbitrary element, I'm afraid they will break things.

src/core/Breadcrumbs/Breadcrumbs.tsx
veekeys: Well, I find mostly info only with "Breadcrumb". SO I think this is valid value for most of the cases. What are others?

mayank99: Breadcrumb is a generic value. Users need to add extra context, e.g. "Primary breadcrumb" or "Table breadcrumb" or something.

src/core/Breadcrumbs/Breadcrumbs.tsx
bentleyvk: Really? 😃
image

mayank99: There was a calculation error which I just fixed. Pull latest

src/core/Breadcrumbs/Breadcrumbs.tsx
bentleyvk: Is it possible that user navigated deeply into one folder, breakpoints were saved, then navigated to completely different folder with different names and old breakpoints will be used with wrong values?

mayank99: Yeah, that's possible. Maybe we need to clear breakpoints when items change?

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Added edge case for one item.

src/core/Breadcrumbs/Breadcrumbs.tsx
veekeys: Do you have any reference for that?
Usually you should have single breadcrumb in the page, no?

mayank99: > Do you have any reference for that?

I can't find any "guidelines" because they all talk about breadcrumbs in isolation, rather than as a part of the site. But I did find a couple examples.

See `nav` on top of github.com 
image

See breadcrumbs on https://www.harvard.edu/about-harvard/harvard-in-the-community/
image

> Usually you should have single breadcrumb in the page, no?

Not necessarily. We already have breadcrumbs in our `Header`. So if this component is used somewhere in the page, it will definitely need some other label.

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Added another effect hook to reset overflow when items change.

It's located above this main overflow effect hook, so this will re-run and calculate correct widths/breakpoints after items change.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,166 @@
...
+export type BreadcrumbsProps = {
+  /**
+   * Index of the currently active breadcrumb.
+   * Defaults to the index of the last breadcrumb item.
+   */
+  currentIndex?: number;
+  /**
+   * Breadcrumb items.
+   */
+  children: JSX.Element[];
```
bentleyvk, line 23: It looks like when its type is `JSX.Element[]`, you don't need to use `React.Children`. Or you can change to `ReactNode`.

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,166 @@
...
+    return (
+      <nav
+        className={cx('iui-breadcrumbs', className)}
+        ref={refs}
+        aria-label='Breadcrumb'
+        {...rest}
+      >
+        <ol className='iui-breadcrumbs-list'>
+          {visibleCount > 1 && <ListItem index={0} />}
+          {visibleCount > 1 && <Separator />}
```
bentleyvk, line 133: Maybe wrap with fragment and put item with separator under same condition.

src/core/Breadcrumbs/Breadcrumbs.tsx
bentleyvk: I still don't like this variable but it works 😃 Is it possible to do the same but saving just a previous width? If it takes too much time then don't bother, leave as it is because it is working and doesn't affect API.

stories/core/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  subcomponents: { BreadcrumbItem },
+  title: 'Core/Breadcrumbs',
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+    id: { control: { disable: true } },
+    children: { control: { disable: true } },
+  },
+} as Meta<BreadcrumbsProps>;
+
```
bentleyvk, line 23: Maybe add story with dropdown menu and folder icon, just to show how to achieve this and what icon to use.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+`Date`
+
+### What's new
+
+- **Added new `Breadcrumbs` component.**
+  - Supports customizing elements in `BreadcrumbItem` and handles overflow based on available space.
```
bentleyvk, line 10: Is this single subitem really needed?

src/core/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,166 @@
+/*---------------------------------------------------------------------------------------------
```
bentleyvk, line 1: I was looking for other examples and found Microsoft lib. It has the same styling for anchor and button breadcrumbs. Maybe we should do the same? @FlyersPh9 https://developer.microsoft.com/en-us/fluentui#/controls/web/breadcrumb

Off topic, but I really like their demo pages with a short description and best practices for the component.

CHANGELOG.md
mayank99: What?

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: > Is it possible to do the same but saving just a previous width? 

No, I need to save all the widths to be able to restore all of them. The alternative would be recalculating and relayout-ing everything on every single resize.

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Done

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: Changed to ReactNodeArray

stories/core/Breadcrumbs.stories.tsx
mayank99: Can we release without this story?

CHANGELOG.md
bentleyvk: image

mayank99: Oh, are you talking about the formatting? I initially had in the same line but it was too long.

Maybe I can break that down into two sub-items

stories/core/Breadcrumbs.stories.tsx
bentleyvk: Is it really that much work? Just show all items and that's all.

mayank99: Yeah I can do that now. I was hoping we'd have released and demoed today

CHANGELOG.md
bentleyvk: Just we are not using sub-items anywhere (maybe we should).

mayank99: https://github.com/iTwin/iTwinUI-react/blob/333640cec2f78702d3cf210aeaf9f60c238c2005/CHANGELOG.md#L80-L81

https://github.com/iTwin/iTwinUI-react/blob/333640cec2f78702d3cf210aeaf9f60c238c2005/CHANGELOG.md#L136-L138

stories/core/Breadcrumbs.stories.tsx
mayank99: Added. See gif in description.

The code is not very nice and does not handle all cases but it's something 🤷‍♀️

mayank99: I'm using simple dropdown menu items for now.
image

@FlyersPh9 Please tell me this atrocity is an outdated standard:
image

stories/core/Breadcrumbs.stories.tsx
FlyersPh9: We could set a height on anchor breadcrumbs so that the height is equal to the buttons breadcrumb if you'd like, should be no issue.

stories/core/Breadcrumbs.stories.tsx
FlyersPh9: That's been removed from our local version of documentation, it can be ignored. New image below:
anatomy-breadcrumbs-select

src/core/Breadcrumbs/Breadcrumbs.tsx
FlyersPh9: Personally I don't mind the different visuals, but we could style the anchor link within a breadcrumb to look and interact more like a button.

stories/core/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,164 @@
...
+          {item}
+        </BreadcrumbItem>
+      )),
+    [items, lastIndex],
+  );
+
+  return (
+    <div style={{ display: 'inline-flex', width: 450 }}>
+      {isEditing ? (
+        <Input
```
bentleyvk, line 127: I thought we will handle editing as it is in standards. Ofc it is not needed in initial implementation so I don't think this should appear in a story.

mayank99: But you asked me to add a story for this folder navigation pattern...

We cannot handle editing in Breadcrumbs because it is literally a different component. This story is just showing how the user might be able to do it.

bentleyvk (approved):

src/core/Breadcrumbs/BreadcrumbItem.tsx
mayank99: Simple div will work now after the latest css changes

stories/core/Breadcrumbs.stories.tsx
mayank99: height is now consistent after latest changes

src/core/Breadcrumbs/BreadcrumbItem.tsx
mayank99: Removed breadcrumb item

src/core/Breadcrumbs/Breadcrumbs.tsx
mayank99: After some discussion, we've decided to keep the different visuals but were able to remove the `BreadcrumbItem` component (by improving the CSS). The user can now pass any kind of element and we'll handle most cases properly.

And I've saved that FluentUI documentation link in #226.

veekeys (approved):

PR #217, opened 2021-07-26
Use inline css variable for `Badge` background-color

### LINKED ISSUE #194: Badge : Add some preset theme color and make backgroundColor props optional
#### Feature
I want to use Badge, I want to use some color when light theme and other color when dark theme. Right now, there is "no way" to do that. If I set a classname, I'm required to leave a backgroundColor and !important my way to override it or I need to code the color props send to the component depending on the theme. 

### My Request : 
- Add default theme to badge. 
  - Something like : Positive, Negative, Warning, Black, White, and the data-viz color as preset option. 
- Add new TextColor and Border props as optional, make backgroundColor props as optional too. 
  - It's bad practice to !important CSS rule but the required props force that.
- If you want to set color, your own classname or style props should do enough.
  - Leave CSS props as shortcut, but not required. 
- Badge should have default color if nothing is set. e.g. `<Badge>Label<Badge>`

### Exemple :
```.tsx
<Badge backgroundColor="#ffffff" border="1px solid black" textColor="black">
  Label
</Badge>
<Badge style={{backgroundColor:"#ffffff", border:"1px solid black", textColor:"black"}}>
  Label
</Badge>
<Badge theme="Positive">
  Label
</Badge>
<Badge className="CustomBadge">
  Label
</Badge>
```

mayank99:
Changed Badge backgroundColor prop to:
- be optional (shows default color through CSS)
- support preset colors for statuses and soft backgrounds (with strong typing)
- set inline css variable instead of inline background-color property. This makes it possible to override the background-color via css.

Also:
- Added new object and function for handling preset badge background values.
- Added fallback for IE.
- Added unit tests.
- Added new story.

## Checklist


src/core/Badge/Badge.tsx
```diff
@@ -28,13 +29,16 @@ export type BadgeProps = {
  */
 export const Badge = (props: BadgeProps) => {
   const { backgroundColor, style, className, children, ...rest } = props;
+
   useTheme();
+
+  const _style =
```
bentleyvk, line 35: Why this change is needed?

bentleyvk: Oh, I saw the linked issue. Won't CSS variable make it to have all Badges the same color?

veekeys: I do not understand this change. I think the issue here was, that we require `backgroundColor `property for it (so people cannot use className on top and override the value). Now we introduce some additional complexity with css variable?? Which as Vilius mentioned, gonna make all badges look same color.

mayank99: CSS variables are scoped, so every badge can have its own.

It does not fully address the problem, but it is the first step towards fixing the linked issue. Inline background-color is impossible to override using css without `!important` so this will help.

veekeys: I do not still understand, how this fixes the issue.. You still cannot use class.
Why not just make prop optional?

mayank99: > You still cannot use class.
?? Sure you can.

> Why not just make prop optional?
Because we don't have a default background-color, so badge looks broken.

veekeys: amm... idk, looks like inner hack to allow simple thing here..

mayank99: I don't see css variable as a hack.

Rather, I would say setting inline styles in a React-only package is more hacky.

veekeys: Well, are we planning to document anything about it? no.
By default we are setting css variable with an empty value, right?
So actually, we are providing from the CSS library the component, without background color.
Now, we are still leaving mandatory `backgroundColor `prop and then say what.. put class to override it? It just looks like something is wrong in there...

mayank99: Something is definitely wrong. That is why I said "this is the first step".

I was talking with @FlyersPh9 about setting a default value, and he suggested skyblue or ash, both of which seemed random to me. I think we could arrive at a more neutral color and use that as default.

veekeys: so whats the "second step" then? Just figure out the default color and make that optional? In that case, would we even need css variable?

mayank99: Yeah I think that is the second step, but @FlyersPh9 and I were also talking about redoing the whole data-viz color system 😅

CSS variable still helps in cases when you want to override the prop through CSS. Inline styles always make it hard.

mayank99: Converting to draft because I'm going to make some big changes in CSS first.

src/core/Badge/Badge.tsx
```diff
@@ -24,17 +59,24 @@ export type BadgeProps = {
...
+ * <Badge backgroundColor="sunglow">Label</Badge>
+ * <Badge backgroundColor="positive">Label</Badge>
  */
 export const Badge = (props: BadgeProps) => {
   const { backgroundColor, style, className, children, ...rest } = props;
+
   useTheme();
+
+  const _style =
+    getWindow()?.CSS &&
```
bentleyvk, line 72: Can you replace with:
```ts
backgroundColor && getWindow()?.CSS?.supports(...)
```

src/core/Badge/Badge.tsx
```diff
@@ -7,13 +7,48 @@ import cx from 'classnames';
...
+    case 'primary':
+      return '#A5D7F5';
+    case 'positive':
+      return '#C3E1AF';
+    case 'negative':
+      return '#EFA9A9';
+    case 'warning':
+      return '#F9D7AB';
+    default:
+      if (isSoftBackground(color)) {
```
bentleyvk, line 30: Wouldn't it be simpler just by doing:
```ts
return SoftBackgrounds[color] ?? color;
```

src/core/Badge/Badge.tsx
```diff
@@ -7,13 +7,48 @@ import cx from 'classnames';
...
+   * If not specified, a default neutral background will be used.
    */
-  backgroundColor: string;
+  backgroundColor?:
+    | 'primary'
+    | 'positive'
+    | 'negative'
+    | 'warning'
+    | keyof typeof SoftBackgrounds
+    | (string & {}); // eslint-disable-line @typescript-eslint/ban-types -- see https://github.com/Microsoft/TypeScript/issues/29729
```
bentleyvk, line 51: Maybe add a comment that it allows predefined values and custom strings, also intellisense is working.

src/core/Badge/Badge.tsx
mayank99: Think this can be resolved now

src/core/Badge/Badge.tsx
mayank99: Changed

src/core/Badge/Badge.tsx
mayank99: Added

src/core/Badge/Badge.tsx
mayank99: The type guard is needed to prevent typescript error

bentleyvk (approved):

PR #220, opened 2021-07-27
Fix Slider bug where onChange was not always called after pointerUp

### LINKED ISSUE #221: Slider does not fire onChange event when last pointerMove value matches pointerUp value
When using new Slider in iModeljs, I found that the Slider does not fire its onChange event when last pointerMove value matches pointerUp value.

#### Expected behavior
onChange event should always be triggered on pointerUp event for slider.

#### Reproduction
Link to a minimal repro: 

##### Steps to reproduce
1. click on slider thumb
2. move thumb
3. release mouse button
4. notice onChange callback is not always fired

bsteinbk:
- Fix bug where pointerUp did not trigger onChange processing when pointerUp matches last pointerMove position. 
- Add Arrow Up/Down support to mimic Arrow Right/Left processing
- Improve eventlistener registration by using new `useEventListener` hook.

Closes #221

## Checklist


src/core/Slider/Slider.tsx
```diff
@@ -12,6 +12,47 @@ import { getBoundedValue } from '../utils/common';
 import { Track } from './Track';
 import { Thumb } from './Thumb';
 
+// based on https://usehooks.com/useEventListener/
+function usePointerEventListener(
```
mayank99, line 16: This could be a generic hook under `utils/`. There is no PointerEvent-specific logic here.

src/core/Slider/Slider.tsx
```diff
@@ -12,6 +12,47 @@ import { getBoundedValue } from '../utils/common';
...
+  handler: (event: PointerEvent) => void,
+  element: HTMLElement | Document | undefined,
+) {
+  // Create a ref that stores handler
+  const savedHandler = React.useRef<(event: PointerEvent) => void>();
+
+  // Update ref.current value if handler changes.
+  // This allows our effect below to always get latest handler ...
+  // ... without us needing to pass it in effect deps array ...
+  // ... and potentially cause effect to re-run every render.
```
mayank99, line 27: This comment can be shortened and combined with the comment above.

src/core/Slider/Slider.tsx
```diff
@@ -12,6 +12,47 @@ import { getBoundedValue } from '../utils/common';
...
+    savedHandler.current = handler;
+  }, [handler]);
+
+  React.useEffect(
+    () => {
+      // Make sure element supports addEventListener
+      const isSupported = element && element.addEventListener;
+      if (!isSupported) {
+        return;
+      }
```
mayank99, line 38: Is this needed? https://caniuse.com/addeventlistener

Could just do:
```suggestion
      if (!element) {
        return;
      }
```

src/core/Slider/Slider.tsx
```diff
@@ -12,6 +12,47 @@ import { getBoundedValue } from '../utils/common';
...
+      const eventListener = (event: PointerEvent) =>
+        savedHandler.current?.(event);
+
+      // Add event listener
+      element?.addEventListener(eventName, eventListener);
+
+      // Remove event listener on cleanup
+      return () => {
+        element?.removeEventListener(eventName, eventListener);
+      };
```
mayank99, line 50: This could be simplified:
```suggestion
      element?.addEventListener(eventName, savedHandler.current);
      return () => element?.removeEventListener(eventName, savedHandler.current);
```

Or does that not work?

src/core/Slider/Slider.tsx
```diff
@@ -272,6 +313,7 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
       (
         event: PointerEvent,
         callback?: (values: ReadonlyArray<number>) => void,
+        skipValueChangeCheck?: boolean,
```
mayank99, line 316: Where will this be false?

src/core/Slider/Thumb.tsx
```diff
@@ -90,9 +90,11 @@ export const Thumb = (props: ThumbProps) => {
       }
       switch (event.key) {
         case 'ArrowLeft':
+        case 'ArrowDown':
           onThumbValueChanged(index, Math.max(value - step, minVal));
           return;
         case 'ArrowRight':
+        case 'ArrowUp':
```
mayank99, line 97: Not sure about this change. Up is not the same as Right, and Down is not the same as Left.

mayank99 (commented):
Thanks for the PR.

Could you also create an issue to link to this PR? Doesn't need to be fully detailed, it's just to help us track tasks.

src/core/Slider/Thumb.tsx
bsteinbk: These new arrow cases were added to mimic what was supported in react-compound-slider. I actually ran into a unit test that used the down arrow to move thumb to left (ie lower the thumb value).

src/core/Slider/Slider.tsx
bsteinbk: I did move it to utils/hook/usePointerEventListener,  but the handler type is 'PointerEvent-specific'

mayank99: Yeah I'm saying make it a generic `useEventListener`

src/core/Slider/Thumb.tsx
mayank99: Okay, I checked the native slider and W3 variant and they both use Up/Down too, so this is fine.

I would add return early if `altKey` is pressed though. See example in https://github.com/iTwin/iTwinUI-react/blob/main/src/core/Menu/MenuItem.tsx#L107

src/core/Slider/Slider.tsx
mayank99: Okay I think see the reasoning now (from the original source): https://overreacted.io/making-setinterval-declarative-with-react-hooks/#refs-to-the-rescue

I would still remove those redundant comments at least.

src/core/Slider/Slider.tsx
bsteinbk: comments combined

src/core/Slider/Slider.tsx
bsteinbk: I should have done this initially... done now.

src/core/Slider/Slider.tsx
bsteinbk: updated

src/core/Slider/Slider.tsx
bsteinbk: comments removed

src/core/Slider/Slider.tsx
bsteinbk: renamed `skipValueChangeCheck` to `forceCallbackToRun`

src/core/Slider/Thumb.tsx
bsteinbk: done

mayank99 (approved):
LGTM!

Would still like @veekeys or @bentleyvk to review this before merging.

src/core/Slider/Thumb.tsx
```diff
@@ -88,11 +88,16 @@ export const Thumb = (props: ThumbProps) => {
       if (disabled) {
```
veekeys, line 88: Can add here `disabled || event.altKey`. Will save few lines

src/core/Slider/Slider.tsx
```diff
@@ -323,7 +328,7 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
 
     const handlePointerUp = React.useCallback(
       (event: PointerEvent) => {
-        updateThumbValue(event, onChange);
+        updateThumbValue(event, onChange, true);
```
veekeys, line 331: If you asked me, I do not think we should raise `onChange` event if the value did not change (for the sliding action, we have `onUpdate` which can inform user that something is happaening, if he cares). Usually if `onChange event is raised, you expect a change and some action, which I do not think here is the case.

bsteinbk: I disagree, because the move action changes the internal value(s) of the slider not the external value(s). This change is required to provide functionality equivalent to the react-compound-slider that was used in iModeljs.

veekeys: If you had value of 50 then you slided to somewhere and without releasing mouse came back to 50, why user should get event of onChange if value did not change at the end???

bsteinbk: As you move the internal array values are being changed maintaining the internal state. for example mouse down at 50 - move - 60 - move - 70 - move 50 - mouse up, the internal array as been updated 3 times. We do not setup an original array of values at mouse down to compare them to values at mouse up.

src/core/Slider/Thumb.tsx
bsteinbk: done

src/core/Slider/Slider.tsx
veekeys: amm.. Ok, I am lost now. This code part does nothing internally:
` if (
            !forceCallbackToRun &&
            pointerValue === currentValues[activeThumbIndex]
          ) {
            return;
          }
          const newValues = [...currentValues];
          newValues[activeThumbIndex] = pointerValue;
          setCurrentValues(newValues);`

It basically sets new array of the same values as it was before (there is a check).
But if you force it, internally it goes through it + calls `onChange` to the user that something has changed (which I assume has not).

bsteinbk: You are right on that account ... I don't need to update the internal state array... but I do need to call the onChange... I can make that change.

bsteinbk: @veekeys I attempted to make the function processing more descriptive... hopefully I did.

src/core/Slider/Slider.tsx
```diff
@@ -163,8 +163,8 @@ export type SliderProps = {
    */
   thumbProps?: (index: number) => React.HTMLAttributes<HTMLDivElement>;
   /**
-   * Callback fired when the value(s) of the slider has changed. This will receive
-   * changes at the end of a slide as well as changes from clicks on rails and tracks.
+   * Callback fired at the end of a thumb move and when there user clicks on rail
```
veekeys, line 166: maybe enough `and when user clicks on rail`? this 'there' does not give anything..does it?

mayank99: I think that's a typo 😆

mayank99: I was also gonna say, maybe we should explicitly mention `pointerUp` in this comment

bsteinbk: I wish I could use the excuse that English is my second language.

src/core/Slider/Slider.tsx
```diff
@@ -284,28 +280,40 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
           pointerValue = roundValueToClosestStep(pointerValue, step, min);
           const [minVal, maxVal] = getAllowableThumbRange(activeThumbIndex);
           pointerValue = getBoundedValue(pointerValue, minVal, maxVal);
-          if (
-            !forceCallbackToRun &&
-            pointerValue === currentValues[activeThumbIndex]
-          ) {
+          if (pointerValue === currentValues[activeThumbIndex]) {
```
veekeys, line 282: `pointerValue === currentValues[activeThumbIndex] && 'onChange' === callbackType` but I am still thinking about the approach here...

bsteinbk: Not sure what you are asking for... I want to return for both callbackTypes, but I just want to fire callback if type is 'onChange'...
```
          if (pointerValue === currentValues[activeThumbIndex]) {
            if ('onChange' === callbackType) {
              onChange?.(currentValues);
            }
            return;
          }
```

veekeys (approved):

PR #228, opened 2021-07-29
Fix for selecting filtered table

### LINKED ISSUE #222: Table: Crashes when selecting a row after filtering
##### Steps to reproduce

1. Select any checkbox 
2. Apply Filter 	
3. Again select any checkbox 	
4. Table component crash and throws exception.

image

bentleyvk:
Closes #222

- Preserving selected state even after filtering.
- Showing indeterminate state for filtered table header checkbox if there are selected rows that are filtered out.
- Added a story with all table features (except lazy-loading).

select_and_filter_fixed

## Checklist


src/types/react-table-config.ts
```diff
@@ -89,7 +89,9 @@ declare module 'react-table' {
       UsePaginationInstanceProps<D>,
       UseRowSelectInstanceProps<D>,
       UseRowStateInstanceProps<D>,
-      UseSortByInstanceProps<D> {}
+      UseSortByInstanceProps<D> {
+    initialRows: Row<D>[];
+  }
```
mayank99, line 94: Formatting messed up here?

src/core/Table/Table.tsx
```diff
@@ -287,13 +287,11 @@ export const Table = <
...
             isRowDisabled?.(row.original),
           );
-          const checked =
-            instance.isAllRowsSelected ||
-            instance.rows.every(
-              (row) =>
-                instance.state.selectedRowIds[row.id] ||
-                isRowDisabled?.(row.original),
-            );
+          const checked = instance.initialRows.every(
```
mayank99, line 290: I can't find any info on `initialRows` in react-table docs. Do you have a link?

bentleyvk: There isn't. Just code:
https://github.com/tannerlinsley/react-table/blob/8ec25cabccf915991ec00cf67bfb26a83aa0c0b2/src/hooks/useTable.js#L213

src/types/react-table-config.ts
bentleyvk: Everything good here.

mayank99: image

src/core/Table/Table.tsx
mayank99: So it's a secret property that magically fixes our bug?

src/types/react-table-config.ts
bentleyvk: It is a property, not an extended interface/type.
image

src/core/Table/Table.tsx
bentleyvk: Kind of. Typings come from different package so they might missed it.

src/types/react-table-config.ts
mayank99: Ohhh the braces are for the main interface, that's confusing lol

src/core/Table/Table.tsx
mayank99: And did you narrow down why the crash was happening? Need to remove both `instance.rows` and `isAllRowsSelected`?

bentleyvk: Of course 😃 All selected rows are being saved to `selectedRowIds` even if table is filtered. Previously I was going through each selected row id and trying to get data from `rowsById` map. However, it only has filtered rows, so it tried to get `.original` from `undefined` row. 
I had to remove `isAllRowsSelected` because it is also checking only in `rowsById` map which is filtered and in order to achieve the same behavior as material-ui, I have to check each initial row.

mayank99 (approved):

PR #229, opened 2021-07-29
Custom table row props

### LINKED ISSUE #216: Table: Add onMouseEnter/onMouseLeave for rows
#### Feature
Please report onMouseEnter and onMouseLeave for rows and let implementers pass callbacks for it to the table component. 

Add to TableProps<T>
```
onMouseEnter?: (event: React.MouseEvent, row: Row<T>) => void 
onMouseLeave?:  (event: React.MouseEvent, row: Row<T>) => void 

```
#### Examples
Highlight elements in a graphical view when hovering the associated data in the table.

bentleyvk:
Closes #216
- Added `rowProps` prop that is a function and user can pass custom props to the row using it.
- Improved `Full` story to add `Tooltip` on each row.
- Fixed `UserIconGroup` `countIconProps` type to support `ref`.
- Fixed `useMergedRefs` hook types to support undefined/null values.
- Tests clean up.

## Checklist


src/core/Table/Table.tsx
```diff
@@ -562,9 +561,10 @@ export const Table = <
                 onBottomReached={onBottomReachedRef}
                 intersectionMargin={intersectionMargin}
                 state={state}
-                key={rowProps.key}
+                key={row.getRowProps().key}
```
bentleyvk, line 564: It is really a lightweight function, so I can call it multiple times:
image

src/core/Table/TableRowMemoized.tsx
```diff
@@ -48,12 +54,26 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
+      className: cx('iui-row', {
+        'iui-selected': row.isSelected,
+        'iui-row-expanded': row.isExpanded && subComponent,
+        'iui-disabled': isRowDisabled,
+      }),
+    }),
+    ...rowProps?.(row.original),
+  };
+
+  const refs = useMergedRefs(rowRef, mergedProps.ref ?? (() => {}));
```
veekeys, line 68: Our hook does not handle undefined ref?

src/core/Table/Table.tsx
```diff
@@ -129,6 +129,11 @@ export type TableProps<
    * If not specified, all rows are enabled.
    */
   isRowDisabled?: (rowData: T) => boolean;
+  /**
+   * Function that should return custom props passed to the each row.
+   * Must be memoized.
+   */
+  rowProps?: (rowData: T) => React.ComponentPropsWithRef<'div'>;
```
veekeys, line 136: hm... you sure about this type?
I see for simple elements examples with `React.DetailedHTMLProps`, just it is really long to write..

src/core/Table/Table.tsx
```diff
@@ -562,9 +561,10 @@ export const Table = <
                 onBottomReached={onBottomReachedRef}
                 intersectionMargin={intersectionMargin}
                 state={state}
-                key={rowProps.key}
+                key={row.getRowProps().key}
                 onClick={onRowClickHandler}
                 subComponent={subComponent}
+                isRowDisabled={!!isRowDisabled?.(row.original)}
```
veekeys, line 567: I guess could call just `disabled` or `isDisabled`, cause its Row component, no need to go long, imo

src/core/Table/TableRowMemoized.tsx
```diff
@@ -48,12 +54,26 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
 
+  const mergedProps = {
+    ...row.getRowProps({
+      className: cx('iui-row', {
+        'iui-selected': row.isSelected,
+        'iui-row-expanded': row.isExpanded && subComponent,
+        'iui-disabled': isRowDisabled,
+      }),
+    }),
+    ...rowProps?.(row.original),
```
veekeys, line 65: Wont this override all our classes, if there is a class passed?

src/core/Table/Table.tsx
bentleyvk: Previously I had `React.HTMLAttributes<HTMLDivElement> & React.RefAttributes<HTMLDivElement>`. Mayank recommended current one: https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#wrappingmirroring-a-html-element.
What `React.DetailedHTMLProps` gives? Because to achieve the same, I have to write: `React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>`

src/core/Table/TableRowMemoized.tsx
bentleyvk: It doesn't. I can add to handle that if we will need the same in the future components.

src/core/Table/Table.tsx
veekeys: Yeah, not sure.. I could not find quickly good info about types.. Just I see `ComponentProps` usually used for components, rather than native elements. But I guess its fine.

src/core/Table/TableRowMemoized.tsx
veekeys: hmm.. that empty func looks not nice here.

src/core/Table/TableRowMemoized.tsx
bentleyvk: Good catch! Fixed.

src/core/Table/Table.tsx
bentleyvk: Renamed.

src/core/Table/TableRowMemoized.tsx
bentleyvk: Handling undefined/null.

veekeys (approved):

src/core/Table/TableRowMemoized.tsx
```diff
@@ -17,14 +18,18 @@ import { CSSTransition } from 'react-transition-group';
  */
 const TableRow = <T extends Record<string, unknown>>(props: {
   row: Row<T>;
-  rowProps: TableRowProps;
```
mayank99, line 20: What was `TableRowProps` giving before?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -17,14 +18,18 @@ import { CSSTransition } from 'react-transition-group';
...
+  rowProps?: (
+    rowData: T,
+  ) => React.HTMLAttributes<HTMLDivElement> &
+    React.RefAttributes<HTMLDivElement>;
```
mayank99, line 24: Update types to `React.ComponentPropsWithRef<'div'>` to match

bentleyvk: Fixed.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -17,14 +18,15 @@ import { CSSTransition } from 'react-transition-group';
  */
 const TableRow = <T extends Record<string, unknown>>(props: {
   row: Row<T>;
-  rowProps: TableRowProps;
+  rowProps?: (rowData: T) => React.ComponentPropsWithRef<'div'>;
```
mayank99, line 21: Now, I wonder if we should also expose `row: Row<T>` instead of just rowData

src/core/Table/TableRowMemoized.tsx
bentleyvk: key, classname, style and role but it only returns key and role

stories/core/Table.stories.tsx
```diff
@@ -1140,17 +1147,42 @@ export const Full: Story<TableProps> = (args) => {
...
+        emptyTableContent='No data.'
+        subComponent={expandedSubComponent}
+        isRowDisabled={isRowDisabled}
+        rowProps={rowProps}
+        isSelectable
+        isSortable
+        {...rest}
+      />
+      <Tooltip
+        reference={hoveredRow ? rowRefMap.current[hoveredRow.name] : undefined}
```
mayank99, line 1181: This seems a bit convoluted. Tippy reference can already handle undefined, so we should be passing `ref` not `ref.current`

stories/core/Table.stories.tsx
```diff
@@ -1088,6 +1089,12 @@ ControlledState.args = { isSelectable: true };
 export const Full: Story<TableProps> = (args) => {
   const { columns, data, ...rest } = args;
 
+  const [hoveredRow, setHoveredRow] = useState<
+    { name: string; description: string } | undefined
```
mayank99, line 1093: If we expose `Row`, we'll be able to save the ids instead of the whole data

stories/core/Table.stories.tsx
bentleyvk: I don't think I understand 😃 I still need map and `ref.current` comes from `useRef`, `ref` just gives us callback prop.

mayank99: Ah well 🤷‍♀️

stories/core/Table.stories.tsx
bentleyvk: Saving index because id comes from inside of react-table and can't be really used outside of it (except if custom id function is passed).

src/core/Table/TableRowMemoized.tsx
bentleyvk: Changed.

mayank99 (approved):

PR #232, opened 2021-07-31
Fix  slider thumb focus issue by reverting to using index as key

### LINKED ISSUE #233: Slider Thumb is not retaining its focus when value changed via keyboard
Slider Thumb is not retaining its focus when value changed via keyboard.

#### Expected behavior
A thumb is focused user should be able to adjust value by pressing arrow key multiple times.

#### Reproduction
```
export const ProcessChange: Story<Partial<SliderProps>> = (args) => {
  const [sliderValues, setSliderValues] = React.useState([20, 40]);

  return (
    <Slider
      {...args}
      values={sliderValues}
      onChange={(values) => setSliderValues([...values])}
    />
  );
};

ProcessChange.args = {
  min: 0,
  max: 60,
  tickLabels: ['0', '20', '40', '60'],
  tooltipProps: (index, val) => {
    return { placement: 'right', content: `\$${val}.00` };
  },
};
```

##### Steps to reproduce
1. Use above code 
1. Set focus to first thumb
1. Press Arrow key and note that after the first key press focus is lost

#### Additional information
See proposed fix  in  https://github.com/iTwin/iTwinUI-react/pull/232

bsteinbk:
If user used keyboard to change thumb value and there was an onChange listener which was creating a new `values` array, then the thumb would lose focus as it was re-rendered in its new position.

This fixes it by changing the Thumb key to user-provided `id` (with `index` as fallback) instead of maintaining an internal ref count.

This also fixes the issue in the unit tests where it was required to keep looking up the Thumb element because a new unique key was generated anytime the values array was updated.

Closes #233

## Checklist

### DIFF
--- a/src/core/Slider/Slider.tsx
+++ b/src/core/Slider/Slider.tsx
@@ -442,7 +442,7 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
             const thisThumbProps = thumbProps?.(index) ?? {};
             return (
               <Thumb
-                key={`${thisThumbProps?.id ?? index}`}
+                key={`${thisThumbProps.id ?? index}`}
                 index={index}
                 disabled={disabled}
                 isActive={activeThumbIndex === index}
--- a/src/core/Slider/Slider.tsx
+++ b/src/core/Slider/Slider.tsx
@@ -439,10 +439,10 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
           <div className='iui-slider-rail' />
           {currentValues.map((thumbValue, index) => {
             const [minVal, maxVal] = getAllowableThumbRange(index);
-            const thisThumbProps = thumbProps?.(index) ?? {};
+            const thisThumbProps = thumbProps?.(index);
             return (
               <Thumb
-                key={`${thisThumbProps.id ?? index}`}
+                key={thisThumbProps?.id ?? index}
                 index={index}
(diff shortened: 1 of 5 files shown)


src/core/Slider/Slider.tsx
```diff
@@ -439,9 +439,13 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
...
           {currentValues.map((thumbValue, index) => {
             const [minVal, maxVal] = getAllowableThumbRange(index);
+            const thisThumbProps = thumbProps?.(index) ?? {};
+            const key = thisThumbProps.id
+              ? thisThumbProps.id
+              : index.toString();
             return (
               <Thumb
-                key={index}
+                key={key}
```
mayank99, line 448: Does this not work?

```suggestion
                key={`${thumbProps?.(index)?.id ?? index}`}
```

bsteinbk: I like being able to see the value in debugger... but I updated to mostly use your suggestion. I see no reason to run `thumbProps?.(index)` twice so I still create `thisThumbProps` and use if for the `key` and for `thumbProps` .

mayank99: Well, if you're assigning `{}` anyway, then you don't need `?.`

So you can change it to `key={thisThumbProps.id ?? index}`

mayank99 (approved):
Thanks!

PR #236, opened 2021-08-02
feat(Table): Virtualization

### LINKED ISSUE #157: Table: row virtualization
#### Feature

In absence of a paginator, it would be nice to have virtualization (only rendering rows that are "visible" in the viewport) for performance reasons.

#### Examples

See this react-window example: https://codesandbox.io/s/3vnx878jk5

veekeys:
~~Initial try, but far from finished. Not only the code needs some cleanup, but also functionality is not fully there.
It was easy and working fine with static child height.. 
Here I was trying to have something covered for dynamic with calculating the heights of visible children and applying translate accordingly (also calculating the visible children).
It is something missing in maths, which I would try to work on when coming back..
Feel free to play around, just do not complete it :D
I think after reviewing the maths again, all should be fine and the most complicated part is to set the full height for the scroll (cause it is not fully known from the initial render). Currently, I just have it set from an avg height, but obviously it is not perfect...~~

UPDATE:
removed all the dynamic part calculations and left the minimal base to work well with static heights.
added story with 100k rows. Notice the initial lag, but it is taking time to generate that list, I think :D After that, it works pretty smooth.
Tried it with ComboBox. Worked pretty well, but.. THere is this issue, when you have selected value and you want to open items with that selected value in the middle.. Obviously, as we do not have full list in there, there is no that value (if it was not at the beggining of the list). So might need to think for some additional functionality, to support this in ComboBox.. But maybe with separate PR. Might need additional options to the VirtualScroll.
Now It has only required children and optional `renderAdditional` prop to play around with the amount of items rendered below visible.. I would expect not to change that too often, but.. Maybe it might help in some weird cases. Naming might be better, though..I have chosen 20 (random) default value, just as it does not cost anything to render such small amount of items.. And theoretically the more, the better.

ISSUES:
~~I still need to check up why recent css changes (flex applied on body) did mess up my scroll. Now I have hardcoded 'block' value, which is obviously not a solution. So this is still in progress.~~ Fixed by using `minHeight` instead of `height` and also needed `width: 100%` cause lazy load and subrows stories broke :(.
Body scroll does not work now.. Need to add some additional logic to handle the margin between first element and the top. Or maybe use IntersectionObserver to catch when first element actually hits the top and ignore everything till then... But maybe for future.
Subrows story has this little visual bug when playing around with expansions and scroll first time.. Couldnt figure out the issue and seenable in Chrome only. Firefox worked fine :/ Firefox has this blank visuals, when dragging scroll over thousands of items.


src/core/Table/Table.tsx
```diff
@@ -648,30 +656,59 @@ export const Table = <
...
-              );
-            })}
+          {data.length !== 0 && (
+            <>
+              {useVirtualization && (
+                <VirtualScroll>
+                  {page.map((row: Row<T>) => {
+                    prepareRow(row);
+                    return (
+                      <TableRowMemoized
```
bentleyvk, line 666: Move row jsx to a variable or smth to avoid duplication.

src/core/Table/Table.tsx
```diff
@@ -648,30 +656,59 @@ export const Table = <
...
-                  onClick={onRowClickHandler}
-                  subComponent={subComponent}
-                  isDisabled={!!isRowDisabled?.(row.original)}
-                  tableHasSubRows={hasAnySubRows}
-                  tableInstance={instance}
-                  expanderCell={expanderCell}
-                />
-              );
-            })}
+          {data.length !== 0 && (
```
bentleyvk, line 659: Try scrolling with keyboard (arrows, page up/down), it stop scrolling after few items. As I understand the problem is that focused row disappears and keyboard actions stop working. Workaround could be adding `tabIndex` to the table body.

stories/core/Table.stories.tsx
```diff
@@ -2117,3 +2122,69 @@ HorizontalScroll.decorators = [
...
+  }, []);
+
+  return (
+    <Table
+      useVirtualization
+      columns={columns}
+      emptyTableContent='No data.'
+      {...args}
+      style={{ height: 440 }}
+      data={generateData(0, 100000)}
```
roluk, line 2183: If this virtualization solution produces a noticeable slowdown on first render, I expect it to be unusable with Tree component which could easily contain 10x more rows. By the way, Tree component would need to re-walk its hierarchy and create a new flat list on each data update (e.g. node expansion or collapse).

I somehow doubt that the slowdown can be directly attributed to the data generation procedure. In my Firefox console I can execute it for million elements in under 70ms.

roluk, line 2183: If this virtualization solution produces a noticeable slowdown on first render, I expect it to be unusable with Tree component which could easily contain 10x more rows. Tree component would also need to re-walk its hierarchy and create a new flat list on each data update (e.g. node expansion or collapse).

I somehow doubt that the slowdown can be directly attributed to the data generation procedure. In my Firefox console I can execute it for million elements in under 70ms.

bentleyvk: The problem might also be `react-table` itself as it is handling each data entry and preparing row and table state.

src/core/Table/Table.tsx
veekeys: Changed, but..looks bad :(

stories/core/Table.stories.tsx
veekeys: @roluk yup, after all it was an issue with children. I assume they are always being rendered before applying the parent comopnent or something... Anyway, changed API now so it is using render function for the item. If you could check it up now, would appreciate.
Story has a little of lag on react-table (not comprable to previous), but it works nice with just plain divs list (so there might be something with react-table still, but I think minor for now).

stories/core/Table.stories.tsx
```diff
@@ -2117,3 +2122,69 @@ HorizontalScroll.decorators = [
...
+              );
+            },
+          },
+        ],
+      },
+    ],
+    [],
+  );
+
+  const generateData = useCallback((start: number, end: number) => {
```
bentleyvk, line 2165: You should have `useMemo` here instead. You are not generating data more than once.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Single item render function, which gives index of the item (0 based) in the data array
+   * and expects to get the JSX of that element to render.
+   * Recommended to memoize the reference of the function.
+   */
+  itemRenderer: (index: number) => JSX.Element;
+  /**
+   * Number of items to be rendered at the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 20
```
bentleyvk, line 87: 20 seems like a big number

veekeys: What is your suggestion? My thought behind it was that we are fine to render even 50 or 100 rows, it should not influence performance. And having 20 in front might save sometimes from unexpected bugs on scroll...

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,249 @@
...
+  let i = startIndex;
+  let sum = 0;
+  while (sum < totalHeight) {
+    sum += childHeight;
+    ++i;
+  }
+  i = sum > totalHeight ? i - 1 : i;
+  return i - startIndex;
+};
+
```
roluk, line 46: When `childHeight` is `0` (e.g. empty div), this will produce an infinite loop and hang the browser tab.

This whole function reduces down to `Math.floor(totalHeight / childHeight)`, I think.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,249 @@
...
+const getTranslateValue = (childHeight: number, startIndex: number) => {
+  let sum = 0;
+  let i = 0;
+  while (i < startIndex) {
+    sum += childHeight;
+    ++i;
+  }
+  return sum;
+};
+
```
roluk, line 56: Performance of this function scales linearly to `startIndex` value. Could be reduced down to `childHeight * startIndex`.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,249 @@
...
+        Math.max(0, i - startNode),
+      ) as HTMLElement;
+      const elementHeight = getElementHeight(htmlElement);
+      heightSum += Number(elementHeight.toFixed(2));
+    }
+    const elementHeight = heightSum / visibleChildren.length;
+    childHeight.current = Math.ceil(elementHeight);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
```
roluk, line 165: There may be no children on first render. The easiest solution is to ask the consumer to provide child height via a prop.

roluk (changes requested):

stories/core/Table.stories.tsx
roluk: You can resolve this comment now (the button to do so isn't there for me).

src/core/utils/components/VirtualScroll.tsx
veekeys: Maybe, but the point is that I have 2nd implementation coming with dynamic heights, where I calculate those, so kinda wanted to have a base for that ready. I will think something about childHeight if there is no children...

src/core/utils/components/VirtualScroll.tsx
veekeys: Yeah, these are also left from the implementation with dynamic heights... There I have a map so taking one by one.
I can simplify these parts for now.

src/core/Table/Table.tsx
veekeys: ok.. So added your suggestion.
Found only one issue: when you have table with selectable rows, select one row and press arrow up/down, seems like a swap in row or something like that :D But I think its minor, compared that previously it was not working at all...

stories/core/Table.stories.tsx
veekeys: fixed

src/core/utils/components/VirtualScroll.tsx
veekeys: Ok, so added guard for 0 value and simplified code

src/core/utils/components/VirtualScroll.tsx
veekeys: Simplified code

src/core/utils/components/VirtualScroll.tsx
veekeys: Ok, so changed.. What i did, added `visibleChildren.length` as dep to trigger calculation of child height and also guard to early quit if there are no visible children.
It runs initially (as only `renderAdditional` amount of elements are rendered) and then on the full render. Scroll does not influence this to be recalculated.

src/core/utils/components/VirtualScroll.tsx
bentleyvk: Oh, I thought it was rendering 20 before and 20 after. But it only adds 20 at the bottom. Why not on the both ends?

stories/core/Table.stories.tsx
```diff
@@ -697,6 +699,7 @@ export const ExpandableSubrows: Story<Partial<TableProps>> = (args) => {
         data={data}
         columns={columns}
         {...args}
+        useVirtualization
```
bentleyvk, line 702: Maybe you can take data generation from `WithPaginator` story and move here to see how virtualization works with big sub-rows list. Or maybe add sub-rows to `Virtualization` sotry.

src/core/utils/components/VirtualScroll.tsx
veekeys: why both? Do you see emptiness at the top whiles scrolling or smth? usually its done when you have virtual scroll with fake top and bottom elements with their heights. This solution I dont think requires it

bentleyvk: I thought you need additional items on top and on the bottom as you mentioned yourself "save sometimes from unexpected bugs on scroll" when you scroll up or down. Also, try tabbing to previous items, you can't because there is none on the top of visible list.

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -0,0 +1,92 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { fireEvent, render } from '@testing-library/react';
+import React from 'react';
+import VirtualScroll from './VirtualScroll';
+
+const component = (
```
bentleyvk, line 10: Why you need to have it as a separate variable?

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      )}
+    />
+  </div>
+);
+
+const renderComponent = () => {
+  return render(component);
+};
+
+const renderEmptyComponent = () => {
```
bentleyvk, line 27: You only use it in one place, is it really needed?

bentleyvk: Ok, I don't understand. Is it really empty list? Seems that it will render 1k divs

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        itemRenderer={(index) => <div key={index} className='element' />}
+      />
+    </div>,
+  );
+};
+
+beforeAll(() => {
+  // return correct values for container 'scroller' and children
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function getBoundingClientRectMock(this: Record<string, any>) {
```
bentleyvk, line 41: Why not inline in `mockImplementation`?
Also you could use `<span>` as a child, that way you could mock `HTMLDivElement.getBoundingClientRect` and `HTMLSpanElement.getBoundingClientRect` but I think it is ok how it is done now.

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+  jest
+    .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
+    .mockImplementation(getBoundingClientRectMock);
+});
+
+it('should render only few elements out of big list', () => {
+  const { container } = renderComponent();
+  expect(container.querySelectorAll('.element').length).toBe(30);
+  expect(container.querySelector('.element')?.textContent).toBe('Element1');
```
bentleyvk, line 55: Maybe also check what is the last element.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+
+const getScrollableParent = (element: HTMLElement | null): HTMLElement => {
```
bentleyvk, line 7: You really don't want to just have scrollable container ref as a prop? That way you could remove all these helper functions, also users would know and could set scrollable container exactly where they want, avoiding bugs when scrollable container is automatically detected in bad place.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+      getElementStyle(element, 'overflow-y'),
+  );
+};
+
+const getElementStyle = (element: HTMLElement, prop: string) => {
+  return getComputedStyle(element, null).getPropertyValue(prop);
+};
+
+const getElementHeight = (element: HTMLElement | undefined) => {
+  return !!element ? element.getBoundingClientRect().height : 0;
```
bentleyvk, line 29: Optional chaining didn't work? Like `element?.getBoundingClientRect().height ?? 0`

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+};
+
+const getCountOfNodesInHeight = (childHeight: number, totalHeight: number) => {
```
bentleyvk, line 32: Strange name 😀 Maybe at least change `count` to `number`.

src/core/utils/components/VirtualScroll.tsx
veekeys: Well, I just needed on one side for the initial calculations (where visible nodes are 0, so I am using only those from additionalItems...
So what you are saying.. I should add that same amount on both top and bottom then...?

src/core/utils/components/VirtualScroll.test.tsx
veekeys: well, I meant empty item...

src/core/utils/components/VirtualScroll.tsx
veekeys: ah man.. I dont know.. Will you always have a reference to scrollable container?
I guess I can change it if you feel thats better and covers more use cases..

src/core/utils/components/VirtualScroll.tsx
bentleyvk: I would say to split it maybe or reduce default value in half.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+   * and expects to get the JSX of that element to render.
+   * Recommended to memoize the reference of the function.
+   */
+  itemRenderer: (index: number) => JSX.Element;
+  /**
+   * Number of items to be rendered at the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 20
+   */
+  renderAdditional?: number;
```
bentleyvk, line 72: This prop name is not clear from the first sight. It sounds to me like a flag or even callback. I see other libs use `buffer` or `overscan` + `amount`/`count`

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+};
+
+/**
+ * `VirtualScroll` component to render only visible items + some additional out of a big list.
```
bentleyvk, line 76: Missing "Use" before `VirtualScroll`?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+ * Good use in expected big number of data either in some of the components (e.g. Table) or just simple list of DOM elements.
```
bentleyvk, line 77: Maybe reread these comments yourself. Sounds a bit like a broken English 😀

src/core/utils/components/VirtualScroll.tsx
bentleyvk: They should have in most of the cases. They might not have if they are using some 3rd party container.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+ */
+export const VirtualScroll = ({
+  itemsLength: childrenLength,
+  itemRenderer,
+  renderAdditional = 20,
+}: VirtualScrollProps) => {
+  const [startNode, setStartNode] = React.useState(0);
+  const [visibleNodeCount, setVisibleNodeCount] = React.useState(0);
+  const scrollContainer = React.useRef<HTMLElement>();
+  const childrenParentRef = React.useRef<HTMLDivElement>(null);
```
bentleyvk, line 101: Maybe just `parentRef`?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+  }, [
+    childrenLength,
+    itemRenderer,
+    renderAdditional,
+    startNode,
+    visibleNodeCount,
+  ]);
+
+  // Get child height when children available
+  React.useLayoutEffect(() => {
```
bentleyvk, line 132: I think we talked that with static height this is not needed anymore. But if you want to save it for dynamic height implementation maybe just create a branch out of this or PR will have history.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+      }
+      const start = getCountOfNodesInHeight(
+        childHeight.current,
+        target.scrollTop,
+      );
+      setStartNode(start);
+      setVisibleNodeCount(
+        getVisibleNodeCount(childHeight.current, start, childrenLength, target),
+      );
+      setTranslateY(getTranslateValue(childHeight.current, start));
```
bentleyvk, line 166: I am thinking, what if you changed translate value directly in the ref? It should be a bit faster as it avoids rerenders from setState. But maybe not a big performance gain as it would only work when startNode and visibleCount is the same.

src/core/utils/components/VirtualScroll.tsx
veekeys: what is not needed? So instead have param to pass height?

src/core/utils/components/VirtualScroll.test.tsx
bentleyvk: Why you need to render empty item?

src/core/utils/components/VirtualScroll.tsx
bentleyvk: This huge check is not needed as you just need to take first child and get its height.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,232 @@
...
+    onScrollRef.current = onScroll;
+    if (!scrollContainer.current || scrollContainer.current === document.body) {
+      document.addEventListener('scroll', onScroll);
+    } else {
+      scrollContainer.current.addEventListener('scroll', onScroll);
+    }
+    return removeScrollListener;
+  }, [onScroll, removeScrollListener]);
+
+  React.useLayoutEffect(() => {
```
bentleyvk, line 195: What it does? Is it for lazy-loading in chunks?

veekeys: It recalculates everything whenever children changes

src/core/utils/components/VirtualScroll.test.tsx
veekeys: removed

src/core/utils/components/VirtualScroll.test.tsx
veekeys: changed, but left divs

src/core/utils/components/VirtualScroll.test.tsx
veekeys: added

src/core/utils/components/VirtualScroll.tsx
veekeys: worked

src/core/utils/components/VirtualScroll.tsx
veekeys: changed

src/core/utils/components/VirtualScroll.tsx
veekeys: changed to `buffer`

src/core/utils/components/VirtualScroll.tsx
veekeys: renamed

src/core/utils/components/VirtualScroll.tsx
veekeys: taking only first child

src/core/utils/components/VirtualScroll.tsx
veekeys: using ref now, seems like still working

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -7,52 +7,38 @@ import { fireEvent, render } from '@testing-library/react';
...
-  function getBoundingClientRectMock(this: Record<string, any>) {
-    if (Object.values(this)[0].memoizedProps.id === 'scroller') {
-      return { height: 400 } as DOMRect;
-    }
-    return { height: 40 } as DOMRect;
-  }
   jest
     .spyOn(HTMLElement.prototype, 'getBoundingClientRect')
-    .mockImplementation(getBoundingClientRectMock);
+    .mockImplementation(function (this: Record<string, any>) {
```
bentleyvk, line 15: Since when we are using functions? 😀

src/core/utils/components/VirtualScroll.test.tsx
bentleyvk: No you don't 😀 It is jsdom and you don't change the mock for client height in this test.

src/core/utils/components/VirtualScroll.test.tsx
bentleyvk: I though you will add last element check in all places you are checking first element too...

src/core/utils/components/VirtualScroll.test.tsx
veekeys: oh please show me, how to do this with arrow func :)

src/core/utils/components/VirtualScroll.tsx
bentleyvk: Why not bufferCoun ir bufferAmount? Buffer alone can mean anything.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  // Get child height when children available
+  React.useLayoutEffect(() => {
+    if (!parentRef.current || !visibleChildren.length) {
+      return;
+    }
+
+    const firstChild = parentRef.current.children.item(0) as HTMLElement;
+    childHeight.current = Number(getElementHeight(firstChild).toFixed(2));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bentleyvk, line 132: Is this eslint disable needed anymore?

src/core/utils/components/VirtualScroll.test.tsx
veekeys: oh fukin shit...................

src/core/utils/components/VirtualScroll.tsx
veekeys: i dont know.. I dont see buffer as good word at all here..
bufferSize is the only one then, but it has such a different meaning and context in general...

src/core/utils/components/VirtualScroll.tsx
bentleyvk: Isn't it doing the same as onScroll?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    if (!parentRef.current) {
+      return;
+    }
+    parentRef.current.style.transform = `translateY(${getTranslateValue(
+      childHeight.current,
+      start,
+    )}px)`;
+  }, [childrenLength]);
+
```
bentleyvk, line 210: I think you might need to recalculate stuff on resize.

src/core/utils/components/VirtualScroll.test.tsx
bentleyvk: Ok, wow, it doesn't allow 😲

src/core/utils/components/VirtualScroll.tsx
veekeys: well, it recalculates same things, yes

bentleyvk: So maybe don't duplicate code?

src/core/utils/components/VirtualScroll.tsx
veekeys: so no, I think?

src/core/utils/components/VirtualScroll.tsx
veekeys: removed

src/core/utils/components/VirtualScroll.tsx
veekeys: rewrote something

src/core/utils/components/VirtualScroll.tsx
veekeys: rewrote

src/core/utils/components/VirtualScroll.test.tsx
veekeys: added :D

src/core/utils/components/VirtualScroll.test.tsx
veekeys: fixed, I think

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,233 @@
...
+   * Number of items to be rendered at the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 20
+   */
+  buffer?: number;
+};
+
+/**
+ * `VirtualScroll` component used to render a huge amount of items in the DOM. It renders only the ones which are visible
+ * and the amount provided through `buffer` prop. Can be used inside other components like `Table`.
```
bentleyvk, line 77: Maybe don't mention Table here as someone will definetely come and ask how to use VirtualizedScroll and Table 😀

src/core/utils/components/VirtualScroll.tsx
bentleyvk: Did you?

stories/core/Table.stories.tsx
veekeys: added new story

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,233 @@
...
+   * Recommended to memoize the reference of the function.
+   */
+  itemRenderer: (index: number) => JSX.Element;
+  /**
+   * Number of items to be rendered at the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 20
+   */
+  buffer?: number;
+};
```
bentleyvk, line 73: Maybe add `CommonProps` too, if someone wants to add custom class or styling e.g. some container class as we are changing DOM (as you mentioned in the comments).

src/core/utils/components/VirtualScroll.tsx
veekeys: stopped duplication

src/core/utils/components/VirtualScroll.tsx
veekeys: now yes :DD

src/core/utils/components/VirtualScroll.tsx
veekeys: amm.. but we have prop there to use it.. so isnt it good?

src/core/utils/components/VirtualScroll.tsx
veekeys: added

stories/core/Table.stories.tsx
```diff
@@ -847,11 +850,13 @@ export const LazyLoading: Story<Partial<TableProps>> = (args) => {
 
   return (
     <Table
+      useVirtualization
       columns={columns}
       emptyTableContent='No data.'
       onBottomReached={onBottomReached}
       isLoading={isLoading}
       {...args}
+      style={{ height: 440 }}
```
bentleyvk, line 859: Maybe change to `90vh` or use decorator as in other stories.

src/core/utils/components/VirtualScroll.tsx
bentleyvk: Idk, I just don't want that someone might misunderstand this comment but now I don't know what example you could give here to, so maybe it is ok.

stories/core/Table.stories.tsx
veekeys: changed

src/core/utils/components/VirtualScroll.tsx
veekeys: renamed again

src/core/utils/components/VirtualScroll.tsx
veekeys: I am struggling to make it work...

stories/core/Table.stories.tsx
```diff
@@ -82,7 +82,16 @@ export default {
...
   },
   parameters: {
-    creevey: { skip: { stories: ['Lazy Loading', 'Row In Viewport'] } },
+    creevey: {
+      skip: {
+        stories: [
+          'Lazy Loading',
+          'Row In Viewport',
+          'Virtualized',
+          'VirtualizedSubRows',
```
bentleyvk, line 91: Need spaces?

stories/core/Table.stories.tsx
```diff
@@ -2117,3 +2128,142 @@ HorizontalScroll.decorators = [
...
+      emptyTableContent='No data.'
+      {...args}
+      style={{ height: 440, maxHeight: '90vh' }}
+      data={data}
+    />
+  );
+};
+
+Virtualized.argTypes = {
+  isLoading: { control: { disable: true } },
```
bentleyvk, line 2196: Also disable `data`. In other story too.

stories/core/Table.stories.tsx
veekeys: upsy

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,229 @@
...
+ *  <div key={index}>
+ *    This is my item #{index}
+ *  </div>
+ * ), [])
+ * <VirtualScroll
+ *  itemsLength={1000}
+ *  itemRenderer={itemRenderer}
+ * />
+ */
+export const VirtualScroll = ({
```
bentleyvk, line 96: Should we mark it with `@beta`? Just in case something needs to be changed? Or not make it public and wait until we apply it to `ComboBox` and `Tree` to see if everything is supported.

stories/core/Table.stories.tsx
veekeys: I think it was removed

src/core/utils/components/VirtualScroll.tsx
veekeys: So I just removed from core barrel, so its like an inner component for now.

stories/core/Table.stories.tsx
veekeys: disabled

src/core/utils/components/VirtualScroll.tsx
veekeys: Changed to 10 and rendering from both sides. Will update description too, which now I realized forgot..

veekeys: done

bentleyvk (approved):
🎈

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,219 @@
...
+};
+
+const getElementStyle = (element: HTMLElement, prop: string) => {
+  return getComputedStyle(element, null).getPropertyValue(prop);
+};
+
+const getElementHeight = (element: HTMLElement | undefined) => {
+  return element?.getBoundingClientRect().height ?? 0;
+};
+
```
roluk, line 32: If I'm not mistaken, parent container height changes are not watched. Container size could change for a variety of reasons, for example, due to user resizing a widget in iTwin.js.

Virtualization libraries usually rely on the component consumer to provide parent container height. This is often done to make libraries smaller and to avoid adding unnecessary ResizeObservers when parent height could be calculated by other means.

veekeys: How resizing widget would change elements container height? I mean, to have a scrollable container, you need to set its height, otherwise, scrollbar would not appear. So to change container height, user would subscribe to resizing event and manually change the css prop. Have you seen many use cases like that?  
Currently, VirtualScroll does not support body scroll (not sure, how valid it is too).

roluk: itwin-viewer uses some virtualized components that fill whole widget area, i.e. resizing the widget alters the amount of visible rows. I don't see any way this could be implemented using the current VirtualScroll component without enormous waste.

You can see how the current implementation fails to handle dynamic container height by altering Virtualized Table story to set Table height to 100% and reducing bufferSize to 1. Then increase window height and observe how no items are being rendered below the initial fold.

src/core/utils/components/VirtualScroll.tsx
mayank99: I think we use `@private` in some of our hooks to explicitly say it's internal.

.vscode/settings.json
```diff
@@ -17,5 +17,6 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": true
   },
-  "files.eol": "\n"
+  "files.eol": "\n",
+  "typescript.tsdk": "node_modules\\typescript\\lib"
```
mayank99, line 21: Is this needed?

src/core/Table/Table.tsx
```diff
@@ -47,6 +47,7 @@ import {
   onTableResizeEnd,
   onTableResizeStart,
 } from './actionHandlers/resizeHandler';
+import VirtualScroll from '../utils/components/VirtualScroll';
```
mayank99, line 50: Could import this from utils barrel in line 25

src/core/Table/Table.tsx
```diff
@@ -218,6 +219,14 @@ export type TableProps<
...
    */
   styleType?: 'default' | 'zebra-rows';
+  /**
+   * Virtualization is used for the scrollable table body.
+   * Height on the table is required for virtualization to work.
+   * @example
+   * <Table useVirtualization style={{height: 400}} {...} />
+   * @default false
+   */
+  useVirtualization?: boolean;
```
mayank99, line 229: I went through all the comments that were previously made on this PR and was surprised not to find any discussion about this prop name. It sounds like a hook. Maybe called it `virtualizeRows` or `enableVirtualization`?

It looks like MUI has a flipped boolean called `disableVirtualization` in their data grid, but I think that disables both column and row virtualization.

src/core/Table/Table.tsx
```diff
@@ -640,38 +692,30 @@ export const Table = <
...
+            <>
+              {useVirtualization && (
+                <VirtualScroll
+                  itemsLength={page.length}
+                  itemRenderer={virtualizedItemRenderer}
                 />
-              );
-            })}
+              )}
+              {!useVirtualization &&
```
mayank99, line 713: nit: Might be more readable as ternary?

src/core/utils/components/VirtualScroll.tsx
mayank99: And the `@beta` flag we might still want to add in the table prop

src/core/Table/Table.tsx
```diff
@@ -640,38 +692,30 @@ export const Table = <
...
-                  onClick={onRowClickHandler}
-                  subComponent={subComponent}
-                  isDisabled={!!isRowDisabled?.(row.original)}
-                  tableHasSubRows={hasAnySubRows}
-                  tableInstance={instance}
-                  expanderCell={expanderCell}
+          {data.length !== 0 && (
+            <>
+              {useVirtualization && (
+                <VirtualScroll
```
mayank99, line 708: I think the extra divs are breaking some styles. Notice the zebra rows are missing and some content is shifting
rows-shift1

stories/core/Table.stories.tsx
```diff
@@ -2117,3 +2128,144 @@ HorizontalScroll.decorators = [
...
+    [generateItem],
+  );
+
+  return (
+    <Table
+      useVirtualization
+      columns={columns}
+      emptyTableContent='No data.'
+      {...args}
+      style={{ height: 440, maxHeight: '90vh' }}
```
mayank99, line 2262: nit: Might want to move this to a decorator and let the wrapper control the height.

mayank99 (commented):
Works pretty well 😮

.vscode/settings.json
veekeys: Yes, this is supposed to make you use our libraries version instead of your machines.
I was having all the time issues, where certain features were marked red. Realized, it was because of TS version.
Hopefully, this fix helps now.

src/core/Table/Table.tsx
veekeys: You know I dont like barrel import.
What is an advantage of barrel import here apart from less chars?

src/core/Table/Table.tsx
```diff
@@ -640,38 +692,30 @@ export const Table = <
             className: cx('iui-table-body', {
               'iui-zebra-striping': styleType === 'zebra-rows',
             }),
+            style: { outline: 0 },
```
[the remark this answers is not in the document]
veekeys: You said you read all the comments :D
https://github.com/iTwin/iTwinUI-react/pull/236#discussion_r778172563
It is a fix for this one, but then it gets focusable and gets default focus style (outline), which I reset here, as you noticed.
Ideally, I guess we would move that to css part.

stories/core/Table.stories.tsx
veekeys: Moved

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Number of items to be rendered at the start and the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 10
+   */
+  bufferSize?: number;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * `VirtualScroll` component used to render a huge amount of items in the DOM. It renders only the ones which are visible
```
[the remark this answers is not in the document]
veekeys: Mentioned

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,219 @@
...
+   * Recommended to memoize the reference of the function.
+   */
+  itemRenderer: (index: number) => JSX.Element;
+  /**
+   * Number of items to be rendered at the start and the end.
+   * Not recommended to go lower than the visible items in viewport.
+   * @default 10
+   */
+  bufferSize?: number;
+} & Omit<CommonProps, 'title'>;
```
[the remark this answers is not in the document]
veekeys: Exposed and forwarded

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -0,0 +1,219 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { CommonProps } from '../props';
+
+const getScrollableParent = (element: HTMLElement | null): HTMLElement => {
+  if (!element || element === document.body) {
+    return document.body;
```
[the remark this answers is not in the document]
veekeys: Added something, even though currently it does not support body scroll anyway..

src/core/Table/Table.tsx
veekeys: changed

src/core/Table/Table.tsx
veekeys: renamed to `enableVirtualization`

stories/core/Table.stories.tsx
veekeys: Actually moved back..  I need height on table, so left just maxHeight. Users can see too, that height is needed to make it work.

src/core/Table/Table.tsx
veekeys: Yes it does and I have created PR for that, but.. THere are other issues. Zebras are flickering, because odd/even index changes with virtualization. It needs js to be handled, so not sure..
Maybe separate PR?

src/core/utils/components/VirtualScroll.tsx
veekeys: Added resize observer and listening to scroll container height changes. Whenever height changes, I recalculate virtual scroll.

roluk: Thanks for adding support for dynamic content height.

However, I want to express a concern regarding the deviation from other windowing libraries, namely measuring parent height internally instead of receiving the height via a prop. Please review the arguments that I gave in my initial comment. 

Additionally, coming from our experience with iTwin.js virtualized components, people **will** forget to set their parent container height and come asking us why their component is not "working". Requiring to set the height prop explicitly solves this problem because the type checker will catch the error, and that's exactly what we've started doing since iTwin.js 3.0.

veekeys: well, if I ask for height prop, I should ask then for scroll container ref, which I was trying to add, but could not figure out a way (issue with ref/value). Cause it does not make sense to get the height, but not ask for the element. Would you have any ideas how I could do that? Then I could get rid of bunch of code to get the scollable container and then take its height.

I do not really get the "forgetting".
I mean, in native html and css, without setting the height, you would not have a scroll. And thats the way our components should also work. Rely on native things as much as possible without having some additional stuff on top of that.
It is not much we can do to help them learn. Developers need to have basic understanding of css or we will always end up in bugs and issues.
I would say then having extended documentation with examples how to achieve things and explain, why it needs to be done that way (which we kinda have with storybook, but also have Wiki and plan to have even separate website with everything at one place).

stories/core/Table.stories.tsx
mayank99: So it doesnt work if height is coming from flex or grid?

mayank99 (approved):
🐱‍💻

stories/core/Table.stories.tsx
veekeys: I tried using flex, it works, but you still need to set the height on flex container.

Issue #225, opened 2021-07-28
Add Id props to SelectOption<T> 


maxime4000:
#### Feature
When writing e2e test, you want to have direct access to element you want to do some actions on it. That mean having some way to click on a specifics item in the select dropdown. I suggest adding Id as optional props in SelectOption<T>.

mayank99:
Do you really need an id just for testing? Playwright even [recommends](https://playwright.dev/docs/selectors/#:~:text=to%20reduce%20the%20maintenance%20burden%2C%20we%20recommend%20prioritizing%20user-facing%20attributes%20and%20explicit%20contracts.) testing with user-facing attributes first.
```ts
await page.$('text="Select option 2"')
```

maxime4000:
It's not because it's recommends that is the only way to get things done. My worries here are that the string can be shown more than once and I can get the wrong element and the test will fail to do what it's suppose to do. Translation can get in the way too, so I don't see anything wrong with ID if it's for improving reliability.

bentleyvk:
I think it is really bad to show select which has items with the same names.

maxime4000:
I think it's not your place to judge how people use the lib. You don't know what is the context, where I use it and I don't need to explain the full situation to justify if adding an optional ID props in the SelectOption type is a good idea or not.

veekeys:
@maxime4000, we do not judge anything here. We are trying to figure out the use case first, before dropping everything to add whatever is needed.
Also, we are trying to help people with the UI/UX of the products, so they could be better.
Can you explain the situation, how you can have same text option elements under a single Select?
You can have selector in playwright to select a specific Select (use id for it) and then choose option by text keeping that Select context. Link which was provided above by @mayank99 has a demo of it.

maxime4000:
I never said it's in the same select, all I'm saying is that id is more reliable than text for selector. In our case is not about making sure a user is finding the right option, it's about simulating interaction with our elements, applying some changes, submit those changes and expecting the same outcome now and two years from now if we don't change this feature. ID won't change, translation might. (Probably not in this case, but that's not the point here)

mayank99:
This is now available in 1.16.0 🎉

PR #240, opened 2021-08-02
Add data-value attribute in Select and Id prop in SelectOption.

### LINKED ISSUE #238: Select/MenuItem add value props as data-value={value}
## Feature
Making a big summaries of what is MenuItem, it's an element in a list of elements shown in a dropdown menu. Each MenuItems represent an action or a value that will be send to a function e.g. onClick. From the definition, value is unknown. In reality, value will be a primitive and I wouldn't expect it to be anything else tbh. Value is kept as memory data and is not represent in the html and that is where I see a problem. Select use MenuItem as alternative to option tag. Fine by that as the select tag does not offer much of CSS customization. But here what is missing about the select that old html select offer. **An value attribute.**

## How to satisfy this missing feature ?
Using data-* attributes ! In my case, MenuItem does have the value but is not shown in the html. I'm looking for a way to get the correct item in the select dropdown to make a click action and change the current value. I can use the selector `.iui-select .iui-menu > .iui-menu-item[data-value="SearchValue"]` to find the exact option I want and make a click on the item. All it's need is 2 line added which I can provide in a PR made for Select changes. For the sake of saving my time and yours as this feature request and #225 are alternative to solve the same issue. I will push only one PR.

maxime4000:
- Data-value attribute added to MenuItem li tag.
- Data-value attribute added to Select div tag.
- Id props added to SelectOption<T>.

Closes #238, #225 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### What's new
 
 - **`Table` accepts custom row props** through `rowProps` prop.
+- **`Select` and `MenuItem` now use data-value attributes** as HTML value representation.
+- **`SelectOption` has support for `id` prop.**
 
 ## [1.13.1]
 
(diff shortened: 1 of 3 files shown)


maxime4000:
Comment for trying to link 225 issues to be close too. 

Closes #225

veekeys:
Hopefully, all the things should be available to you with #275 after the release. We not gonna handle ourselves the data-value, but you are free to do it at your end as now all the props needed should be available.

PR #246, opened 2021-08-05
Fix issue where ArrowKey on Slider propagated to scrollbars

bsteinbk:
Fixed issue where ArrowKey on Slider's Thumb propagated to scrollbars of parent container.

Closes # 247

## Checklist

### DIFF
--- a/src/core/Slider/Thumb.tsx
+++ b/src/core/Slider/Thumb.tsx
@@ -107,7 +107,6 @@ export const Thumb = (props: ThumbProps) => {
           return;
       }
       event.preventDefault();
-      event.stopPropagation();
     },
     [disabled, onThumbValueChanged, index, value, step, minVal, maxVal],
   );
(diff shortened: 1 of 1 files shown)


src/core/Slider/Thumb.tsx
```diff
@@ -92,18 +92,22 @@ export const Thumb = (props: ThumbProps) => {
...
-          return;
+          break;
         case 'End':
           onThumbValueChanged(index, maxVal);
+          break;
+        default:
           return;
       }
+      event.preventDefault();
+      event.stopPropagation();
```
mayank99, line 110: stopPropagation should not be needed, right? Just preventDefault should be enough

bsteinbk: I have always blindly done both, but you are corrected, it seems I only need preventDefault....so I made that change.

mayank99 (approved):
Looks good!

bentleyvk (approved):

PR #249, opened 2021-08-05
Add fieldset component

### LINKED ISSUE #234: Add fieldset component
Add `Fieldset` component in React consuming css (https://itwin.github.io/iTwinUI/backstop/minified/fieldset.html)

- Component should have `legend` prop (name discussable)
- Should be possible to have `disabled` state (can extend all native props)
- Create stories to demonstrate use case. Can take same as it is in html example.

saaaaaally:
Added Fieldset Component

image
image

## Checklist


src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <fieldset
+      className={cx('iui-fieldset', className)}
+      disabled={disabled}
+      {...rest}
+    >
+      {legend && <legend>{legend}</legend>}
+
+      {children && disabled
```
mayank99, line 34: I think this should just check for `disabled`, not `children`.

Right now, this component is returning `children` (in line 40) even if it's undefined.

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,45 @@
...
+        ? childrenArray.map((child) =>
```
mayank99, line 35: If you use `React.Children.map`, you won't need to declare childrenArray in line 22
```suggestion
        ? React.Children.map((child) =>
```

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,45 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+
+import cx from 'classnames';
+import '@itwin/itwinui-css/css/fieldset.css';
+
+export type FieldsetProps = {
+  legend?: React.ReactNode;
```
mayank99, line 12: Add a comment above this explaining what it does.

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,45 @@
...
+} & React.ComponentPropsWithoutRef<'fieldset'>;
+
+/**
+ * Fieldset component
+ * @example
+ * Example usages go here!
```
mayank99, line 18: Update comments.

src/core/index.ts
```diff
@@ -182,3 +182,6 @@ export { getUserColor } from './utils/common';
 
 export { useTheme } from './utils/hooks/useTheme';
 export type { ThemeType } from './utils/hooks/useTheme';
+
+export { Fieldset } from './Fieldset';
+export type { FieldsetProps } from './Fieldset';
```
mayank99, line 187: Alphabetically sort this.

stories/core/Fieldset.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  Radio,
+  ToggleSwitch,
+} from '../../src/core';
+
+export default {
+  title: 'Input/Fieldset',
+  component: Fieldset,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
```
mayank99, line 21: I think these two lines can be removed.

src/core/Fieldset/Fieldset.test.tsx
```diff
@@ -0,0 +1,20 @@
...
+  // TODO: Make sure all required props are passed in here
+  const { container } = render(<Fieldset />);
+  expect(container.querySelector('div')).toBeTruthy();
+});
+
+// TODO: Write tests here!
+
+it('should be improved', () => {
+  expect(false).toBe(true);
+});
```
mayank99, line 20: Test for:
1. basic state (check classname and element name)
2. disabled state (also check that it sets disabled on children)
3. custom props passed by user (e.g. id, style)

mayank99 (commented):
This is a good start! Some minor comments to tackle, and write tests, and we should be good!

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * The caption or title for the fieldset
+   */
+  legend?: React.ReactNode;
+} & React.ComponentPropsWithoutRef<'fieldset'>;
+
+/**
+ * Basic fieldset component to group different components together
+ * @example
+ * <Fieldset legend='General Settings' {...args}>
```
bentleyvk, line 21: It is a bit too big example for such component 😃 Maybe use this one from CSS:
image
Of course without disabled state.

stories/core/Fieldset.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <ToggleSwitch label='Share usage statistics' />
+        <ToggleSwitch label='Share crash logs' />
+        <ToggleSwitch disabled label='Advanced settings' />
+      </InputGroup>
+    </Fieldset>
+  );
+};
+Basic.args = { legend: 'General Settings', disabled: false };
+
+export const Disabled: Story<FieldsetProps> = (args) => {
```
bentleyvk, line 58: You can reuse `Basic` story:
```ts
export const Disabled = Basic.bind({});
```

stories/core/Fieldset.stories.tsx
bentleyvk: Really? We are disabling these fields for every story.

mayank99: They don't show up, so they don't need to be disabled 🤔

src/core/Fieldset/Fieldset.tsx
mayank99: I was thinking the same 😄 

And maybe we don't even need to specify optional props. So:
```jsx
<Fieldset legend='Settings'>
  <Input />
  <InputGroup>
    <ToggleSwitch />
    <ToggleSwitch />
  </InputGroup>
</Fieldset>
```

package.json
```diff
@@ -60,6 +60,7 @@
     "@storybook/preset-typescript": "^3.0.0",
     "@storybook/react": "~6.3.6",
     "@storybook/theming": "~6.3.6",
+    "@testing-library/jest-dom": "^5.14.1",
```
bentleyvk, line 63: Why you needed this one?

stories/core/Fieldset.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+      <InputGroup>
+        <ToggleSwitch label='Share usage statistics' />
+        <ToggleSwitch label='Share crash logs' />
+        <ToggleSwitch disabled label='Advanced settings' />
+      </InputGroup>
+    </Fieldset>
+  );
+};
+Basic.args = { legend: 'General Settings', disabled: false };
```
bentleyvk, line 56: Args that have the same value in each story, can be defined in global `args` like in `LabeledInput.story.tsx`:
image

bentleyvk: `disabled` prop is false by default no need to define it here.

mayank99: Storybook does not infer default values anymore, so it's good to define `disabled` here

package.json
mayank99: This is for the `toBeDisabled` matcher.

See this issue: https://github.com/testing-library/dom-testing-library/issues/1003

src/core/Fieldset/Fieldset.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const inputContainers = container.querySelectorAll(
+    '.iui-input-container.iui-disabled',
+  );
+  expect(inputContainers.length).toBe(1);
+
+  const inputs = container.querySelectorAll('input');
+  inputs.forEach((input) => expect(input).toBeDisabled());
+});
+
+it('should render disabled select group', () => {
```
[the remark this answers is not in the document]
mayank99: but they're testing kind of different things.

Select is a custom component, so it's testing the `cloneElement` logic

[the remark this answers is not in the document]
bentleyvk: Aren't they both testing `cloneElement`?

package.json
bentleyvk: Why not `expect(input.disabled).toBe(true)`? It is the first time I see `input.matches`.

mayank99: `disabled` attribute is not set explicitly on the inputs, but on the fieldset

src/core/Fieldset/Fieldset.test.tsx
```diff
@@ -0,0 +1,105 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+import { Fieldset } from './Fieldset';
+import { InputGroup } from '../InputGroup';
+import { ToggleSwitch } from '../ToggleSwitch';
+import '@testing-library/jest-dom/extend-expect';
```
bentleyvk, line 10: Instead of adding this import to every test file, it could be moved to setup:
https://github.com/testing-library/jest-dom#usage
image

saaaaaally: I tried moving the jest-dom import to setupTest.js but I wasn't able to figure out how to do it because I kept running into this syntax error. Any ideas how to get around this? 
image

mayank99: This will be fixed in #258

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,53 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+
+import cx from 'classnames';
+import '@itwin/itwinui-css/css/fieldset.css';
+
+export type FieldsetProps = {
+  /**
+   * The caption or title for the fieldset
```
mayank99, line 13: Can you add period at the end?

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,53 @@
...
+   */
+  legend?: React.ReactNode;
+} & React.ComponentPropsWithoutRef<'fieldset'>;
+
+/**
+ * Basic fieldset component to group different components together
```
mayank99, line 19: Maybe update the wording here and add period at the end too.

```suggestion
 * Fieldset component to group several inputs, controls and labels within a form.
```

src/core/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,53 @@
...
+      disabled={disabled}
+      {...rest}
+    >
+      {legend && <legend>{legend}</legend>}
+
+      {disabled
+        ? React.Children.map(children, (child) =>
+            React.cloneElement(child as JSX.Element, {
+              disabled: true,
+            }),
```
mayank99, line 46: Forgot to mention but maybe only clone if the child is a valid element.
```suggestion
            React.isValidElement(child) ? (
              React.cloneElement(child as JSX.Element, {
                disabled: true,
              })
            ) : child
```

mayank99 (approved):
Looks good!

Lets maybe wait for #258 before merging.

yarn.lock
```diff
@@ -2081,9 +2081,9 @@
     regenerator-runtime "^0.13.4"
 
 "@babel/runtime@^7.9.2":
-  version "7.15.3"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.15.3.tgz#2e1c2880ca118e5b2f9988322bd8a7656a32502b"
-  integrity sha512-OvwMLqNXkCXSz1kSm58sEsNuhqOx/fKpnUnKnFB5v8uDda5bLNEHNgKPvhDN6IU0LDcnHQ90LlJ0Q6jnyBSIBA==
+  version "7.14.8"
```
bentleyvk, line 2084: You need to revert your changes in `yarn.lock` as Mayank's PR updated `jest` and other deps.

saaaaaally: Yes, I think it should be fixed now

bentleyvk (approved):

PR #254, opened 2021-08-06
Update to itwinui-css 0.25.0

mayank99:
Updated itwinui-css to 0.25.0.

Added iui-text-spacing class to Typography components to preserve margins from before.

Updated badge color to the new skyblue background. Test images changed.


stories/core/Tile.stories.tsx
```diff
@@ -76,7 +76,7 @@ Basic.args = {
       </TagContainer>
     </>
   ),
-  badge: <Badge backgroundColor='steelblue'>Badge</Badge>,
+  badge: <Badge backgroundColor='hsl(197, 71%, 83%)'>Badge</Badge>,
```
bentleyvk, line 79: I missed that we now have separate foreground and background colors that needs to be set. Couldn't we achieve the same accessibility with opacity?

mayank99: > I missed that we now have separate foreground and background colors that needs to be set. Couldn't we achieve the same accessibility with opacity?

I don't plan on exposing the foreground part though.

> Also, we need to expose some enum/object with data viz colors
This PR is just for updating the package. Bigger changes are coming in #217.

> It would be great to update UX site with new values.
@FlyersPh9

bentleyvk: >> I missed that we now have separate foreground and background colors that needs to be set. Couldn't we achieve the same accessibility with opacity?
>
>I don't plan on exposing the foreground part though.

Then I don't understand why you spent time finding colors that has good contrast 😃

mayank99: The default foreground color changed from #FFF to #000 at 0.6 opacity, which looks better on the new colors.

bentleyvk: But then it is still not the same and nice as we have in CSS examples. So either we need to remove those foreground variables from CSS or we have to tackle that in React, too. Because I thought the main reason for these changes were to have better contrast and appearance.

bentleyvk (approved):

PR #255, opened 2021-08-06
Add new polymorphic `Text` component

mayank99:
Polymorphic typography component to render any kind of text as any kind of element, using `variant` and `as` props, similar to [Material-UI](https://material-ui.com/components/typography/#typography).

## Checklist


src/core/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,96 @@
...
+   * Use it if you are still loading the content.
+   * @default false
+   */
+  isSkeleton?: boolean;
+
+  /**
+   * Does the element have a bottom margin?
+   * @default false
+   */
+  hasMargin?: boolean;
```
bentleyvk, line 46: I am not sure about this prop. Do we really need it?

src/core/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,96 @@
...
+import React from 'react';
+import { useTheme } from '../../utils/hooks/useTheme';
+import { CommonProps } from '../../utils/props';
+import '@itwin/itwinui-css/css/text.css';
+
+type TextOwnProps<T extends React.ElementType | React.ComponentType = 'div'> = {
+  /**
+   * What element should the text be rendered as?
+   *
+   * @default 'div'
```
bentleyvk, line 15: If default variant is `body`, then default element should be `<p>`

src/core/Typography/Text/Text.tsx
```diff
@@ -0,0 +1,96 @@
...
+  T extends React.ElementType | React.ComponentType = 'div'
+> = TextOwnProps<T> &
+  Omit<React.ComponentPropsWithoutRef<T>, keyof TextOwnProps<T>>;
+
+/**
+ * Polymorphic typography component to render any kind of text as any kind of element.
+ * @example
+ * <Text>Some normal paragraph text rendered as a div</Text>
+ * @example
+ * <Text variant='title' as='h3'>I'm some h2 text rendered as an h3 element!</Text>
```
bentleyvk, line 59: Do we really need this kind of magic - rendering one element as another?

src/core/Typography/Text/Text.tsx
mayank99: I think the ultimate goal would to replace all Typography components with this. Since many of them have margin, it should be possible to achieve the same here.

src/core/Typography/Text/Text.tsx
mayank99: `<p>` doesn't make sense to me as a default. Text can be anywhere, and it's not usually a paragraph. I could go with `<span>`.

src/core/Typography/Text/Text.tsx
mayank99: Yes we do. That's the whole point of this component 😄 

Without this, it would be less useful as it's always rendered as div or something

bentleyvk: So can we just handle that ourselves? If `variant` is `title` then render `<h2>`.

src/core/Typography/Text/Text.tsx
bentleyvk: I was also thinking about `<span>` but as `body` is default variant and it is `<p>`.

src/core/Typography/Text/Text.tsx
mayank99: That's much less useful unless we allow customizing. Sometimes you want small headings as `h2` (or larger headings as `h3`) for semantically and stylistically correct pages.

bentleyvk (approved):

PR #256, opened 2021-08-10
feat: Split Header Button

### LINKED ISSUE #253: Add "Split Button" type of Header Button
#### Feature
Current header buttons are either dropdown buttons (click on the full button and get the dropdown) or simple buttons (click on the button and do something.)

In the header, we would need the "main" part of the button to do something (For example, bring me to the currently "active" project and do something with it) and the drop down part to select a different project and do likely the same thing.

Current connect implementation does this but is rather tricky to click only on the carret image, which does a different thing than the rest of the button.

Not 100% sure of how it should look, but the split button might be a starting point, where the 2 different click targets  highlights separately, I dont know if the current | in the middle of the component would do well

### Basic

image

### Main hover

image

### Drop down

image

(Examples made by pasting header button code in split button "left" button, so styling is not great)

gretanausedaite:
- Added split header button
	When user provides both onClick and MenuItems to Header button, header renders split header button

image
image

## Checklist


src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +85,21 @@ export const HeaderButton = (props: HeaderButtonProps) => {
...
+        onClick={() => buttonProps.onClick}
+      >
+        {buttonProps.children}
+      </SplitButton>
+    );
+  }
+  if (menuItems) {
+    return <DropdownButton {...buttonProps} menuItems={menuItems} />;
+  }
+  return <Button {...buttonProps} />;
```
mayank99, line 102: This doesn't work?
```suggestion
  return menuItems ? (
    isSplitButton ? (
      <SplitButton {...buttonProps} menuItems={menuItems} 
    ) : (
      <DropdownButton {...buttonProps} menuItems={menuItems} />
    )
  ) : (
    <Button {...buttonProps} />
  );
```

The main things here is that `buttonProps` already includes onClick and children, so you don't need to specify them.

src/core/Header/HeaderButton.test.tsx
```diff
@@ -55,6 +55,41 @@ it('render isActive correctly', () => {
   expect(activeButton?.getAttribute('aria-current')).toEqual('location');
 });
 
+it('render split button correctly', () => {
```
mayank99, line 58: Test case should start with "should"

src/core/Header/HeaderButton.test.tsx
```diff
@@ -55,6 +55,41 @@ it('render isActive correctly', () => {
...
+  const itemOneOnClick = jest.fn();
+  const buttonOnClick = jest.fn();
```
mayank99, line 60: You're not really doing anything with these mock handlers 🤔

src/core/Header/HeaderButton.tsx
```diff
@@ -67,7 +69,12 @@ export const HeaderButton = (props: HeaderButtonProps) => {
         })
       : undefined,
     styleType: 'borderless',
-    className: cx('iui-header-button', { 'iui-active': isActive }, className),
+    className: cx(
+      { 'iui-header-button': !isSplitButton },
+      { 'iui-header-split-button': isSplitButton },
+      { 'iui-active': isActive },
```
mayank99, line 75: You can combine these into one object.

stories/core/Header.stories.tsx
```diff
@@ -278,3 +278,66 @@ export const CenterContent: Story<HeaderProps> = (args) => {
     </Header>
   );
 };
+
+export const SplitButtonHeader: Story<HeaderProps> = (args) => {
```
mayank99, line 282: Wonder if we should just modify existing stories, since split button really makes more sense in most cases.

src/core/Header/HeaderButton.tsx
bentleyvk: I prefer the current one. I really hate ternary inside ternary 😃 Even it is considered best practice to return component earlier if some condition is met.

gretanausedaite: Split button props has mandatory onClick and children 
image
I get an error when not providing those 
image

mayank99: Maybe you can change the check to `if (buttonProps.menuItems && buttonProps.onClick)`.

stories/core/Header.stories.tsx
gretanausedaite: Modified stories to have mix of default and split buttons.

src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +87,21 @@ export const HeaderButton = (props: HeaderButtonProps) => {
     ...rest,
   };
 
-  return menuItems ? (
-    <DropdownButton {...buttonProps} menuItems={menuItems} />
-  ) : (
-    <Button {...buttonProps} />
-  );
+  if (menuItems && props.onClick) {
```
bentleyvk, line 90: You can use `isSplitButton` here.

src/core/Header/HeaderButton.test.tsx
```diff
@@ -55,7 +55,42 @@ it('render isActive correctly', () => {
...
+        </MenuItem>,
+      ]}
+    />,
+  );
+
+  const splitButton = container.querySelector('.iui-header-split-button');
+  expect(splitButton).toBeTruthy();
+
+  const innerButtons = splitButton?.querySelectorAll('.iui-borderless');
+  expect(innerButtons?.length).toBe(2);
```
bentleyvk, line 90: I would like to also check that clicking first menu `buttonOnClick` is called, clicking arrow button opens the menu like in the test `should render menuItems correctly`.

src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +87,21 @@ export const HeaderButton = (props: HeaderButtonProps) => {
...
+    return (
+      <SplitButton
+        {...buttonProps}
+        menuItems={menuItems}
+        onClick={() => buttonProps.onClick}
```
bentleyvk, line 95: Just write:
```tsx
onClick={buttonProps.onClick}
```

gretanausedaite: Do I need to add something else? I'm getting this error:
image

bentleyvk: You can set the correct type for SplitButton onClick 😃 It should be `onClick: MouseEventHandler<HTMLButtonElement>`

src/core/Buttons/SplitButton/SplitButton.tsx
```diff
@@ -34,7 +34,7 @@ export type SplitButtonProps = {
    * Style of the button.
    * @default 'default'
    */
-  styleType?: 'cta' | 'high-visibility' | 'default';
+  styleType?: 'cta' | 'high-visibility' | 'default' | 'borderless';
```
mayank99, line 37: You don't need to define this anymore, it is the same as the one from `ButtonProps` (so don't omit it)

src/core/Buttons/SplitButton/SplitButton.tsx
```diff
@@ -95,7 +95,7 @@ export const SplitButton = (props: SplitButtonProps) => {
         onShow={React.useCallback(() => setIsMenuOpen(true), [])}
         onHide={React.useCallback(() => setIsMenuOpen(false), [])}
       >
-        <IconButton styleType={styleType} size={size}>
+        <IconButton styleType={styleType} size={size} disabled={props.disabled}>
```
mayank99, line 98: You should also add `iui-disabled` to the span className.

It belongs in this file, rather than the header button

src/core/Header/HeaderButton.test.tsx
```diff
@@ -55,7 +55,42 @@ it('render isActive correctly', () => {
   expect(activeButton?.getAttribute('aria-current')).toEqual('location');
 });
 
-it('render startIcon correctly', () => {
+it('should render split button correctly', () => {
+  const itemOneOnClick = () => {};
+  const buttonOnClick = () => {};
```
mayank99, line 60: You're not doing anything with these. Why define them?

src/core/Header/HeaderButton.tsx
```diff
@@ -67,7 +69,15 @@ export const HeaderButton = (props: HeaderButtonProps) => {
         })
       : undefined,
     styleType: 'borderless',
-    className: cx('iui-header-button', { 'iui-active': isActive }, className),
+    className: cx(
+      {
+        'iui-header-button': !isSplitButton,
+        'iui-header-split-button': isSplitButton,
+        'iui-active': isActive,
+        'iui-disabled': props.disabled,
```
mayank99, line 77: Remove from here once you add to SplitButton

src/core/Header/HeaderButton.tsx
```diff
@@ -57,6 +58,7 @@ export const HeaderButton = (props: HeaderButtonProps) => {
 
   useTheme();
 
+  const isSplitButton = menuItems && props.onClick;
   const buttonProps: ButtonProps & { styleType: 'borderless' } = {
```
mayank99, line 62: This type should be updated to include `menuItems`. And then you don't have to pass `menuItems` individually.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### What's new
 
 - **Added new `Fieldset` component** for grouping inputs.
+- **Added Split Header Button**
```
mayank99, line 10: Add more wording about how it's enabled. "Split Header Button" is not a component so the user would have no idea.

src/core/Header/HeaderButton.tsx
mayank99: Once this type is fixed, I think we'll be able to use it just from `...buttonProps`.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### What's new
 
 - **Added new `Fieldset` component** for grouping inputs.
+- **Added Split Header Button** when user provides both onClick and menuItems for Header Button component.
```
bentleyvk, line 10: Can you wrap `onClick` and `menuItems` with `?

src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +89,17 @@ export const HeaderButton = (props: HeaderButtonProps) => {
...
 
-  return menuItems ? (
-    <DropdownButton {...buttonProps} menuItems={menuItems} />
-  ) : (
-    <Button {...buttonProps} />
-  );
+  if (menuItems && onClick) {
+    return (
+      <SplitButton {...buttonProps} onClick={onClick} menuItems={menuItems}>
+        {buttonProps.children}
```
bentleyvk, line 95: `children` might not be needed as `{...buttonProps}` should provide it.

src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +89,17 @@ export const HeaderButton = (props: HeaderButtonProps) => {
     ...rest,
   };
 
-  return menuItems ? (
-    <DropdownButton {...buttonProps} menuItems={menuItems} />
-  ) : (
-    <Button {...buttonProps} />
-  );
+  if (menuItems && onClick) {
```
bentleyvk, line 92: Previously you had `isSplitButton` variable, it seems you could bring it back and reuse in several places.

src/core/Header/HeaderButton.tsx
bentleyvk: We tried to pass only `{...buttonProps}` but it seems that Typescript doesn't understand that `onClick` and `menuItems` exist inside `buttonProps`, it only is happy when props are passed explicitly.
If you know the way, we are happy to see it 😃

mayank99: Type guard will help here
image

src/core/Header/HeaderButton.tsx
```diff
@@ -78,11 +98,13 @@ export const HeaderButton = (props: HeaderButtonProps) => {
     ...rest,
   };
 
-  return menuItems ? (
-    <DropdownButton {...buttonProps} menuItems={menuItems} />
-  ) : (
-    <Button {...buttonProps} />
-  );
+  if (isSplitButton(buttonProps)) {
```
gretanausedaite, line 101: @mayank99 Did you have in mind something like this?

src/core/Header/HeaderButton.tsx
```diff
@@ -58,10 +58,19 @@ export const HeaderButton = (props: HeaderButtonProps) => {
...
+
+  const isSplitButton = (
+    props: Partial<SplitButtonProps>,
+  ): props is SplitButtonProps => {
+    return !!props.menuItems && !!props.onClick;
+  };
+
   const buttonProps: ButtonProps & {
     styleType: 'borderless';
+    menuItems: ((close: () => void) => JSX.Element[]) | undefined;
```
mayank99, line 70:
```suggestion
    menuItems?: (close: () => void) => JSX.Element[];
```

src/core/Header/HeaderButton.tsx
```diff
@@ -58,10 +58,19 @@ export const HeaderButton = (props: HeaderButtonProps) => {
...
   } = {
+    menuItems: props.menuItems,
+    onClick: onClick,
```
mayank99, line 73: Are these two lines needed?

src/core/Header/HeaderButton.tsx
```diff
@@ -58,10 +58,19 @@ export const HeaderButton = (props: HeaderButtonProps) => {
   } = props;
 
   useTheme();
-  const isSplitButton = menuItems && onClick;
+
+  const isSplitButton = (
```
mayank99, line 62: Maybe move this up outside the component

src/core/Header/HeaderButton.tsx
```diff
@@ -14,7 +14,7 @@ import {
 
 import { useTheme } from '../utils/hooks/useTheme';
 import '@itwin/itwinui-css/css/header.css';
-import { SplitButton } from '../Buttons/SplitButton';
+import { SplitButton, SplitButtonProps } from '../Buttons/SplitButton';
```
mayank99, line 17: Could combine these imports with the rest of the `./Buttons`.

src/core/Header/HeaderButton.tsx
mayank99: Yeah that looks right

src/core/Header/HeaderButton.tsx
gretanausedaite: apparently not. Removed. Should I add similar function to isSplitButton for DropdownButton to remove menuItems={menuItems}?

mayank99: Yeah I guess DropdownButton is fine, idk

veekeys: Weird :D so go fully same approach or none...

src/core/Header/HeaderButton.test.tsx
```diff
@@ -56,12 +56,21 @@ it('should render isActive correctly', () => {
...
       name={'MockName'}
-      onClick={() => {}}
+      onClick={buttonOnClick}
       menuItems={(close) => [
-        <MenuItem key={0} onClick={() => {}}>
+        <MenuItem
+          key={0}
+          onClick={() => {
+            itemOneOnClick();
+            close;
```
bentleyvk, line 71:
```suggestion
            close();
```

bentleyvk (approved):
🎉

PR #258, opened 2021-08-11
Add jest-dom and upgrade jest + testing library

mayank99:
Sequence of steps taken:
1. Added `jest-dom` for new matchers (like `toBeVisible` and `toBeDisabled`).
2. Renamed `setupTests.js` to `setupTests.ts` so that ESM import for jest-dom is transformed by `ts-jest`.
3. Moved setup into `setupFilesAfterEnv` (otherwise jest-dom errors out).
4. Changed jest environment from `node` to `jsdom` (otherwise setupFilesAfterEnv errors out).
5. Finally got tests to run. Fixed failing tests in a few places.
6. Added check for `CSS.supports` in Badge to fix test in Tile.
7. Upgraded jest and testing library to fix remaining failures.


src/core/Select/Select.test.tsx
```diff
@@ -202,14 +202,14 @@ it('should respect visible prop', () => {
   expect(select).toBeTruthy();
```
bentleyvk, line 203: Can you try changing our focus checks with this one `toHaveFocus()`?

mayank99: Changed in this file

src/core/Tile/Tile.test.tsx
```diff
@@ -37,6 +37,7 @@ it('should render new and selected states', () => {
 });
 
 it('should render main text content correctly', () => {
+  window.CSS = { supports: () => true, escape: (i) => i };
```
bentleyvk, line 40: Why this line is needed? For badges?

mayank99: yes. I'm surprised it was not failing before

bentleyvk: Yeah... Maybe it is worth moving to the global setup?

mayank99: Removed this line and added `supports?.` to Badge

bentleyvk (approved):

PR #262, opened 2021-08-12
Table cleanup before sub-rows PR

bentleyvk:
Moved hooks and action handlers to separate files as there will be more code inside them after sub-rows PR.
No code inside has changed.

## Checklist


src/core/Table/Table.tsx
```diff
@@ -231,202 +231,45 @@ export const Table = <
...
-          ...newState,
-          selectedRowIds: { [action.id]: true } as Record<string, boolean>,
-        };
-      }
-      case TableActions.toggleRowSelected:
-      case TableActions.toggleAllRowsSelected:
-      case TableActions.toggleAllPageRowsSelected: {
-        onSelectHandler(newState, instance);
-        break;
+  const tableStateReducer = React.useCallback(
```
bentleyvk, line 234: I am not sure whether to move it to separate file or not. The only down side having it in separate file is that function will have a lot of props.

mayank99: I'd say leave it here.

src/core/Table/hooks/index.ts
```diff
@@ -0,0 +1,6 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+export { EXPANDER_CELL_ID, useExpanderCell } from './useExpanderCell';
+export { SELECTION_CELL_ID, useSelectionCell } from './useSelectionCell';
```
mayank99, line 6: Do we need to export those ids?

Not sure if you plan to use them in your sub-rows PR or if it's just for futureproofing

mayank99 (approved):

src/core/Table/hooks/index.ts
bentleyvk: Yeah, I plan to use them. But if turns out to be a bad idea, I will remove them.

PR #266, opened 2021-08-16
feat: Table sub-rows

### LINKED ISSUE #192: Table: expandable `subRows`
#### Feature
A full integration of subRows. Like what they have in [React Table](https://react-table.tanstack.com/docs/examples/sub-components-lazy).

For our need, only one child depth is required.
It should work with selection and with sorting.

Please note that in our case, the parent and child have different selection.
For example, if I select the parent it shouldn't toggle the selection on all child.
We also think it would be nice to have an icon in the header to toggle the expansion of all rows.

#### Examples
Here is [a sandbox](https://codesandbox.io/s/itwinui-react-example-forked-84l4t?file=/src/App.tsx:1381-1382) by @mayank99 illustrating a partial implementation.

bentleyvk:
Closes iTwin/iTwinUI-react#192

`react-table` is handling sub-rows differently than we want to so I had to write our own filtering and selection handling.

image

## Checklist


jest.config.js
```diff
@@ -69,7 +69,14 @@ module.exports = {
 
   // A set of global variables that need to be available in all test environments
 
-  // globals: {},
+  globals: {
+    'ts-jest': {
+      tsconfig: {
+        allowJs: true,
+        sourceMap: true,
```
bentleyvk, line 76: This line fixes the bug that came with `jest` v27 that it shows `import` and empty lines as uncovered also showing wrong line for failed test.

jest.config.js
```diff
@@ -69,7 +69,14 @@ module.exports = {
 
   // A set of global variables that need to be available in all test environments
 
-  // globals: {},
+  globals: {
+    'ts-jest': {
+      tsconfig: {
+        allowJs: true,
```
bentleyvk, line 75: This line is for default `react-table` filter functions import to work.

src/core/Table/Table.tsx
```diff
@@ -134,6 +140,11 @@ export type TableProps<
    * Must be memoized.
    */
   rowProps?: (row: Row<T>) => React.ComponentPropsWithRef<'div'>;
+  /**
+   * Flag whether show sub-rows or not.
+   * @default false
+   */
+  showSubRows?: boolean;
```
bentleyvk, line 147: Previously I was checking whether there are any rows with `subRows` but I thought people might want to have lazy-loading and first rows might not have any `subRows` and if they appear later it would cause layout shift.

src/core/Table/Table.tsx
```diff
@@ -207,6 +218,8 @@ export const Table = <
     expanderCell,
     isRowDisabled,
     rowProps,
+    selectSubRows = true,
```
bentleyvk, line 221: In `react-table` its default value is treated as `true` but I wanted to set it explicitly.

src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -18,14 +18,81 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+      if (!result) {
+        isAllSubSelected = false;
+      }
+    });
+
+    // If `selectSubRows` is false, then no need to select sub-rows and just check current selection state.
+    // If a row doesn't have sub-rows then check its selection state.
+    // If it has sub-rows then check whether all of them are selected.
+    if (
+      (!instance.selectSubRows && newState.selectedRowIds[row.id]) ||
```
bentleyvk, line 40: Maybe any of you know how to write this if condition better?

src/core/Table/Table.tsx
```diff
@@ -278,13 +291,16 @@ export const Table = <
...
     },
     useFlexLayout,
     useFilters,
+    useSubRowFiltering(showSubRows),
     useSortBy,
     useExpanded,
     useRowSelect,
+    useSubRowSelection,
     useExpanderCell(subComponent, expanderCell, isRowDisabled),
     useSelectionCell(isSelectable, isRowDisabled),
```
bentleyvk, line 305: There is an issue with filtered sub-rows selection when selecting parent row and selecting again. This is caused that nothing really changed for the row and it is not being rerendered but checkbox does what it has to do - it unmarks indeterminate state. 
table-subrows-filtered-selection-bug

The only workaround that I can think of is making checkbox cell to rerender:
image

src/core/Table/Table.tsx
```diff
@@ -231,44 +244,44 @@ export const Table = <
...
-        }
-        case TableActions.toggleRowSelected:
-        case TableActions.toggleAllRowsSelected:
-        case TableActions.toggleAllPageRowsSelected: {
-          onSelectHandler(newState, instance, onSelect, isRowDisabled);
-          break;
-        }
-        default:
-          break;
+  const tableStateReducer = (
```
veekeys, line 247: Why removed `useCallback` ? no use?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -70,6 +79,38 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
   const refs = useMergedRefs(rowRef, mergedProps.ref);
 
+  const subRowExpanderCellIndex = row.cells.findIndex(
+    (c) => c.column.id !== SELECTION_CELL_ID,
+  );
+
+  const getSubRowExpander = (cell: Cell<T>) =>
+    expanderCell ? (
+      expanderCell({
+        ...tableInstance,
```
veekeys, line 89: hmm.. I wonder if we could just apply tableInstance props in expanderCell function in `Table`, so we dont need to pass whole instance to the memoized row.

bentleyvk: But I still need to pass `cell` which I don't have outside of row.

src/types/react-table-config.ts
```diff
@@ -149,5 +149,7 @@ declare module 'react-table' {
     extends UseExpandedRowProps<D>,
       UseGroupByRowProps<D>,
       UseRowSelectRowProps<D>,
-      UseRowStateRowProps<D> {}
+      UseRowStateRowProps<D> {
+    initialSubRows: Row<D>[];
```
veekeys, line 153: so this prop can be used by the user for the initial table state, right?
But then you use same prop for the actual sub rows in there.. Where the name `initial` naming is kinda confusing..but yeah, dont have better suggestion

src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -18,14 +18,81 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+      (!row.initialSubRows.length && newState.selectedRowIds[row.id]) ||
+      (row.initialSubRows.length && isAllSubSelected)
+    ) {
       newSelectedRowIds[row.id] = true;
-      selectedData.push(row.original);
     }
-  });
+    return !!newSelectedRowIds[row.id];
```
veekeys, line 46: I dont see you use the return value of it..

src/types/react-table-config.ts
bentleyvk: Nope, it is `Row` object. It has nothing to do with initial state. Here is the explanation: https://github.com/iTwin/iTwinUI-react/pull/266/files#diff-da3bb33793f0bcf735d6e37ffcb60a8d246b78615fb85e7c3c0a840c01e8d28bR21

src/core/Table/actionHandlers/selectHandler.ts
bentleyvk: Huh? Lines 28-34

veekeys: right...

src/core/Table/TableRowMemoized.tsx
veekeys: Yeah, but you would avoid passing whole table instance.. Just would be not nice to add it in table as this is prop passed by the user...

src/core/Table/Table.tsx
veekeys: I think it should mark to checked, when you click..

src/core/Table/actionHandlers/selectHandler.ts
veekeys: I think maybe leave like that together with the comments? Even tohugh some checks are repeated, but this way maybe will be clearer..

src/core/Table/Table.tsx
mayank99: why mark to checked? some rows are hidden

veekeys: clickint on a checkbox to get indetermiante state is just wrong (not valid). Indeterminate state cannot be achieved interactively by design.

jest.config.js
mayank99: If we're allowing JS, I guess there is no need to make setupTests use TS.

Maybe all of this should go in a new PR?

src/core/Table/TableRowMemoized.tsx
mayank99: so instead of passing `instance` and verbatim `expanderCell` to `TableRow`, you would pass a wrapper around `expanderCell`? The wrapper would apply `instance` from Table?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -70,6 +79,38 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
+        ...{ cell, row: cell.row, value: cell.value, column: cell.column },
+      })
+    ) : (
+      <IconButton
+        style={{ marginLeft: cell.row.depth * 43, marginRight: 16 }}
```
mayank99, line 94: where does 43 come from?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -70,6 +79,38 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
+        className='iui-row-expander'
+        styleType='borderless'
+        size='small'
+        onClick={(e) => {
```
mayank99, line 98: This should be memoized, no?

src/core/Table/TableRowMemoized.tsx
mayank99: ok I see comment below. Do we really need it in two different places?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -80,13 +121,27 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
       >
-        {row.cells.map((cell) => {
+        {row.cells.map((cell, index) => {
           const cellProps = cell.getCellProps({
             className: cx('iui-cell', cell.column.cellClassName),
-            style: getCellStyle(cell.column),
+            style: {
+              ...getCellStyle(cell.column),
+              ...{
+                paddingLeft:
```
mayank99, line 130: This is the only part I don't like tbh

src/core/Table/TableRowMemoized.tsx
```diff
@@ -80,13 +121,27 @@ const TableRow = <T extends Record<string, unknown>>(props: {
           onClick?.(event, row);
         }}
       >
-        {row.cells.map((cell) => {
+        {row.cells.map((cell, index) => {
           const cellProps = cell.getCellProps({
             className: cx('iui-cell', cell.column.cellClassName),
```
mayank99, line 126: Unrelated but we need to provide a way for the user to set cell class name.

src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -18,14 +18,81 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
-    if (newState.selectedRowIds[row.id] && !isRowDisabled?.(row.original)) {
+
+  const handleRow = (row: Row<T>) => {
+    if (isRowDisabled?.(row.original)) {
+      return false;
+    }
+
+    let isAllSubSelected = true;
+    row.initialSubRows.forEach((subRow) => {
+      const result = handleRow(subRow);
```
mayank99, line 30: Recursion... 😬

Is react-table using similar recursive logic by default?

src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -18,14 +18,81 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+  };
+  instance.initialRows.forEach((row) => handleRow(row));
+
+  const selectedData = getSelectedData(newSelectedRowIds, instance);
+
   newState.selectedRowIds = newSelectedRowIds;
   onSelect?.(selectedData, newState);
 };
+
+export const onSingleSelectHandler = <T extends Record<string, unknown>>(
```
mayank99, line 56: The difference between these handlers is not immediately obvious to me. Can you add comments?

src/core/Table/hooks/useSubRowFiltering.tsx
```diff
@@ -0,0 +1,133 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { Hooks, IdType, Row, TableInstance } from 'react-table';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import * as defaultFilterTypes from '../../../../node_modules/react-table/src/filterTypes.js';
```
mayank99, line 9: What is this atrocious import path?

src/core/Table/hooks/useSubRowFiltering.tsx
```diff
@@ -0,0 +1,133 @@
...
+) => (hooks: Hooks<T>) => {
+  hooks.useInstance.push(useInstance.bind({}, showSubRows));
+};
+
+const useInstance = <T extends Record<string, unknown>>(
+  showSubRows: boolean,
+  instance: TableInstance<T>,
+) => {
+  // There is a bug in react-table that filtered out sub-rows won't be shown after clearing the filter.
+  // Therefore saving original sub-rows separately.
```
mayank99, line 22: Would be nice to create an issue in react-table and link here.

src/core/Table/actionHandlers/selectHandler.ts
mayank99: I agree, it's fine with comments

src/core/Table/TableRowMemoized.tsx
```diff
@@ -70,6 +79,38 @@ const TableRow = <T extends Record<string, unknown>>(props: {
 
   const refs = useMergedRefs(rowRef, mergedProps.ref);
 
+  const subRowExpanderCellIndex = row.cells.findIndex(
```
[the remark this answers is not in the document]
mayank99: separate expander component would be nice but it still needs 43px paddingLeft outside it :/

src/core/Table/Table.tsx
bentleyvk: If lazy-loading is also used e.g. first 10 rows won't have sub-rows, then another batch of 10 rows are loaded that has sub-rows, then previous rows 10 would shift to the right to match the alignment.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -70,6 +79,38 @@ const TableRow = <T extends Record<string, unknown>>(props: {
...
+    (c) => c.column.id !== SELECTION_CELL_ID,
```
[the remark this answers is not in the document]
bentleyvk: First not selection column.

src/core/Table/Table.tsx
bentleyvk: I did the same as we are handling simple filtered rows. And previously we were discussing that it should show indeterminate state, as react-table was showing checked. So I don't know if it is correct to show checked here. And maybe you have any link about showing indeterminate being a bad practice?

src/core/Table/TableRowMemoized.tsx
bentleyvk: We have: just pass `cellClassName` property to column.

src/core/Table/TableRowMemoized.tsx
bentleyvk: It is small component, not really worth.

src/core/Table/TableRowMemoized.tsx
bentleyvk: Do you have a better idea? 😀 Previously we 4 were trying to manage to position everything correctly, so I was happy to make it work like this 😀

src/core/Table/actionHandlers/selectHandler.ts
bentleyvk: Yeah, I copied a lot of ideas from there.

src/core/Table/hooks/useSubRowFiltering.tsx
bentleyvk: To get default `reat-table` filter functions. https://github.com/tannerlinsley/react-table/blob/master/src/filterTypes.js

src/core/Table/TableRowMemoized.tsx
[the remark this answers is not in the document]
bentleyvk: Yeah, I don't like setting paddingLeft in two different files.

src/core/Table/Table.tsx
veekeys: Well, indeterminate can be achieved only through js. So user click on a checkbox to make it indeterminate looks weird. There is css-tricks article, where they made rotating checkbox UI example...https://css-tricks.com/indeterminate-checkboxes/
I guess the only argument is that if user clicks and gets indetermiante, it is not clear, what is actually selected, what is not. Cause common UX is when you click, you check or uncheck.
Dont know... But in that GIF, if you uncheck indeterminate, nothing really changes in children selection (they should go unchecked..?).

I guess need to try to find same examples in other libs and see how they handle..

src/core/Table/TableRowMemoized.tsx
veekeys: Yeah, that was the idea, but not sure it is better now :(

src/core/Table/Table.tsx
bentleyvk: Or we could change how our checkbox works: instead of going from indeterminate to true, it should go from indeterminate to false.

jest.config.js
veekeys: Didnt need to allowJs in our tsconfig???

src/core/Table/Table.tsx
veekeys: thats a big one...

src/core/Table/TableRowMemoized.tsx
mayank99: That applies the same class name to all cells in the column. What if there is a need to customize per row?

bentleyvk: You could use `rowProps` to pass custom class that has `nth-child` selector.

src/core/Table/hooks/useSubRowFiltering.tsx
mayank99: Cant import from `react-table/src/filterTypes`?

src/core/Table/Table.tsx
mayank99: Ah ok, horizontal shift.

Can we at least make the default true (more useful and common)? It seems wrong that everyone needs to manually set this to true (just for the few _potential_ cases of lazy loading where first set doesn't have subrows).

veekeys: I dont think it should default to true. Its not that often case to show subrows at all :/

src/core/Table/Table.tsx
bentleyvk: Removed during merge, brought it back.

src/core/Table/actionHandlers/selectHandler.ts
bentleyvk: Added, idk if it helps.

src/core/Table/hooks/useSubRowFiltering.tsx
bentleyvk: Apparently I can 😮 I tried previously and intellisense didn't show anything.

mayank99: Are you also able to remove `* as`?

bentleyvk: Apparently no.

src/core/Table/TableRowMemoized.tsx
bentleyvk: Nice, used your suggestion!

jest.config.js
bentleyvk: For some reason no. I also checked in demo app and everything seems to be working.

src/core/Table/hooks/useSubRowFiltering.tsx
bentleyvk: Added this issue to comment: https://github.com/tannerlinsley/react-table/issues/2928

src/core/Table/Table.tsx
bentleyvk: Created an issue for this: https://github.com/iTwin/iTwinUI-react/issues/271

veekeys (approved):

mayank99 (approved):

PR #270, opened 2021-08-18
fix(HeaderButton): allow passing undefined `menuItems`

### LINKED ISSUE #269: HeaderButton pass menuItems prop to underlying button resulting in React warning
When using `HeaderButton` component, if I pass an `undefined` `menuItems` prop, react is complaining that it does not recognise `menuItems` as valid for the `button`:
image

#### Expected behavior
No warning is expected when I provide a prop with an `undefined` value to an optional prop, while I dont have statistics, I think this is a common use of optional props.

#### Reproduction
Link to a minimal repro: https://codesandbox.io/s/itwinui-react-example-forked-s1y94?file=/src/App.tsx

##### Steps to reproduce
In the repro above, 
Open the console drawer, notice that it is empty.

Click on the first "Toggle menu items" link, notice that the error shown above will display.

Reload the page to "reset" the errors. (Otherwise, it will not display more error, but that doesnt mean anything 😄, it would not show a second error even if we have 2 of the same behavior )

Click on the second "Toggle menu items" link, notice that no error shows up. 

#### Additional information
The second method I used to do the optional prop actually destructure an object, which contains or not the `menuItems` property.
This way, when the underlying component applies the `...rest` to the `button` component, the `menuItems` prop does not exist and it does not complain. (Howerver, I dont think this is a common way of applying conditional props to a component)

mayank99:
Closes #269

## Checklist

### DIFF
--- a/src/core/Header/HeaderButton.tsx
+++ b/src/core/Header/HeaderButton.tsx
@@ -65,6 +65,7 @@ export const HeaderButton = (props: HeaderButtonProps) => {
     isActive = false,
     className,
     startIcon,
+    menuItems,
     ...rest
   } = props;
 
@@ -98,6 +99,7 @@ export const HeaderButton = (props: HeaderButtonProps) => {
         {description && <div className='iui-description'>{description}</div>}
       </>
     ),
+    ...(!!menuItems && { menuItems }),
     ...rest,
   };
 


src/core/Header/HeaderButton.tsx
```diff
@@ -98,6 +99,7 @@ export const HeaderButton = (props: HeaderButtonProps) => {
         {description && <div className='iui-description'>{description}</div>}
       </>
     ),
+    ...(!!menuItems && { menuItems }),
```
veekeys, line 102: Interesting..if I removed your changes and passed undefined in story, there is no error...
Anyway, something like `...{ menuItems }` would not work too?

mayank99: Maybe storybook does something about undefined values internally then.

`...{ menuItems }` would not change current behavior because it's already there (with `undefined` value).

veekeys (approved):

PR #275, opened 2021-08-23
feat: Added native props to Select and SelectOption

### LINKED ISSUE #108: Select: add onBlur support and maintain focus
The Select component in iTwinUI-react does not support an onBlur prop & callback. When the drop-down is opened, the Select component seems to lose focus; the focus indicator is no longer around the button. When an option is chosen, focus does not return to the button.

Can the Select component maintain focus while the drop-down is open?
Can support for an onBlur prop & callback be added?

### LINKED ISSUE #225: Add Id props to SelectOption<T>

### LINKED ISSUE #238: Select/MenuItem add value props as data-value={value}

veekeys:
- Added native div props to Select (checked that Select native props would not work on our component)
- Added `CommonProps` to option and added support to any other prop passed in there

Closes #108, closes #238, closes #225


src/core/Select/Select.tsx
```diff
@@ -109,7 +113,18 @@ export type SelectProps<T> = {
    */
   popoverProps?: Omit<PopoverProps, 'onShow' | 'onHide' | 'disabled'>;
 } & Pick<PopoverProps, 'onShow' | 'onHide'> &
-  CommonProps;
+  CommonProps &
+  Omit<
+    React.HTMLAttributes<HTMLDivElement>,
```
mayank99, line 118: Better to use `React.ComponentPropsWithoutRef<'div'>`

src/core/Select/Select.tsx
```diff
@@ -51,7 +51,11 @@ export type SelectOption<T> = {
    * Item is disabled.
    */
   disabled?: boolean;
-};
+  /**
+   * Any other props.
+   */
+  [key: string]: unknown;
+} & CommonProps;
```
mayank99, line 58: This should also be added to `MenuItemProps` then

src/core/Select/Select.tsx
mayank99: Or maybe even `React.ComponentPropsWithoutRef<'select'>`

src/core/Select/Select.tsx
veekeys: only there or keep both?

src/core/Select/Select.tsx
veekeys: Select no, I can change to div..
Whats the diff again?

src/core/Select/Select.tsx
mayank99: might not work if it's only there

src/core/Select/Select.tsx
mayank99: Well, for one, you can be explicit about ref.

Second, it's easier to read/remember I guess

mayank99 (approved):

PR #278, opened 2021-08-23
feat: Add new `VerticalTabs` component

### LINKED ISSUE #279: Tabs: vertical orientation
Need vertical tabs in this lib. CSS is ready: https://github.com/iTwin/iTwinUI/blob/main/src/tabs/classes.scss#L13

image

mayank99:
- Added new `VerticalTabs` component.
- Converted `HorizontalTabs` into a new internal `Tabs` component.
  - Added new prop `orientation`, which defaults to `'horizontal'` and supports `'vertical'`.
    - Setting to `'horizontal'` uses default behavior as before.
    - Setting to `'vertical'` adds `iui-vertical` modifer on wrapper div.
    - Setting to `'vertical'` also changes stripe inline styles (for positioning).
  - Added new story and tests.
- Deprecated `HorizontalTab` and renamed to `Tab`. Nothing else changed inside.
- Updated itwinui-css to 0.27.1

image

## Checklist


src/core/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,301 @@
...
+        </div>
+      )}
+    </div>
+  );
+};
+
+/**
+ * @deprecated Use `Tabs` instead.
+ */
+export const HorizontalTabs = Tabs;
```
mayank99, line 299: @veekeys I know you're not a fan of this approach but I think it's time 😄 
@bentleyvk I heard you're the one responsible for the original name here, so I'm interested in your thoughts too 😄 

---

If we are scared of deprecations, then one alternative would be to make these new `Tabs` components internal, and then expose `HorizontalTabs` and `VerticalTabs` on top of it. Any other way would lead to too much duplicate code.

bentleyvk: It was one of the first components that I did and I think I went with keeping the same name also iModel.js I think had the same naming.
The bigger question for me is why you named not `Tab` but `HorizontalTab`? 😀 Because I came back from vacation and it was a surprise for me as we already had an item for vertical tabs.
I would go with separate horizontal and vertical tabs comoponents as vertical doesn't support pill style. Also, might help to avoid another breaking change in the future if there will be some bigger differences between these two. But I agree that we could have them as wrappers to `Tabs` component.

src/core/Tabs/Tabs.tsx
```diff
@@ -172,6 +194,7 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
 
     let newIndex = focusedIndex ?? currentActiveIndex;
     switch (event.key) {
+      case 'ArrowDown':
```
bentleyvk, line 197: Shouldn't we check for orientation too?

stories/core/Tabs.stories.tsx
```diff
@@ -6,26 +6,22 @@ import SvgStar from '@itwin/itwinui-icons-react/cjs/icons/Star';
...
     color: 'blue',
   },
   argTypes: {
     children: { control: { disable: true } },
   },
-} as Meta<HorizontalTabsProps>;
+} as Meta<TabsProps>;
 
-const Template: Story<HorizontalTabsProps> = (args) => {
+const Template: Story<TabsProps> = (args) => {
```
bentleyvk, line 24: Can you make that instead of `Template`, `Default` would be reused? It seems that then it would be enough just to change the type in other stories.

src/core/Tabs/Tabs.tsx
veekeys: I vote for separate component.. And we can have any inner code reuse, if needed.

src/core/Tabs/Tabs.tsx
mayank99: I thought it would be ok to have both in both orientation, but [W3 disagrees](https://www.w3.org/TR/wai-aria-practices-1.1/#:~:text=if%20the%20tab%20list%20is%20horizontal%2C%20it%20does%20not%20listen%20for%20down%20arrow%20or%20up%20arrow%20so%20those%20keys%20can%20provide%20their%20normal%20browser%20scrolling%20functions%20even%20when%20focus%20is%20inside%20the%20tab%20list.).

src/core/Tabs/Tabs.tsx
```diff
@@ -165,26 +187,51 @@ export const HorizontalTabs = (props: HorizontalTabsProps) => {
...
       case 'ArrowLeft': {
-        do {
-          newIndex = (newIndex - 1 + labels.length) % labels.length;
-        } while (isTabDisabled(newIndex) && newIndex !== focusedIndex);
-        focusTabAt(newIndex);
+        if (orientation === 'horizontal') {
+          focusPreviousTab();
+          event.preventDefault();
+        }
         event.preventDefault();
```
bentleyvk, line 235: Leftover

mayank99: Fixed

src/core/Tabs/Tabs.tsx
mayank99: Added `VerticalTabs` wrapper. Kept deprecation on individual `HorizontalTab`.

stories/core/Tabs.stories.tsx
mayank99: Think I need new story using `VerticalTabs`

stories/core/Tabs.stories.tsx
```diff
@@ -50,7 +57,7 @@ DefaultTabs.args = {
   ],
 };
 
-export const BorderlessTabs = Template.bind({});
+export const BorderlessTabs = DefaultTabs.bind({});
 BorderlessTabs.args = {
   labels: [
```
bentleyvk, line 62: Can you remove `labels`?

mayank99: 🤔 It will not work if I do. The story does not have labels.

bentleyvk: Shouldn't it take them from `DefaultTabs`?

bentleyvk: tested and saw 😀

bentleyvk (approved):

PR #282, opened 2021-08-24
feat(Modal): focus is now trapped inside modal

### LINKED ISSUE #195: Can use tab key to focus out of modal
I can use the tab key to send the focus outside of the modal.  See the focus is on the side navigation in this screenshot.
Screen Shot 2021-07-08 at 5 24 47 PM

veekeys:
Closes #195


src/core/utils/FocusTrap.tsx
```diff
@@ -0,0 +1,65 @@
...
+    ) {
+      (lastElement as HTMLElement).focus();
+    } else {
+      (firstElement as HTMLElement).focus();
+    }
+  };
+
+  return (
+    <div>
+      <div tabIndex={0} onFocus={onFirstFocus} />
```
mayank99, line 53: So I was hoping that these two divs would simplify the implementation and you wouldn't need to find all focusable elements.

Since that is not the case, maybe just use first/last elements to loop focus? I imagine you'd have to add `onKeyDown` (or maybe `onBlur`). https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex#modals_and_keyboard_traps


There is also an accessibility benefit to removing the empty divs, because tabbable elements are announced by screen readers and should not be hidden.

src/core/Modal/Modal.tsx
```diff
@@ -109,9 +110,11 @@ export const Modal = (props: ModalProps) => {
...
   const originalBodyOverflow = React.useRef('');
 
+  const childrenRef = React.useRef<HTMLDivElement>(null);
+
   // Give focus to overlay for key handling to work.
-  React.useEffect(() => {
-    if (isOpen) {
+  React.useLayoutEffect(() => {
+    if (isOpen && !overlayRef.current?.contains(document.activeElement)) {
       overlayRef.current?.focus();
```
mayank99, line 118: When modal is opened using keyboard, there is a visible outline around the overlay. So we should focus the dialog instead, and update the focus styling for the dialog.

src/core/utils/FocusTrap.tsx
```diff
@@ -0,0 +1,65 @@
...
+      event.relatedTarget === firstElement
+    ) {
+      (lastElement as HTMLElement).focus();
+    } else {
+      (firstElement as HTMLElement).focus();
+    }
+  };
+
+  return (
+    <div>
```
mayank99, line 52: Do you need this extra div or can you just use fragment?

src/core/utils/FocusTrap.tsx
veekeys: I had it done, none of you liked it. Code is less nice.
image
It was with hook, but pretty much same..

src/core/utils/FocusTrap.tsx
veekeys: I addded it, cause I noticed we have test which is checking modal creation under container with the children count...
So if I dont have wrapper, 1 modal creates 3 elements under the container, which might be not that nice, cause if you have 2 modals, then is like 6 children elements and hard to get what is what..

src/core/utils/FocusTrap.tsx
veekeys: I added hook now too..
Could you please have a look and cast your vote: Hook or Component.
@bentleyvk

src/core/Modal/Modal.tsx
veekeys: Amm.. idk about this one.. Not sure, which one should be focused (we can adjust styling of both, of course)

src/core/utils/FocusTrap.tsx
bentleyvk: My vote goes to.... divs. It looks better and easier to understand. Just add `aria-hidden` to those divs and it should be ok with those screen readers.

mayank99: `aria-hidden` on focusable elements is considered a bad practice for both accessibility and SEO

https://www.w3.org/TR/using-aria/#fourth
https://web.dev/aria-hidden-focus/

src/core/Modal/Modal.tsx
mayank99: how would you style focus on overlay?

src/core/utils/FocusTrap.tsx
mayank99: Shouldn't we fix the test then?

mayank99 (approved):

PR #290, opened 2021-08-24
feat(Wizard): improve accessibility and forward ref

### LINKED ISSUE #277: Wizard: improve a11y and keyboard navigation
Wizard should have keyboard navigation similar to HorizontalTabs, and some aria attributes.

mayank99:
- Added tabIndex and onKeyDown to clickable Steps.
- Changed aria-current=true to aria-current=step.
- Forwarded ref to allow user to modify aria attributes on any level.
- Updated itwinui-css to 0.27.3.

Closes #277

### DIFF
--- a/src/core/Wizard/Step.tsx
+++ b/src/core/Wizard/Step.tsx
@@ -77,7 +77,8 @@ export const Step = (props: StepProps) => {
         ...style,
       }}
       onClick={onCompletedClick}
-      aria-current={isActive}
+      aria-current={isActive ? 'step' : undefined}
+      tabIndex={0}
       {...rest}
     >
       <div className='iui-wizard-track-content'>
--- a/src/core/Wizard/Step.tsx
+++ b/src/core/Wizard/Step.tsx
@@ -55,10 +55,21 @@ export const Step = (props: StepProps) => {
 
   const isPast = type !== 'workflow' && currentStepNumber > index;
   const isActive = type !== 'workflow' && currentStepNumber === index;
+  const isClickable = type !== 'workflow' && isPast && !!onClick;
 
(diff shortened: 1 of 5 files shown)


src/core/Wizard/Wizard.tsx
```diff
@@ -55,43 +55,43 @@ const defaultWizardLocalization: WizardLocalization = {
 };
 
 /**
- * A Wizard.
- * Give it some steps.
- * Tell it what step to display.
- * Set the type if you'd like.
- * That's about it.
+ * A wizard displays progress through a sequence of logical and numbered steps.
```
veekeys, line 58: Kinda liked that previous description.. Was fun to read..But maybe this is more professional.

src/core/Wizard/Step.tsx
```diff
@@ -77,7 +77,8 @@ export const Step = (props: StepProps) => {
         ...style,
       }}
       onClick={onCompletedClick}
-      aria-current={isActive}
+      aria-current={isActive ? 'step' : undefined}
+      tabIndex={0}
```
veekeys, line 81: I do not think all he steps should be focusable.. Only the interactive ones, IMO (previous).
Now styling is awful for current and next ones:
image
image
Workflow should not have it at all.

image

veekeys: Also, keyboard click does not work :(

veekeys: Test needed

mayank99: Fixed all

veekeys (approved):

saaaaaally (approved):

PR #292, opened 2021-08-25
fix(Table): Fixed failing user tests by copying filter functions from `react-table`

bentleyvk:
- `react-table` bundled code does not export `filterTypes` so can't use it from there.
 - `instance.filterTypes` does not have default filter functions, only ours.
 - I tried adding `filterTypes.js` to `tsconfig.build.json` `files` property. It compiled that file but after running `yarn install` in a test app, that compiled `filterTypes.js` is removed. Also, it changes folder structure by wrapping everything in `src` what could be a "breaking change" for our users as some of them are still importing using direct paths.

## Checklist

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -33,11 +33,9 @@ module.exports = {
 
   // An array of regexp pattern strings used to skip coverage collection
 
-  // coveragePathIgnorePatterns: [
-
-  //   "\\\\node_modules\\\\"
-
-  // ],
+  coveragePathIgnorePatterns: [
+    'src/core/Table/filters/defaultFilterFunctions.ts',
+  ],
 
   // A list of reporter names that Jest uses when writing coverage reports
 
@@ -72,7 +70,6 @@ module.exports = {
   globals: {
     'ts-jest': {
       tsconfig: {
-        allowJs: true,
         sourceMap: true,
       },
     },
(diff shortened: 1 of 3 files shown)


jest.config.js
```diff
@@ -33,11 +33,9 @@ module.exports = {
 
   // An array of regexp pattern strings used to skip coverage collection
 
-  // coveragePathIgnorePatterns: [
-
-  //   "\\\\node_modules\\\\"
-
-  // ],
+  coveragePathIgnorePatterns: [
+    'src/core/Table/filters/defaultFilterFunctions.ts',
```
bentleyvk, line 37: I don't want to write tests for each function 😀 So not to reduce our coverage, I am excluding that file.

src/core/Table/hooks/useSubRowFiltering.tsx
```diff
@@ -57,7 +55,17 @@ const useInstance = <T extends Record<string, unknown>>(
         return;
       }
 
-      const filterTypes = { ...defaultFilterTypes, ...instance.filterTypes };
+      const filterTypes: Record<
```
bentleyvk, line 58: As it is not of type `any`, TS is crying that `string` can't be used as a key.

jest.config.js
```diff
@@ -236,13 +234,15 @@ module.exports = {
   // A map from regular expressions to paths to transformers
 
   transform: {
-    '^.+\\.(ts|tsx|js)$': 'ts-jest',
+    '^.+\\.(ts|tsx)$': 'ts-jest',
```
veekeys, line 237: WHy js needed to remove?

bentleyvk: I reverted my changes that was done to make imported `filterTypes.js` to work. Here is how it was before:
image
https://github.com/iTwin/iTwinUI-react/pull/266/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1L232

veekeys: ok..
As long as you tested and it works, I would suggest releasing this as patch.

veekeys (approved):

mayank99 (approved):

PR #300, opened 2021-08-30
fix(Tooltip): Forward ref to allow nesting

### LINKED ISSUE #299: Tooltip component should be a forwardRef component
Tooltips isn't a forwardRef component and cannot be add in a DropdownMenu. If added, tooltip work, but no dropdown. 
```
<DropdownMenu menuItems={items}>
  <Tooltip content={"text"}>
    <IconButton>
      <SvgSomething/>
    </IconButton>
  </Tooltip>
</DropdownMenu>
```
`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

Can work with this workflow but unnecessary div wrapper is needed : 
```
<DropdownMenu menuItems={items}>
  <div>
    <Tooltip content={"text"}>
      <IconButton>
        <SvgSomething/>
      </IconButton>
    </Tooltip>
  </div>
</DropdownMenu>
```

#### Expected behavior
No wrapper div is necessary in this case. Seem like Tooltip was design to allow ref as props. Maybe this package could help : react-merge-refs ?

#### Suggestion : 
Having some props of tooltip in Button component ?

mayank99:
Closes #299

## Checklist


src/core/Tooltip/Tooltip.tsx
```diff
@@ -33,17 +33,8 @@ export type TooltipProps = {
...
-    content,
-    children,
-    className,
-    style,
-    visible,
-    ref,
-    id,
-    ...rest
-  } = props;
+export const Tooltip = React.forwardRef<Element, TooltipProps>((props, ref) => {
```
veekeys, line 36: We always used `HTMLElement` as type. Any reason you chose Element here?

src/core/Tooltip/Tooltip.tsx
```diff
@@ -33,17 +33,8 @@ export type TooltipProps = {
...
+  const { content, children, className, style, visible, id, ...rest } = props;
```
veekeys, line 37: So even though we pass ref on the `Popover` component, at the end it appears on children? (according to your test)
Then I am lost a little, whats the point of this ref if user can add ref on his provided children himself...

src/core/Tooltip/Tooltip.tsx
mayank99: I kept the same type as the one from tippy props

src/core/Tooltip/Tooltip.tsx
mayank99: I was a little surprised too, that it appears on children.

I think the point is to allow nesting tippys. See linked issue and also see #296

veekeys: So now the scenario mentioned is working well? without the need of div?

mayank99: Yes, I tested this
image
image

src/core/Tooltip/Tooltip.tsx
veekeys: Why I ask, cause usually ppl create refs of `HTMLElement` and then access certain props, which `Element` might not have.. And I wildly guess, that if you create `HTMLElement` ref it wont allow you to set it on Tooltip then..

mayank99: But it won't let me use HTMLElement
image

src/core/Tooltip/Tooltip.tsx
```diff
@@ -68,6 +59,6 @@ export const Tooltip = (props: TooltipProps) => {
       {children && React.cloneElement(children, { title: undefined })}
     </Popover>
```
maxime4000, line 60: Does adding the ref in reference={ref} instead of ref in popover make it work ?

mayank99: That's not how `reference` is supposed to be used

PR #307, opened 2021-09-01
feat(Table): Editable cell

bentleyvk:
- Moved all cell related stuff to `TableCell`. 
- Callback is only called when `onBlur` is fired to avoid many `Table` rerenders when typing that causes refocus and caret position reset.
 - Made `isCellEditable` a function if someone might need conditionally enable edit. Just a safe bet.

editable-cell3

## Checklist


src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+export type OnCellEditCallbackParams<T extends Record<string, unknown>> = {
+  columnId: string;
+  value: string;
+  rowData: T;
+};
+
+export type EditableCellProps<T extends Record<string, unknown>> = {
+  cell: Cell<T>;
+  cellProps: TableCellProps;
+  onCellEdit?: (cellEditParams: OnCellEditCallbackParams<T>) => void;
```
bentleyvk, line 17: I am not sure whether to make params object or just each param separately `(columnId, value, rowData) => void`.
Currently made it as a object because I saw it somewhere like that 😀

bentleyvk:
Visual tests should not fail after https://github.com/iTwin/iTwinUI-react/pull/306 is merged.

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { getCellStyle } from './utils';
+import EditableCell, { OnCellEditCallbackParams } from './EditableCell';
+import SubRowExpander from './SubRowExpander';
+import { SELECTION_CELL_ID } from './hooks';
+
+export type TableCellProps<T extends Record<string, unknown>> = {
+  cell: Cell<T>;
+  cellIndex: number;
+  onCellEdit?: (cellEditParams: OnCellEditCallbackParams<T>) => void;
+  isRowDisabled: boolean;
```
mayank99, line 17: `isDisabled` would make more sense, no? Why do we care about row?

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,87 @@
...
+  tableHasSubRows: boolean;
+  tableInstance: TableInstance<T>;
+  expanderCell?: (cellProps: CellProps<T>) => React.ReactNode;
+};
+
+const TableCell = <T extends Record<string, unknown>>(
```
mayank99, line 23: export this?

src/core/index.ts
```diff
@@ -122,6 +122,7 @@ export type {
   TableFilterValue,
   DateRangeFilterOptions,
   FilterButtonBarProps,
+  OnCellEditCallbackParams,
```
mayank99, line 125: Not sure we need to export this.

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [isDirty, setIsDirty] = React.useState(false);
+
+  return (
+    <div
+      {...cellProps}
+      contentEditable
+      suppressContentEditableWarning
+      onInput={(e) => {
+        setValue(
+          (e.target as HTMLElement).innerText.replace(/(\r\n|\n|\r)/gm, ''),
```
mayank99, line 36: Maybe add a comment saying you're removing new lines here

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+  children: React.ReactNode;
+};
+
+const EditableCell = <T extends Record<string, unknown>>(
+  props: EditableCellProps<T>,
+) => {
+  const { cell, cellProps, onCellEdit, children } = props;
+
+  const [value, setValue] = React.useState(cell.value);
+  const [isDirty, setIsDirty] = React.useState(false);
```
mayank99, line 27: Is this a standard name? Something like `didChange` or `isChanged` or `isModified` might be more clear.

Or add a comment.

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { SELECTION_CELL_ID } from './hooks';
+
+export type TableCellProps<T extends Record<string, unknown>> = {
+  cell: Cell<T>;
+  cellIndex: number;
+  onCellEdit?: (cellEditParams: OnCellEditCallbackParams<T>) => void;
+  isRowDisabled: boolean;
+  tableHasSubRows: boolean;
+  tableInstance: TableInstance<T>;
+  expanderCell?: (cellProps: CellProps<T>) => React.ReactNode;
```
mayank99, line 20: This is not really a "cell" if it's part of another cell.

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+};
+
+export type EditableCellProps<T extends Record<string, unknown>> = {
+  cell: Cell<T>;
+  cellProps: TableCellProps;
+  onCellEdit?: (cellEditParams: OnCellEditCallbackParams<T>) => void;
+  children: React.ReactNode;
+};
+
+const EditableCell = <T extends Record<string, unknown>>(
```
mayank99, line 21: export module

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+      suppressContentEditableWarning
+      onInput={(e) => {
+        setValue(
+          (e.target as HTMLElement).innerText.replace(/(\r\n|\n|\r)/gm, ''),
+        );
+        setIsDirty(true);
+      }}
+      onBlur={() =>
+        isDirty &&
+        onCellEdit?.({
```
mayank99, line 42: So I was thinking that we might handle the cell editing internally, i.e. update table data on blur.

But that might be more complicated and less flexible, which makes me wonder... Why not just provide this `EditableCell` component (with styling) for the user to pass into `Cell` or `defaultCell`? Or is there some data/state that the user may not have access to?

src/core/Table/EditableCell.tsx
mayank99: object is less clear to me

src/core/Table/EditableCell.tsx
bentleyvk: I know Angular forms had this field `isDirty`

src/core/Table/TableCell.tsx
bentleyvk: Just left the same naming as it comes from the `Table` 🤷‍♂️

src/core/Table/EditableCell.tsx
mayank99: Yeah, I think `cellRenderer` would be a better approach than adding all these extra props. It would also future proof us for other kinds of cells (e.g. status cell).

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+  props: EditableCellProps<T>,
+) => {
+  const { cell, cellProps, onCellEdit, children } = props;
+
+  const [value, setValue] = React.useState(cell.value);
+  const [isDirty, setIsDirty] = React.useState(false);
+
+  return (
+    <div
+      {...cellProps}
```
mayank99, line 31: I think the user should have some way to override or add to `onInput` and `onBlur`.

The current approach of holding update until onBlur is a bit fragile. What if onBlur is never called because the end-user stays on the cell for too long? In such a case, it might be helpful to update the table after a set delay.

stories/core/Table.stories.tsx
```diff
@@ -1324,3 +1325,100 @@ Full.args = {
...
+            isCellEditable: (rowData) => rowData.description !== 'Fetching...',
+            Filter: tableFilters.TextFilter(),
+          },
+        ],
+      },
+    ],
+    [],
+  );
+
+  const onCellEdit = useCallback(
```
mayank99, line 1372: Not sure if you plan to keep this approach, but if you do, then you should add an example like this callback in the JSDoc comment of `onCellEdit` prop. Because it's not very obvious that the user needs to update the data on their end.

src/core/Table/Table.test.tsx
```diff
@@ -1486,3 +1486,97 @@ it('should render sub-rows with custom expander', () => {
...
+          id: 'description',
+          Header: 'description',
+          accessor: 'description',
+        },
+        {
+          id: 'view',
+          Header: 'view',
+          Cell: () => {
+            return <span>View</span>;
+          },
```
mayank99, line 1513: Huh?

mayank99: Oh ok I see you copied from other test and removed `onClick`. Maybe could remove this entire column as it doesn't do anything for this test.

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,87 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { Cell, CellProps, TableInstance } from 'react-table';
+import cx from 'classnames';
+import { getCellStyle } from './utils';
+import EditableCell, { OnCellEditCallbackParams } from './EditableCell';
+import SubRowExpander from './SubRowExpander';
```
mayank99, line 10: Forgot to mention in previous PR but probably should export ES module for `SubRowExpander` instead of using default export.

src/core/Table/EditableCell.tsx
```diff
@@ -0,0 +1,54 @@
...
+import React from 'react';
+import { Cell, TableCellProps } from 'react-table';
+
+export type OnCellEditCallbackParams<T extends Record<string, unknown>> = {
+  columnId: string;
+  value: string;
+  rowData: T;
+};
+
+export type EditableCellProps<T extends Record<string, unknown>> = {
```
gretanausedaite, line 14: Maybe some comments or description is missing.

bentleyvk: Added

src/core/Table/Table.test.tsx
bentleyvk: `assertRowsData` expects that there will be `View` column, so I need to add it.

src/core/Table/TableCell.tsx
bentleyvk: Renamed

bentleyvk: Changed to just params

src/core/Table/TableCell.tsx
bentleyvk: Exported

src/core/index.ts
bentleyvk: Removed

src/core/Table/EditableCell.tsx
bentleyvk: Added

src/core/Table/EditableCell.tsx
bentleyvk: Exported

stories/core/Table.stories.tsx
bentleyvk: Added a comment on `onCellEdit` prop.

src/core/Table/TableCell.tsx
bentleyvk: Added

src/core/Table/EditableCell.tsx
bentleyvk: Extended div props and calling user `onInput` and `onBlur` inside ours.

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,93 @@
...
+          isDisabled={isDisabled}
+          cellProps={cellProps}
+          expanderCell={expanderCell}
+        />
+      )}
+      {cell.render('Cell')}
+    </>
+  );
+
+  const cellRendererProps = {
```
veekeys, line 76: Maybe define the type?

src/core/Table/TableCell.tsx
```diff
@@ -0,0 +1,93 @@
...
+export type TableCellProps<T extends Record<string, unknown>> = {
+  cell: Cell<T>;
+  cellIndex: number;
+  isDisabled: boolean;
+  tableHasSubRows: boolean;
+  tableInstance: TableInstance<T>;
+  expanderCell?: (cellProps: CellProps<T>) => React.ReactNode;
+};
+
+export const TableCell = <T extends Record<string, unknown>>(
```
veekeys, line 22: Why this is not in the cells folder?

src/core/Table/Table.test.tsx
```diff
@@ -1486,3 +1487,51 @@ it('should render sub-rows with custom expander', () => {
...
+  const rows = container.querySelectorAll('.iui-table-body .iui-row');
+  assertRowsData(rows);
+
+  const editableCells = container.querySelectorAll(
+    '.iui-cell[contenteditable]',
+  );
+  expect(editableCells).toHaveLength(3);
+
+  fireEvent.input(editableCells[1], {
+    target: { innerText: 'test-data\n\r\r\n' },
```
veekeys, line 1533: why new line is needed in inner text?

src/core/Table/cells/DefaultCell.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { CellRendererProps } from 'react-table';
+
+export type DefaultCellProps<
+  T extends Record<string, unknown>
+> = CellRendererProps<T> & React.ComponentPropsWithoutRef<'div'>;
+
+/**
+ * Default cell.
+ * It should be passed to `cellRenderer`.
+ * In most cases you shouldn't need to pass this component.
```
veekeys, line 15: Added this so users could add some native attributes on the Cell itself, right? So maybe instead of this comment, we could say that `Use this component to add native html attributes to the cell container` or something like that...

stories/core/Table.stories.tsx
```diff
@@ -1324,3 +1327,117 @@ Full.args = {
...
+    (columnId: string, value: string, rowData: TableStoryDataType) => {
+      action('onCellEdit')({ columnId, value, rowData });
+      setData((oldData) => {
+        const newData = [...oldData];
+        const index = oldData.indexOf(rowData);
+        const newObject = { ...newData[index] };
+        newObject[columnId as keyof TableStoryDataType] = value;
+        newData[index] = newObject;
+        return newData;
```
veekeys, line 1358: This is weird... Interface expect cellEdit with void return, but here you return something.. does it have influence?

src/core/Table/EditableCell.tsx
bentleyvk: Changed to just params

src/core/Table/Table.test.tsx
bentleyvk: image

veekeys: oh...
Could test both cases then, but minor, I guess..

stories/core/Table.stories.tsx
bentleyvk: Huh? It is a callback inside `setData`.

veekeys: oh crap... yeah, alignment didnt help :D

src/core/Table/TableCell.tsx
bentleyvk: I wanted to avoid confusion between this component and `DefaultCell`.

src/core/Table/TableCell.tsx
bentleyvk: Added

src/core/Table/Table.test.tsx
bentleyvk: Added another test with edge cases.

src/core/Table/cells/DefaultCell.tsx
bentleyvk: Changed.

src/core/Table/cells/EditableCell.tsx
```diff
@@ -41,39 +41,59 @@ export type EditableCellProps<
...
+      onKeyDown={(e) => {
+        // Prevents from adding HTML elements (div, br) inside a cell on Enter press
+        if (e.key === 'Enter') {
+          e.preventDefault();
+        }
+        props.onKeyDown?.(e);
+      }}
+      onPaste={(e) => {
+        e.preventDefault();
+        document.execCommand(
```
veekeys, line 85: image

stories/core/Table.stories.tsx
```diff
@@ -1324,3 +1327,120 @@ Full.args = {
...
+      });
+    },
+    [],
+  );
+
+  const cellRenderer = useCallback(
+    (props: CellRendererProps<TableStoryDataType>) => (
+      <>
+        {!isRowDisabled(props.cellProps.row.original) &&
+        props.cellProps.value !== 'Fetching...' ? (
```
veekeys, line 1368: why cursor changes on `fetching` cell even though its not editable?

veekeys: ahh, seems it was like that before everywhere...nvm

src/core/Table/cells/EditableCell.tsx
bentleyvk: I know... but there isn't any alternatives in new APIs.

veekeys (approved):

Issue #137, opened 2021-06-15
LabeledInput: inline display does not show any message (Proposal : Message in Tooltip)


maxime4000:
#### Environment
- __Package version(s)__: 1.6.0
- __Browser name and version__: Chrome latest

#### Steps to reproduce
```tsx
<LabeledInput
  type="text"
  value={value}
  status="negative"
  message="Some message"
  displayStyle="inline"
  />
```

#### Actual behavior
When LabeledInput use inline, there is not place for the message. When not inline, message can appear on the text under. (Thought, there might be other reason why this is happening. So ignore this issue.)

![image](https://user-images.githubusercontent.com/4070713/122080229-1e482880-cdcc-11eb-969c-f7555a14d565.png)

#### Expected behavior
When LabeledInput is "inline", message should appear as tooltip on status icon.

mayank99:
This is not really a bug but a feature request (and a good one!). We'll look into adding this, but there's a css refactor imminent for inline inputs, so it could take a while.

In the meantime, you can use the `svgIcon` prop. See https://codesandbox.io/s/itwinui-react-example-forked-rry56?file=/src/App.tsx

PR #309, opened 2021-09-03
feat: Add `iconDisplayStyle` prop to LabeledInput and LabeledTextarea

mayank99:
Added new `iconDisplayStyle` prop to LabeledInput and LabeledTextarea for hybrid layouts, and made all the other required changes based on iTwin/iTwinUI#287.

Moved all common InputContainer logic into internal wrapper component.

image
image

## Checklist


mayank99:
This can be reviewed if you link the CSS repo. Visual tests will fail in CI until CSS is released.

Some notes to help review:

Tried many different approaches for new prop/layouts and decided to go with `iconDisplayStyle` for simplicity and flexibility.

Message does not currently show up in inline displayStyle (so #137 is still pending). To add message as tooltip, we would need to wrap the icon in a div or span. But then it will break the borderless button case. We could either have logic to only add the wrapper for non-button elements, or we could improve the CSS so that button works with a wrapper.

Some extra images had to be updated where we have input containers without labels, as `grid-row-gap` adds 3px even when label is not present. I tried to change CSS to use margin but it got really messy so I left it.

src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -102,11 +117,13 @@ export const LabeledInput = React.forwardRef<
         ref={ref}
         {...rest}
       />
-      {(message || icon) && (
-        <div className='iui-message'>
-          {icon}
-          {displayStyle === 'default' && message}
-        </div>
+      {icon &&
+        React.cloneElement(icon, {
```
bentleyvk, line 121: Everywhere it is the same. Maybe we need `Label` type component? At first it can even be inner, later we could make it public.

veekeys: `Label` or `InputIcon` ? dont understand which part you mean

bentleyvk: Label with icon and message.

src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -79,7 +92,9 @@ export const LabeledInput = React.forwardRef<
         {
           'iui-disabled': disabled,
           [`iui-${status}`]: !!status,
-          [`iui-${displayStyle}`]: displayStyle !== 'default',
+          'iui-inline-label': displayStyle === 'inline',
+          'iui-with-message': !!message && displayStyle !== 'inline',
```
bentleyvk, line 96: Don't our hybrid cases support having message with inline styling?

stories/core/LabeledInput.stories.tsx
```diff
@@ -146,3 +147,29 @@ Inline.args = {
...
+    <>
+      <LabeledInput
+        placeholder='Enter text here'
+        label='This is a label'
+        displayStyle='inline'
+        svgIcon={<div ref={ref}>{svgIcon}</div>}
+        message='Block layout with inline icon'
+        {...rest}
+      />
+      <Tooltip reference={ref} content='Placeholder' />
```
bentleyvk, line 166: Does it work passing `Tooltip` with icon as a children to `svgIcon`?

stories/core/LabeledInput.stories.tsx
```diff
@@ -146,3 +147,29 @@ Inline.args = {
   status: 'negative',
   displayStyle: 'inline',
 };
+
+export const HybridLayout: Story<LabeledInputProps> = ({
```
bentleyvk, line 151: I think we also need to show input + button.

src/core/LabeledInput/LabeledInput.tsx
veekeys: Where message should appear then?

stories/core/LabeledInput.stories.tsx
veekeys: no, because svgIcon component does not forward ref.

stories/core/LabeledInput.stories.tsx
veekeys: you think create another story?
Maybe put another input in the same..?

stories/core/LabeledInput.stories.tsx
bentleyvk: So have it wrapped with `div` like it is now.

stories/core/LabeledInput.stories.tsx
bentleyvk: I would say separate story. We don't really have several cases in one story.

src/core/LabeledInput/LabeledInput.tsx
veekeys: not sure, how label with icon and message could look like..?

src/core/LabeledInput/LabeledInput.tsx
bentleyvk: Like in one of these, no?
image

stories/core/LabeledInput.stories.tsx
veekeys: yeah, I just started to think to introduce some :D Might be easier to have a look, but..

src/core/LabeledInput/LabeledInput.tsx
bentleyvk: Just HOC/wrapper component, you won't use it alone.

stories/core/LabeledInput.stories.tsx
bentleyvk: 😁
image

src/core/LabeledInput/LabeledInput.tsx
veekeys: It works,
image

veekeys: oh wait, need inline label..

veekeys: I think I know what Mayank had in mind here.. He didnt do that message could appear as tooltip, so didnt want to display it at all, if its inline type...

stories/core/LabeledInput.stories.tsx
veekeys: It does not work like that...

stories/core/LabeledInput.stories.tsx
```diff
@@ -173,3 +173,32 @@ HybridLayout.args = {
...
+        message='Block layout with inline borderless button'
+        iconDisplayStyle='inline'
+        {...args}
+      />
+    </>
+  );
+};
+
+HybridLayoutWithButton.args = {
+  svgIcon: (
```
bentleyvk, line 197: Is it needed here as `svgIcon` control is disabled?

stories/core/LabeledInput.stories.tsx
veekeys: added

src/core/LabeledInput/LabeledInput.tsx
veekeys: So I removed inline need for the message. Tests were not updated, so they are failing. But now I am concerned, what we shold do. This is kinda a breaking change...

stories/core/LabeledInput.stories.tsx
veekeys: yeah, can remove, I guess

src/core/LabeledInput/LabeledInput.tsx
mayank99: My initial intent was to preserve current behavior for `displayStyle: inline`, and then add a new prop for `messageDisplayStyle`.

src/core/utils/InputContainer.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export type InputContainerProps<T extends React.ElementType = 'div'> = {
+  as?: T;
+  label?: React.ReactNode;
+  disabled?: boolean;
+  required?: boolean;
+  status?: 'positive' | 'warning' | 'negative';
+  message?: React.ReactNode;
+  icon?: JSX.Element;
+  withInlineLabel?: boolean;
```
bentleyvk, line 16: These prop names sound like some custom HOC 😀 Maybe rename to `hasInlineLabel` or something with `is*`.

mayank99: Done

src/core/LabeledInput/LabeledInput.tsx
mayank99: Reverted to preserve current behavior and prevent breaking visual change.

src/core/LabeledInput/LabeledInput.tsx
```diff
@@ -15,15 +15,15 @@ export type LabeledInputProps = {
    */
   label?: React.ReactNode;
   /**
-   * Message below the input. Does not apply to 'inline' input.
+   * Message below the input.
```
bentleyvk, line 18: Revert here too

src/core/utils/InputContainer.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <Element
+      className={cx(
+        'iui-input-container',
+        {
+          'iui-disabled': disabled,
+          [`iui-${status}`]: !!status,
+          'iui-inline-label': isLabelInline,
+          'iui-inline-icon': isIconInline,
+          'iui-with-message': !!message,
```
bentleyvk, line 51: Don't we need to check whether it is inline too?

mayank99: Good catch

bentleyvk (approved):

PR #310, opened 2021-09-03
feat: Add Basic ColorPicker Component

### LINKED ISSUE #312: Provide a color-picker control
#### Feature
AWMDS is a map configuration tool that needs to allow users to define map layers. When defining a layer, aspects of the layer presentation are detailed. Those details include setting the colors of items drawn on the map. As a result there is a need for the user to be able to pick colors. Setting color by RGB value is not user friendly, a color picker capability would be very useful.

###### Possible components on github
https://uiwjs.github.io/react-color/
https://casesandberg.github.io/react-color/

###### How a color picker would be used in AWMDS
The user is editing/creating a layer, and whist defining the **Symbology** they click the **Color** control and a suitably styled color picker control appears to allow alteration of the value:

Color-Picker-Layer-Map-Config

saaaaaally:
Added Basic Color Picker with options to add custom click handler for the color swatches and tooltips

image

Closes # 

## Checklist


src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+import { getWindow } from '../utils/common';
+import { CommonProps } from '../utils/props';
+import cx from 'classnames';
+import '@itwin/itwinui-css/css/color-picker.css';
+
+export type ColorProps = {
+  /**
+   * Color
```
veekeys, line 14: Maybe end all comments with dot `.`, so its consistent.

veekeys: here could be just `Color code`

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * Is color selected
+   */
+  isActive?: boolean;
+  /**
+   * User friendly color name
+   */
+  name?: string;
+  /**
+   * User Count Icon props.
```
veekeys, line 30: Not sure what is this comment about

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export type ColorProps = {
+  /**
+   * Color
+   */
+  color: string;
+  /**
+   * Handler for clicking a color
+   */
+  onColorClicked?: () => void;
```
veekeys, line 20: it can be just `onClick`, self explanatory prop and most commonly used.

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+    color && getWindow()?.CSS?.supports?.(`--swatch-color: ${color}`)
+      ? { '--swatch-color': color, ...style }
+      : { backgroundColor: color, ...style };
+
+  return (
+    <span
+      className={cx('iui-color-swatch', { [`iui-active`]: isActive })}
+      style={_style}
+      onClick={onColorClicked}
+      tabIndex={0}
```
veekeys, line 57: Not sure about this.. I guess expectation is, that you can move between colors with arrow keys.. Usually tab is used to focus whole element, but inner switch arrow keys could be better. 
So `Color` component would get tabIndex 0 only when its active and you will need to implement keyboard navigation yourself in ColorPicker...
There is example in DatePicker to check.
@bentleyvk do you feel tabing between colors is better than arrow keys?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React, { ReactNode } from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+import Color from './Color';
+import cx from 'classnames';
+import '@itwin/itwinui-css/css/color-picker.css';
+
+export const DefaultColors = [
```
veekeys, line 11: oh, I meant a little bit simpler constant here :D
You could just have simple list of color info object, not give the `Color` component itself. People might want to reuse this list for some different purposes...
`export const DefaultColors = [{color='#FFFFFF' name='WHITE'}, {color='#5A6973' name='GREY'}, ...]`
Then need to think if we want to export this through barrel file in here or maybe move this constant to common...

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+  <Color key={26} color='#EEF6E8' name='VERDE PASTEL' />,
+  <Color key={27} color='#9BA5AF' name='SERYI GREY' />,
+  <Color key={28} color='#CF0000' name='RED EPIPHYLLUM' />,
+  <Color key={29} color='#FF6300' name='SAFETY ORANGE' />,
+  <Color key={30} color='#FFC335' name='RISE-N-SHINE' />,
+];
+
+export type ColorPickerProps = {
+  /**
+   * List of colors to show in color palette
```
veekeys, line 46: again, missing some dots at the end of the sentences.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+   * Can pass in strings of the color values for simple input or a list of Colors to customize onColorClick handler etc.
+   */
+  colors?: ReactNode[];
+  /**
+   * Style of the color picker - basic or advanced.
+   * @default 'basic'
+   */
+  type?: 'basic' | 'advanced';
```
veekeys, line 54: This is not needed for now as we have only basic one.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * Handler for clicking a color
+   * This will be applied to all colors in the palette unless specifically passed in to a specific color in the colors list
+   */
+  onColorClicked?: (index: number) => void;
+  /**
+   * Any custom nodes that will be added within the color palette
+   */
+  children?: React.ReactNode;
+};
```
veekeys, line 68: Lets take `CommonProps` here too for any custom class or style, but maybe exclude the `title` property from there.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+    activeColorIndex,
+    onColorClicked,
+    children,
+    ...rest
+  } = props;
+
+  useTheme();
+
+  // Set active color swatch
+  const [currentActiveIndex, setCurrentActiveIndex] = React.useState(
```
veekeys, line 83: YOu need to also have `React.useEffect` hook to listen to the changes of `activeColorIndex`, cause user might change it in is own way, so we need to be sure, that ColorPicker updates its state after it.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+  return (
+    <div
+      className={cx('iui-color-picker', {
+        [`iui-${type}`]: type !== 'basic',
+      })}
+      {...rest}
+    >
+      <div className='iui-color-palette'>
+        {colors &&
+          colors.length > 0 &&
```
veekeys, line 99: no need to check for length. Map will not fail with empty array.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+          colors.map((color, index) => {
+            const onClick = () => {
+              if (onColorClicked != null) {
+                onColorClicked(index);
```
veekeys, line 103: you could have optional chaining in one liner like `onColorClicked?.(index)`

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+    ...rest
+  } = props;
+
+  useTheme();
+
+  // Set active color swatch
+  const [currentActiveIndex, setCurrentActiveIndex] = React.useState(
+    activeColorIndex,
+  );
+  const onColorChanged = (index: number) => {
```
veekeys, line 86: you use it only once here, no need to wrap in function.. you can just call directly `setCurrentActiveIndex(index)`

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+            const onClick = () => {
+              if (onColorClicked != null) {
+                onColorClicked(index);
+              }
+              onColorChanged(index);
+            };
+            if (typeof color === 'string') {
+              return (
+                <Color
+                  key={index}
```
veekeys, line 110: There is a recommendation not to use index for key in react, so instead, you could have something like ``{index}+{color}``.
https://reactjs.org/docs/lists-and-keys.html

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  color={color}
+                  onColorClicked={onClick}
+                  isActive={index === currentActiveIndex}
+                />
+              );
+            } else {
+              {
```
veekeys, line 117: why so many curly brackets needed?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+                return React.isValidElement(color)
+                  ? React.cloneElement(color, {
+                      // Use the same onClick handler for all colors in palette unless a custom handler was passed in to specific color
+                      onColorClicked: color.props['onColorClicked']
+                        ? color.props['onColorClicked']
+                        : onClick,
+                      isActive: index === currentActiveIndex,
```
veekeys, line 124: I guess here we should firstly take users input and only if its undefined, apply ours?
`isActive: props.isActive ?? index === currentActiveIndex`

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+    onColorClicked,
+    isActive,
+    tooltipRefProp,
+    ...rest
+  } = props;
+
+  useTheme();
+
+  const _style =
+    color && getWindow()?.CSS?.supports?.(`--swatch-color: ${color}`)
```
veekeys, line 48: Add a test for this part of the code (can be in `Color.test.tsx`)

stories/core/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const [currentColorName, setCurrentColorName] = useState(
+    DefaultColors[activeColorIndex].props['name'],
+  );
+  const [currentColorValue, setCurrentColorValue] = useState(
+    DefaultColors[activeColorIndex].props['color'],
+  );
+
+  const onClick = (index: number) => {
+    action(`Clicked color #${index}`)();
+    setCurrentColorName(DefaultColors[index].props['name']);
```
bentleyvk, line 41: Better save index and reuse it. Also, you need to pass it to the `ColorPicker` because when you close and reopen it always show blue cola.

src/core/ColorPicker/Color.tsx
bentleyvk: I agree that arrow navigation here is better because otherwise you would need to press tab e.g. for 20 times to go out.

src/core/ColorPicker/Color.tsx
bentleyvk: We could also extend `span` props and this won't be needed.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+    type = 'basic',
+    activeColorIndex,
+    onColorClicked,
+    children,
+    ...rest
+  } = props;
+
+  useTheme();
+
+  // Set active color swatch
```
bentleyvk, line 82: I don't think comment is needed.

src/core/ColorPicker/ColorPicker.tsx
bentleyvk, line 86: This wrapper function is not really needed as it is just one function call.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+        {colors &&
+          colors.length > 0 &&
+          colors.map((color, index) => {
+            const onClick = () => {
+              if (onColorClicked != null) {
+                onColorClicked(index);
+              }
+              onColorChanged(index);
+            };
+            if (typeof color === 'string') {
```
bentleyvk, line 107: I would go with just expecting `Color` elements. That way we would avoid having this if.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,136 @@
...
+                  onColorClicked={onClick}
+                  isActive={index === currentActiveIndex}
+                />
+              );
+            } else {
+              {
+                return React.isValidElement(color)
+                  ? React.cloneElement(color, {
+                      // Use the same onClick handler for all colors in palette unless a custom handler was passed in to specific color
+                      onColorClicked: color.props['onColorClicked']
```
bentleyvk, line 121: What if we called both callbacks? What do you think @veekeys?

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+   * User friendly color name
+   */
+  name?: string;
+  /**
+   * User Count Icon props.
+   */
+  tooltipRefProp?: React.ComponentPropsWithRef<'span'>;
+} & CommonProps;
+
+export const Color = (props: ColorProps) => {
```
bentleyvk, line 35: Isn't `Color` too generic name? Maybe `ColorSwatch`?

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useTheme();
+
+  const _style =
+    color && getWindow()?.CSS?.supports?.(`--swatch-color: ${color}`)
+      ? { '--swatch-color': color, ...style }
+      : { backgroundColor: color, ...style };
+
+  return (
+    <span
+      className={cx('iui-color-swatch', { [`iui-active`]: isActive })}
```
bentleyvk, line 54: It can be just `{ 'iui-active': isActive }`

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: I fixed it.
I'm not sure if it is used anywhere else so I just left it in color picker for now.

src/core/ColorPicker/Color.tsx
saaaaaally: Fixed comments

src/core/ColorPicker/Color.tsx
saaaaaally: Fixed comment

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Fixed comments

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Added CommonProps to ColorPicker

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Changed

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Changed and removed onColorChanged

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Changed to key={index + color}

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed excess curly brackets

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed comment.
Changed to activeIndex

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  children?: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
+
+export const ColorPicker = (props: ColorPickerProps) => {
+  const { colors, activeColorIndex, onColorClicked, children, ...rest } = props;
+
+  useTheme();
+
+  const [activeIndex, setactiveIndex] = React.useState(activeColorIndex);
```
bentleyvk, line 70: It should be `setActiveIndex` with capital A.

src/core/ColorPicker/Color.tsx
veekeys: Yeah, but this has nothing to do with the tooltip.. this is the reference of the Color component, which can be used for whatever...
Check my other comment about ref, maybe that would make more sense.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <div className={'iui-color-picker'} {...rest}>
```
veekeys, line 73: Now here you need to merge classes, because user can override you `iui-color-picker` class how its written now...
Check other components, but you need to destruct the `className` from props and use here with cx like `cx('iui-color-picker', className)`

src/core/ColorPicker/ColorPicker.tsx
veekeys: I guess we can.. One would be always same, just returning the index of clicked color and other would be.. users custom one on every Color? Looks weird, but.. maybe there is a use case :D

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <div className='iui-color-palette'>
+        {colors &&
+          colors.map((color, index) => {
+            const onClick = () => {
```
gretanausedaite, line 77: Maybe move this handler outside return and just call as a function here? Confused me a bit having a const statement here.

stories/core/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const onClick = (index: number) => {
+    action(`Clicked color #${index}`)();
+    setCurrentColorName(DefaultColors[index].name);
+    setCurrentColorValue(DefaultColors[index].color);
+  };
+
+  return (
+    <>
+      <IconButton onClick={() => setOpened(!opened)}>
```
gretanausedaite, line 47: White color swatch needs some border
image
Hover: 
image

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: I added this to allow a simpler way to use by just passing a list of strings. But if we were to remove this option then maybe we don't need the colors prop at all, I'm not sure how useful it is since you can do the same thing with the children prop. Maybe we should just have the colors passed in as children only and remove the colors prop all together??

src/core/ColorPicker/Color.tsx
saaaaaally: I refactored Colors to use extend span props and removed onColorClicked

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Is this still needed after the refactoring where I removed the colors prop? Since now setting the active color is being handled in the story instead

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: This was removed and setting active color is now handled in the story

stories/core/ColorPicker.stories.tsx
saaaaaally: This is fixed now, when you close and open the color picker the selected color remains

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Fixed

src/core/ColorPicker/Color.tsx
saaaaaally: Renamed to ColorSwatch

src/core/ColorPicker/ColorPicker.tsx
veekeys: Yeah, now I have a question, why have you removed all of that? Dont remember we talked about removing all the logic.. Now every user will need to handle this himself.. and most of the cases, same.
Even if we moved colors to children, need to handle clicks and setting active color. You can convert children to array by using `React.Children.toArray(children)`. We have such examples with children in components like `UserIconGroup`. Or even `Fieldset` you were working with children directly.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,62 @@
...
+  { color: '#EEF6E8', name: 'VERDE PASTEL' },
+  { color: '#9BA5AF', name: 'SERYI GREY' },
+  { color: '#CF0000', name: 'RED EPIPHYLLUM' },
+  { color: '#FF6300', name: 'SAFETY ORANGE' },
+  { color: '#FFC335', name: 'RISE-N-SHINE' },
+];
+
+export type ColorPickerProps = {
+  /**
+   * Any custom nodes that will be added within the color palette.
```
veekeys, line 45: Here is not any custom nodes.. Maybe mention, we expect `ColorSwatch` components in here.

src/core/ColorPicker/Color.tsx
saaaaaally: I removed this prop and use forwardRef instead

src/core/ColorPicker/Color.tsx
```diff
@@ -0,0 +1,64 @@
...
+   */
+  isActive?: boolean;
+  /**
+   * User friendly color name
+   */
+  name?: string;
+  /**
+   * User Count Icon props.
+   */
+  tooltipRefProp?: React.ComponentPropsWithRef<'span'>;
```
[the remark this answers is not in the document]
saaaaaally: Added forwardRef

src/core/ColorPicker/Color.tsx
saaaaaally: Changed

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Added

stories/core/ColorPicker.stories.tsx
saaaaaally: I added a border around the color swatch in the Icon Button

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -22,33 +23,34 @@ export type ColorProps = {
...
-export const Color = (props: ColorProps) => {
-  const { color, style, onClick, isActive, tooltipRefProp, ...rest } = props;
+export const Color = React.forwardRef<HTMLSpanElement, ColorProps>(
+  (props, ref) => {
+    const { color, style, onClick, isActive, ...rest } = props;
+
+    useTheme();
 
-  useTheme();
+    const spanRef = React.useRef<HTMLSpanElement>(null);
```
veekeys, line 35: No need to create `spanRef` here as you are not using it in the component anywhere.
Just take the `ref` which you have forwarded and add it to the span (no need to merge those refs then).

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: I removed all of it because if we try to set the active color automatically it only works if the children are ALWAYS ColorSwatches, if they are anything else it will not work. 
For example in the WithTooltip story the color swatches are wrapped in tooltips so the automatic active color will not work because the children passed in were not ColorSwatches only.
I also looked at some other examples of color picker in react and it looks like they do not have the color picker automatically setting the active color in there either, they just have the onClick handler and leave that up to the user.

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: updated the comment

src/core/ColorPicker/ColorSwatch.tsx
saaaaaally: I removed it

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,63 @@
...
+} & Omit<CommonProps, 'title'>;
+
+export const ColorPicker = (props: ColorPickerProps) => {
+  const { children, className, ...rest } = props;
+
+  useTheme();
+
+  return (
+    <div className={cx('iui-color-picker', className)} {...rest}>
+      <div className='iui-color-palette'>{children}</div>
```
mayank99, line 58: I was expecting color palette logic to be part of this component.
- changing active color on click/enter/space
- `onChange` prop to know when active color changes
- arrow key navigation

I see you have some basic click handling in the story. Why not handle everything here?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -43,7 +43,7 @@ export const DefaultColors = [
 
 export type ColorPickerProps = {
   /**
-   * Any custom nodes that will be added within the color palette.
+   * Color Swatches that will be added within the color palette.
```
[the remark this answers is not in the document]
mayank99: Why does the user need to use the swatch component? We don't even export it in the barrel.

Shouldn't we handle strings for the user by wrapping them in `ColorSwatch`?

stories/core/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+export const WithTooltip: Story<ColorPickerProps> = (args) => {
+  const [opened, setOpened] = useState(false);
+
+  const [activeIndex, setCurrentActiveIndex] = React.useState(10);
+  const [currentColorName, setCurrentColorName] = useState(
+    DefaultColors[activeIndex].name,
+  );
+  const [currentColorValue, setCurrentColorValue] = useState(
+    DefaultColors[activeIndex].color,
+  );
```
mayank99, line 88: So we have 3 variables for managing the same state? I think we can show a cleaner example.

stories/core/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,132 @@
...
+import React, { useState } from 'react';
+import {
+  ColorPicker,
+  ColorPickerProps,
+  IconButton,
+  Tooltip,
+} from '../../src/core';
+import { action } from '@storybook/addon-actions';
+import ColorSwatch from '../../src/core/ColorPicker/ColorSwatch';
+import { DefaultColors } from '../../src/core/ColorPicker/ColorPicker';
```
mayank99, line 15: Now we have a mix of different kinds of imports:
- module import from barrel (ColorPicker, etc)
- default import from file (ColorSwitch)
- module import from file (DefaultColors)

Here we should think about how and what we expect the user to access. If the user needs something, it should be in the barrel. Also the default import is hiding that bug where the file name does not match the component name.

stories/core/ColorPicker.stories.tsx
```diff
@@ -0,0 +1,132 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { Story, Meta } from '@storybook/react';
+import React, { useState } from 'react';
```
mayank99, line 6: Very first item in the style guide:

https://github.com/iTwin/iTwinUI-react/blob/fc2b477bdf2cfc77328c358c9e76aa86d7be7007/STYLEGUIDE.md#L11-L12

Also you probably want to import `useState` from storybook instead.

src/core/ColorPicker/Color.tsx
saaaaaally: I added arrow key navigation

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: This was removed and moved to the story so it can be done there now

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: This was moved to the story now but it is needed within the map because I need to pass in the index value for the current color so I'm not really sure how to do this outside of the map

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
saaaaaally: I updated the comment.

Also passing in strings was my initial approach, but it got tricky to do something more complex like have a color name, custom click handler, tooltips, etc. 
So we decided to remove that option, making it more flexible and leaving it up to the user to pass in the ColorSwatches as children and let them handle the clicks and setting active color in the story.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { useTheme } from '../utils/hooks/useTheme';
+import '@itwin/itwinui-css/css/color-picker.css';
+import { CommonProps } from '../utils/props';
+import cx from 'classnames';
+
+export const DefaultColors = [
+  { color: '#FFFFFF', name: 'WHITE' },
```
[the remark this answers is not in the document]
saaaaaally: It was just like that where I copied it from. I can change it if it matters

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: I had originally tried to handle all of that within the component but decided to remove it.
After looking at some other existing examples of color picker in react it looked like they did not have the color picker automatically setting the active color, they only had the onClick handler when clicking on a color and leaving the setting active up to the user. 
Also setting active automatically got tricky trying to handle situations when they passed in anything other than a ColorSwatch. For example in the WithTooltip story the ColorSwatches are wrapped in Tooltips so the automatic active color will not work because the ColorPicker was looking for ColorSwatches in the children but found Tooltips. There will be the same problem if any other custom component is passed in.

mayank99: > After looking at some other existing examples of color picker in react it looked like they did not have the color picker automatically setting the active color

Well, you will need to handle the color selection for advanced picker (by providing an `onChange` prop). And the saved color palette in advanced picker would most certainly need to behave consistently with the slider. So you may want to do some more research and experimentation (just so that we don't need to reverse any decisions we make now).

> setting active automatically got tricky trying to handle situations when they passed in anything other than a ColorSwatch.

From what I can tell, custom components are really only needed for tooltips, so maybe we could try making tooltips part of the API and see if it offers any benefits.

> they only had the onClick handler when clicking on a color and leaving the setting active up to the user.

If you want to keep this approach, you should still be able to improve the keyboard aspect of it, by calling `onClick` whenever enter/spacebar is pressed.

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
mayank99: where did you copy it from?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
 export type ColorPickerProps = {
   /**
    * Color Swatches that will be added within the color palette.
+   * Recommended to use `ColorSwatch` components.
    */
   children?: React.ReactNode;
+  /**
+   * Index of the selected color on the color palette.
+   */
+  activeColor?: number;
```
mayank99, line 53: I don't understand... According to the story, we expect the user to handle selection state. Then why was this prop added again?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
...
   useTheme();
 
+  const ref = React.useRef<HTMLDivElement>(null);
+  const [focusedColor, setFocusedColor] = React.useState(activeColor);
+
+  React.useEffect(() => {
+    setFocusedColor(activeColor);
+  }, [activeColor]);
+
+  const handleCalendarKeyDown = (
```
mayank99, line 68: calendar?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
...
+    if (!colorSwatches?.length) {
+      return;
+    }
+
+    // If focusedColor is undefined, assume it is at index 0 and move from there
+    const currentIndex = focusedColor ?? 0;
+    let newIndex = -1;
+    switch (event.key) {
+      case 'ArrowDown':
+        newIndex = currentIndex + 10;
```
mayank99, line 81: will this really always be 10? The user might change the width or use a different swatch size.

So you might need to calculate this dynamically.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
...
+        break;
+      case 'Enter':
+      case ' ':
+      case 'Spacebar':
+        const element = colorSwatches[currentIndex] as HTMLSpanElement;
+        element?.click();
+        event.preventDefault();
+        break;
+    }
+    if (newIndex < colorSwatches.length && newIndex >= 0) {
```
mayank99, line 100: This block will be called even if Enter/Spacebar is pressed. I don't think we need that, so you could return early.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
...
+      setFocusedColor(newIndex);
+      const element = colorSwatches[newIndex] as HTMLSpanElement;
```
mayank99, line 102: This will not always be a span element, maybe use `HTMLElement`. Same thing above.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -44,17 +44,74 @@ export const DefaultColors = [
...
+        break;
+      case 'ArrowRight':
+        newIndex = currentIndex + 1;
+        break;
+      case 'Enter':
+      case ' ':
+      case 'Spacebar':
+        const element = colorSwatches[currentIndex] as HTMLSpanElement;
+        element?.click();
+        event.preventDefault();
```
mayank99, line 97: Why do you need preventDefault here? It makes sense for arrow keys but not sure about here.

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+   */
+  isActive?: boolean;
+  /**
+   * User friendly color name.
+   */
+  name?: string;
+} & CommonProps &
+  React.ComponentPropsWithRef<'span'>;
+
+export const Color = React.forwardRef<HTMLSpanElement, ColorProps>(
```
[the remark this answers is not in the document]
saaaaaally: It was originally called Color but then I renamed it and missed it here. The name should be updated now

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
saaaaaally: @gretanausedaite sent them to me, I'm not sure where she got them from.

stories/core/ColorPicker.stories.tsx
saaaaaally: Ok I cleaned it up

stories/core/ColorPicker.stories.tsx
saaaaaally: Removed

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Yeah I needed it to access the current active color, I'm not sure how to do that without this prop. Is there an example of roving tabindex somewhere else that I can refer to?

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: I was following the example of DatePicker handleCalendarKeyDown, would it not be needed here?

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Hmm I'm not sure how to do it dynamically, any ideas?

src/core/ColorPicker/ColorPicker.tsx
mayank99: Not sure if we are using roving tabindex in some other component, but it's a pretty standard pattern so you can find tons of examples online. The idea is pretty simple: [set tabindex=-1 for all children except currently focused one (which gets 0)](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex#:~:text=roving%20tabindex%20works%20by%20setting%20tabindex%20to%20-1%20for%20all%20children%20except%20the%20currently-active%20one). And then you can find 0 easily, without needing an `activeColor` prop.

Alternatively, you might be able to just use [`Document.activeElement`](https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement) if you force focus on active swatch when color picker is opened.

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Fixed name

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Added a return statement for Enter/spacebar case

stories/core/ColorPicker.stories.tsx
saaaaaally: I added ColorSwatch to index.ts so now you can import from src/core like ColorPicker. 

I'm not sure what to do about Default Colors.

mayank99: Well, you could just export DefaultColors too 😄

src/core/ColorPicker/ColorPicker.tsx
mayank99: You don't need the `onClick` const anymore, you can just inline the `onColorClick` in ColorSwatch

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed for loop and used findIndex instead

stories/core/ColorPicker.stories.tsx
saaaaaally: Done

src/core/ColorPicker/ColorPicker.tsx
mayank99: Somehow the code became even worse. Now you have two instances of `findIndex`, in addition to 3 branches in your if-else. Why is it so complicated?

src/core/ColorPicker/ColorPicker.tsx
mayank99: Well you should have a reason for every single line of code. I tested a little bit and it seems that this line prevents scrolling to the current element when Enter/Spacebar is pressed. Not sure that's how it should be, but I guess it's fine if it's in DatePicker too.

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
mayank99: It seems that these random colors were pulled from the old bwc palette for prototyping purposes, and the names came from https://colornamer.robertcooper.me/ (which is why are they are uppercase).

I (and @FlyersPh9) think we will need to come up with a proper standard color list if we are going to export it to users.

For now I would suggest removing this list from the component and hardcoding it in the story. Unless @bentleyvk or @veekeys has a different opinion.

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,138 @@
...
+      case 'ArrowUp':
+        // Look backwards for next ColorSwatch with same offsetLeft value
+        for (let i = currentIndex - 1; i >= 0; i--) {
+          const element = colorSwatches[i] as HTMLElement;
+          if (element.offsetLeft == currentColor.offsetLeft) {
+            newIndex = i;
+            break;
+          }
+        }
+        break;
```
mayank99, line 106:
```suggestion
      case 'ArrowDown': {
        // Look for next ColorSwatch with same offsetLeft value
        const newIndex = colorSwatches.findIndex(
          (swatch, index) =>
            index > currentIndex &&
            swatch.offsetLeft === colorSwatches[currentIndex].offsetLeft,
        );
        setFocusedColor(newIndex > -1 ? newIndex : currentIndex);
        event.preventDefault();
        break;
      }
      case 'ArrowUp':
        // Look backwards for next ColorSwatch with same offsetLeft value
        let newIndex = -1;
        for (let i = currentIndex - 1; i >= 0; i--) {
          if (
            colorSwatches[i].offsetLeft ==
            colorSwatches[currentIndex].offsetLeft
          ) {
            newIndex = i;
            break;
          }
        }
        setFocusedColor(newIndex > -1 ? newIndex : currentIndex);
        event.preventDefault();
        break;
```

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+      color && getWindow()?.CSS?.supports?.(`--swatch-color: ${color}`)
+        ? { '--swatch-color': color, ...style }
+        : { backgroundColor: color, ...style };
+
+    return (
+      <span
+        className={cx('iui-color-swatch', { 'iui-active': isActive })}
+        style={_style}
+        onClick={onClick}
+        tabIndex={isActive ? 0 : -1}
```
mayank99, line 44: aria-selected will be useful here
```suggestion
        tabIndex={isActive ? 0 : -1}
        aria-selected={isActive}
```

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+   */
+  isActive?: boolean;
+  /**
+   * User friendly color name.
+   */
+  name?: string;
+} & CommonProps &
+  React.ComponentPropsWithRef<'span'>;
+
+export const ColorSwatch = React.forwardRef<HTMLSpanElement, ColorSwatchProps>(
```
mayank99, line 28: Since this component is exported, you might want to add a JSDoc comment here

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -0,0 +1,52 @@
...
+   */
+  color: string;
+  /**
+   * Is color selected.
+   */
+  isActive?: boolean;
+  /**
+   * User friendly color name.
+   */
+  name?: string;
```
mayank99, line 24: this prop can be removed, unless you plan on experimenting with tooltips

src/core/ColorPicker/ColorPicker.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const colorSwatches = container.querySelectorAll('.iui-color-swatch');
+
+  fireEvent.click(colorSwatches[0]);
+  fireEvent.click(colorSwatches[1]);
+  fireEvent.click(colorSwatches[2]);
+  fireEvent.click(colorSwatches[3]);
+
+  expect(onColorClick).toHaveBeenCalledTimes(4);
+});
```
mayank99, line 61: Add a test for keyboard navigation

mayank99 (commented):
The whole keyboard navigation code can be improved so here are some suggestions to make it robust. The big change I'm suggesting is the addition of useEffect hook.

src/core/ColorPicker/ColorSwatch.tsx
mayank99: Also you need some tests for this component too

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
veekeys: I like these colors, Pallete looks nice and because its color picker its not that those need to be coming from standard colors. I know @FlyersPh9 worked on those a little, so not sure what is bad about it?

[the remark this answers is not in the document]
bentleyvk: It would be nice to release with predefined and approved colors so users wouldn't have to add some random colors. We currently have `SoftColors` in UX documentation site I see there is System colors that even has an example with color picker.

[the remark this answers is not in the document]
mayank99: > I know @FlyersPh9 worked on those a little, so not sure what is bad about it?

The work was for prototyping purposes and was never meant to be the final list we offer to users. The three biggest problems with the list are: names (random and uppercase), values (hex only), and count (why only 30 colors? why not 20 or 40?). Then there is also the problem of how the list is currently exported: it's hard to pick and choose from that list without knowing the index of each color.

> We currently have SoftColors

Those are not exported to the user.

> I see there is System colors in UX documentation site that even has an example with color picker.

There are not enough colors in that list. I think the current list is an extended version of that, with more shades from the bwc palette.

---

If we release with the current list, it would be hard to change it in the future. That's why we should make sure that it's proper if we want to export it.

src/core/ColorPicker/ColorPicker.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+      <ColorSwatch key={2} color='#23450b' />
+    </ColorPicker>,
+  );
+
+  expect(
+    container.querySelectorAll(`.iui-color-swatch.iui-active`).length,
+  ).toBe(1);
+
+  const colorSwatches = container.querySelectorAll('.iui-color-swatch');
+  expect(colorSwatches[0].classList).toContain('iui-active');
```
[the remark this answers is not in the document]
saaaaaally: Added

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
saaaaaally: I can clean up the names to not be uppercase and the hex values to hsl. 
I can also change DefaultColors to be like SoftBackgrounds so that you can call the color by name (DefaultColors.White, DefaultColors.Grey) instead of by index only. 

I might need help coming up with some better color names, if that's something we want to change as well.

src/core/ColorPicker/ColorPicker.test.tsx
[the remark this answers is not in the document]
mayank99: you don't really need this one because you're testing for `iui-active` on the swatch itself
image

That's why I removed it in the suggested change above

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
mayank99: If the names were neutral and a bit more useful, I could see this format working.
```ts
export const SystemColors = {
  white: {
    name: 'white',
    hex: '#FFFFFF',
    r: 255, g: 255, b: 255, rgb: 'rgb(255 255 255)',
    h: 0, s: 0.0, l: 100.0, hsl: 'hsl(0deg 0% 100%)',
  },
  // ...
} as const;
```

Would still need to curate the list with @FlyersPh9's help when he returns.

src/core/ColorPicker/ColorPicker.tsx
saaaaaally: Removed

src/core/ColorPicker/ColorSwatch.tsx
saaaaaally: Removed

src/core/ColorPicker/ColorPicker.test.tsx
saaaaaally: Added tests

src/core/ColorPicker/ColorSwatch.tsx
saaaaaally: Added comment and tests

Issue #326, opened 2021-09-21
Select: handle max menu width


veekeys:
We already set min width on menu, need to also add max width as it can be only twice bigger than the select button according to our standards.

![image](https://user-images.githubusercontent.com/81580355/134158983-73a98160-2d15-4b13-ae2c-2e83dc384b09.png)

Test also on small width.

PR #330, opened 2021-09-22
feat(Select): Truncate text with ellipsis in the middle

### LINKED ISSUE #235: Utility to truncate the middle of a string
Could be hook, or wrapper component, or...
But nice to have thing.
image

> _This style of truncating is especially useful for filenames or directory paths. Sometimes it is important to see the end of the text string. A good example of when this form of truncation is needed is when comparing a bunch of files with the same filename that are appended by a number. Where you truncate the string at depends upon your use case. We recommend starting with displaying the last 3 characters of the string and expanding from there._

veekeys:
Added component to truncate the text leaving the end visible. Chose to go with native css versus js calculations, cause latter did not work well with our resize observer hook. The only downside to the css is extra space after ellipsis (and few DOM elements).
Decided not to build it in into Select not to get complains that something now is different. So for now, having a story and exporting component.
Also added maxWidth to the options as per our standard, but it does not look that well with really small screens...

UPDATE:
Changed to js approach. Code is still interesting, but looks better visually, I guess...

LAST UPDATE:
Using `useOverflow` hook and getting pretty good results. Added support for strings in there and should be fine.

Closes #235 #326 

## Checklist


src/core/utils/TextTruncation.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+import TextTruncation from './TextTruncation';
+
+it('should truncate the text', () => {
+  const { container } = render(
+    <TextTruncation text='This is some text to truncate' />,
+  );
+
+  const containerDiv = container.querySelector('div') as HTMLDivElement;
+  expect(containerDiv.style.display).toBe('flex');
```
bentleyvk, line 15: Are all those style checks really needed? Looks like a noise.

src/core/utils/TextTruncation.tsx
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Text to truncate.
+   */
+  text: string;
+  /**
+   * Number of characters to leave at the end.
+   * @default 6
+   */
+  endCharsCount?: number;
+};
```
bentleyvk, line 17: If we expose it public use, maybe extent `CommonProps`.

src/core/utils/TextTruncation.tsx
```diff
@@ -0,0 +1,44 @@
...
+        style={{
+          textOverflow: 'ellipsis',
+          whiteSpace: 'nowrap',
+          overflow: 'hidden',
+        }}
+      >
+        {text.substr(0, text.length - endCharsCount)}
+      </span>
+      <span style={{ flexShrink: 0 }}>
+        {text.substr(text.length - endCharsCount)}
```
bentleyvk, line 38: What happens with an empty string?

src/core/Select/Select.tsx
```diff
@@ -288,7 +288,11 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
         placement='bottom-start'
         className={cx('iui-scroll', menuClassName)}
         style={{
-          minWidth,
+          minWidth: selectButtonWidth,
+          maxWidth: `${Math.min(
```
bentleyvk, line 292: I think no need use string interpolation here.

mayank99:
>  Chose to go with native css versus js calculations, cause latter did not work well with our resize observer hook. The only downside to the css is extra space after ellipsis (and few DOM elements).

The extra space looks bad, and the code is also kinda messy. What was the problem with ResizeObserver? Maybe we can fix that.

bentleyvk:
> > Chose to go with native css versus js calculations, cause latter did not work well with our resize observer hook. The only downside to the css is extra space after ellipsis (and few DOM elements).
> 
> The extra space looks bad, and the code is also kinda messy. What was the problem with ResizeObserver? Maybe we can fix that.

It throws error that too many setState is called because it removes one letter and tries to render, it doesn't fit, then removes another and on and on and then error... In button group you don't have so many items, so it was not a problem but here it can be at least 256 chars long string.

mayank99:
> It throws error that too many setState is called because it removes one letter and tries to render, it doesn't fit, then removes another and on and on and then error... In button group you don't have so many items, so it was not a problem but here it can be at least 256 chars long string.

There is no `setState` in `useResizeObserver`. Maybe you meant `useOverflow` (which is not meant for individual characters), in which case maybe we could break the string down into smaller chunks, or just use different custom logic.

stories/core/Select.stories.tsx
```diff
@@ -302,3 +302,58 @@ Custom.args = {
...
+          <TextTruncation text={option.label} />
+        )}
+        {...rest}
+      />
+    </div>
+  );
+};
+
+MiddleTextTruncation.args = {
+  placeholder: 'Placeholder text',
```
saaaaaally, line 346: Why do these values need to be set twice?

stories/core/Select.stories.tsx
```diff
@@ -302,3 +302,58 @@ Custom.args = {
...
+        label:
+          'MyFileWithAReallyLongNameThatWillBeTruncatedBecauseItIsReallyThatLongSoHardToBelieve_FinalVersion_V2.html',
+      },
+      { value: 'ShortNameFile.jpg', label: 'ShortNameFile.jpg' },
+      { value: 'SomeOtherFile.dgn', label: 'SomeOtherFile.dgn' },
+    ],
+    placeholder = 'Placeholder text',
+    ...rest
+  } = args;
+  const [selectedValue, setSelectedValue] = useState<string | undefined>(
```
saaaaaally, line 321: The selected value is not being updated when selecting a different option
text truncation

stories/core/Select.stories.tsx
veekeys: There is this issue with storybook.. It is going to override with undefined values as we take rest from the args..
Good catch, I moved rest above our inner set props and removed all the args

src/core/utils/TextTruncation.test.tsx
veekeys: Yeah, changed. Will need to write some proper tests now

src/core/utils/TextTruncation.tsx
veekeys: added

stories/core/Select.stories.tsx
veekeys: Yeah, was issue with truncation component. Should be fixed

src/core/utils/components/TextTruncation.tsx
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+
+    const { paddingLeft, paddingRight } = getComputedStyle(parentElement);
+    const actualWidth =
+      parentElement.clientWidth -
+      parseFloat(paddingLeft) -
+      parseFloat(paddingRight);
+    const charWidth = Math.ceil(parentElement.scrollWidth / text.length);
+    const totalVisibleCount = Math.floor(actualWidth / charWidth);
```
bentleyvk, line 52: It takes average char width that might cause some random errors when there are wide chars and smaller ones that might cause double ellipsis - in the middle and in the end.

src/core/utils/components/TextTruncation.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { render } from '@testing-library/react';
+import React from 'react';
+import TextTruncation from './TextTruncation';
+
+it('should truncate the text', () => {
+  const { container } = render(
+    <div style={{ width: 20 }}>
```
bentleyvk, line 11: Don't you need to mock `scrollWidth` and `clientWidth`?

src/core/Select/Select.tsx
```diff
@@ -285,8 +285,9 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
         placement='bottom-start'
         className={cx('iui-scroll', menuClassName)}
         style={{
-          minWidth,
-          maxHeight: `300px`,
+          minWidth: selectButtonWidth,
+          maxWidth: Math.min(selectButtonWidth * 2, document.body.offsetWidth),
```
bentleyvk, line 289: Use `ownerDocument`

src/core/utils/components/TextTruncation.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * @default 6
+   */
+  endCharsCount?: number;
+} & CommonProps;
+
+/**
+ * Truncates text with the ellipsis in the middle,
+ * leaving defined number of chars at the end.
+ */
+export const TextTruncation = (props: TextTruncationProps) => {
```
bentleyvk, line 26: Maybe rename something with `middle` to avoid misunderstandings with regular truncation.

src/core/utils/components/TextTruncation.tsx
```diff
@@ -0,0 +1,71 @@
...
+   * Number of characters to leave at the end.
+   * @default 6
+   */
+  endCharsCount?: number;
+} & CommonProps;
+
+/**
+ * Truncates text with the ellipsis in the middle,
+ * leaving defined number of chars at the end.
+ */
```
bentleyvk, line 25: Hm what is better component or hook?

mayank99:
Lets try using the refactored `useOverflow` hook here.

Also, lets remove the "max menu width" part (#326) as it was already handled in a different PR.

veekeys:
> Lets try using the refactored `useOverflow` hook here.
> 
> Also, lets remove the "max menu width" part (#326) as it was already handled in a different PR.

Added to use `useOverflow` hook now.

bentleyvk (approved):

src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
+  const { text, endCharsCount = 6, style, ...rest } = props;
+
+  const [ref, visibleCount] = useOverflow(text);
+
+  const truncatedText = React.useMemo(() => {
+    if (visibleCount < text.length) {
+      return `${text.substr(
+        0,
+        visibleCount - endCharsCount - ELLIPSIS_CHAR.length,
```
mayank99, line 38: `ELLIPSIS_CHAR.length` will always be 1 so what's the use

src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -0,0 +1,61 @@
...
+   * @default 6
+   */
+  endCharsCount?: number;
+} & CommonProps;
+
+/**
+ * Truncates text with the ellipsis in the middle,
+ * leaving defined number of chars at the end.
+ * @example
+ * <MiddleTextTruncation text='Here can be a very long text' />
```
mayank99, line 27: Maybe show a more useful example here, like a file name with numbers/extension at the end

src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <span
+      style={{
+        display: 'flex',
+        minWidth: 0,
+        whiteSpace: 'nowrap',
+        ...style,
+      }}
+      ref={ref}
```
mayank99, line 53: In the future we might want to forward the ref but also might want to have an `as` prop. But we can release without it

src/core/utils/components/TextTruncation.tsx
mayank99: Hook would be more useful as an API, but the text needs to actually be rendered to truncate, so component might be more intuitive

src/core/utils/components/MiddleTextTruncation.tsx
veekeys: whats the use of what?

src/core/utils/components/MiddleTextTruncation.tsx
veekeys: might

src/core/utils/components/MiddleTextTruncation.tsx
veekeys: Updated :D

src/core/utils/components/MiddleTextTruncation.tsx
veekeys: I am not sure at the end about api and approach, but I guess fir the version first should be ok :?

src/core/utils/components/MiddleTextTruncation.tsx
mayank99: what's the use of `ELLIPSIS_CHAR.length`? It will always be 1 because it is just one character 😆

PR #337, opened 2021-09-27
feat(Table): Added a flag `selectRowOnClick`.

### LINKED ISSUE #335: Selection of rows not optional on row click event and cannot be disabled
On the select table , in case user clicks on any row the entire selection is reset as in clicking on the row makes it to select that row and deselect others. 
When we have an editable control inside a cell and try to edit the value, the selected row is deselected or if its not selected then it deselects all rows and selects only that row.

Tried adding event.stopPropagation and event.preventDefault to the onRowClick function, but that did not resolve the issue.

#### Expected behavior
For a select table , selection of rows should be disabled or should be optional on row click event.

#### Reproduction
Here is an example with an editable cell:https://codesandbox.io/s/itwinui-react-example-forked-umlk1?file=/src/App.tsx
Here is an example using regular input control inside a cell:https://codesandbox.io/s/itwinui-react-example-forked-umlk1?file=/src/App.tsx
##### Steps to reproduce
1. Select a few items from the table
1. Try to edit the value of 'Weight' in any of the non selected rows, selection of previously selected rows is lost 
1. Try to edit the value of 'Weight' in any selected row , the row is deselected

bentleyvk:
- Added a flag `selectRowOnClick`. Better name suggestions are welcome.
 - Fixed a bug in `EditableCell` when table is selectable and click goes through.

Closes #335  

## Checklist


src/core/Table/Table.tsx
```diff
@@ -145,6 +145,8 @@ export type TableProps<
    * @default 'default'
    */
   density?: 'default' | 'condensed' | 'extra-condensed';
+  /** Disables row selection when clicked anywhere on the row. */
+  disableOnRowClickSelection?: boolean;
```
mayank99, line 149: Maybe call it `selectRowOnClick` (default to `true`).

bentleyvk: Sounds better. I will change to that.

bentleyvk: Changed.

mayank99: Update PR title too

bentleyvk: Done.

mayank99 (approved):

PR #338, opened 2021-09-27
feat: Add new `ComboBox` component

### LINKED ISSUE #146: Typeahead/searchable select
Bentley's BWC library currently features a Typeahead (searchable select) component. However, this does not currently exist in iTwinUI. In order to make moving over to IUI from BWC smoother, it would be nice to have this component.

mayank99:
Added new `ComboBox` component to provide basic "searchable select" functionality. Closes #146 

Uses `iui-focused` class (from https://github.com/iTwin/iTwinUI/pull/307)  and [`aria-activedescendant`](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html#kbd_label_listbox) to emulate focus on menu items through arrow keys. But it requires id, so I generate a random one if not provided. Also had to add `setFocus` prop to `Menu` to prevent focus being moved automatically.

Arrow icon has ripple effect using CSS from https://github.com/iTwin/iTwinUI/pull/313

Input props (e.g. size), selectOption props (e.g. disabled, sublabel) and popover props (e.g. animation, appendTo) are supported.

typeahead

## Checklist


src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+  );
+
+  // Update input value when option is selected
+  React.useEffect(() => {
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    selectedOption && setInputValue(selectedOption.label);
+  }, [selectedValue, options]);
+
+  // Filter options when input value changes
+  React.useEffect(() => {
```
bentleyvk, line 109: Do we want to filter list items? I see you are converting strings to lower case but users might want to use exact match.

mayank99: I was actually thinking of allowing a filter function as a prop. Because the default logic will never satisfy all use cases

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+        (option) => option === options[focusedIndex],
+      );
+      switch (event.key) {
+        case 'ArrowDown':
+          const nextIndex = Math.min(
+            focusedIndexInFilteredList + 1,
+            filteredOptions.length - 1,
+          );
+          setFocusedIndex(options.indexOf(filteredOptions[nextIndex]));
+          !isOpen && setIsOpen(true); // reopen menu if closed when typing
```
bentleyvk, line 165: Why arrow up it is not needed?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+      setInputValue(event.target.value);
+      inputProps?.onChange?.(event);
+    },
+    [inputProps],
+  );
+
+  // Update input value when option is selected
+  React.useEffect(() => {
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    selectedOption && setInputValue(selectedOption.label);
```
bentleyvk, line 105: Could you use `if` instead? 😀 For me it takes a moment to realize that it is a condition.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+  }, [selectedValue, options]);
+
+  // Filter options when input value changes
+  React.useEffect(() => {
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    if (
+      inputValue &&
+      inputValue.length > 0 &&
+      isOpen &&
+      selectedOption?.label !== inputValue // don't filter list if option is selected
```
bentleyvk, line 115: I don't understand this line.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+    if (!filteredOptions.find(({ value }) => value === selectedValue)) {
+      setFocusedIndex(options.indexOf(filteredOptions[0]));
+    } else {
+      setFocusedIndex(
+        options.findIndex(({ value }) => value === selectedValue),
+      );
+    }
+  }, [filteredOptions, options, selectedValue]);
+
+  // Update input value and focused value when menu is closed
```
bentleyvk, line 140: I also don't understand what this whole useEffect does.

src/core/ComboBox/ComboBox.tsx
mayank99: if an option is selected, we want to show the whole list instead of just that one option. If we only show the one value, then it would be very hard to select a different value

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+    selectedOption && setInputValue(selectedOption.label);
+  }, [selectedValue, options]);
+
+  // Filter options when input value changes
+  React.useEffect(() => {
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    if (
+      inputValue &&
+      inputValue.length > 0 &&
+      isOpen &&
```
bentleyvk, line 114: Do you really need to check whether it is opened or not?

src/core/ComboBox/ComboBox.tsx
bentleyvk: Makes sense. Maybe change a comment then a bit.

src/core/ComboBox/ComboBox.tsx
mayank99: If a value is selected, it closes the menu so I update the `input` value to be the same.

And then I set focused value too, but I might change that

src/core/ComboBox/ComboBox.tsx
mayank99: Yes, we don't want to filter list when it's closed

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+      selectedOption?.label !== inputValue // don't filter list if option is selected
+    ) {
+      const _filteredOptions = options.filter((option) =>
+        option.label.toLowerCase().includes(inputValue?.trim().toLowerCase()),
+      );
+      setFilteredOptions(_filteredOptions);
+      if (!_filteredOptions.find(({ value }) => value === selectedValue)) {
+        setSelectedValue(undefined);
+      }
+    } else {
```
bentleyvk, line 124: You could move this to the beggingin of the function and check whether there is inputValue, if not set to full list and return.

src/core/ComboBox/ComboBox.tsx
bentleyvk: Can you just do this on menu item click?

src/core/Menu/Menu.tsx
```diff
@@ -23,14 +23,26 @@ export type MenuProps = {
...
 /**
  * Basic menu component. Can be used for select or dropdown components.
  */
 export const Menu = React.forwardRef<HTMLUListElement, MenuProps>(
   (props, ref) => {
-    const { children, role = 'menu', className, style, ...rest } = props;
+    const {
+      children,
+      role = 'menu',
+      bringFocusInside = true,
```
bentleyvk, line 41: In other inputs we have `setFocus`. Maybe use the same name here.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+      }
+      {...rest}
+      id={id}
+    >
+      <Popover
+        content={
+          <Menu
+            id={`${id}-list`}
+            className='iui-scroll'
+            style={{ minWidth, maxHeight: '300px' }}
```
bentleyvk, line 249: You should do the same as in #330 :
image

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+  // Maintain internal selected value state synced with `value` prop
+  const [selectedValue, setSelectedValue] = React.useState<T | undefined>();
+  React.useEffect(() => {
+    setSelectedValue(value);
+  }, [value]);
+
+  // Fire onChange callback every time selected value changes
+  React.useEffect(() => {
+    onChange?.(selectedValue ?? null);
```
bentleyvk, line 89: Can it be moved to `MenuItem` `onClick` handler?

src/core/ComboBox/ComboBox.tsx
mayank99: My initial thought was: you go "down" from the input into the first/selected item, but you can't go "up" from the input into the menu.

src/core/ComboBox/ComboBox.tsx
mayank99: Inverted the if condition and updated comment

src/core/ComboBox/ComboBox.tsx
mayank99: there is click and there is enter and there is also external `value` prop

src/core/ComboBox/ComboBox.tsx
```diff
@@ -102,27 +102,29 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   // Filter options when input value changes
   React.useEffect(() => {
     const selectedOption = options.find(({ value }) => value === selectedValue);
     if (
-      inputValue &&
-      inputValue.length > 0 &&
-      isOpen &&
-      selectedOption?.label !== inputValue // don't filter list if option is selected
+      !inputValue ||
+      inputValue.length === 0 ||
```
bentleyvk, line 115: I think this check is unnecessary.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -102,27 +102,29 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      !isOpen ||
+      selectedOption?.label === inputValue // if a value is selected, show whole list to allow selecting a different value
     ) {
+      setFilteredOptions(options);
+    } else {
```
bentleyvk, line 120: Just return, no need for `else`.

src/core/ComboBox/ComboBox.tsx
mayank99: I'd need to call `onClick` when Enter is pressed, but can be done

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+ *     { label: 'Item 3', value: 3 },
+ *   ]}
+ *   onChange={() => {}}
+ * />
+ */
+export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
+  const { options, value, onChange, className, inputProps, ...rest } = props;
+
+  // Generate a stateful random id if not specified
+  const [id] = React.useState(() => props.id ?? `iui-${nanoid(10)}`);
```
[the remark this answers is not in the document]
mayank99: That doesn't sound very nice

src/core/ComboBox/ComboBox.tsx
mayank99: I am not sure about this. It's using `document` instead of `ownerDocument`. I would have just used the CSS `min` function with `100vw`.

src/core/ComboBox/ComboBox.tsx
mayank99: Done

src/core/ComboBox/ComboBox.tsx
mayank99: Added `filterFunction` prop. Name is debatable

src/core/ComboBox/ComboBox.tsx
mayank99: I was able to move it into the `onHide` callback of Popover

src/core/Menu/Menu.tsx
mayank99: Done but it's weird because it does not really mean "set focus on menu"

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,318 @@
...
+          break;
+      }
+    },
+    [filteredOptions, focusedIndex, isOpen, options],
+  );
+
+  const menuItems = React.useCallback(
+    (close: () => void) => {
+      return filteredOptions.map((option) => {
+        const index = options.findIndex(({ value }) => option.value === value);
```
bentleyvk, line 213: I see everywhere you are using `options` instead of `filteredOptions` and then searching for the index inside `options`. Can you just use `filteredOptions` everywhere instead and `focusedIndex` just be current index?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,318 @@
...
+      );
+    }
+  }, [selectedValue, onChange, options]);
+
+  // Filter options and update focus when input value changes
+  React.useEffect(() => {
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    if (
+      !inputValue ||
+      !isOpen ||
```
bentleyvk, line 132: To optimize even more, you can check if it is open and if not then you don't need to do anything, including searching for `selectedOption`

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,318 @@
...
+            className={cx({ 'iui-focused': focusedIndex === index })}
+            onClick={(value: T) => {
+              setSelectedValue(value);
+              close();
+            }}
+            isSelected={selectedValue === value}
+            ref={(el) => {
+              if (
+                focusedIndex === index ||
+                (focusedIndex === -1 && selectedValue === value)
```
bentleyvk, line 229: How it is possible?

src/core/Menu/Menu.tsx
```diff
@@ -48,8 +60,8 @@ export const Menu = React.forwardRef<HTMLUListElement, MenuProps>(
       const selectedIndex = items.findIndex(
         (el) => el.getAttribute('aria-selected') === 'true',
       );
-      setFocusedIndex(selectedIndex > -1 ? selectedIndex : 0);
-    }, [focusedIndex]);
+      setFocus && setFocusedIndex(selectedIndex > -1 ? selectedIndex : 0);
```
bentleyvk, line 63: Convert to `if` condition, it is not an inline function or anything.

src/core/ComboBox/ComboBox.tsx
mayank99: The big problem with that is that the id is tied to the index. And id of the focused option is used for setting activedescandant.

src/core/Menu/Menu.tsx
mayank99: Done

src/core/ComboBox/ComboBox.tsx
mayank99: Done

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,282 @@
...
+      inputValue.length > 0 &&
+      isOpen &&
+      selectedOption?.label !== inputValue // don't filter list if option is selected
+    ) {
+      const _filteredOptions = options.filter((option) =>
+        option.label.toLowerCase().includes(inputValue?.trim().toLowerCase()),
+      );
+      setFilteredOptions(_filteredOptions);
+      if (!_filteredOptions.find(({ value }) => value === selectedValue)) {
+        setSelectedValue(undefined);
```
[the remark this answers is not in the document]
mayank99: So I added `emptyStateMessage` but now I'm not so sure
image
https://w3c.github.io/aria-practices/#combobox

src/core/ComboBox/ComboBox.tsx
mayank99: Finally found the right source for defined behavior. Says it's optional so maybe I'll leave it out
image
https://w3c.github.io/aria-practices/#combobox

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
bentleyvk: @FlyersPh9 What is your opinion here from UX perspective?

src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -59,12 +64,185 @@ it('should render with selected value', () => {
 
   input.focus();
   container.querySelectorAll('.iui-menu-item').forEach((item, index) => {
-    expect(item).toHaveTextContent(`Item ${index + 1}`);
-    if (index === 1) {
+    expect(item).toHaveTextContent(`Item ${index}`);
+    if (index === 2) {
```
bentleyvk, line 68: You could also check that others won't have the class and attribute.

src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -59,12 +64,185 @@ it('should render with selected value', () => {
...
+  expect(items[2]).not.toHaveClass('iui-focused');
+
+  // 1 -> 0
+  fireEvent.keyDown(input, { key: 'ArrowUp' });
+  expect(input).toHaveAttribute('aria-activedescendant', `${id}-option0`);
+  expect(items[0]).toHaveClass('iui-focused');
+
+  // select 0
+  fireEvent.keyDown(input, { key: 'Enter' });
+  // expect(mockOnChange).toHaveBeenCalledWith(0); // FIXME
```
bentleyvk, line 222: ?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -211,6 +217,9 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
 
   const menuItems = React.useCallback(
     (close: () => void) => {
+      if (filteredOptions.length === 0) {
+        return <MenuItem disabled>{emptyStateMessage}</MenuItem>;
```
bentleyvk, line 221: Is it valid to do this from accessibility perspective?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+      id={id}
+    >
+      <Popover
+        placement='bottom-start'
+        visible={isOpen}
+        onClickOutside={() => setIsOpen(false)}
+        {...dropdownMenuProps}
+        content={
+          <Menu
+            id={`${id}-list`}
```
bentleyvk, line 290: Maybe move `id` to the end, looks weird having it at the beginning.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+            className='iui-scroll'
+            style={{
+              minWidth,
+              maxWidth: `min(${minWidth * 2}px, 90vw)`,
```
bentleyvk, line 294: Why `90vw`?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,337 @@
...
+    if (inputRef.current) {
+      setMinWidth(inputRef.current.offsetWidth);
+    }
+  }, [isOpen]);
+
+  const [filteredOptions, setFilteredOptions] = React.useState(options);
+  const [focusedIndex, setFocusedIndex] = React.useState<number>(-1);
+
+  // Maintain internal selected value state synced with `value` prop
+  const [selectedValue, setSelectedValue] = React.useState<T | undefined>();
```
bentleyvk, line 108: Just started to think, do we really want to save the selected value internally? Shouldn't we leave this to the user?

src/core/ComboBox/ComboBox.test.tsx
mayank99: Is it really worth it? I already check for normal state in the basic test

src/core/ComboBox/ComboBox.test.tsx
mayank99: For some reason, Enter key is closing the menu but not calling `onChange` here

src/core/ComboBox/ComboBox.tsx
mayank99: if the id was "my-combobox", this would be "my-combobox-list" which makes perfect sense to me (and "list-my-combobox" doesn't)

src/core/ComboBox/ComboBox.tsx
mayank99: 100vw goes outside viewport
image

src/core/ComboBox/ComboBox.tsx
mayank99: How would it help? It's much nicer to support both

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
FlyersPh9: I'm thinking that showing the menu with an empty state message might be better than showing nothing at all. If a user typed in 3 characters & got results, then typed a 4th character & got no results _may_ think that the menu is still loading.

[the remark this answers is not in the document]
mayank99: but there is no loading anywhere. The results are instantaneous, so they could just delete the last character and get the results back.

I do see the benefit in "No options" message but it has an accessibility downside as mentioned in https://github.com/iTwin/iTwinUI-react/pull/338#discussion_r718458424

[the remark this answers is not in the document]
bentleyvk: So yeah we need to have a loading state here too. Because it can be used to query some e.g. user list based on the input.

src/core/ComboBox/ComboBox.tsx
bentleyvk: I imagine id to be `1234-asd-789` and for me it looks weird `123-asd-qwe-list` 😀

mayank99: um, id cannot start with a number. So it would be `iui-1234-asd-qwe-list`.

If id is provided by user, it will likely be something more useful too. `country-select-list`

bentleyvk: It really doesn't change the fact if it starts with `123` or `asd`.

> If id is provided by user, it will likely be something more useful too.

That's why I said to make it mandatory, I don't really see a big problem here.

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: If we add a loading state, we might also need something for when the menu is not open.
image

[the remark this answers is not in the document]
bentleyvk: If we plan to have loading state inside input, maybe we could also add clear icon:
image

src/core/ComboBox/ComboBox.tsx
mayank99: Fixed by initializing selectedValue based on value

src/core/ComboBox/ComboBox.test.tsx
mayank99: Fixed

src/core/ComboBox/ComboBox.tsx
```diff
@@ -145,8 +146,8 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
 
   // Fire onChange callback and update inputValue every time selected value changes
   React.useEffect(() => {
-    if (selectedValue) {
-      onChange?.(selectedValue);
+    onChange?.(selectedValue ?? null);
```
bentleyvk, line 149: Huh? Maybe better change `onChange` param type.

mayank99: Done

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,370 @@
...
+    const selectedOption = options.find(({ value }) => value === selectedValue);
+    if (!inputValue || selectedOption?.label === inputValue) {
+      setFilteredOptions(options);
+      return;
+    }
+
+    const _filteredOptions =
+      filterFunction?.(options, inputValue) ??
+      options.filter((option) =>
+        option.label.toLowerCase().includes(inputValue?.trim().toLowerCase()),
```
saaaaaally, line 178: Not sure but can you make this also find the inputValue in option.value instead of just option.label? 

For example if I search for American Samoa by typing AS shouldn't it show as a result?

image
image

mayank99: That would not work because `option.value` will not always be a string. But for that use case, I expose the `filterFunction` so it is still possible on the user's side.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,370 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import { Input, InputProps, Menu, MenuItem, SelectOption, Text } from '../..';
```
bentleyvk, line 7: I don't remember but importing like this `../..` might cause the issue for the client apps because of circular dependencies.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,370 @@
...
+  } = props;
+
+  // Generate a stateful random id if not specified
+  const [id] = React.useState(
+    () => props.id ?? `${inputProps?.id}-cb` ?? `iui-cb-${getRandomValue(10)}`,
+  );
+
+  useTheme();
+
+  /** Generates an memoized id for an option, given the index from original list */
```
bentleyvk, line 88: a or an memomoized?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,370 @@
...
+   */
+  value?: T;
+  /**
+   * Callback fired when selected value changes.
+   */
+  onChange?: (value: T) => void;
+  /**
+   * Function to customize the default filtering logic.
+   */
+  filterFunction?: (
```
bentleyvk, line 36: Test

src/core/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,370 @@
...
+  /**
+   * Native input element props.
+   */
+  inputProps?: Omit<InputProps, 'setFocus'>;
+  /**
+   * Props to customize dropdown menu behavior.
+   */
+  dropdownMenuProps?: PopoverProps;
+  /**
+   * Message shown when no options are available (after filtering).
```
bentleyvk, line 49: Not only after filtering because if initial options are empty too, the same message will be shown.

src/core/Menu/MenuItem.tsx
```diff
@@ -147,7 +147,7 @@ export const MenuItem = React.forwardRef<HTMLLIElement, MenuItemProps>(
         ref={refs}
         style={style}
         role={role}
-        tabIndex={disabled ? undefined : -1}
+        tabIndex={disabled || role === 'presentation' ? undefined : -1}
```
bentleyvk, line 150: Looks hacky but maybe ok until we complete this https://github.com/iTwin/iTwinUI-react/issues/349

src/core/ComboBox/ComboBox.tsx
mayank99: Fixed

src/core/ComboBox/ComboBox.tsx
mayank99: Added

src/core/ComboBox/ComboBox.tsx
mayank99: Fixed

src/core/ComboBox/ComboBox.tsx
mayank99: So I need to import everything individually?

mayank99: ok changed

bentleyvk (approved):

PR #339, opened 2021-09-27
chore(utils): Split into separate files and add index

mayank99:
- Split up `common.tsx` into separate files with descriptive names.
- Added `getRandomValue` function to generate random alphanumeric value using Math.random.
- Added `index.ts` under utils/, utils/functions, utils/hooks, and utils/components.
- Updated all components to import from `utils`.

## Checklist


src/core/utils/numbers.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Returns a random value using nanoid (which uses crypto API).
+ * Contains IE fallback to return a uuid using Math.random.
+ */
+export const getRandomValue = (length?: number) => {
+  if (getWindow()?.crypto) {
+    return nanoid(length);
+  }
+  /** uuid function from https://gist.github.com/jed/982883 */
+  const getUuid = (a = '') =>
```
bentleyvk, line 24: So maybe nanoid is not really needed anymore?

src/core/utils/components/InputContainer.tsx
```diff
@@ -37,6 +37,7 @@ export const InputContainer = <T extends React.ElementType = 'div'>(
     children,
     className,
     style,
+    ...rest
```
bentleyvk, line 40: Not related to this PR.

src/core/utils/common.spec.tsx
```diff
@@ -10,7 +10,7 @@ import {
   getTabbableElements,
   getUserColor,
   getWindow,
-} from './common';
+} from '.';
```
bentleyvk, line 13: I would say to split tests to separate files too.

src/core/utils/numbers.ts
mayank99: just look at that uuid, it's very long and ugly. nanoid is shorter and faster and very lightweight

bentleyvk: Have you checked the inside of nanoid? 😀 Because we would still see `getRandomValue`.

mayank99: I'm not talking about the insides. The generated uuid is 36 characters long and ugly. It's just too much for our needs

src/core/utils/components/InputContainer.tsx
mayank99: Removed

src/core/utils/common.spec.tsx
mayank99: Done

src/core/utils/numbers.ts
bentleyvk: Do you really need to generate GUID? Why not some value that is 4-6 chars long? We are not working in databases where collision could happen, if we also add some prefix/suffix depending on the element, then it will be always unique.

mayank99: replaced guid and nanoid with a non-secure implementation

bentleyvk (approved):

PR #346, opened 2021-10-01
feat: Add new `ColorPicker` component

### LINKED ISSUE #312: Provide a color-picker control

saaaaaally:
Closes #312

image
image

## Checklist


src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,546 @@
...
+
+  return {
+    hsv: { h: hsv.h, s: hsv.s, v: hsv.v, displayString: c.toHsvString() },
+    hsl: { h: hsl.h, s: hsl.s, l: hsl.l, displayString: c.toHslString() },
+    rgb: { r: rgb.r, g: rgb.g, b: rgb.b, displayString: c.toRgbString() },
+    hex: { hex: c.toHex() },
+  };
+};
+
+export type ColorPickerProps = {
```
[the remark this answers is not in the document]
veekeys: Would need to think about the naming, but I feel we could still have colors as children and get rid of `onAdd`. Instead, we could support ability to have that (passing add button together with `ColorSwatch` components as children). That would be more flexible for what ever people might want to have there. Do you guys think it would make sense? @bsteinbk @saaaaaally @mayank99

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+    return `${hsv.h}, ${hsv.s}%, ${hsv.v}%`;
+  }
+
+  /** Convert this ColorValue to a string in the form "hsv(h,s,v)" - i.e hsv(120,50%,50%). */
+  public toHsvString(): string {
+    return `hsv(${this.getHsvString()})`;
+  }
+
+  /** Convert this ColorValue to a string in the form "hsva(h,s,v,a)" - i.e hsva(120,50%,50%,0.5). */
+  public toHsvaString(): string {
```
veekeys, line 705: Hmm... Maybe we could save alpha value from the input and if it is defined, return it with all schemes. That would lessen our code and require just to have `toHSVString`, etc...

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+      }
+    } else {
+      h = 0;
+    }
+
+    return { h, s, v };
+  }
+
+  /** Create a ColorValue from an HsvColor */
+  public static fromHSV(hsv: HsvColor, transparency = 0): ColorValue {
```
veekeys, line 605: Not sure about having separate methods to get `ColorValue` and having constructor with only tbgr support. Could we have just constructor, which could accept any of the supported formats as input and create `ColorValue` object that way? Would be consistent between different color schemes and possibly easier to use.

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+        hsl.h / 360,
+        hsl.s / 100,
+        hsl.l / 100,
+        transparency,
+      ),
+    );
+  }
+
+  /** Compute the 0xTTBBGGRR color corresponding to the specified hue, saturation, lightness values. */
+  public static computeTbgrFromHSL(
```
veekeys, line 445: Wouldnt it simplify, if we just ask to create `ColorValue` instance from what ever color input and then have method `toTbgr`? Similar as with other color schemes

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+    return 255 - scratchBytes[3];
+  }
+
+  /** True if this ColorValue is fully opaque. */
+  public get isOpaque(): boolean {
+    return ColorValue.isOpaque(this._tbgr);
+  }
+
+  /** True if the specified 0xTTBBGGRR color is fully opaque. */
+  public static isOpaque(tbgr: number): boolean {
```
veekeys, line 353: Not sure this is needed.. Pretty simple and users can easily check that on their end. We provide ability to get alpha, so..

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+
+  /** Create a ColorValue from a string representation. The following representations are supported:
+   * *"rgb(255,0,0)"*
+   * *"rgba(255,0,0,.2)"*
+   * *"rgb(100%,0%,0%)"*
+   * *"hsl(120,50%,50%)"*
+   * *"#rrbbgg"*
+   * *"blanchedAlmond"* (see possible values from [[ColorByName]]). Case-insensitive.
+   */
+  public static fromString(val: string): ColorValue {
```
veekeys, line 163: Yeah, same here. We could just support all of that in constructor.

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+ * ColorValue is immutable. To obtain a modified copy of a ColorValue, use methods like [[adjustedForContrast]], [[inverse]], or [[withTransparency]]. For example:
+ * ```ts
+ *  const semiTransparentBlue = ColorValue.blue.withTransparency(100);
+ * ```
+ * @public
+ */
+export class ColorValue {
+  private readonly _tbgr: number;
+
+  public constructor(tbgr: number) {
```
veekeys, line 80: I think would be great and simplify things a little, if we just accept any format in constructor (rhb, hsv, hsl, stringed of those, tbgr??). Would be also great, if `ColorValue` would use rgba (or hsv?? which is easier @mayank99) as base for all the calculations, as we all are more familiar with those color schemes than tbgr (mb not familiar, but jus used to that order than inversed :D). I think we could have method to get tbgr if needed (even though users could just use rgb for that and inverse).

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
bsteinbk: @saaaaaally - That sounds OK to me. I plan to get back to making these updates to ColorValue later today or early Monday.

src/core/utils/color/ColorValue.ts
mayank99: using `type Color = HslColor | RgbColor | HsvColor | string` from https://github.com/iTwin/iTwinUI-react/pull/346#discussion_r729926621, the constructor would change to:
```ts
public contructor(color: Color) {}
```

not sure which format would be easier for maintaining internal state, but I think rgba would be easier to avoid a full rewrite

src/core/utils/color/ColorValue.ts
bsteinbk: What if showAlpha is requested but it is set to 1.... in this case tinycolor2 would just return the hsv()/ rgb() functions and not the hsva/rgba - and that may confuse the user who request to see the alpha value.... and I think that is the same functionality you are requesting.

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+        return val;
+      case 'string':
+        return this.computeTbgrFromString(val);
+      default:
+        return 0;
+    }
+  }
+
+  /** Convert this ColorValue to a 32 bit number representing the 0xTTBBGGRR value */
+  public toJSON(): ColorDefProps {
```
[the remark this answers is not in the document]
bsteinbk: got rid of this and other "confusing" methods.

src/core/utils/color/ColorValue.ts
bsteinbk: see new ColorValue.create method.  constructor method is now private and compatible with class in imodeljs... its just as easy to use `ColorValue.create()` instead of `new ColorValue ()`

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+    return this.withAlpha(tbgr, 255 - transparency);
+  }
+
+  /** The "known name" for this ColorValue. Will be undefined if color value is not in [[ColorByName]] list */
+  public get name(): string | undefined {
+    return ColorValue.getName(this.tbgr);
+  }
+
+  /** Obtain the name of the color in the [[ColorByName]] list associated with the specified 0xTTBBGGRR value, or undefined if no such named color exists. */
+  public static getName(tbgr: number): string | undefined {
```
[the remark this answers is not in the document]
bsteinbk: gone

src/core/utils/color/ColorValue.ts
```diff
@@ -0,0 +1,709 @@
...
+    return ColorValue.toRgbString(this.tbgr);
+  }
+
+  /** Convert the 0xTTBBGGRR color to a string in the form "rgb(r,g,b)" where each component is specified in decimal. */
+  public static toRgbString(tbgr: number): string {
+    return `rgb(${this.getColorsString(tbgr)})`;
+  }
+
+  /** Convert this ColorValue to a string in the form "rgba(r,g,b,a)" where color values are decimal digits and a is a fraction */
+  public toRgbaString(): string {
```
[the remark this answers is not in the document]
bsteinbk: remove static and instance methods with same name.

src/core/utils/color/ColorValue.ts
bsteinbk: used with displaying values. see ColorValue.create for all supported input types.

bsteinbk (approved):
Looks good to me.

mayank99 (approved):

PR #351, opened 2021-10-04
feat(Table): Added pagination support 

### LINKED ISSUE #159: Table: Pagination & Manual Handling of Paginator State
#### Feature
Add pagination to table.
Also the ability to manually manipulate the state of the paginator. 

#### Examples
As discussed in teams, very large datasets can crash a page without pagination.

There was a manual mode in v6 react table that let you pass in total page count, current page number and whatever segment of your full dataset you needed to have displayed. On large data sets that were stored in state, only a portion of the data would be passed into the table at one time otherwise the page crashes/lags. At least this was how the v6 table worked, I don't know if this is applicable to v7 or not.

bentleyvk:
image

Closes iTwin/iTwinUI-react#159 

## Checklist


src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+   */
+  focusActivationMode?: 'auto' | 'manual';
+  /**
+   * Array of possible page size options. When provided then shows current rows info and page size selection.
+   */
+  rowsPerPage?: number[];
+  /**
+   * Object of functions to localize pagination labels.
+   */
+  localization?: {
```
bentleyvk, line 77: Not sure about the naming of this and functions inside.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+  onPageChange: (page: number) => void;
+  /**
+   * Callback when page size is changed.
+   */
+  onPageSizeChange?: (size: number) => void;
+  /**
+   * Modify the density of the pagination (adjusts the height).
+   * @default 'default'
+   */
+  density?: TableProps['density'];
```
bentleyvk, line 58: I don't know is it good to accepts `density` when only there are two sizes.

src/core/Table/actionHandlers/expandHandler.ts
```diff
@@ -12,13 +12,13 @@ export const onExpandHandler = <T extends Record<string, unknown>>(
...
 ) => {
-  if (!instance?.rows.length) {
+  if (!instance?.page.length) {
     onExpand?.([], newState);
     return;
   }
 
   const expandedData: T[] = [];
-  instance.rows.forEach((row) => {
+  instance.page.forEach((row) => {
```
bentleyvk, line 21: I will take a look at this change later.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+  /**
+   * Control whether focusing tabs (using arrow keys) should automatically select them.
+   * Use 'manual' if tab panel content is not preloaded.
+   * @default 'auto'
+   */
+  focusActivationMode?: 'auto' | 'manual';
+  /**
+   * Array of possible page size options. When provided then shows current rows info and page size selection.
+   */
+  rowsPerPage?: number[];
```
mayank99, line 73: Name can be improved as it's unclear. Maybe `pageSizes` or something.

Also we could add default value too.
image

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+  const ellipsis = (
+    <span className={cx('iui-ellipsis', { 'iui-small': size === 'small' })}>
+      …
+    </span>
+  );
+
+  return (
+    <div className={cx('iui-paginator', className)} style={style} {...rest}>
+      <div className='iui-left' />
+      <div className='iui-center' style={{ minWidth: 0 }} ref={overflowRef}>
```
mayank99, line 245: what's `minWidth: 0` for?

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+        <IconButton
+          styleType='borderless'
+          disabled={currentPage === 0 || hasNoRows}
+          onClick={() => onPageChange(currentPage - 1)}
+          size={size}
+          aria-label='Previous page'
```
mayank99, line 251: Could add this to localization object maybe? Same with `Next page`

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+  );
+
+  return (
+    <div className={cx('iui-paginator', className)} style={style} {...rest}>
+      <div className='iui-left' />
+      <div className='iui-center' style={{ minWidth: 0 }} ref={overflowRef}>
+        <IconButton
+          styleType='borderless'
+          disabled={currentPage === 0 || hasNoRows}
+          onClick={() => onPageChange(currentPage - 1)}
```
mayank99, line 249: No need to memoize this?

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+    pageSize,
+    onPageChange,
+    focusActivationMode = 'auto',
+    isLoading = false,
+    density = 'default',
+    rowsPerPage,
+    onPageSizeChange,
+    localization,
+    className,
+    style,
```
mayank99, line 126: Don't think you need to destruct `style`

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+    ...rest
+  } = props;
+
+  useTheme();
+
+  const localizationFunctions = { ...defaultLocalization, ...localization };
```
mayank99, line 132: I think you might need to add more keys to this object, so the name `localizationFunctions` may not be accurate.

Check this out for examples of keys: https://material.angular.io/components/paginator/api

src/core/Table/TablePaginator.tsx
mayank99: Check this out: https://material.angular.io/components/paginator/api

src/core/Table/Table.tsx
```diff
@@ -373,118 +428,162 @@ export const Table = <
...
-                          />
-                        )}
-                      </div>
-                    )}
-                  </div>
-                );
-              })}
-            </div>
-          );
+    <>
```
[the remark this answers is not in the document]
bentleyvk: In CSS package it is the same 🤷‍♂️

src/core/Table/Table.tsx
```diff
@@ -373,118 +428,162 @@ export const Table = <
     [isRowDisabled, isSelectable, selectRowOnClick, dispatch, onRowClick],
   );
 
+  React.useEffect(() => {
+    setPageSize(pageSize);
+  }, [pageSize, setPageSize]);
```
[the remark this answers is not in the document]
bentleyvk: It is a Table prop.
image

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+        {index + 1}
+      </Button>
+    ),
+    [size, currentPage, focusedIndex, onPageChange],
+  );
+
+  const totalPagesCount = Math.ceil(totalRowsCount / pageSize);
+  const pageList = React.useMemo(
+    () =>
+      [...Array(totalPagesCount).fill(null)].map((_, index) =>
```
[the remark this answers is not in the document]
bentleyvk: `[...new Array]` doesn't work for some reason, need to use `fill`.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+   * Flag whether data is still loading and total rows count is not known.
+   * @default false
+   */
+  isLoading?: boolean;
+  /**
+   * Control whether focusing tabs (using arrow keys) should automatically select them.
+   * Use 'manual' if tab panel content is not preloaded.
+   * @default 'auto'
+   */
+  focusActivationMode?: 'auto' | 'manual';
```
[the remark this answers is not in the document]
bentleyvk: But you already have your data preloaded. At least it is nice to have ability to change as I already implemented it 😁 We can change the default behavior.

src/core/Table/TablePaginator.tsx
bentleyvk: Without it, center element takes more than a third of paginator. Later I will add these changes to the CSS.
image

src/core/Table/TablePaginator.tsx
bentleyvk: `IconButton` is really small component and still when `currentPage` changes everything will be rerendered.

src/core/Table/TablePaginator.tsx
bentleyvk: We are doing it everywhere, I think.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+   * Callback when page is changed.
+   */
+  onPageChange: (page: number) => void;
+  /**
+   * Callback when page size is changed.
+   */
+  onPageSizeChange?: (size: number) => void;
+  /**
+   * Modify the density of the pagination (adjusts the height).
+   * @default 'default'
```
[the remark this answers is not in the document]
bentleyvk: If I rename to `size`, should I still use `Table['density']` as a value? Because it looks odd to me 😀

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+  currentRowsInfoLabel: (
+    startIndex: number,
+    endIndex: number,
+    totalRows: number,
+    isLoading: boolean,
+  ) =>
+    isLoading
+      ? `${startIndex}-${endIndex}…`
+      : `${startIndex}-${endIndex} of ${totalRows}`,
+};
```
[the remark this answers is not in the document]
bentleyvk: What it will give in this case?

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Table paginator component. Recommended to pass to the `Table` as `paginatorRenderer` prop.
+ * @example
+ * <Table
+ *   // ...
+ *   paginatorRenderer={(props) => <TablePaginator {...props} />}
+ * />
+ */
+export const TablePaginator = (props: TablePaginatorProps) => {
```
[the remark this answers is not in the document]
bentleyvk: We discussed that this approach could help us avoid the need for adding `nativePaginatorProps` or anything similar. Because every time we go with options approach, we end up having a lot of native props.

src/core/Table/TablePaginator.tsx
bentleyvk: I was afraid it can be easily confused with `pageSize`, so I took the name from antd.

src/core/Table/TablePaginator.tsx
bentleyvk: Ok, I just went trough our project and found out that we are using `localization` and `translated...` for our props. We really need to decide which one to use.

src/core/Table/TablePaginator.tsx
bentleyvk: Maybe worth adding this too?
image

src/core/Table/actionHandlers/expandHandler.ts
bentleyvk: Changed to `preFilteredFlatRows` because when expanded `rows` list doesn't have expanded/visible sub-rows.

src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -20,7 +20,7 @@ export type ButtonGroupProps = {
    * and returns the `ReactNode` to render.
    */
   overflowButton?: (firstOverflowingIndex: number) => React.ReactNode;
-} & Omit<CommonProps, 'title'>;
+} & React.ComponentPropsWithoutRef<'div'>;
```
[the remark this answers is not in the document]
bentleyvk: Changed.

src/core/Table/Table.tsx
```diff
@@ -317,12 +366,15 @@ export const Table = <
       selectSubRows,
       data,
       getSubRows,
+      initialState: { pageSize, ...props.initialState },
+      pageCount: !paginatorRenderer ? -1 : undefined,
```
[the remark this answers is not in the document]
bentleyvk: Removed, it might have left from testing stuff (I really hope so).

src/core/Table/Table.tsx
```diff
@@ -44,6 +45,35 @@ import { onSingleSelectHandler } from './actionHandlers/selectHandler';
 
 const singleRowSelectedAction = 'singleRowSelected';
 
+export type TablePaginatorRendererProps = {
```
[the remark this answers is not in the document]
bentleyvk: It is a type for `paginatorRenderer` so it make sense to have it here. Also, used good old inheritance to avoid duplication.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,335 @@
...
+import SvgChevronLeft from '@itwin/itwinui-icons-react/cjs/icons/ChevronLeft';
+import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
+import { ButtonGroup } from '../ButtonGroup';
+import { Button } from '../Buttons/Button';
+import { useOverflow } from '../utils/hooks/useOverflow';
+import { ProgressRadial } from '../ProgressIndicators';
+import { TableProps } from './Table';
+import { DropdownButton } from '../Buttons';
+import { MenuItem } from '../Menu';
+import { CommonProps, getBoundedValue, useTheme } from '../utils';
```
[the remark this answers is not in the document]
bentleyvk: Changed,

src/core/Table/TablePaginator.tsx
bentleyvk: Added.

src/core/Table/TablePaginator.tsx
bentleyvk: Renamed.

src/core/Table/Table.tsx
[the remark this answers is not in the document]
mayank99: What about `setPageSize`?

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
mayank99: just a narrow type with better intellisense

src/core/Table/Table.tsx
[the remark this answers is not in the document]
bentleyvk: It comes from `instance`:
image

And `pageSize` comes from 'state`:
image

src/core/Table/TablePaginator.tsx
mayank99: What do you mean "only two sizes"?

There are 3 paginator heights here:
https://itwin.github.io/iTwinUI/backstop/minified/table.html#demo-extras

bentleyvk: It comes from table:
image

In paginator I only apply `small` to buttons when it is `condensed` or `extra-condensed`.

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
bentleyvk: Added.

src/core/Table/Table.tsx
```diff
@@ -149,6 +183,19 @@ export type TableProps<
...
    */
   selectRowOnClick?: boolean;
+  /**
+   * Function that returns pagination component. Recommended to use `TablePaginator`.
+   * @example
+   * (props: TablePaginatorRendererProps) => (
+   *   <TablePaginator {...props} />
+   * )
+   */
+  paginatorRenderer?: (props: TablePaginatorRendererProps) => React.ReactNode;
```
mayank99, line 193: so if you're sticking with this approach, I would recommend adding more details in the comment of this prop and the comments of the individual props.

src/core/Table/Table.tsx
```diff
@@ -149,6 +183,19 @@ export type TableProps<
...
+  /**
+   * Number of rows per page.
+   * @default 10
```
mayank99, line 196: Should be 25, not 10

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,324 @@
...
+        <IconButton
+          styleType='borderless'
+          disabled={currentPage === 0 || hasNoRows}
+          onClick={() => onPageChange(currentPage - 1)}
+          size={size}
+          aria-label={localization.previousPage}
+        >
+          <SvgChevronLeft />
+        </IconButton>
+        <ButtonGroup onKeyDown={onKeyDown} ref={pageListRef}>
```
mayank99, line 244: I think the logic inside this button group could be cleaned up so it's more readable/maintainable, because right now I'm having a tough time even reviewing

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  const halfVisibleCount = Math.floor(visibleCount / 2);
+  let startPage = focusedIndex - halfVisibleCount;
+  let endPage = focusedIndex + halfVisibleCount + 1;
+  if (startPage < 0) {
+    endPage = Math.min(totalPagesCount, endPage + Math.abs(startPage));
+    startPage = 0;
+  }
+  if (endPage > totalPagesCount) {
+    startPage = Math.max(0, startPage - (endPage - totalPagesCount));
```
mayank99, line 219: What is this weird math? If values go out of bounds, shouldn't it just be `startPath = 0` and `endPage = totalPagesCount`

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const localization = React.useMemo(
+    () => ({ ...defaultLocalization, ...userLocalization }),
+    [userLocalization],
+  );
+
+  const pageListRef = React.useRef<HTMLDivElement | null>(null);
+
+  const [focusedIndex, setFocusedIndex] = React.useState<number>(currentPage);
+  React.useEffect(() => {
+    setFocusedIndex(currentPage);
```
mayank99, line 126: Is this only needed for keeping track internally? Because it doesn't actually focus the new `currentPage` automatically when it changes.

stories/core/Table.stories.tsx
```diff
@@ -1445,3 +1446,92 @@ Editable.parameters = {
...
+      Array(195)
+        .fill(null)
+        .map((_, index) => generateItem(index)),
+    [generateItem],
+  );
+
+  const rowsPerPage = useMemo(() => [10, 25, 50], []);
+  const paginator = useCallback(
+    (props: TablePaginatorRendererProps) => (
+      <TablePaginator {...props} rowsPerPage={rowsPerPage} />
```
mayank99, line 1509: So `rowsPerPage` is required to be passed or is this just an example?

src/core/Table/Table.tsx
```diff
@@ -373,118 +431,162 @@ export const Table = <
...
+      totalRowsCount: rows.length,
+      density,
+      isLoading,
+      onPageChange: (page) => {
+        gotoPage(page);
+        tableRef.current?.scrollIntoView({
+          block: 'start',
+          inline: 'nearest',
+          behavior: 'smooth',
+        });
```
mayank99, line 451: Table should already be in view, no?

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
mayank99: I'm ok with keeping it and changing the default to `manual`

src/core/Table/TablePaginator.tsx
bentleyvk: Ugh, I thought I need to have some comment here but I don't know how to put that in readable English.
It helps to maintain all possible visible pages. If current page is e.g. 2 and visible count is 8, instead of showing pages from 1 to 6 (2 + (8 / 2)), I am showing from 1 to 8. I am taking the overflow part (if you can call it like that) and bringing it to the other side.

stories/core/Table.stories.tsx
bentleyvk: Just an example.

src/core/Table/Table.tsx
bentleyvk: It is but if you check the story yourself, it will scroll to the top of the table. Not sure if it is ok or not but personally I don't like when I switch the page and have to scroll to the top myself 😀

src/core/Table/TablePaginator.tsx
bentleyvk: If it would, then page would scroll down to the paginator. As I am not saving the current page inside, I thought that when navigating with keyboard in auto mode, it shouldn't focus next page if the user did not changed the prop, otherwise it will be inconsistent state between focused index and current index.

bentleyvk: Ok, it is not working well with `auto`...

src/core/Table/TablePaginator.tsx
bentleyvk: Tried some refactoring. Please tell if it is still not clear and maybe some comments in the code are still needed.

bentleyvk: Two slices because that's how it worked in my head 😀
I am using IIFE here to have if/else because for me several wrapped ternaries are the most hard thing to read.
I can't simplify like that because it won't work when seeing last pages:
image

src/core/Table/Table.tsx
mayank99: It doesn't work when using keyboard. Or is that intentional?
paginator

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
bentleyvk: Done.

src/core/Table/Table.tsx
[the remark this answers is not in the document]
bentleyvk: Created issue: https://github.com/iTwin/iTwinUI/issues/330

src/core/Table/TablePaginator.tsx
mayank99: Maybe something like this?
```suggestion
  if (startPage < 0) {
    endPage = Math.min(totalPagesCount, endPage + Math.abs(startPage)); // if no room at the beginning, show extra pages at the end
    startPage = 0;
  }
  if (endPage > totalPagesCount) {
    startPage = Math.max(0, startPage - (endPage - totalPagesCount)); // show extra pages at the beginning
```

bentleyvk: Added.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,343 @@
...
+        }
+        break;
+      }
+      default:
+        break;
+    }
+  };
+
+  const halfVisibleCount = Math.floor(visibleCount / 2);
+  let startPage = focusedIndex - halfVisibleCount;
```
veekeys, line 223: useMemo wouldnt be useful here for these calculations to be run?

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,343 @@
...
+  const pageListRef = React.useRef<HTMLDivElement | null>(null);
+
+  const [focusedIndex, setFocusedIndex] = React.useState<number>(currentPage);
+  React.useEffect(() => {
+    setFocusedIndex(currentPage);
+  }, [currentPage]);
+
+  const isMounted = React.useRef(false);
+  React.useEffect(() => {
+    if (isMounted.current) {
```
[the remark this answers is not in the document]
mayank99: I was actually thinking this should be a separate utility hook called `useEffectAfterMount` but thought that might be too much

src/core/Table/actionHandlers/expandHandler.ts
```diff
@@ -12,13 +12,13 @@ export const onExpandHandler = <T extends Record<string, unknown>>(
...
 ) => {
-  if (!instance?.rows.length) {
+  if (!instance?.preFilteredFlatRows.length) {
     onExpand?.([], newState);
     return;
   }
 
   const expandedData: T[] = [];
-  instance.rows.forEach((row) => {
+  instance.preFilteredFlatRows.forEach((row) => {
```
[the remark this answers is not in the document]
bentleyvk: `rows` previously had only visible rows which included expanded subrows but with pagination `rows` only has top rows without any subrow so I had to use `preFilteredFlatRows` which is a list of all rows with all subrows.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,343 @@
...
+    startPage = 0;
+  }
+  if (endPage > totalPagesCount) {
+    startPage = Math.max(0, startPage - (endPage - totalPagesCount)); // If no room at the end, show extra pages at the beginning
+    endPage = totalPagesCount;
+  }
+
+  const hasNoRows = totalPagesCount === 0;
+
+  const ellipsis = (
```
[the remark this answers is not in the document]
bentleyvk: But then `iui-ellipsis` must be in some common CSS file. Maybe later (never 😁)...

src/core/Table/TablePaginator.tsx
bentleyvk: Renamed to `pageSizeList`

src/core/Table/actionHandlers/expandHandler.ts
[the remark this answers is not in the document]
veekeys: oh wow.. if there is a reference in react-table about it, maybe link would be useful for future use?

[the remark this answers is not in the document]
bentleyvk: I don't really know if there is a proper link as I just found this with the help of IDE and going through react-table source files.

src/core/Table/TablePaginator.tsx
```diff
@@ -0,0 +1,343 @@
...
+  const pageList = React.useMemo(
+    () =>
+      new Array(totalPagesCount)
+        .fill(null)
+        .map((_, index) => pageButton(index)),
+    [pageButton, totalPagesCount],
+  );
+  const [overflowRef, visibleCount] = useOverflow(pageList);
+
+  const onKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
```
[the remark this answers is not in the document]
bentleyvk: Fixed.

veekeys (approved):

PR #364, opened 2021-10-08
feat: Add new `InformationPanel` component

### LINKED ISSUE #318: Add InfoPanel
#### Feature
Add InfoPanel component that exists in the bwc-react library as we can't find any alternative to it inside the itwinui-react.
**bwc-react/innersource/InfoPanel**

image

mayank99:
Added new `InformationPanel` component. Closes #318

Supports props for header content (pass `InformationPanelHeader`), body content (pass `children`), resizing, orientation and more. Requires to be wrapped in `InformationPanelWrapper` for positioning correctly.

infopanel

## Checklist


src/core/InformationPanel/InformationPanel.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const resizer = infoPanel.querySelector('.iui-resizer') as HTMLElement;
+
+  act(() => {
+    fireEvent.pointerDown(resizer, { button: 0 });
+    fireEvent.pointerMove(document, { x: 500 });
+    fireEvent.pointerUp(document);
+  });
+
+  // FIXME
+  // expect(infoPanel).toHaveAttribute('style', 'width: 500px');
```
mayank99, line 147: Currently trying to fix this test.

mayank99: Fixed using `clientX` instead of `x`.

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+   * Custom header action buttons shown alongside the close button.
+   * @example
+   * [
+   *   <IconButton styleType='borderless' onClick={() => {}}><SvgEdit /></IconButton>,
+   *   <IconButton styleType='borderless' onClick={() => {}}><SvgDelete /></IconButton>,
+   * ]
+   */
+  headerActions?: React.ReactNode[];
+  /**
+   * Set to false to make the panel non-resizable.
```
veekeys, line 41: But default is also falso, so maybe you meant `set to true to make the panel resizable` ?

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+  } = props;
+
+  useTheme();
+
+  const infoPanelRef = React.useRef<HTMLDivElement>(null);
+  const refs = useMergedRefs(ref, infoPanelRef);
+
+  const [isOpen, setIsOpen] = React.useState(isOpenProp ?? false);
+  React.useEffect(() => {
+    setIsOpen(isOpenProp ?? false);
```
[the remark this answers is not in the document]
mayank99: It's not really a value of the prop though.

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+          }}
+        >
+          <div className='iui-resizer-bar' />
+        </div>
+      )}
+      <div className='iui-information-header'>
+        <span className='iui-information-header-label'>{label}</span>
+        <div className='iui-information-header-actions'>
+          {headerActions}
+          <IconButton
```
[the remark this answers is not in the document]
mayank99: it's mandatory according to UX Standards and Jon

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,174 @@
...
+            if (e.button != undefined && e.button !== 0) {
+              return;
+            }
+            e.preventDefault();
+            e.stopPropagation();
+            infoPanelRef.current.ownerDocument.addEventListener(
+              'pointermove',
+              resizeCb,
+            );
+            infoPanelRef.current.ownerDocument.addEventListener(
```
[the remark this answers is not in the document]
mayank99: image

src/core/InformationPanel/InformationPanel.tsx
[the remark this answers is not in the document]
mayank99: removed internal state to make prop more consistent

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+  onClose?: () => void;
+  /**
+   * Custom header action buttons shown alongside the close button.
+   * @example
+   * [
+   *   <IconButton styleType='borderless' onClick={() => {}}><SvgEdit /></IconButton>,
+   *   <IconButton styleType='borderless' onClick={() => {}}><SvgDelete /></IconButton>,
+   * ]
+   */
+  headerActions?: React.ReactNode[];
```
bentleyvk, line 39: We might have the same problem like in `Header` when users wants custom stuff that we haven't thought of. What if we have a prop for header and have a sub-component `InfomartaionPanelHeader`?

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+          onPointerDown={(e) => {
+            if (!infoPanelRef.current) {
+              return;
+            }
+            if (e.button != undefined && e.button !== 0) {
+              return;
+            }
+            e.preventDefault();
+            e.stopPropagation();
+            infoPanelRef.current.ownerDocument.addEventListener(
```
bentleyvk, line 131: Can't you use our hook for this?

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const onResize = React.useCallback(
+    (e: PointerEvent) => {
+      e.preventDefault();
+      if (!infoPanelRef.current) {
+        return;
+      }
+      const { right, bottom } = infoPanelRef.current.getBoundingClientRect();
+      if (orientation === 'vertical') {
+        infoPanelRef.current.style.width = `${right - e.clientX}px`;
```
bentleyvk, line 97: Is it ok to set styles to node and not through `styles` prop? Is it possible that after rerender it might reset its size?

stories/core/InformationPanel.stories.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const data = React.useMemo(
+    () =>
+      [...Array(10).fill(null)].map((_, index) => ({ name: `Row${index}` })),
+    [],
+  );
+
+  const lorem100 = `Lorem ipsum dolor sit, amet consectetur adipisicing elit. Accusamus veniam dicta error doloremque libero sit est. Voluptatum nam modi, ex illum veritatis nobis omnis porro quod harum optio minus magnam tenetur quia dolor quis natus, eius, suscipit hic? Nobis deleniti obcaecati, sequi mollitia vero magnam error quidem, voluptatem asperiores repudiandae, molestias sit et voluptatibus magni. Sequi delectus, sunt eaque corrupti architecto modi suscipit? Quos in itaque dolore voluptas saepe natus repellat ad qui dolores. Incidunt temporibus ut, unde maxime nam explicabo saepe aspernatur molestiae iste libero neque, alias corporis laboriosam fugiat ad. Dicta neque quos fuga odit quae sequi dolore!`;
+
+  return (
+    <InformationPanelWrapper>
```
bentleyvk, line 85: Would it shift the Table if there would be no wrapper? If yes, can we have a story with that?

stories/core/InformationPanel.stories.tsx
```diff
@@ -0,0 +1,256 @@
...
+        <Text>{lorem100}</Text>
+      </InformationPanel>
+    </InformationPanelWrapper>
+  );
+};
+Basic.decorators = [
+  (Story) => (
+    <>
+      <Text isMuted style={{ marginBottom: 11 }}>
+        <em>Click on Details to open InformationalPanel</em>
```
bentleyvk, line 105: These are barely visible.

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+        },
+        className,
+      )}
+      ref={refs}
+      {...rest}
+    >
+      {resizable && (
+        <div
+          className='iui-resizer'
+          onPointerDown={(e) => {
```
bentleyvk, line 122: Maybe move to separate function.

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+        infoPanelRef.current.style.width = `${right - e.clientX}px`;
+      } else {
+        infoPanelRef.current.style.height = `${bottom - e.clientY}px`;
+      }
+    },
+    [orientation],
+  );
+
+  return (
+    <div
```
bentleyvk, line 106: Shouldn't we use `react-transition-group` here to prevent rendering stuff even when it is not visible? Users might have some heavy content here e.g. map.

mayank99: I'm starting to dislike that library after trying to use it for SideNav submenu 😆 Plus it would require CSS changes.

I am not sure of the performance difference between re-rendering hidden content vs unmounting/remounting. And I think browsers do have the ability to lazy load stuff like iframes (maps) with just one property. Not to mention, the user can add their own performance optimizations.

src/core/InformationPanel/InformationPanel.tsx
mayank99: Don't know. But changed to style 🤷‍♀️

src/core/InformationPanel/InformationPanel.tsx
mayank99: Moved

src/core/InformationPanel/InformationPanel.tsx
mayank99: I tried causing rerender (after updating internal state) and it didnt reset

src/core/InformationPanel/InformationPanel.tsx
bentleyvk: I know, you would have it outside. But now started to think whether you need `addEventListener`, can't you have like a flag `isResizing` and have react props `onPointerDown` and `onPointerMove`?

stories/core/InformationPanel.stories.tsx
bentleyvk: So is it needed if we don't want anyone to see it?

src/core/InformationPanel/InformationPanel.tsx
mayank99: `onPointerDown` and `onPointerMove` would need to be applied on some element. In this case, I'm adding the listeners to document, so it wouldn't work.

Fwiw, I added `{ once: true }` so there is no concern of the listener not getting removed.

src/core/InformationPanel/InformationPanel.tsx
bentleyvk: I don't think we should have mixed sub-component namings right now.

mayank99: so you would also want me to create `SidenavSubmenuHeader`? weird naming

src/core/InformationPanel/InformationPanelHeader.tsx
```diff
@@ -0,0 +1,69 @@
...
+ *     <IconButton styleType='borderless' onClick={() => {}}><SvgEdit /></IconButton>,
+ *     <IconButton styleType='borderless' onClick={() => {}}><SvgDelete /></IconButton>,
+ *   ]}
+ * >
+ *   <Text variant='subheading'>InfoPanel heading</Text>
+ * </InformationPanelHeader>
+ */
+export const InformationPanelHeader = (props: InformationPanelHeaderProps) => {
+  const { children, onClose, actions, className, ...rest } = props;
+
```
bentleyvk, line 49: `useTheme` and CSS import is missing, if they are needed...

bentleyvk (approved):

PR #381, opened 2021-10-15
fix(ComboBox): Reset filteredOptions if options are updated

### LINKED ISSUE #377: Getting error (React.cloneElement(...): The argument must be a React element, but you passed undefined.) in Combobox when options changes
I'm getting this error (React.cloneElement(...): The argument must be a React element, but you passed undefined.) in my combobox when the options changes.  The use case is I have a LabeledSelect and a ComboBox. Depending on the selection from the LabeledSelect, the ComboBox  options changes.

#### Expected behavior
ComboBox options should change.

#### Reproduction
Please see: https://codesandbox.io/s/combobox-h4f9u

##### Steps to reproduce
1. Select a skill from the dropdown.

#### Additional information

### LINKED ISSUE #382: Combobox onChange events running unnecessarily
I have the following code with comboboxes. When an asset is selected from the first combobox, the second combobox should appear and allow the user to select a treatment. If the asset changes, the treatment should be reset.
image
When selecting an asset for the first time, the code runs as expected. When selecting a treatment afterwards, the onChange for the first combobox is also run (in the console, hit2, hit1, hit2 is printed) even though there is no code to change the state of the first box if anything in the second one changes. If the asset is changed, it results in an infinite loop and the first and second onChanges keep firing and triggering each other. It seems that either adding a value property fires off an onChange, combobox onChanges are linked, or both.

#### Expected behavior
onChange events should only be fired if the user selects a new value and should only activate for the relevant combobox.

#### Reproduction
See above image

mayank99:
Added useEffect to reset state of filtered options when options change. Also added early return in `menuItems` if option is not found. Closes #377

Also fixed a bug where `undefined` is ignored if set in `value` prop. Closes #382 

## Checklist


src/core/ComboBox/ComboBox.tsx
```diff
@@ -136,6 +136,10 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
   }, [isOpen]);
 
   const [filteredOptions, setFilteredOptions] = React.useState(options);
+  React.useEffect(() => {
```
veekeys, line 139: Comment mentions added `useLayoutEffect`, but here its not that one.

mayank99: Yeah I changed it as it does not have an *effect* 😆

veekeys (approved):

PR #387, opened 2021-10-18
Bsteinbk/colorpicker updates

bsteinbk:
## From earlier PR #384
- Update ColorValue.create method to use standard types in standard rgba notation
- left internal use of tbgr for stability and to make it easy to merge in any fixes made to ColorDef class in imodel.js
- made many methods private since they are used only to process tbgr value
- all toXXXString methods now allow a boolean argument to include the alpha value.
- added more unit test
- Removed unpopular ColorByNames file

## From latest branch
- Removed fillColor 
- Removed Color
- Updated Swatch and ColorPicker to use ColorValue
- Update stories


src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -85,7 +61,7 @@ export type ColorPickerProps = {
    * This can be on pointerUp when thumb is done dragging,
    * or when user clicks on color builder components, or when user clicks on color swatch
    */
-  onChangeCompleted?: (color: ColorValue) => void;
+  onChangeCompleted: (color: ColorValue) => void;
```
veekeys, line 64: Any specific reason its mandatory?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -58,20 +38,16 @@ export type ColorBuilderProps = {
...
 export type ColorPaletteProps = {
-  /**
-   * Available color values to show in palette
-   * */
-  colors?: Color[];
-  /**
-   * Title shown above color palette
-   */
+  /** Available color values to show in palette */
+  colors?: ColorValue[];
```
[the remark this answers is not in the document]
bsteinbk: Disagree... one type of Color in / one type of Color out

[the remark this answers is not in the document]
veekeys: So its same type in, same out. In is the `SelectedColor`, out is `onChangeCompleted`.
Color list is a part of `ColorPicker` and that should not be tied to our input/output type.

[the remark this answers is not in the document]
bsteinbk: If you are talking about the color Palette colors, then in our use case, those colors are stored in a database as an array of tbgr values, I will convert this array of values to ColorValues so the conversion is done once. I don't want to have to convert to some other format that you like pass it in to the list just to have to convert them back to ColorValues internally.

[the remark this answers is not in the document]
mayank99: > In every current imodeljs use case we would want to pass the current color is via ColorValue

> I need to support an array of ColorValues.

I don't understand. Converting ColorValue to Color type is trivial, and makes the API more user friendly.

The user should not ever have to type `ColorValue` if they are not already using it imo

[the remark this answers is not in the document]
bsteinbk: To me it makes sense to use ColorValue since the onChange will give you a ColorValue back. It is the single source of truth for the color.  The old Color type have multiple different specifications in it.... it was NOT a single source of truth.  A user could specify {rgb: {r:255 g:255: b:255, colorString: 'rgb(0,0,0)'} hsl: {h:180, s:50, l:50, colorString: '#ocffdd'} where is the single source of truth there and which of the bogus values should be used?

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -578,6 +579,7 @@ export const ColorPicker = (props: ColorPickerProps) => {
             {inputType === 'HEX' && (
               <div className='iui-color-input-fields'>
                 <Input
+                  key='hex'
```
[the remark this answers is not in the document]
bsteinbk: No.  There was a bug in Saally code and I added those keys trying to debug the issue.

src/core/utils/color/ColorValue.ts
```diff
@@ -291,55 +452,28 @@ export class ColorValue {
...
-   * @returns The 0xTTBBGGRR value equivalent to `tbgr` but with the specified alpha.
-   */
-  public static withAlpha(tbgr: number, alpha: number): number {
-    scratchUInt32[0] = tbgr;
-    scratchBytes[3] = 255 - (alpha | 0);
-    return scratchUInt32[0];
-  }
-
   /** Get the alpha value for this ColorValue. Will be between 0-255 */
   public getAlpha(): number {
```
veekeys, line 471: Is Opaque still needed, when we have alpha?

src/core/utils/color/ColorValue.ts
```diff
@@ -147,36 +232,114 @@ export class ColorValue {
...
-   * *"#rrbbgg"*
-   * *"blanchedAlmond"* (see possible values from [[ColorByName]]). Case-insensitive.
+   * *"#rrggbb"*
    */
   public static fromString(val: string): ColorValue {
     return this.fromTbgr(this.computeTbgrFromString(val));
   }
 
+  /** Create a ColorValue from hue, saturation, lightness values.  */
+  public static fromHSL(hsl: HslColor): ColorValue {
```
veekeys, line 247: Now thinking if we really need all these `from...`. You can just use `ColorValue.create`

bsteinbk: What if I just make them private to keep the create method from getting too big?

src/core/utils/color/ColorValue.ts
```diff
@@ -525,10 +487,15 @@ export class ColorValue {
...
   }
 
-  /** Create an HslColor from this ColorValue */
+  /** Get the alpha value for this ColorDef. Will be between 0-255 */
   public toHsl(): HslColor {
+    return ColorValue.toHsl(this._tbgr);
+  }
+
+  /** Create an HslColor from this ColorValue */
+  private static toHsl(tbgr: number): HslColor {
```
mayank99: Two things:
1. Probably don't want to repeat the same function name. Maybe the static one should be called `toHsvFromTbgr` or something.

2. The comments for both are wrong.
   - toHsl has nothing to do with alpha value
   - the static method says "this ColorValue", but there is no "this"
   - Also, there is no `ColorDef` anywhere

src/core/utils/color/ColorValue.ts
```diff
@@ -573,9 +540,14 @@ export class ColorValue {
...
 
-  /** Create an [[HsvColor]] from this ColorValue */
+  /** Get the alpha value for this ColorDef. Will be between 0-255 */
   public toHsv(): HsvColor {
-    const { r, g, b } = ColorValue.getColors(this._tbgr);
+    return ColorValue.toHsv(this._tbgr);
+  }
+
+  /** Create an [[HsvColor]] from this ColorValue */
+  private static toHsv(tbgr: number): HsvColor {
```
mayank99: Same comment here as `toHsl`

src/core/ColorPicker/ColorPicker.tsx
```diff
@@ -12,14 +12,29 @@ import {
...
+  }
+  if (undefined === color) {
+    return ColorValue.fromTbgr(0);
+  }
+
+  try {
+    return ColorValue.create(color);
+  } catch (_e) {
+    return ColorValue.fromTbgr(0);
+  }
```
mayank99, line 35: Is all this validation really needed? ColorValue.create should already return a default color if it fails to parse the input

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -22,15 +22,16 @@ export type ColorSwatchProps = {
 /**
  * ColorSwatch component to display a color within a basic color picker.
  * @example
- * <ColorSwatch color='#23450b'  onClick={onClick}/>
+ * <ColorSwatch color={ColorValue.create('#23450b'}  onClick={onClick}/>
  */
 export const ColorSwatch = React.forwardRef<HTMLDivElement, ColorSwatchProps>(
   (props, ref) => {
     const { color, style, onClick, isActive, className, ...rest } = props;
+    const inColor = getColorValue(color);
```
mayank99, line 30: Using this function here might not be correct behavior actually. This function (and `ColorValue.create`) don't support all valid css color formats. Examples include keywords like `transparent` and `currentColor`, as well as the new [space-separated functional notation from CSS Colors Level 4 spec](https://www.w3.org/TR/css-color-4/#:~:text=new%20syntax%20consisting%20of%20space-separated%20arguments%20and%20an%20optional%20slash-separated%20opacity).

So it might be better to just assign the value to the style if it's not an instance of `ColorValue`.

src/core/utils/color/ColorValue.ts
```diff
@@ -47,82 +55,114 @@ export type HsvColor = {
...
+  private constructor(tbgr: number) {
     scratchUInt32[0] = tbgr; // Force to be a 32-bit unsigned integer
     this._tbgr = scratchUInt32[0];
   }
 
   /**
    * Create a new ColorValue.
    * @param val value to use.
-   * If a number, it is interpreted as a 0xTTBBGGRR (Red in the low byte, high byte is transparency 0==fully opaque) value.
+   * If a number, it is interpreted as a 0xRRGGBB or 0xRRGGBBAA value.
```
bentleyvk, line 123: I guess not needed anymore.

src/core/utils/color/ColorValue.ts
```diff
@@ -147,38 +190,120 @@ export class ColorValue {
...
+   * Create a ColorValue from a string representation. The following representations are supported:
    * *"rgb(255,0,0)"*
    * *"rgba(255,0,0,.2)"*
    * *"rgb(100%,0%,0%)"*
    * *"hsl(120,50%,50%)"*
-   * *"#rrbbgg"*
-   * *"blanchedAlmond"* (see possible values from [[ColorByName]]). Case-insensitive.
+   * *"#rrggbb"*
    */
   public static fromString(val: string): ColorValue {
```
bentleyvk, line 201: Is it still needed when we have `create` that also accepts string?

src/core/ColorPicker/ColorPicker.tsx
bsteinbk: Yes, because ColorValue.fromString will now throw instead of returning a default color. This was needed in the hex input to know if the user put in a string that can't be converted.

src/core/utils/color/ColorValue.ts
bsteinbk: I will make private, it is used be create method.

src/core/utils/color/ColorValue.ts
bsteinbk: removed

src/core/utils/color/ColorValue.ts
bsteinbk: removed

src/core/ColorPicker/ColorPicker.tsx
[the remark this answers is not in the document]
bsteinbk: removed

src/core/ColorPicker/ColorPicker.tsx
bsteinbk: I could remove the validation here if a make ColorValue.fromString public again so it could be called when the user is typing in a value and it cannot be converted the value would not automatically jump to a default color value.... they may have just have mistyped one character.  I could then add `| undefined` to `ColorType` so .create would just return a default black color, then this function could be removed.

src/core/utils/color/ColorValue.ts
```diff
@@ -4,24 +4,31 @@
...
 
 // portions adapted from Three.js Copyright © 2010-2017 three.js authors
 
 const scratchBytes = new Uint8Array(4);
 const scratchUInt32 = new Uint32Array(scratchBytes.buffer);
 
+/**
+ * An unsigned 32-bit integer in 0xTTBBGGRR format
+ */
+export type TbgrColorValue = number;
```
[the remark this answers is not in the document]
bsteinbk: its gone

PR #389, opened 2021-10-19
chore(Table): Proper full height handling

bentleyvk:
Applying changes from https://github.com/iTwin/iTwinUI/pull/335

### DIFF
--- a/src/core/Table/TableRowMemoized.tsx
+++ b/src/core/Table/TableRowMemoized.tsx
@@ -55,7 +55,7 @@ const TableRow = <T extends Record<string, unknown>>(props: {
 
   const userRowProps = rowProps?.(row);
   const mergedProps = {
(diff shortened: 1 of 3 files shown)


src/core/Table/TableRowMemoized.tsx
```diff
@@ -55,7 +55,7 @@ const TableRow = <T extends Record<string, unknown>>(props: {
 
   const userRowProps = rowProps?.(row);
   const mergedProps = {
-    ...row.getRowProps(),
+    ...row.getRowProps({ style: { flex: `0 0 auto` } }),
```
bentleyvk, line 58: Body now has `flex-direction: column` and `react-table` sets row style to `flex: 1 0 auto` which makes row grow in height when table has `height: 100%`:
image

mayank99 (approved):

PR #409, opened 2021-11-01
feat(Button): Allow rendering as anchor element

### LINKED ISSUE #332: Ability to have Button as anchor
In certain Link cases, there might be a need to have our Button displayed as anchor (accessbility) when it is navigatin to some page. Need to think if it makes sense to provide `as` prop and render button as anchor..

mayank99:
Added new utility type that adds `as` prop to a component and merges its native props with the component props, all while preserving strong typing.

image

Closes #332

## Checklist


src/core/utils/props.ts
```diff
@@ -26,3 +28,15 @@ export type CommonProps = {
    */
   id?: string;
 } & StylingProps;
+
+export type PolymorphicComponentProps<
```
bentleyvk, line 32: Comment?

src/core/Typography/Text/Text.tsx
```diff
@@ -53,34 +56,38 @@ export type TextProps<
...
-) => {
-  const {
-    variant = 'body',
-    as: Element = 'div',
-    className,
-    isMuted = false,
-    isSkeleton = false,
-    ...rest
-  } = props;
+export const Text: TextComponent = React.forwardRef(
```
bentleyvk, line 59: Why cast to `TextComponent`? Will it still allow to pass ref?

src/core/Buttons/Button/Button.tsx
```diff
@@ -5,10 +5,10 @@
 import cx from 'classnames';
 import React from 'react';
 
-import { useTheme } from '../../utils';
+import { PolymorphicComponentPropsWithRef, useTheme } from '../../utils';
```
bentleyvk, line 8: Add a story with button as an anchor.

src/core/Typography/Text/Text.tsx
mayank99: Read PR description 😄 and yes, TextProps includes ref

src/core/Buttons/Button/Button.tsx
mayank99: Read PR description. Would be weird to publish a story with a visual bug

Also rendering button as anchor is a bad practice so I'm not even sure we want to make it obvious that it's possible

veekeys: Link to bad practice?

mayank99: There are two problems:
- screen readers will announce button as a link.
- clicking spacebar on a link scrolls the page, so it will confuse even normal users.

https://marcysutton.com/links-vs-buttons-in-modern-web-applications

Of course there can be some valid use cases like a download link or a "get started" link, but those are less common.

veekeys: What about buttons in header, which navigate you to some page? You want them to be anchors, to support middle clicka nd show url.. but you need them to look as buttons.

src/core/utils/props.ts
mayank99: Added

src/core/Buttons/Button/Button.tsx
mayank99: not sure we want to use our normal buttons for those. But like I said, there are some valid use cases. It's just that if we advertise this feature, it has a great potential for misuse.

src/core/utils/props.ts
```diff
@@ -48,3 +48,10 @@ export type PolymorphicComponentPropsWithRef<
 > = PolymorphicComponentProps<T, Props> & {
   ref?: React.ComponentPropsWithRef<T>['ref'];
 };
+
+export type AsProp<T> = {
+  /**
+   * What element should this component be rendered as?
+   */
+  as?: T;
```
bentleyvk, line 56: What was the issue that you need this type?

mayank99: we are extending `ButtonProps` in a few other components, where it would not work without adding logic

src/core/Buttons/Button/Button.tsx
bentleyvk: But if we don't show it, then it will be missed in dozens of native props.

veekeys: I would say lets add a story so users are aware that if there is a need, we support that. .otherwise, it is not that obvious to even try to find it. 
And we can complete this PR.

mayank99: Added story

mayank99:
This PR is now ready for review. I've changed the approach to use different types for the component and props. The prop types inside the component are now inferred automatically.

src/core/utils/props.ts
```diff
@@ -26,3 +28,40 @@ export type CommonProps = {
...
+  T extends React.ElementType,
+  Props = Record<string, unknown>
+> = Merge<React.ComponentPropsWithoutRef<T>, Props>;
+
+/**
+ * Merges original OwnProps and the inferred props from `as` element.
+ *
+ * T should be the default element that is used for the `as`  prop.
+ */
+export interface PolymorphicForwardRefComponent<
```
bentleyvk, line 47: Any chance you can explain all this magic?

src/core/Header/HeaderButton.tsx
```diff
@@ -14,40 +14,43 @@ import {
...
-} & Omit<
-  Partial<DropdownButtonProps> & Partial<ButtonProps>,
-  'children' | 'styleType' | 'name'
->;
+} & Partial<Pick<DropdownButtonProps, 'menuItems'>> &
+  Pick<ButtonProps, 'startIcon' | 'endIcon'>;
 
 const isSplitButton = (
-  props: Omit<Partial<SplitButtonProps>, 'name'>,
+  props: Partial<HeaderButtonProps> & { [k: string]: unknown },
```
bentleyvk, line 38: Why you needed to change this?

src/core/Buttons/SplitButton/SplitButton.tsx
```diff
@@ -11,14 +11,10 @@ import { Placement } from 'tippy.js';
 import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
 import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
 
-import { useTheme } from '../../utils';
+import { PolymorphicForwardRefComponent, useTheme } from '../../utils';
 import '@itwin/itwinui-css/css/button.css';
 
-export type SplitButtonProps = {
-  /**
-   * Callback fired on clicking the primary button.
```
bentleyvk, line 19: I think we have `onClick` here for the documentation.

src/core/utils/props.ts
```diff
@@ -26,3 +28,40 @@ export type CommonProps = {
    */
   id?: string;
 } & StylingProps;
+
+/**
+ * Merges provided Props with the props of T.
+ *
+ * T can be any native HTML element or a custom component.
+ */
+export type PolymorphicComponentProps<
```
bentleyvk, line 37: Don't we need `Omit` here? Saw in a video:
image
https://www.youtube.com/watch?v=uZ8GZm5KEXY

src/core/SideNavigation/SidenavButton.tsx
```diff
@@ -23,43 +23,47 @@ export type SidenavButtonProps = {
...
->((props, ref) => {
-  const {
-    className,
-    children,
-    isActive = false,
-    disabled = false,
-    isSubmenuOpen = false,
-    ...rest
-  } = props;
+export const SidenavButton: SideNavButtonComponent = React.forwardRef(
```
bentleyvk, line 35: Tests for this one?

src/core/Typography/Text/Text.tsx
```diff
@@ -53,9 +53,7 @@ export type TextProps<
  * <Text isMuted>Some muted text.</Text>
  * <Text isSkeleton>Skeleton text</Text>
  */
-export const Text = <T extends React.ElementType | React.ComponentType = 'div'>(
-  props: TextProps<T>,
-) => {
+export const Text: TextComponent = React.forwardRef((props, ref) => {
```
bentleyvk, line 56: We can't have `as` without `forwardRef`?

mayank99: I think we could, but it's better to forward the ref wherever possible. Our tooltip, for example, only works on elements that forward their ref

src/core/utils/props.ts
mayank99: The merge utility takes care of that 
image

src/core/SideNavigation/SidenavButton.tsx
mayank99: Hmm, maybe I need to create a `SidenavButton.test.tsx`. 

But it just delegates the `as` prop to `Button` so maybe it's fine

src/core/Buttons/SplitButton/SplitButton.tsx
mayank99: I removed it because the type will not be correct with polymorphism

src/core/Header/HeaderButton.tsx
mayank99: Dont remember. Reverted.

src/core/utils/props.ts
mayank99: First, we extend `React.ForwardRefExoticComponent` so that ref gets correct type. We also add the `as` prop here.

After that we progressively narrow down the type using chained ternaries, checking whether it's extending intrinsic element vs component and merging props accordingly.

src/core/Buttons/SplitButton/SplitButton.tsx
```diff
@@ -47,66 +51,74 @@ export type SplitButtonProps = {
...
+        </div>
+        <div>
+          <DropdownMenu
+            placement={menuPlacement}
+            menuItems={menuItems}
+            style={{ minWidth: menuWidth }}
+            onShow={React.useCallback(() => setIsMenuOpen(true), [])}
+            onHide={React.useCallback(() => setIsMenuOpen(false), [])}
+          >
+            <IconButton
```
gretanausedaite, line 110: Split button `IconButton` is not rendered as anchor.

gretanausedaite: But yeah, that doesn't make sense if it would. Irrelevant comment, just observation, sorry.

src/core/Buttons/Button/Button.test.tsx
```diff
@@ -159,3 +159,15 @@ it('should render without label correctly', () => {
...
+it('should render as anchor element', () => {
+  const { container } = render(
+    <Button as='a' href='https://example.com/'>
+      label
+    </Button>,
+  );
+
+  const button = container.querySelector('a.iui-button') as HTMLButtonElement;
+  expect(button).toHaveTextContent('label');
+  expect(button).toHaveAttribute('href', 'https://example.com/');
```
gretanausedaite, line 172: Why different approaches? In some tests you test if `button.href equals 'link text'` and in others you have this `.toHaveAttribute`

mayank99: Ah it's because I am using the wrong type in line 170 so `button.href` is not valid. Fixed here

I don't know if there's any difference in the two approaches, but I remember noticing `toHaveAttribute` is more reliable in some cases

src/core/utils/props.ts
bentleyvk: Maybe add this explanation as a comment. Also thinking maybe it is worth to add an example how to use these types but this one is up to you.

mayank99: Updated jsdoc

bentleyvk (approved):

PR #420, opened 2021-11-05
fix(Tabs): Add styling props to tabs wrapper

mayank99:
Without these props, users need to target our internal class names for layouts.

### DIFF
--- a/src/core/Tabs/Tabs.tsx
+++ b/src/core/Tabs/Tabs.tsx
@@ -2,17 +2,17 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import cx from "classnames";
-import React from "react";
+import cx from 'classnames';
+import React from 'react';
 import {
   useTheme,
   useResizeObserver,
   useMergedRefs,
   getBoundedValue,
   CommonProps,
-} from "../utils";
-import "@itwin/itwinui-css/css/tabs.css";
-import { Tab } from "./Tab";
+} from '../utils';
+import '@itwin/itwinui-css/css/tabs.css';
+import { Tab } from './Tab';
 
 export type TabsProps = {
   /**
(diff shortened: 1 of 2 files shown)


src/core/Tabs/Tabs.tsx
```diff
@@ -244,7 +248,11 @@ export const Tabs = (props: TabsProps) => {
   };
 
   return (
-    <div className={cx('iui-tabs-wrapper', `iui-${orientation}`)}>
+    <div
+      className={cx('iui-tabs-wrapper', `iui-${orientation}`, className)}
+      style={style}
+      id={id}
```
bentleyvk, line 254: Why not just `{...rest}`?

bentleyvk: Hm, then all `data-*` and `aria-*` attributes are passed not on the parent element.

mayank99: Yeah... this wrapper should have been a separate component

bentleyvk: Should it? Isn't it here because of selection animation?

mayank99: which one makes more sense?

```tsx
<Tabs {...billionProps} />
```
or

```tsx
<TabsWrapper>
  <Tabs />
  <TabPanel />
</TabsWrapper>
```

bentleyvk: When you write like this 😁 But it still would be `<Tabs {...sameProps} />`

mayank99: We would be able to get rid of these props.
image

And we would gain extra customization on all of those parts. Right now you can't do anything with the panel content other than apply class name.

mayank99:
Closing this to reduce complexity. As a workaround, a flex wrapper can be used to control layouts.  In some cases (e.g. overflow) it might be necessary to apply `min-width: 0` but that can be solved using `> *`.

```tsx
<div className='custom-tabs-container'>
  <HorizontalTabs /> // or <VerticalTabs />
</div>
```

```scss
.custom-tabs-container {
  display: flex;
  // ...layout styles, e.g. height: 200px

  > * {
    min-width: 0;
  }
}
```

In v2 we should rethink the API and not forget to consider layouts.

PR #425, opened 2021-11-09
fix(ColorPicker): Upgrade color conversion to remove rounding

bsteinbk:
Avoid most of the round off issues when going from rgb -> hsl -> rgb

Closes # 

## Checklist


mayank99:
@bsteinbk Could you update your branch? There are merge conflicts

bsteinbk:
> @bsteinbk Could you update your branch? There are merge conflicts

I don't know how .. I tried the following yesterday before the PR and again just now and here is what I see.
```
E:\github\bsteinbk\iTwinUI-react>git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

E:\github\bsteinbk\iTwinUI-react>git pull
Already up to date.

E:\github\bsteinbk\iTwinUI-react>git checkout bsteinbk/betterRgbHslProcessing
Switched to branch 'bsteinbk/betterRgbHslProcessing'
Your branch is up to date with 'origin/bsteinbk/betterRgbHslProcessing'.

E:\github\bsteinbk\iTwinUI-react>git pull
Already up to date.

E:\github\bsteinbk\iTwinUI-react>git merge main
Already up to date.
```

mayank99:
> > @bsteinbk Could you update your branch? There are merge conflicts
> 
> I don't know how .. I tried the following yesterday before the PR and again just now and here is what I see.

That's because your main branch is also out-of-date and you need to fetch upstream. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork

bsteinbk:
But I did Fetch upstream in GitHub Desktop
image

bsteinbk:
looks like `git merge upstream/main` not just `git merge main` is what I needed. It is a pain in the but working with forks!

src/core/ColorPicker/ColorInputPanel.tsx
```diff
@@ -250,6 +284,38 @@ export const ColorInputPanel = React.forwardRef(
...
+                ? 'negative'
+                : undefined
+            }
+          >
+            <Input
+              size='small'
+              type='number'
+              min='0'
+              max='1'
+              step='.01'
```
saaaaaally, line 300: Need to add step to the other inputs too, otherwise you get this message when hovering over the input
image

bsteinbk: done

src/core/ColorPicker/ColorInputPanel.tsx
```diff
@@ -65,18 +65,22 @@ export const ColorInputPanel = React.forwardRef(
...
         const hsl = activeColor.toHslColor();
         setInput([
-          hsvColor.h.toString(), // use hsvColor to preserve hue for 0,0,0 edge case
-          hsl.s.toString(),
-          hsl.l.toString(),
-          hsl.a?.toFixed(2) ?? (activeColor.getAlpha() / 255).toFixed(2),
+          ColorValue.getFormattedColorNumber(hsvColor.h), // use hsvColor to preserve hue for 0,0,0 edge case
+          ColorValue.getFormattedColorNumber(hsl.s),
+          ColorValue.getFormattedColorNumber(hsl.l),
+          ColorValue.getFormattedColorNumber(
```
saaaaaally, line 71: The input is messed up, on certain colors you can't input the hsl value that you want because it changes it to random decimal values. Also note how the S value also changes slightly when inputting the L value. I think this has to do with all the converting back and forth
input

bsteinbk: This is a by-product of using RGB (tbgr) as the single source of truth. But I want to keep that as it is the cornerstone of all colors in imodeljs. If you go from rgb -> hsl -> rgb then there is no loss but going from hsl -> rgb -> hsl is not foolproof. Same as before but at least now you can do the rgb -> hsl -> rgb.

src/core/utils/color/ColorValue.ts
```diff
@@ -649,25 +676,36 @@ export class ColorValue {
     return this._tbgr === other._tbgr && this._hue === other._hue;
   }
 
+  public static getFormattedColorNumber(value: number, precision = 2) {
```
saaaaaally, line 679: This still doesn't fix the problem where `ColorValue.equals()` doesn't work correctly. 

In line 164 of ColorPalette.tsx, it is setting the `ColorSwatch` to active using `isActive={ color.toHslString(true) === activeColor.toHslString(true)}`, which works ok for now, but ideally we want to use `isActive={color.equals(activeColor)}`. 

However, when using is `equals` the active color is not set correctly when the input is changed:
isActive

Can you look to see if there's a way to use equals instead of toHslString?

src/core/ColorPicker/ColorInputPanel.tsx
mayank99: I think it would help to save the entire original input instead of just `_hue`

bsteinbk: @saaaaaally @mayank99 - just pushed updates that I believe improves the results -- I cleaned the hsv conversion - but there is still some changing just due to hsl -> rgb -> hsl conversion.

src/core/utils/color/ColorValue.ts
bsteinbk: The latest updates use .equals instead of comparing strings.

src/core/ColorPicker/ColorInputPanel.tsx
mayank99: Lets move forward with this then. We can think about saving original input in the future.

src/core/utils/color/ColorValue.ts
saaaaaally: Yeah but it still has the same bug from above where the color swatch is not set to active even when the input should match. Maybe we should just leave comparing strings?

src/core/utils/color/ColorValue.ts
```diff
@@ -613,70 +613,50 @@ export class ColorValue {
...
+    };
   }
 
   /** True if the value of this ColorValue is the same as another ColorValue. */
   public equals(other: ColorValue): boolean {
     return this._tbgr === other._tbgr && this._hue === other._hue;
   }
 
-  public static getFormattedColorNumber(value: number, precision = 2) {
+  public static getFormattedColorNumber(value: number, precision = 1) {
```
saaaaaally, line 659: I don't think alpha values should be rounding to 1 decimal place because the alpha slider is kind of mess up now

alphavalue

bsteinbk: Yes, you are correct ... I will fix that

bsteinbk: Fixed

src/core/utils/color/ColorValue.ts
bsteinbk: try this again after latest push.

saaaaaally (approved):

src/core/ColorPicker/ColorInputPanel.tsx
saaaaaally: Created https://github.com/iTwin/iTwinUI-react/issues/429#issue-1052036750 to address this last remaining bug

src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -30,14 +30,28 @@ export const ColorSwatch = React.forwardRef<HTMLDivElement, ColorSwatchProps>(
...
+    );
+
+    React.useEffect(() => {
+      const newString =
+        typeof color === 'string'
+          ? color
+          : getColorValue(color).toHslString(true);
+
+      setColorString(newString);
+    }, [color]);
```
mayank99, line 46: This is new. Was the style not updating without this useEffect?

src/core/utils/color/ColorValue.ts
```diff
@@ -209,7 +211,7 @@ export class ColorValue {
         hsl.h / 360,
         hsl.s / 100,
         hsl.l / 100,
-        (1 - alpha) * 255,
+        Math.round((1 - alpha) * 255),
```
mayank99, line 214: I thought we were removing all rounding? What's different about opacity?

mayank99: Is it that we want the percentage to be an integer, and then the value between 0-1 will be 2 decimal points at most?

bsteinbk: for RGBA all values have to be integer between 0-255

src/core/ColorPicker/ColorSwatch.tsx
bsteinbk: The problem was that getColorValue(color).toHslString(true) way too many times of every render for every swatch.  Now it only gets called if the color value changes.

mayank99: why not use `useMemo`?

bsteinbk: good point ... I was too fast to reach for useEffect... I will update.

bsteinbk: changed

PR #427, opened 2021-11-11
feat(Table): Add `rowsPerPageLabel` to paginator localization

mayank99:
Added `rowsPerPageLabel` property to paginator localization to show label for the page size selector. Used hook from #436 to hide label on small widths. User can also pass empty string if they don't want the label.

## Checklist


src/core/Table/TablePaginator.tsx
```diff
@@ -311,6 +331,11 @@ export const TablePaginator = (props: TablePaginatorProps) => {
         </IconButton>
       </div>
       <div className='iui-right'>
+        {localization.rowsPerPageLabel && currentWidth >= 1024 && (
```
mayank99, line 334: I am not sure if this should be 1024. Might be a bit too much. But 768 might be too little, idk

src/core/Table/TablePaginator.test.tsx
```diff
@@ -296,12 +297,21 @@ it('should render with custom localization', () => {
...
   const pageSizeSelector = container.querySelector(
     '.iui-dropdown',
   ) as HTMLButtonElement;
   expect(pageSizeSelector).toBeTruthy();
   expect(pageSizeSelector.textContent).toEqual('1-10 of test 195');
 
+  // TODO: fix this
+  // expect(
+  //   container.querySelector('.iui-right .iui-text-block'),
+  // ).toHaveTextContent('Items per test page');
```
mayank99, line 313: I don't know an easy way to test this

src/core/Table/TablePaginator.tsx
```diff
@@ -177,6 +186,13 @@ export const TablePaginator = (props: TablePaginatorProps) => {
   );
   const [overflowRef, visibleCount] = useOverflow(pageList);
 
+  const paginatorRef = React.useRef<HTMLDivElement>(null);
+  const [currentWidth, setCurrentWidth] = React.useState(
+    () => paginatorRef.current?.offsetWidth ?? 0,
+  );
+  const [resizeRef] = useResizeObserver(({ width }) => setCurrentWidth(width));
+  const paginatorRefs = useMergedRefs(paginatorRef, resizeRef);
```
mayank99, line 194: This is actually buggy, which is why the test is failing. Once #436 is merged, I can just reuse the logic from there

src/core/Table/TablePaginator.tsx
```diff
@@ -311,6 +325,11 @@ export const TablePaginator = (props: TablePaginatorProps) => {
         </IconButton>
       </div>
       <div className='iui-right'>
+        {localization.rowsPerPageLabel && paginatorWidth >= 1024 && (
+          <Text isMuted as='span'>
+            {localization.rowsPerPageLabel}
```
saaaaaally, line 330: May need to do something to handle very long strings
image
image

src/core/Table/TablePaginator.tsx
```diff
@@ -311,6 +325,11 @@ export const TablePaginator = (props: TablePaginatorProps) => {
         </IconButton>
       </div>
       <div className='iui-right'>
+        {localization.rowsPerPageLabel && paginatorWidth >= 1024 && (
```
saaaaaally, line 328: Label shouldn't be shown if pageSizeList is undefined

src/core/Table/TablePaginator.tsx
```diff
@@ -89,6 +92,11 @@ export type TablePaginatorProps = {
      * @default (page: number) => `Go to page ${page}`
      */
     goToPageLabel?: (page: number) => string;
+    /**
+     * A label shown next to the page size selector. Use `null` to hide.
+     * @default 'Rows per page'
+     */
+    rowsPerPageLabel?: string | null;
```
saaaaaally, line 99: How come this label doesn't change the visual test images?

src/core/Table/TablePaginator.tsx
mayank99: Fixed by merging main

src/core/Table/TablePaginator.tsx
mayank99: It's because the table is not wide enough in the test environment.

src/core/Table/TablePaginator.tsx
mayank99: this is easily fixed with `text-overflow: ellipsis` but that will need to be in CSS

mayank99: Added `.iui-paginator-page-size-label` from https://github.com/iTwin/iTwinUI/pull/381. Doesn't do anything currently as it needs CSS update (Not updating CSS in this PR because it will break buttons)

src/core/Table/TablePaginator.tsx
```diff
@@ -89,6 +92,11 @@ export type TablePaginatorProps = {
      * @default (page: number) => `Go to page ${page}`
      */
     goToPageLabel?: (page: number) => string;
+    /**
+     * A label shown next to the page size selector. Use `null` to hide.
```
saaaaaally, line 96: Maybe add test that label is hidden when passing in null or empty string

src/core/Table/TablePaginator.tsx
```diff
@@ -320,31 +334,38 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
-            }
-          >
-            {localization.rangeLabel(
-              currentPage * pageSize + 1,
-              Math.min(totalRowsCount, (currentPage + 1) * pageSize),
-              totalRowsCount,
-              isLoading,
+          <>
+            {localization.rowsPerPageLabel && paginatorWidth >= 1024 && (
+              <Text as='span' className='iui-paginator-page-size-label'>
```
saaaaaally, line 339: Why not just use span here?

mayank99: changed to span

src/core/Table/TablePaginator.tsx
mayank99: added test 🤷‍♀️

saaaaaally (approved):

src/core/Table/TablePaginator.tsx
```diff
@@ -28,6 +34,7 @@ const defaultLocalization = {
   previousPage: 'Previous page',
   nextPage: 'Next page',
   goToPageLabel: (page: number) => `Go to page ${page}`,
+  rowsPerPageLabel: 'Rows per page',
```
bentleyvk, line 37: Why stories screenshot did not changed?

src/core/Table/TablePaginator.tsx
```diff
@@ -89,6 +91,11 @@ export type TablePaginatorProps = {
      * @default (page: number) => `Go to page ${page}`
      */
     goToPageLabel?: (page: number) => string;
+    /**
```
bentleyvk, line 94: Story would be nice because people will miss whether it is possible to hide this label. Also, is it possible to increase story screenshot width that label would be visible there?

PR #428, opened 2021-11-12
fix(useOverflow): Fixed the hook to handle more items

### LINKED ISSUE #400: Table: Paginator crashes with 70+ pages
Need to refactor `useOverflow` hook.

bentleyvk:
Calculating average width of the item and using that to find the item count that fits. It usually takes 7-9 rerenders, even with 2.5k items.
It would be nice that reviewers could run and test `Table` with paginator and `Breadcrumbs` to see if there are any bugs left.

Closes #400 

## Checklist


src/core/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -98,29 +98,18 @@ it('should overflow when there is not enough space', () => {
   expect(breadcrumbs[1].textContent).toEqual('…');
   expect(breadcrumbs[1].firstElementChild?.classList).toContain('iui-ellipsis');
   expect(breadcrumbs[2].textContent).toEqual('Item 2');
-
-  scrollWidthSpy.mockRestore();
-  offsetWidthSpy.mockRestore();
 });
 
 it('should restore hidden items when there is enough space again', () => {
```
bentleyvk, line 103: Not sure this test is needed as it was testing the inside logic of `useOverflow` hook but I leaved it.

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -6,6 +6,8 @@ import React from 'react';
 import { useMergedRefs } from './useMergedRefs';
 import { useResizeObserver } from './useResizeObserver';
 
+const STARTING_ITEMS_COUNT = 20;
```
bentleyvk, line 9: Starting by rendering 20 items (just kind of a random number) because rendering all of the items e.g. 2.4k is too much and it becomes laggy.

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -33,22 +35,29 @@ export const useOverflow = <T extends HTMLElement>(
 ) => {
   const containerRef = React.useRef<T>(null);
 
-  const [visibleCount, setVisibleCount] = React.useState(items.length);
-  const overflowBreakpoints = React.useRef<number[]>([]);
+  const [visibleCount, setVisibleCount] = React.useState(
+    Math.min(items.length, STARTING_ITEMS_COUNT),
```
mayank99, line 39: Maybe add `() =>` so it's lazy init

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +69,26 @@ export const useOverflow = <T extends HTMLElement>(
...
-      overflowBreakpoints.current.pop();
+    if (availableWidth < requiredWidth) {
+      const avgItemWidth = requiredWidth / visibleCount;
+      const visibleItems = Math.floor(availableWidth / avgItemWidth);
+      setVisibleCount(visibleItems);
+    } else if (needsFullRerender.current) {
+      const childrenWidth = Array.from(containerRef.current.children).reduce(
+        (sum: number, child: HTMLElement) => sum + child.offsetWidth,
+        0,
+      );
```
mayank99, line 80: If you add type to the array, I think you won't need them in the in callback.
```suggestion
      const childrenWidth = Array.from<HTMLElement>(containerRef.current.children).reduce(
        (sum, { offsetWidth }) => sum + offsetWidth,
        0,
      );
```

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +69,26 @@ export const useOverflow = <T extends HTMLElement>(
...
+      const avgItemWidth = childrenWidth / visibleCount;
```
mayank99, line 81: Why not use `requiredWidth`?

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +69,26 @@ export const useOverflow = <T extends HTMLElement>(
...
+      const visibleItems = Math.floor(availableWidth / avgItemWidth);
+      // Doubling the visible items to overflow the container. Just to be safe.
+      setVisibleCount(Math.min(items.length, visibleItems * 2));
```
mayank99, line 84: Oh wow this is very smart!

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -33,22 +35,29 @@ export const useOverflow = <T extends HTMLElement>(
...
   const updateContainerWidth = React.useCallback(
     ({ width }) => setContainerWidth(width),
     [],
   );
   const [resizeRef, observer] = useResizeObserver<T>(updateContainerWidth);
   const resizeObserverRef = React.useRef(observer);
 
+  React.useLayoutEffect(() => {
+    // Reset when width was increased
+    if (previousContainerWidth.current < containerWidth) {
```
mayank99, line 55: This hook will run even when `disabled` is true, so maybe add `!disabled` to this if condition

src/core/utils/hooks/useOverflow.tsx
mayank99: Maybe should have "MAX" in the name, as it's the maximum (items.length will be used if its lower).

Also maybe make it a power of 2 since we are multiplying by 2 repeatedly 😆 so 16 would be a nice value

src/core/utils/hooks/useOverflow.tsx
bentleyvk: I tried that previously:
image

src/core/utils/hooks/useOverflow.tsx
bentleyvk: If container fits 50 items and in the beginning I add 20 then single item width will be way too big because `offsetWidth` and `scrollWidth` should be all the posible container width.

src/core/utils/hooks/useOverflow.tsx
bentleyvk: ☺

bentleyvk: 😊

src/core/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -79,14 +86,7 @@ it('should accept currentIndex prop', () => {
...
 it('should overflow when there is not enough space', () => {
-  const scrollWidthSpy = jest
-    .spyOn(HTMLElement.prototype, 'scrollWidth', 'get')
-    .mockReturnValueOnce(250)
-    .mockReturnValue(200);
-  const offsetWidthSpy = jest
-    .spyOn(HTMLElement.prototype, 'offsetWidth', 'get')
-    .mockReturnValue(200);
-
+  jest.spyOn(UseOverflow, 'useOverflow').mockReturnValue([jest.fn(), 2]);
```
veekeys, line 89: You are already spying above. Cant you save that spy in variable and just change the return value?

src/core/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -131,20 +120,10 @@ it('should restore hidden items when there is enough space again', () => {
...
 it('should hide first item on very small widths', () => {
-  const scrollWidthSpy = jest
-    .spyOn(HTMLElement.prototype, 'scrollWidth', 'get')
-    .mockReturnValueOnce(250)
-    .mockReturnValueOnce(150)
-    .mockReturnValue(100);
-  const offsetWidthSpy = jest
-    .spyOn(HTMLElement.prototype, 'offsetWidth', 'get')
-    .mockReturnValue(100);
+  jest.spyOn(UseOverflow, 'useOverflow').mockReturnValue([jest.fn(), 1]);
```
veekeys, line 126: Kinda does not make sense these tests anymore as there is nothing being mocked/done with width as test says..

src/core/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -8,6 +8,7 @@ import { render } from '@testing-library/react';
 import { Breadcrumbs, BreadcrumbsProps } from './Breadcrumbs';
 import { Button } from '../Buttons';
 import { SvgChevronRight } from '@itwin/itwinui-icons-react';
+import * as UseOverflow from '../utils/hooks/useOverflow';
```
veekeys, line 11: Why need to import all?

bentleyvk: Because for `spyOn` you need object and its property name. Or you know a better way to mock a function?

src/core/Breadcrumbs/Breadcrumbs.test.tsx
bentleyvk: But it has a special handling when there is only 1 item visible.

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +69,26 @@ export const useOverflow = <T extends HTMLElement>(
     const availableWidth = containerRef.current.offsetWidth;
```
veekeys, line 69: Havent you ever noticed that offsetWidth is not giving you full width? I do remember ending up using `getoundingClientRect` and taking width from there

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +69,26 @@ export const useOverflow = <T extends HTMLElement>(
...
-    // restore items when there's enough space again
-    else if (
-      overflowBreakpoints.current.length > 0 &&
-      availableWidth >
-        overflowBreakpoints.current[overflowBreakpoints.current.length - 1]
-    ) {
-      setVisibleCount((count) => count + 1);
-      overflowBreakpoints.current.pop();
+    if (availableWidth < requiredWidth) {
+      const avgItemWidth = requiredWidth / visibleCount;
```
veekeys, line 73: hmm..Interesting part. So initially, if there is more than 20 items, `visibleCount` would be 20, so avgItem is not really right one. Unless idea behind it is different?

src/core/utils/hooks/useOverflow.test.tsx
```diff
@@ -34,21 +38,66 @@ it('should overflow when there is not enough space', () => {
...
   );
 
-  expect(container.querySelectorAll('span')).toHaveLength(4);
+  await waitFor(() => {
+    expect(container.querySelectorAll('span')).toHaveLength(4);
+  });
 
   scrollWidthSpy.mockRestore();
   offsetWidthSpy.mockRestore();
+  childOffsetWidthSpy.mockRestore();
```
veekeys, line 47: `restoreAllMocks` wont do the job? Also, move this to `afterEach`

src/core/Breadcrumbs/Breadcrumbs.test.tsx
veekeys: I see there is way if you have default export.. you can just pass path to spy, but.. Maybe this is giving better readability

src/core/utils/hooks/useOverflow.tsx
veekeys: well ok, maybe kinda got it. looks smart, gj

src/core/Breadcrumbs/Breadcrumbs.test.tsx
veekeys: hmm.. thought test name is kinda wrong then, but minor.. leave it.

src/core/Breadcrumbs/Breadcrumbs.test.tsx
bentleyvk: Removed.

src/core/utils/hooks/useOverflow.tsx
bentleyvk: Added

src/core/utils/hooks/useOverflow.tsx
bentleyvk: Added

src/core/utils/hooks/useOverflow.tsx
bentleyvk: Idk if I can to use it because `offsetWidht` and `scrollWidth` are rounded values so I can end up in a situation where client rect returns `300.5px` and `scrollWidth` is `301px` and then it would remove one item although it might be ok.

src/core/utils/hooks/useOverflow.test.tsx
bentleyvk: Changed.

src/core/utils/hooks/useOverflow.tsx
bentleyvk: For me it is easier to work with numbers that divide by 10, maybe I am not a real programmer 😕

veekeys (approved):

src/core/utils/hooks/useOverflow.tsx
bentleyvk: Added `MAX`. He is a really great guy!

mayank99:
Maybe unrelated to this PR but I noticed two different issues

1. touch target
paginator1

2. content shift
paginator2

bentleyvk:
> I noticed two different issues that are not present in the live version.
> 
> 1. touch target
>    paginator1
>    image
> 2. content shift
>    paginator2

Fixed the first one, another one needs CSS changes and is unrelated to the current PR.

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -60,21 +70,26 @@ export const useOverflow = <T extends HTMLElement>(
...
+        0,
+      );
+      const avgItemWidth = childrenWidth / visibleCount;
+      const visibleItems = Math.floor(availableWidth / avgItemWidth);
+      // Doubling the visible items to overflow the container. Just to be safe.
+      setVisibleCount(Math.min(items.length, visibleItems * 2));
     }
-  }, [containerWidth, visibleCount, disabled]);
+    needsFullRerender.current = false;
+  }, [containerWidth, visibleCount, disabled, items.length]);
```
mayank99, line 88: Just noticed `containerWidth` in deps array but it's not used

src/core/utils/hooks/useOverflow.tsx
```diff
@@ -33,22 +35,30 @@ export const useOverflow = <T extends HTMLElement>(
...
+  React.useLayoutEffect(() => {
+    if (disabled) {
+      setVisibleCount(items.length);
+    } else {
+      setVisibleCount(Math.min(items.length, STARTING_MAX_ITEMS_COUNT));
+      needsFullRerender.current = true;
+    }
+  }, [containerWidth, disabled, items]);
+
   const mergedRefs = useMergedRefs(containerRef, resizeRef);
```
mayank99, line 62: Might be more efficient if we don't merge `resizeRef` when `disabled` is true. Ref wont be attached and ResizeObserver won't be created

src/core/utils/hooks/useOverflow.tsx
bentleyvk: Needed

mayank99 (approved):

PR #431, opened 2021-11-12
feat(Label): Add standalone `Label` component

mayank99:
Added polymorphic `Label` component with `as` prop. Uses `iui-input-label` class from https://github.com/iTwin/iTwinUI/pull/351.

Added stories in both `ComboBox` and `Typgraphy/Label`:

image

## Checklist


stories/core/ComboBox.stories.tsx
```diff
@@ -365,3 +365,25 @@ export const DisabledItems: Story<Partial<ComboBoxProps<string>>> = (args) => {
   );
 };
 DisabledItems.args = { ...Basic.args };
+
+export const WithLabel: Story<Partial<ComboBoxProps<string>>> = (args) => {
+  const options = React.useMemo(() => countriesList, []);
+
+  return (
+    <>
+      <Label htmlFor='combo-input'>Country</Label>
```
mayank99, line 374: I am not sure if I should add a separate story for this. There is no use for a label on its own.

src/core/Label/Label.tsx
```diff
@@ -0,0 +1,73 @@
...
+   * @default 'label'
+   */
+  as?: T;
+  /**
+   * Set the display style of the label.
+   *   - 'block' - default, no extra spacing (label and input will be on separate lines)
+   *   - 'inline' - label gets a right margin so it can be placed inline
+   * @default 'block'
+   */
+  displayStyle?: 'block' | 'inline';
```
mayank99, line 22: I am not sure about this prop name. Maybe it should just be called `inline` or something. Suggestions welcome

gretanausedaite: In CSS it's just display, maybe that's enough?

mayank99: Well, the problem is, it doesn't really change the actual `display` value (it stays `inline-block` for both cases). It just changes the margin

I only called it `displayStyle` because that's the name we use in our inputs

stories/core/ComboBox.stories.tsx
gretanausedaite: Maybe add story for `Label` in Typography section. Have examples with combobox and other inputs

src/core/Label/Label.tsx
gretanausedaite: I vote for consistency. If we already use it in inputs, lets use same name.

bentleyvk: Shouldn't it be `display: block` to take all width when it is `block` type?

stories/core/ComboBox.stories.tsx
bentleyvk: I agree with Greta. If we don't have a dedicated story then people might not even know about this component.

src/core/Label/Label.tsx
mayank99: span in pink
image

stories/core/ComboBox.stories.tsx
mayank99: Added stories under Typography

src/core/Label/Label.tsx
mayank99: Fixed CSS and bumped version

stories/core/Typography/Label.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+Inline.args = {
+  displayStyle: 'inline',
+  required: true,
+};
+
+export const Polymorphic: Story<LabelProps<'div'>> = (args) => {
+  return (
+    <div style={{ display: 'flex' }}>
+      <Label displayStyle='inline' as='div' {...args}>
+        <Text isMuted>Name:</Text>
```
mayank99, line 58: I was not able to use `<Label as={Text} isMuted>`. I think there is a problem in the polymorphic approach where it doesn't take the props of custom component

stories/core/Typography/Label.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import { Input, Label, LabelProps, Text } from '../../../src/core';
+
+export default {
+  component: Label,
+  args: {
+    children: 'Name',
+    displayStyle: 'block',
```
bentleyvk, line 13: Is it possible to have stories where switching `displayStyle` changes the layout? Because now the control for it is useless.

stories/core/Typography/Label.stories.tsx
bentleyvk: I think it is not a big issue right now for `Label`. But you check this article: https://blog.andrewbran.ch/polymorphic-react-components/. It handles custom components and has an alternative way for allowing custom types.

stories/core/Typography/Label.stories.tsx
mayank99: It shows the actual behavior tho, i.e. it's not enough by itself.

I could toggle flex-direction based on arg but I think that adds too much noise

bentleyvk (approved):

stories/core/Typography/Label.stories.tsx
mayank99: I already read that article when I first implemented polymorphism. Their conclusion doesn't really help:
image

mayank99: Turns out it's because Text is also a polymorphic component, so I just had to add a wrapper like this: `as={(args: TextProps) => <Text {...args} />} isMuted>`
```suggestion
      <Label displayStyle='inline'as={(args: TextProps) => <Text {...args} />} isMuted {...args}>
        Name:
```

But it's not reliable for styling because the components have same specificity (notice the Text font-weight gets overridden by Label)
image

PR #433, opened 2021-11-15
feat(InformationPanel): Add `InformationPanelContent` component for inner rows

mayank99:
Added `InformationPanelContent` to use inside `InformationPanelBody`. Supports toggling responsive layout (two columns vs one) using `displayStyle` prop.

responsive-panel

## Checklist


src/core/InformationPanel/InformationPanelContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+      )}
+      ref={resizeRef}
+      {...rest}
+    >
+      {children}
+    </div>
+  );
+};
+
+export const useContainerWidth = <T extends HTMLElement>(
```
mayank99, line 80: This hook will be imported after #436 is merged

src/core/InformationPanel/InformationPanelContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Maximum width of the panel below which layout will shift from two columns to one,
+   * i.e. label and input will display on separate lines for widths lower than this value.
+   *
+   * If set to 0, the layout will always be two columns.
+   * If set to `Number.POSITIVE_INFINITY`, the layout will always be one column.
+   *
+   * @default 0
+   */
+  responsiveBreakpoint?: number;
```
mayank99, line 25: Need feedback on this part

src/core/InformationPanel/InformationPanelContent.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Content of the component.
+   * Should be pairs of `Label` and input components.
```
veekeys, line 24: I dont think we should be that specific.. it should support anything in there with key, value lets say.. Or even more. Or our css is only handling those?

src/core/InformationPanel/InformationPanelContent.tsx
veekeys: oh wow...
Couldnt we somehow give this control to the user, so he handles, when something happens to content??

src/core/InformationPanel/InformationPanel.tsx
```diff
@@ -42,7 +42,10 @@ export type InformationPanelProps = {
  *       <Text variant='subheading'>InfoPanel heading</Text>
  *     </InformationPanelHeader>
  *     <InformationPanelBody>
- *       <span>Info panel content</span>
+ *       <Text>Metadata</Text>
```
bentleyvk, line 45: Why not `<Label>`?

stories/core/InformationPanel.stories.tsx
```diff
@@ -102,7 +102,34 @@ export const Basic: Story<InformationPanelProps> = (args) => {
...
+            <Input size='small' id='year-input' defaultValue='2021' readOnly />
+
+            <Label htmlFor='path-input'>Path</Label>
+            <Input
+              size='small'
+              id='path-input'
+              defaultValue='/Shared/Music/'
+              readOnly
+            />
+          </InformationPanelContent>
```
bentleyvk, line 132: Can you also add splitter and sub-title? So users could see the proper example.
image

src/core/InformationPanel/InformationPanelContent.tsx
```diff
@@ -0,0 +1,76 @@
...
+   */
+  children: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * The `InformationPanelContent` component should be used inside `InformationPanelBody`
```
bentleyvk, line 30: Too long description, no one will read, sorry 😀

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: You mean like prop to toggle 2 columns vs 1?

i doubt users will add resize observers to toggle it responsively

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: Cant hurt to have. All the info is useful to understand behavior

stories/core/InformationPanel.stories.tsx
mayank99: That "proper example" has tons of overrides, which I'm not sure I should add.

1. readonly inputs have border hidden
2. `hr` has custom margin, color and thickness

bentleyvk: Oh, I just meant the `<hr>` so you just need padding to it as we already provide color from CSS and then "Attributes" which is just leading text.

src/core/InformationPanel/InformationPanelContent.tsx
veekeys: Lets rename this prop to something like `displayType`, `styleType` or something from that family to basically give option `block` or `inline`. With that, we just apply `iui-inline` class when needed. And then this example with using our custom hook we can give in both: stories and @example above this component. This will give us more flexibility and we leave user to what ever custom logic he might want to have to arrange this layout.
There is still a risk of having some custom layout, which we wont support and will have issues, but for now I think that would be a better solution than having this custom prop.
And lets merge this, cause its just staying for too long.

stories/core/InformationPanel.stories.tsx
veekeys: Update story too

src/core/InformationPanel/InformationPanelContent.tsx
veekeys: Opinion?

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: It's an internal hook. Sounds like a bad idea to show it in examples.

stories/core/InformationPanel.stories.tsx
mayank99: >  we already provide color from CSS

That color is not enough because it is relying on border. This is what I had to do to make it look the same as our CSS example.
image

Without those overrides, it's too thick.
image

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: the grid already supports pairs of any two components, but using `Label` gives it some extra styling

veekeys: So if we put something else than Label, it should still be aligned? So I would suggest rephrasing the message from 'should' to 'could', for example..

src/core/InformationPanel/InformationPanelContent.tsx
veekeys: Ok, dont show then. Just rename the prop and lets merge this this year somehow..

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: added the word "ideally"

src/core/InformationPanel/InformationPanelContent.tsx
mayank99: ok changed

veekeys (approved):

PR #436, opened 2021-11-16
chore(Tabs): Move responsive width logic into reusable hook

mayank99:
Added new hook `useContainerWidth` to allow reusing the logic for responsive width from Tabs. This hook will be used in `InformationPanelContent` in my next PR.

## Checklist

### DIFF
--- a/src/core/utils/hooks/useContainerWidth.test.tsx
+++ b/src/core/utils/hooks/useContainerWidth.test.tsx
@@ -1,0 +1,85 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+// import React from 'react';
+import { useContainerWidth } from './useContainerWidth';
+import * as UseResizeObserver from './useResizeObserver';
+import { renderHook, act } from '@testing-library/react-hooks';
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+afterEach(() => {
+  jest.restoreAllMocks();
+});
+
+it('should set initial size', () => {
+  const { result } = renderHook(() => useContainerWidth());
+  const element = document.createElement('div');
+  Object.defineProperties(element, {
+    getBoundingClientRect: { value: () => ({ width: 100 }) },
+  });
+  act(() => {
+    result.current[0](element);
+  });
+
+  expect(result.current[1]).toEqual(100);
+});
+
+it('should update value when resized', () => {
+  let triggerResize: (size: DOMRectReadOnly) => void = jest.fn();
+  jest
+    .spyOn(UseResizeObserver, 'useResizeObserver')
+    .mockImplementation((onResize) => {
+      triggerResize = onResize;
+      return [
+        jest.fn(),
+        ({ disconnect: jest.fn() } as unknown) as ResizeObserver,
+      ];
+    });
+  const { result } = renderHook(() => useContainerWidth());
+  const element = document.createElement('div');
+  Object.defineProperties(element, {
+    getBoundingClientRect: { value: () => ({ width: 100 }) },
+  });
+  act(() => {
+    result.current[0](element);
+  });
+
+  expect(result.current[1]).toEqual(100);
+
+  act(() => triggerResize({ width: 125 } as DOMRect));
+  expect(result.current[1]).toEqual(125);
+
+  act(() => triggerResize({ width: 150 } as DOMRect));
+  expect(result.current[1]).toEqual(150);
+});
+
+it('should not update value when resized if disabled', () => {
+  let triggerResize: (size: DOMRectReadOnly) => void = jest.fn();
+  const disconnect = jest.fn(() => {
+    triggerResize = jest.fn();
+  });
+  jest
+    .spyOn(UseResizeObserver, 'useResizeObserver')
+    .mockImplementation((onResize) => {
+      triggerResize = onResize;
+      return [jest.fn(), ({ disconnect } as unknown) as ResizeObserver];
+    });
+  const { result, waitFor } = renderHook(() => useContainerWidth(false));
+  const element = document.createElement('div');
+  Object.defineProperties(element, {
+    getBoundingClientRect: { value: () => ({ width: 100 }) },
+  });
+  act(() => {
+    result.current[0](element);
+  });
+
+  expect(result.current[1]).toEqual(100);
+
+  act(() => triggerResize({ width: 125 } as DOMRect));
+  waitFor(() => expect(result.current[1]).toEqual(100));
+  expect(disconnect).toHaveBeenCalled();
+});
--- a/src/core/utils/hooks/useContainerWidth.test.tsx
+++ b/src/core/utils/hooks/useContainerWidth.test.tsx
@@ -68,7 +68,7 @@ it('should not update value when resized if disabled', () => {
       triggerResize = onResize;
       return [jest.fn(), ({ disconnect } as unknown) as ResizeObserver];
     });
-  const { result, waitFor } = renderHook(() => useContainerWidth(false));
+  const { result } = renderHook(() => useContainerWidth(false));
   const element = document.createElement('div');
   Object.defineProperties(element, {
     getBoundingClientRect: { value: () => ({ width: 100 }) },
@@ -80,6 +80,6 @@ it('should not update value when resized if disabled', () => {
   expect(result.current[1]).toEqual(100);
 
   act(() => triggerResize({ width: 125 } as DOMRect));
-  waitFor(() => expect(result.current[1]).toEqual(100));
+  expect(result.current[1]).toEqual(100);
   expect(disconnect).toHaveBeenCalled();
 });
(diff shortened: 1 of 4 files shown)


src/core/utils/hooks/useContainerWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @private
+ * @param watchResizes If false, ResizeObserver will not be connected and only the initial width will be returned
+ * @returns [ref to attach to the element, stateful width of the element]
+ *
+ * @example
+ * const [ref, width] = useContainerWidth();
+ * ... // do something with width
+ * return <div ref={ref}>...</div>;
+ */
+export const useContainerWidth = <T extends HTMLElement>(
```
mayank99, line 24: Not sure about the naming here. Maybe should be `useResponsiveWidth` or `useContentWidth` or `useResizableWidth`.

Also not sure if we should include height too. Usually for responsive design, width is the main concern.

src/core/utils/hooks/useContainerWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+  // callback ref to set initial contentWidth only once
+  const isInitialized = React.useRef(false);
+  const ref = React.useCallback((element: T) => {
+    if (!element) {
+      return;
+    }
+    if (!isInitialized.current) {
+      setContentWidth(element.getBoundingClientRect().width);
+    } else {
+      isInitialized.current = true;
```
veekeys, line 38: Amm... I think you wanted to set this in the above block :D

src/core/utils/hooks/useContainerWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }, []);
+
+  const updateWidth = React.useCallback(
+    ({ width }) => setContentWidth(width),
+    [],
+  );
+
+  const [resizeRef, resizeObserver] = useResizeObserver(updateWidth);
+
+  if (!watchResizes) {
```
veekeys, line 49: Didnt want to add this in useEffect?

src/core/utils/hooks/useContainerWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @param watchResizes If false, ResizeObserver will not be connected and only the initial width will be returned
+ * @returns [ref to attach to the element, stateful width of the element]
+ *
+ * @example
+ * const [ref, width] = useContainerWidth();
+ * ... // do something with width
+ * return <div ref={ref}>...</div>;
+ */
+export const useContainerWidth = <T extends HTMLElement>(
+  watchResizes = true,
```
veekeys, line 25: so is there a point to use this hook if you dont want to listen to resizes?? Kinda looks too heavy

src/core/utils/hooks/useContainerWidth.tsx
veekeys: hmm.. yeah, I think this hook is more about giving you the width of the element on screen change, but then naturally i have a question, why not just use resize observer for that.. so then maybe would vote for `useResizableWidth`, but.. kinda sounds weird.

src/core/utils/hooks/useContainerWidth.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    .spyOn(UseResizeObserver, 'useResizeObserver')
+    .mockImplementation((onResize) => {
+      triggerResize = onResize;
+      return [
+        jest.fn(),
+        ({ disconnect: jest.fn() } as unknown) as ResizeObserver,
+      ];
+    });
+  const { result } = renderHook(() => useContainerWidth());
+  const element = document.createElement('div');
```
veekeys, line 43: cant you have this element mocked before test and reuse it in every test case?

src/core/utils/hooks/useContainerWidth.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    getBoundingClientRect: { value: () => ({ width: 100 }) },
+  });
+  act(() => {
+    result.current[0](element);
+  });
+
+  expect(result.current[1]).toEqual(100);
+
+  act(() => triggerResize({ width: 125 } as DOMRect));
+  waitFor(() => expect(result.current[1]).toEqual(100));
```
veekeys, line 83: why waitfor here is needed?

src/core/utils/hooks/useContainerWidth.tsx
```diff
@@ -0,0 +1,58 @@
...
+) => {
+  const [contentWidth, setContentWidth] = React.useState(0);
+
+  // callback ref to set initial contentWidth only once
+  const isInitialized = React.useRef(false);
+  const ref = React.useCallback((element: T) => {
+    if (!element) {
+      return;
+    }
+    if (!isInitialized.current) {
```
bentleyvk, line 35: Will it trigger rerender with the same value?

src/core/utils/hooks/useContainerWidth.test.tsx
bentleyvk: Or you can just have an object with prop `getBoundingClientRect`.

src/core/utils/hooks/useContainerWidth.test.tsx
mayank99: leftover, removed

src/core/utils/hooks/useContainerWidth.tsx
mayank99: Yes I think there is. The first callback ref to set initial width is very useful, and there is also the ability to turn off listening to resizes later (we do this in Tabs for default vs rest)

src/core/utils/hooks/useContainerWidth.tsx
mayank99: I had `useResizableWidth` previously but then I thought it could be useful in non-resizable cases too so I changed it

src/core/utils/hooks/useContainerWidth.tsx
mayank99: what do you mean?

mayank99: oh i see. I only added this because I never ever want to use the width from getBoundingClientRect again after the initial mount, because ResizeObserver should take preference.

src/core/utils/hooks/useContainerWidth.test.tsx
mayank99: it's only used in two tests so I thought abstraction would not be that useful

src/core/utils/hooks/useContainerWidth.test.tsx
```diff
@@ -0,0 +1,85 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+// import React from 'react';
```
veekeys, line 5: ?

src/core/utils/hooks/useContainerWidth.test.tsx
veekeys: I count 3 uses, so..

src/core/utils/hooks/useContainerWidth.tsx
veekeys: What is the use of just initial width? Why would you need to use it, if you can control content width with css (using display inline)

src/core/utils/hooks/useContainerWidth.tsx
veekeys: So it runs when watchResizes change only?

mayank99: This if condition is just to disconnect the stale resize observer

The real magic happens below where I'm not merging the resize ref:
image

src/core/utils/hooks/useContainerWidth.tsx
mayank99: well, sometimes you only care about initial width because you want to do something with it (like set a CSS variable)

src/core/utils/hooks/useContainerWidth.test.tsx
mayank99: ok added helper

src/core/utils/hooks/useContainerWidth.test.tsx
mayank99: !

src/core/utils/hooks/useContainerWidth.tsx
veekeys: but if `watchResizes` is never changed, we expect just garbage collector to take care of disconnecting it?

src/core/utils/hooks/useContainerWidth.tsx
veekeys: for that you can have just ref and take its width, same 1 liner basically

src/core/utils/hooks/useContainerWidth.tsx
mayank99: if `watchResizes` is never changed (meaning it's false) then `resizeObserver` would be undefined, so it wouldn't disconnect.

whereas if it changes from true to false, the manual disconnect is needed

src/core/utils/hooks/useContainerWidth.tsx
mayank99: Well that's what I had before, it doesn't work because ref will be null initially
image

veekeys: oh, I would have expected you use it directly from ref, where you need it.. but yeah

veekeys (approved):

PR #441, opened 2021-11-19
feat(Table): Don't show paginator when there is only one page

### LINKED ISSUE #392: Table: Don't show paginator when there is only one page
Don't show paginator when there is only one page.

gretanausedaite:
Closes: #392 

paginator1page

### DIFF
--- a/src/core/Table/TablePaginator.test.tsx
+++ b/src/core/Table/TablePaginator.test.tsx
@@ -76,7 +76,7 @@ it('should render currently visible rows info and page size selector', () => {
   expect(onPageSizeChange).toHaveBeenCalledWith(25);
 });
 
-it('should render without paginator when rows count is less than page size', () => {
+it('should render without page list', () => {
   const { container } = renderComponent({
     totalRowsCount: 10,
     pageSizeList: [10, 25, 50],
@@ -89,6 +89,13 @@ it('should render without paginator when rows count is less than page size', () 
   expect(pageSizeSelector.textContent).toEqual('1-10 of 10');
 
   const paginator = container.querySelector('.iui-center') as HTMLElement;
+  expect(paginator).not.toBeTruthy();
+});
+
+it('should render without page list and page size list', () => {
+  const { container } = renderComponent({ totalRowsCount: 10 });
+
+  const paginator = container.querySelector('.iui-paginator') as HTMLElement;
   expect(paginator).not.toBeTruthy();
 });
 
--- a/src/core/Table/TablePaginator.test.tsx
+++ b/src/core/Table/TablePaginator.test.tsx
@@ -76,6 +76,21 @@ it('should render currently visible rows info and page size selector', () => {
   expect(onPageSizeChange).toHaveBeenCalledWith(25);
 });
 
+it('should render without paginator when rows count is less than page size', () => {
+  const { container } = renderComponent({
+    totalRowsCount: 10,
+  });
+
+  const pageSizeSelector = container.querySelector(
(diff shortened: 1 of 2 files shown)


src/core/Table/TablePaginator.tsx
```diff
@@ -257,59 +258,61 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
-          styleType='borderless'
-          disabled={currentPage === totalPagesCount - 1 || hasNoRows}
-          onClick={() => onPageChange(currentPage + 1)}
-          size={buttonSize}
-          aria-label={localization.nextPage}
-        >
-          <SvgChevronRight />
-        </IconButton>
-      </div>
+      {showPaginator && (
```
bentleyvk, line 261: If there is no page size selector then it showing just an empty gray footer. Might need to check that too.

gretanausedaite: Added condition for page list and page size list

src/core/Table/TablePaginator.test.tsx
```diff
@@ -131,31 +154,6 @@ it('should render loading state when there is no data', () => {
   expect(container.querySelector('.iui-dropdown')).toBeFalsy();
 });
 
-it('should render disabled state when there is no data', () => {
```
mayank99, line 134: why removed this test?

src/core/Table/TablePaginator.tsx
```diff
@@ -254,64 +256,70 @@ export const TablePaginator = (props: TablePaginatorProps) => {
     </>
   );
 
+  if (!showPagesList && !showPageSizeList) {
+    return null;
```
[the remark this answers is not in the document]
bentleyvk: I am not sure what is better: showing paginator with everything disabled or not showing paginator at all. I checked some other sites and they are not showing paginator at all when there is just one page.

src/core/Table/TablePaginator.tsx
```diff
@@ -235,7 +235,8 @@ export const TablePaginator = (props: TablePaginatorProps) => {
   }
 
   const hasNoRows = totalPagesCount === 0;
-  const showPaginator = totalRowsCount > pageSize || isLoading;
+  const showPagesList = totalRowsCount > pageSize || isLoading;
```
bentleyvk, line 238: Maybe use `totalPagesCount` instead.

src/core/Table/TablePaginator.test.tsx
gretanausedaite: Because we are not showing pages list when there is no data.

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
gretanausedaite: Not sure how to solve this issue without adding additional prop for total rows count or filtered boolean

src/core/Table/TablePaginator.tsx
gretanausedaite: Changed to totalPagesCount

src/core/Table/TablePaginator.tsx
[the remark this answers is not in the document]
FlyersPh9: I am ok with the paginator disappearing if no pages are available after a filter is applied.  As long as the table's body extends all the way down to where the paginator was, so if there is content under the table we avoid content shift.

bentleyvk (approved):

mayank99 (approved):

PR #448, opened 2021-11-30
feat(Table): Added column resizing

### LINKED ISSUE #84: Table: Resizable Columns
#### Feature
Table that will have Resizable columns (gripper on header ) so that user will be able to increase/decrease width of each column as per his requirement.

bentleyvk:
column-resize-done

Closes #84 

 - In order to resize columns, we need to know their widths and make every column static width therefore I am saving widths inside `column.resizeWidth` that I got from ref.
 - When one column width is increasing another one is decreasing in the same amount. If next column has disabled resize then it is looking for the next one that is resizable.
 - It is favoring `minWidth` and `maxWidth` therefore you can't resize past those.
 - Because columns widths are static, I need to handle table resize too. For this I am temporally making columns kind of dynamic width again by setting `flex-grow` and `flex-shrink`, then getting new widths and making columns static again.
 - `useResizeColumns` hook is taken from `react-table` and I make it work with `ownerDocument` also cleaned up code a bit and made it to resize sibling columns.

## Checklist


src/core/Table/Table.tsx
```diff
@@ -478,22 +559,39 @@ export const Table = <
...
                       {...columnProps}
                       key={columnProps.key}
                       title={undefined}
+                      ref={(el) => {
+                        if (el) {
+                          columnRefs.current[column.id] = el;
+                          column.resizeWidth = el.offsetWidth;
+                        }
+                      }}
+                      onMouseDown={onSortClick}
```
bentleyvk, line 594: Prevents from triggering sort when resizing the column.

stories/core/Table.stories.tsx
```diff
@@ -1335,12 +1336,16 @@ export const Full: Story<Partial<TableProps>> = (args) => {
...
         isSelectable
         isSortable
+        isResizable
         {...args}
       />
       <Tooltip
         reference={rowRefMap.current[hoveredRowIndex]}
         content={`Hovered over ${data[hoveredRowIndex].name}.`}
         placement='bottom'
+        followCursor
```
bentleyvk, line 1346: Slipped in story update from the discussion.

mayank99: So are you going to undo it? 😄

bentleyvk: Too much work, I'll just leave it here and hope nobody sees it 😁

stories/core/Table.stories.tsx
```diff
@@ -1676,3 +1682,140 @@ WithManualPaginator.argTypes = {
...
+
+  const data = useMemo(
+    () => [
+      {
+        index: 1,
+        name: 'Name1',
+        description: 'Description1',
+        id: '111',
+        startDate: new Date('May 1, 2021'),
+        endDate: new Date('Jun 1, 2021'),
```
saaaaaally, line 1759: If there is no min width maybe the text should be hidden and show "N..." instead of extending the cell vertically
image

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+  Hooks,
+  makePropGetter,
+  TableInstance,
+  TableKeyedProps,
+  TableState,
+  useGetLatest,
+} from 'react-table';
+
+export const useResizeColumns = <T extends Record<string, unknown>>(
+  ownerDocument: Document | undefined,
```
veekeys, line 47: Maybe this could be optional and default to just `document`

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+    return props;
+  }
+
+  const { dispatch, flatHeaders } = instance;
+
+  const onResizeStart = (
+    e: React.TouchEvent & React.MouseEvent,
+    header: HeaderGroup,
+  ) => {
+    let isTouchEvent = false;
```
veekeys, line 76: `const isTouchEvent = e.type === 'touchstart``

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+    if (e.type === 'touchstart') {
+      // lets not respond to multiple touches (e.g. 2 or 3 fingers)
+      if (e.touches && e.touches.length > 1) {
```
veekeys, line 79: And then you can have just single if here `if (isTouchEvent && e.touches && e.touches.length) {return;}`

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+
+    const headerIdWidths = getLeafHeaders(header).map((d) => [d.id, d.width]);
+    const nextHeaderIdWidths = nextHeader
+      ? getLeafHeaders(nextHeader).map((d) => [d.id, d.width])
+      : [];
+
+    const clientX = isTouchEvent ? Math.round(e.touches[0].clientX) : e.clientX;
+
+    const dispatchMove = (clientXPos: number) =>
+      dispatch({ type: actions.columnResizing, clientX: clientXPos });
```
veekeys, line 100: is this cought by react-table and need only clientx? or this might go to the user too..? in that case, could there be a situation, that we need more than just clientX .. ?

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+      },
+      touch: {
+        moveEvent: 'touchmove',
+        moveHandler: (e: TouchEvent) => {
+          if (e.cancelable) {
+            e.preventDefault();
+            e.stopPropagation();
+          }
+          dispatchMove(e.touches[0].clientX);
+          return false;
```
veekeys, line 131: Sure you need return boolean? mouse move handler is void, so types are different

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+    getHooks,
+    state: { columnResizing },
+  } = instance;
+
+  const getInstance = useGetLatest(instance);
+
+  flatHeaders.forEach((header, index) => {
+    const resizeWidth = columnResizing.columnWidths[header.id];
+    header.width = resizeWidth || header.width || header.originalWidth;
+    header.isResizing = columnResizing.isResizingColumn === header.id;
```
veekeys, line 355: so header id is same as column id, right?

bentleyvk: yup

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+
+const getPreviousResizableHeader = <T extends Record<string, unknown>>(
+  headerColumn: ColumnInstance<T>,
+  instance: TableInstance<T>,
+) => {
+  const headerIndex = instance.flatHeaders.findIndex(
+    (h) => h.id === headerColumn.id,
+  );
+  return [...instance.flatHeaders]
+    .slice(1, headerIndex)
```
veekeys, line 386: why start index 1?

src/core/Table/Table.tsx
```diff
@@ -456,10 +480,59 @@ export const Table = <
...
+      flatHeaders.forEach((header) => {
+        if (columnRefs.current[header.id]) {
+          header.resizeWidth = columnRefs.current[
+            header.id
+          ].getBoundingClientRect().width;
+        }
+      });
+
+      // Leave resize handling to the flex
+      if (Object.keys(state.columnResizing.columnWidths).length === 0) {
```
veekeys, line 502: could you explain more, what this does?

src/core/Table/Table.tsx
```diff
@@ -478,22 +551,39 @@ export const Table = <
...
+                    style: { ...getCellStyle(column, state.isTableResizing) },
                   });
                   return (
                     <div
                       {...columnProps}
                       key={columnProps.key}
                       title={undefined}
+                      ref={(el) => {
+                        if (el) {
+                          columnRefs.current[column.id] = el;
```
veekeys, line 582: I guess we dont need to do that if resize is disabled?

src/core/Table/Table.tsx
veekeys: Prevents by calling onSort ?

src/core/Table/TableRowMemoized.tsx
```diff
@@ -14,7 +14,7 @@ import { TableCell } from './TableCell';
  * Although state is not used it is needed for `React.memo` to check state that changes row state e.g. selection.
  * When adding new features check whether it changes state that affects row. If it does then add equality check to `React.memo`.
  */
-const TableRow = <T extends Record<string, unknown>>(props: {
+export const TableRow = <T extends Record<string, unknown>>(props: {
```
veekeys, line 17: why export now?

src/core/Table/utils.ts
```diff
@@ -6,14 +6,20 @@ import { ColumnInstance } from 'react-table';
 
 export const getCellStyle = <T extends Record<string, unknown>>(
   column: ColumnInstance<T>,
+  isTableResizing = false,
```
veekeys, line 9: maybe make it mandatory to always pass it from table instance?

src/core/Table/utils.ts
```diff
@@ -6,14 +6,20 @@ import { ColumnInstance } from 'react-table';
...
   const style = {} as React.CSSProperties;
   style.flex = `1 1 145px`;
   if (column.width) {
     const width =
       typeof column.width === 'string' ? column.width : `${column.width}px`;
     style.width = width;
-    style.flex = `0 0 ${width}`;
+    // This allows flexbox to handle the width of the column on table resize
+    if (isTableResizing && column.canResize) {
+      style.flex = `${Number(column.width)} 1 ${width}`;
```
veekeys, line 19: what you expect to get from `Number(column.width)` ? I tohught first value should be from 0 to 1

src/types/react-table-config.ts
```diff
@@ -35,10 +35,10 @@ import {
   UsePaginationInstanceProps,
   UsePaginationOptions,
   UsePaginationState,
-  // UseResizeColumnsColumnOptions,
+  UseResizeColumnsColumnOptions,
   UseResizeColumnsColumnProps,
-  // UseResizeColumnsOptions,
-  UseResizeColumnsState,
+  UseResizeColumnsOptions,
+  // UseResizeColumnsState,
```
veekeys, line 41: sure not needed?

stories/core/Table.stories.tsx
veekeys: yeah, i agree.. we need to set min width on columns...

veekeys: Also, different, but looks weird that I want to resize date, but description is resizing
resize

src/core/Table/Table.tsx
bentleyvk: Prevents from calling `onClick` when resizer is dragged then released and mouse is over some column.

stories/core/Table.stories.tsx
veekeys: so as we talked, that behavior might be ok, but need to handle last column

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: We can't use `document` because of SSG/SSR frameworks, we can only use `getDocument()` which also has `Document | undefined`  type.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: We talked about this in chat that `dispatch` is being used to handle internal `Table` state.

src/types/react-table-config.ts
bentleyvk: I am pretty sure about this:
image

src/core/Table/utils.ts
bentleyvk: I expect to get width which is plain number because I saved it that way. `flex-grow` can be any positive number.

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+/**
+ * Copied from react-table as useResizerColumns does not have ownerDocument check also
```
mayank99, line 6: Maybe add a link to the source. Also you misspelled the hook name.

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+ * made some changes to resize sibling/next column on resize.
```
mayank99, line 7: "some changes" is not very helpful.

I tried to do a side-by-side diff and it was not obvious what changed (TS+lint changes basically every line), so I would suggest adding more info here or even inline about everything you changed/added.

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+      clientX,
+    });
+  };
+
+  return [
+    props,
+    {
+      onMouseDown: (e: React.TouchEvent & React.MouseEvent) => {
+        e.persist();
+        e.stopPropagation();
```
mayank99, line 181: I see you added this. Is it to prevent text selection?

I also saw you added stopPropagation to the onClick of iui-resizer. Are both needed?

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+      headerIdWidths,
+      nextHeaderIdWidths,
+      clientX,
+    });
+  };
+
+  return [
+    props,
+    {
+      onMouseDown: (e: React.TouchEvent & React.MouseEvent) => {
```
mayank99, line 179: is this type correct? should only be MouseEvent i imagine

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+    flatHeaders,
+    getHooks,
+    state: { columnResizing },
+  } = instance;
+
+  const getInstance = useGetLatest(instance);
+
+  flatHeaders.forEach((header, index) => {
+    const resizeWidth = columnResizing.columnWidths[header.id];
+    header.width = resizeWidth || header.width || header.originalWidth;
```
mayank99, line 354: react-table had `originalWidth` before `width`. Wonder what's the impact

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+          dispatchEnd();
+        },
+      },
+    };
+
+    const events = isTouchEvent
+      ? handlersAndEvents.touch
+      : handlersAndEvents.mouse;
+    const passiveIfSupported = passiveEventSupported()
+      ? { passive: false }
```
mayank99, line 152: If we are just setting passive to false, then do we really need the utility? Isn't false default?

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -0,0 +1,440 @@
...
+
+    const deltaX = clientX - startX;
+
+    const newColumnWidths = getColumnWidths(
+      headerIdWidths,
+      deltaX / columnWidth,
+    );
+    const newNextColumnWidths = getColumnWidths(
+      nextHeaderIdWidths,
+      -deltaX / nextColumnWidth,
```
mayank99, line 266: From what I understand, this approach is to preserve the total width of the table.

What if the table is unbounded and needs to be resized beyond the screen?

src/core/Table/utils.ts
```diff
@@ -6,14 +6,20 @@ import { ColumnInstance } from 'react-table';
...
+    } else {
+      style.flex = `0 0 ${width}`;
```
mayank99, line 21: Would it be bad if we always set the flex to column width even if it's not resizable?

stories/core/Table.stories.tsx
mayank99: at least add `initial`. It is considered outdated UX practice for tooltip to follow cursor.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: > some browsers (specifically, Chrome and Firefox) have changed the default value of the passive option to true for the touchstart and touchmove events on the document-level nodes Window, Document, and Document.body
I guess this is for the touch events also found that `d3-drag` lib also says that passive event is not appropriate for the drag: https://github.com/d3/d3-drag/issues/35#issuecomment-314798648

src/core/Table/utils.ts
bentleyvk: Yup because we would need to always handle table resize in js.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: I think I did this on purpose, don't remember now. Because when resizing `width` and `originalWidth` will be different as `originalWidth` is the one set by the dev.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Devs can increase table width or ask us for resize mode 😀

stories/core/Table.stories.tsx
bentleyvk: Do you have a link for this guideline?

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Changed

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Changed

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Removed

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: It was a "hack" to omit wrapper column that is required for the table by accident. Changed to firstly look at the same parent if missing then at all columns.

src/core/Table/Table.tsx
bentleyvk: Added

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Added

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Added changes list.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: It is to prevent sort from firing when resizing columns. onClick is not needed, removed.

src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: Fixed types

src/core/Table/TableRowMemoized.tsx
bentleyvk: Idk, it was in Mayank's PR so I did the same 😀 We also do that in other components.

stories/core/Table.stories.tsx
bentleyvk: Handling last column with disabled resizing.

src/core/Table/utils.ts
bentleyvk: Changed

veekeys (approved):

stories/core/Table.stories.tsx
mayank99: Check our standards.

The guideline was updated in BWC standards first:
> The tooltip should be anchored to the item that is being hovered and should not follow the cursor as it moves.

And then refined in iTwinUI standards:
> It is anchored to the hovered item and fades out when the cursor moves away from the item.

bentleyvk: Removed for now.

src/core/Table/Table.test.tsx
```diff
@@ -1672,3 +1673,520 @@ it('should change page size', () => {
...
+  const { container } = renderComponent({
+    columns,
+    isResizable: true,
+  });
+
+  const rows = container.querySelectorAll('.iui-table-body .iui-row');
+  expect(rows.length).toBe(3);
+
+  const resizer = container.querySelector('.iui-resizer') as HTMLDivElement;
+  expect(resizer).toBeTruthy();
```
saaaaaally, line 1712: Maybe add test that there no `iui-resizer` when `isResizable` is false

bentleyvk: Added

saaaaaally (approved):

PR #451, opened 2021-12-02
feat(Tabs): Add `wrapperClassName` prop

### LINKED ISSUE #450: Tabs: allow setting className on wrapper
Currently there is no way to set className on tabs-wrapper. It is needed for layout purposes. Without a custom className, users need to override `.iui-tabs-wrapper` which is not nice.

Related discussions: #417, #420

mayank99:
Take two after #420 where I tried adding `className`. This time I'm adding `wrapperClassName` as it's more consistent with the other props, even if we remove it in 2.0.

Closes #450 

## Checklist

### DIFF
--- a/src/core/Tabs/Tabs.tsx
+++ b/src/core/Tabs/Tabs.tsx
@@ -58,6 +58,10 @@ export type TabsProps = {
    */
   contentClassName?: string;
   /**
+   * Custom CSS class name for the tabs wrapper.
+   */
+  wrapperClassName?: string;
+  /**
    * Content inside the tab panel.
    */
   children?: React.ReactNode;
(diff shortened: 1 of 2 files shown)


src/core/Tabs/Tabs.tsx
```diff
@@ -57,6 +57,10 @@ export type TabsProps = {
    * Custom CSS class name for tab panel.
    */
   contentClassName?: string;
+  /**
+   * Custom CSS class name for the tabs wrapper.
+   */
+  wrapperClassName?: string;
```
bentleyvk, line 63: Hm, will users will understand what "wrapper" stand for. What if we name this prop "containerClassName"?

mayank99: wrapper and container kinda mean different things though. container implies that it contains multiple instances of something. That's why we have `TagContainer`

bentleyvk (approved):

PR #455, opened 2021-12-03
fix(utils): MiddleTextTruncation responds to resize

### LINKED ISSUE #454: MiddleTextTruncation flex-grow bug
MiddleTextTruncation does not grow back when screen is resized from small to larger.

gretanausedaite:
MiddleTextTruncation now responds to resize.
Closes: #454

### DIFF
--- a/src/core/utils/components/MiddleTextTruncation.tsx
+++ b/src/core/utils/components/MiddleTextTruncation.tsx
@@ -47,6 +47,7 @@ export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
       style={{
         display: 'flex',
         minWidth: 0,
+        flexGrow: 1,
         whiteSpace: 'nowrap',
         ...style,
       }}


src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -47,6 +47,7 @@ export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
       style={{
         display: 'flex',
         minWidth: 0,
+        flexGrow: 1,
```
mayank99, line 50: @veekeys Can you verify this is fine, since you implemented this component?

mayank99 (approved):
Looks good. PR title can be improved for changelog.

src/core/utils/components/MiddleTextTruncation.tsx
veekeys: Yeah, it is needed.. It is weird, why I dont have it, cause i was sure I added it. But it worked well without it in select examples somehow though... Might be the case only with the elements size change, though..

PR #458, opened 2021-12-03
chore: Add lint-staged to husky hook

### LINKED ISSUE #442: Lint only staged files on commit to reduce commit time

mayank99:
Added `lint-staged` v12 and upgraded `husky` to v4 (didn't do v7 because it's like a completely different library).
Autofixing on commit seems to be working but I can't tell if the commit time is faster. 🤷‍♀️

Closes #442

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "eslint-plugin-react": "^7.25.1",
     "eslint-plugin-react-hooks": "^4.2.0",
     "fast-glob": "^3.2.5",
-    "husky": "4.0.0",
+    "husky": "^4.0.0",
     "inquirer": "^6.2.2",
     "jest": "^27.0.6",
     "jest-cli": "^27.0.6",
--- a/package.json
+++ b/package.json
@@ -81,11 +81,12 @@
     "eslint-plugin-react": "^7.25.1",
     "eslint-plugin-react-hooks": "^4.2.0",
     "fast-glob": "^3.2.5",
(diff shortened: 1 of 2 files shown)


mayank99:
I'm not sure why `yarn lint` is failing now. I didn't touch eslint at all. Will need to investigate.

bentleyvk:
Huh?
image

package.json
```diff
@@ -81,14 +81,15 @@
...
+    "husky": "^4.0.0",
     "inquirer": "^6.2.2",
     "jest": "^27.0.6",
     "jest-cli": "^27.0.6",
     "jest-junit": "^12.0.0",
+    "lint-staged": "^12.1.2",
     "markdown-to-jsx": "6.11.4",
     "mkdirp": "^1.0.4",
-    "prettier": "^2.2.1",
+    "prettier": "^2.5.0",
```
bentleyvk, line 92: Can you try reverting prettier version changes and checking whether build passes? Maybe they did some breaking changes or added some rules that we need to override.

mayank99: Bingo, that fixed it

bentleyvk: Can you check what has been changed if we need to update to the newest version for some reason?

mayank99: I'm not sure. I tried to check their release notes but didn't see anything https://prettier.io/blog/

bentleyvk (approved):

PR #468, opened 2021-12-08
feat: Add Tree Component

### LINKED ISSUE #475: Tree Component

saaaaaally:
Closes #475

image

## Checklist


src/core/Tree/Tree.tsx
```diff
@@ -3,9 +3,21 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
-import { TreeNode } from './TreeNode';
 import { CommonProps, useTheme } from '../utils';
 import 'D:/itwinUI/iTwinUI/lib/css/tree.css'; //'@itwin/itwinui-css/css/tree.css';
```
bentleyvk, line 7: You can now import from iTwinUI.

src/core/Tree/TreeNode.tsx
```diff
@@ -88,24 +68,34 @@ export const TreeNode = (props: TreeNodeProps) => {
...
+          'iui-active': isSelected,
           'iui-disabled': isDisabled,
           className,
         })}
         style={style_level}
-        onClick={() => {
-          setActive(!active);
-          onSelected?.();
+        onClick={(e) => {
+          if (e.currentTarget != e.target) {
```
bentleyvk, line 87: Why you need this check?

src/core/Tree/Tree.tsx
```diff
@@ -30,31 +42,30 @@ export type TreeProps = {
...
-              isDisabled={node.props['isDisabled']}
-            >
-              {node.props['children']}
-            </TreeNode>
-          ) : (
-            node
-          ),
-        )}
+    <TreeContext.Provider
+      value={{
```
bentleyvk, line 58: You can just write:
```tsx
value={{
  selectedNode,
  setSelectedNode,
  onNodeSelected,
  showCheckboxes,
}}
```

src/core/Tree/TreeNode.tsx
```diff
@@ -64,17 +48,13 @@ export type TreeNodeProps = {
  */
 export const TreeNode = (props: TreeNodeProps) => {
   const {
-    title,
-    caption,
-    isExpanded = false,
-    isActive,
+    label,
+    sublabel,
     depthLevel = 0,
```
bentleyvk, line 53: Maybe `depthLevel` can also be moved to the Context? What do you think @mayank99 @veekeys ?

src/core/Tree/TreeNode.tsx
```diff
@@ -88,24 +68,34 @@ export const TreeNode = (props: TreeNodeProps) => {
...
-          onSelected?.();
+        onClick={(e) => {
+          if (e.currentTarget != e.target) {
+            return;
+          }
+          if (isSelected) {
+            context?.setSelectedNode('');
+          } else if (!isDisabled) {
+            context?.setSelectedNode(label);
+            context?.onNodeSelected?.();
```
bentleyvk, line 94: I think we need to pass something to `onNodeSelect` because people might want to know what node was selected e.g. to highlight element in a viewer. And I start to think that `onNodeSelected` should go into `TreeNode` as a prop or we need to have some id prop like Material UI:
On tree:
image
On node:
image

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+      >
+        {context?.showCheckboxes && (
+          <Checkbox className='iui-tree-node-checkbox' disabled={isDisabled} />
+        )}
+        <div className='iui-tree-node-content'>
+          {children && (
+            <IconButton
+              styleType='borderless'
+              size='small'
+              onClick={() => {
```
bentleyvk, line 106: We might also need `onExpanded` callback prop.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+    children,
+    style,
+    className,
+    icon,
+    isDisabled = false,
+    ...rest
+  } = props;
+  useTheme();
+
+  const style_level = React.useMemo(
```
bentleyvk, line 63: We are using camelCase.

src/core/Tree/TreeNode.tsx
bentleyvk: @mayank99 @veekeys Would be great to get your input on this one too.

bentleyvk:
We might need to check iModel.js implementation. We shouldn't implement everything they have on our initial Tree component but we should at least create issues for the missing stuff.

src/core/Tree/TreeNode.tsx
mayank99: Could just move it to the individual nodes.

src/core/Tree/TreeNode.tsx
bentleyvk: So wrap TreeNode children with context and it will override parent context.

grigasp:
@roluk please review

src/core/Tree/Tree.tsx
saaaaaally: Merged

src/core/Tree/TreeNode.tsx
saaaaaally: This check prevents the node from being selected when clicking on the expander button (without this it was expanding and selecting when clicking the button). However, I noticed that now this doesn't select the node when clicking over the label/sublabel so I will need to rethink this

src/core/Tree/Tree.tsx
saaaaaally: Done

src/core/Tree/TreeNode.tsx
saaaaaally: Added `onNodeExpanded`

src/core/Tree/TreeNode.tsx
saaaaaally: Fixed

src/core/Tree/TreeNode.tsx
saaaaaally: Added `nodeDepth` to context

src/core/Tree/TreeNode.tsx
bentleyvk: Oh, in that case you need to add `e.stopPropagation()` in expander `onClick` handler.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <div className='iui-tree-node-content-title'>{label}</div>
+            <div className='iui-tree-node-content-caption'>{sublabel}</div>
+          </span>
+        </div>
+      </div>
+      {children && expanded && (
+        <ul className='iui-sub-tree' role='group'>
+          <TreeContext.Provider
+            value={{
+              selectedNode: context?.selectedNode,
```
bentleyvk, line 141: You can write like this instead:
```tsx
value={{
  ...context,
  nodeDepth: currentDepth + 1
}}
```

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,158 @@
...
+              setSelectedNode: context?.setSelectedNode,
+              onNodeSelected: context?.onNodeSelected,
+              onNodeExpanded: context?.onNodeExpanded,
+              checkbox: context?.checkbox,
+              onNodeCheckboxSelected: context?.onNodeCheckboxSelected,
+              nodeDepth: currentDepth + 1,
+            }}
+          >
+            {React.Children.map(children, (node) => node)}
```
bentleyvk, line 150: You no longer need to map as you don't provide any props or anything.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onNodeExpanded?: () => void;
+  /**
+   * Callback fired when checking a TreeNode's checkbox.
+   */
+  onNodeCheckboxSelected?: () => void;
+  /**
+   * Checkbox to be shown before TreeNodes, if undefined checkbox will not be shown.
+   * Recommended to use Checkbox component.
+   */
+  checkbox?: React.ReactNode;
```
bentleyvk, line 42: If we decide to move all callbacks to each node props then we might want to move checkbox too because that way users can handle all the checkbox callbacks and checked status themselves.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,158 @@
...
+            React.cloneElement(icon, {
+              className: cx('iui-tree-node-content-icon', icon.props.className),
+            })}
+          <span className='iui-tree-node-content-label'>
+            <div className='iui-tree-node-content-title'>{label}</div>
+            <div className='iui-tree-node-content-caption'>{sublabel}</div>
+          </span>
+        </div>
+      </div>
+      {children && expanded && (
```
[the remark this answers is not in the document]
bentleyvk: Would you recommend having a flat list? Because we were fallowing W3 guidelines for having nested structure with all required roles. Also, Material UI is having the same. https://www.w3.org/TR/wai-aria-practices-1.2/#TreeView

src/core/Tree/TreeNode.tsx
mayank99: I think it might be better to use a separate context for this

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
roluk: A flat list is much easier to virtualize, but list virtualization always comes at the cost of accessibility. I cannot find any resources that talk about virtualized list accessibility, and I expect all accessibility tools to have poor support for this pattern. I raised this point because I saw virtualization is being implemented for the table component.

If iTwinUI team chose not to support virtualization in any of its components, I would consider it as a big win for all end users, because scrolling an enormous list with a tiny scrollbar is always a horrible experience. Absence of out-of-the-box virtualization may lead developers to rethink data presentation strategy and come up with something that has friendlier UX.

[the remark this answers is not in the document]
veekeys: @roluk  that is a really good comment, thank you! but.. We already have virtualized trees used in our itwin.js applications to show those complicated data structures and I do not think they would replace that UI into something else. But we also want our tree to be used there as we take care of visual issues which are present currently and ensure quality.
We will have virtualization in at least table, I think, cause this is waited and required. Even though after your comment, we might rethink something and try to see if people are up to using pagination :D
So not sure, what would be ok way to go here :/

src/core/Tree/Tree.tsx
saaaaaally: So you would have to pass in checkbox and onNodeCheckboxSelected into each TreeNode?

src/core/Tree/TreeNode.tsx
veekeys: Yeah, depends.. If we still going to end up having ids (or some other identifiers) over nodes, having a callback on Tree component might make sense.

src/core/Tree/TreeNode.tsx
veekeys: I think depth in context might be alright.. not sure, maybe technically better have separate, but..if it works, maybe not a big deal..

src/core/Tree/TreeNode.tsx
veekeys: it does not make sense now empty callbacks on `Tree` component... they give no info at all to the user, so whats the point of having them like that? or its not finished implementation?

saaaaaally: Yes, it is not finished implementation yet. 
I added `Array<string>` to `onNodeSelected` callback so now you can have a list of the currently selected nodeIds, this allows to have single/multi selection of nodes. If we like this approach I can do something similar for checkboxes/expansion to have list of which nodes are checked/expanded.

src/core/Tree/TreeNode.tsx
saaaaaally: fixed

src/core/Tree/TreeNode.tsx
saaaaaally: Removed

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+type FlatNode = {
+  data: TreeData;
+  id: string;
+  depth: number;
+  subNodeIds: string[];
+  parent?: FlatNode;
+};
+
+export type TreeProps = {
```
bentleyvk, line 43: I think Saally you will hate me for this 😀 I am thinking maybe we could get best of what others have by having a prop `getNode: (node: TreeData, index: number) => TreeData` and we could call in in `flatNodesList`. This way users could set `isExpanded` and `isSelected` whatever way they want: have separate lists where expanded node ids are saved or saving inside their node data properties. It may also work for Robertas too e.g. passing `data={new Array(500)}` and then according `getNode`.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,228 @@
...
+        const nodeWrapper: FlatNode = {
+          id,
+          data: node,
+          depth,
+          subNodeIds: [],
+          parent: parent ?? undefined,
+        };
+        flatList.push(nodeWrapper);
+        nodeIdList.push(id);
+        if (node.subnodes && node.subnodes.length) {
```
bentleyvk, line 189: If node is collapsed, you don't need to add its subnodes to the flat list.
If you add `getNode` then this should be easy.

src/core/Tree/Tree.tsx
bentleyvk: @mayank99 @veekeys would be nice to hear your opinion.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,228 @@
...
+      setExpandedNodes?: (
+        node: Array<string> | ((prev: Array<string>) => Array<string>),
+      ) => void;
+      nodeDepth?: number;
+      parentNode?: FlatNode;
+    }
+  | undefined
+>(undefined);
+
+export type TreeData = {
```
bentleyvk, line 26: If we implement `getNode`, I am thinking we could change typings to this:
```suggestion
export type TreeData<T> = T & {
  subnodes?: Array<TreeData<T>>;
};

export type FlatNode<T> = {
  data: TreeData<T>;
  id: string;
  depth: number;
  subNodeIds: string[];
  parent?: FlatNode<T>; // still not sure about this
  isExpanded: boolean;
  isDisabled: boolean;
};
```

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Checkbox to be shown before TreeNode.
+   * If undefined checkbox will not be shown.
+   * Recommended to use Checkbox component.
+   */
+  nodeCheckbox?: React.ReactNode;
+  /**
+   * Callback fired when checking a TreeNode's checkbox.
+   */
+  onNodeCheckboxSelected?: (nodeId: string, checked: boolean) => void;
```
bentleyvk, line 57: This callback is not needed as users pass checkbox themselves therefore they can have `onChanged` props on it.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+export type TreeNodeProps = {
+  /**
+   * Unique id of the node.
+   */
+  nodeId: string;
+  /**
+   * The main text displayed on the node.
+   */
+  label?: React.ReactNode;
```
bentleyvk, line 21: Shouldn't this be mandatory?

src/core/Tree/Tree.tsx
bentleyvk: Also, maybe rename to `NodeData` or `TreeNodeData`

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,228 @@
...
+          data: node,
+          depth,
+          subNodeIds: [],
+          parent: parent ?? undefined,
+        };
+        flatList.push(nodeWrapper);
+        nodeIdList.push(id);
+        if (node.subnodes && node.subnodes.length) {
+          const subNodeIds = flatNodes(node.subnodes, nodeWrapper, depth + 1);
+          nodeWrapper.subNodeIds = subNodeIds;
```
bentleyvk, line 191: Not sure anymore whether we should generate `subNodeIds` here.

bentleyvk: Oh, I see you are generating them inside `TreeNode`, so yeah, you can remove it from here.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,257 @@
...
+              },
+              className,
+            )}
+            style={styleLevel}
+            onClick={() => onNodeClick()}
+            tabIndex={isSelected ? 0 : -1}
+          >
+            {nodeCheckbox && React.isValidElement(nodeCheckbox)
+              ? React.cloneElement(nodeCheckbox, {
+                  className: 'iui-tree-node-checkbox',
```
bentleyvk, line 205: I think we should only apply the class name and leave the rest to the user. Also, you need to merge `className` with its original one, check `IconButton how to do that.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    'iui-tree-node-content-icon',
+                    icon.props.className,
+                  ),
+                })}
+              <span className='iui-tree-node-content-label'>
+                <div className='iui-tree-node-content-title'>{label}</div>
+                <div className='iui-tree-node-content-caption'>{sublabel}</div>
+              </span>
+            </div>
+            {children}
```
bentleyvk, line 242: Maybe we should move `children` inside `iui-tree-node-content` div.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,228 @@
...
+    switch (event.key) {
+      case 'ArrowDown': {
+        newIndex = currentIndex + 1;
+        break;
+      }
+      case 'ArrowUp': {
+        newIndex = currentIndex - 1;
+        break;
+      }
+      case 'ArrowLeft':
```
bentleyvk, line 128: Hm, Maybe we can handle left and right like in `MenuItem`: call `onExpand` if it has subrows, if not we can focus its parent. Maybe we can use parent `nodeId` to focus it.

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+  component: Tree,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/Tree',
+} as Meta<TreeProps>;
+
+export const Basic: Story<TreeProps> = (args) => {
+  const onSelectedNodeChange = (nodeId: string, selected: boolean) => {
```
bentleyvk, line 22: Wrap with `useCallback`

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,199 @@
...
+        .fill(null)
+        .map((_, index) => generateItem(index)),
+    [generateItem],
+  );
+
+  return (
+    <Tree
+      data={data}
+      nodeRenderer={(props) => (
+        <TreeNode
```
bentleyvk, line 85: I feel like it would be nice to show an example with `React.memo` to prevent rerendering on each action. But maybe after everything is cleared up we can help you with that.

src/core/Tree/TreeNode.tsx
saaaaaally: Removed

src/core/Tree/TreeNode.tsx
saaaaaally: Made prop required

src/core/Tree/Tree.tsx
saaaaaally: Removed

src/core/Tree/TreeNode.tsx
saaaaaally: Fixed

src/core/Tree/TreeNode.tsx
saaaaaally: Moved

stories/core/Tree.stories.tsx
saaaaaally: Added `useCallback`

src/core/Tree/Tree.tsx
mayank99: I find the current api already confusing. There is `NodeData` and `FlatNode`, which seem to be inexplicably separated out.

If `getNode` is added, it would create more confusion with `nodeRenderer` which already has some handlers. How does `getNode` help exactly?

I think it would make sense to separate out the state from the UI and then expose two props: `initialState` and `stateReducer`. And we can keep the `nodeRenderer` but it should only be used for the UI, and not have any handlers.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,225 @@
...
+  depth: number;
+  subNodeIds: string[];
+  parent?: FlatNode;
+};
+
+export type TreeProps = {
+  /**
+   * Node renderer.
+   */
+  nodeRenderer?: (props: NodeData) => JSX.Element;
```
gretanausedaite, line 47: If node rendered is not provided by user, nodes are not being displayed. Maybe make it mandatory parameter or give it default value?

src/core/Tree/Tree.tsx
bentleyvk: But how users are going to pass extended nodes to `initialState`? Than we might have to make `id` prop required in their data.

saaaaaally: I added `getNode` to Tree props, so user can set isExpanded. I also combined `NodeData` and `FlatNode` since I think both are not needed anymore.

src/core/Tree/Tree.tsx
saaaaaally: Subnodes of collapsed nodes are no longer added to flat list.

src/core/Tree/Tree.tsx
saaaaaally: Added default value for nodeRenderer

src/core/Tree/Tree.tsx
saaaaaally: I changed left to either collapse node or go to parent node if already collapsed

src/core/Tree/Tree.tsx
gretanausedaite: Current default nodeRenderer is not functional at all, user needs to provide their own renderer if they want ANY functions (expand, select, etc). My idea for default renderer was to have it same as in Basic story example so that story wouldn't have any renderer provided and add another story with "custom" renderer (eg. visibility checkbox added). Let me know if this makes sense.

saaaaaally: node renderer is now a required prop

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,214 @@
...
+        nodeIdList.push(flatNode.nodeId);
+        flatList.push(flatNode);
+        flatNode.depth = depth;
+        if (flatNode.isExpanded) {
+          const subNodeIds = flatNodes(flatNode.subNodes, depth + 1);
+          flatNode.subNodeIds = subNodeIds;
+        } else {
+          const subNodeIds = Array<string>();
+          flatNode.subNodes?.forEach((subNode) => {
+            subNodeIds.push(subNode.id);
```
saaaaaally, line 179: Build is failing because here  Property 'id' does not exist on type 'T'. These subnode ids are needed for the `aria-owns` label on each parent node. Not sure if I should just call `getNode` on all of the subnodes to get `nodeId` (even if they should be collapsed)

bentleyvk: Aren't you getting subNodeIds from this part?
image

saaaaaally: Yes only for expanded parent nodes, the else covers collapsed nodes

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,214 @@
...
+export const TreeContext = React.createContext<
+  | {
+      nodeDepth: number;
+      subNodeIds?: string[];
+    }
+  | undefined
+>(undefined);
+
+export type NodeData<T> = {
+  subNodes?: Array<T>;
```
bentleyvk, line 19: To support lazy loading of sub-nodes, we need a flag that tells whether node has any subnodes like `hasSubnodes` or `isLeaf`.

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+export default {
+  component: Tree,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/Tree',
+} as Meta<TreeProps<TreeData>>;
+
+type TreeData = {
```
bentleyvk, line 21: Maybe rename it to `StoryData` or smth, because at first I thought it is some data type from the Tree. And move it inside the story, so users could see it in code tab.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+  isExpanded?: boolean;
+  /**
+   * Is TreeNode selected.
+   * @default false
+   */
+  isSelected?: boolean;
+  /**
+   * Callback fired when expanding or closing a TreeNode.
+   */
+  onNodeExpanded?: (nodeId: string, expanded: boolean) => void;
```
veekeys, line 48: I am just thinking.. As we are not controlling state at all, it means user actually needs to do something to have Tree working, i.e. expanding node. I do not think it might make sense Tree without expansion. 
Following this, shoulld this prop be required then?

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+   */
+  nodeCheckbox?: React.ReactNode;
+  /**
+   * Content shown after TreeNode.
+   */
+  children?: React.ReactNode;
+} & CommonProps;
+
+/**
+ * TreeNode component to display within a Tree.
```
veekeys, line 66: Would prefer to have more detailed description and example below (maybe together with Tree component or something more understandable than just repeatance of same props name). Description could consist of explanation, how Tree works, that we expect all the state to be controlled by user and some other important points. Maybe try to put that info in the way, that user would be able to use Tree with TreeNode without actually looking into Storybook. 
Just need to separate that info between TreeNode and Tree.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+   * Checkbox to be shown before TreeNode.
+   * If undefined checkbox will not be shown.
+   * Recommended to use Checkbox component.
+   */
+  nodeCheckbox?: React.ReactNode;
+  /**
+   * Content shown after TreeNode.
+   */
+  children?: React.ReactNode;
+} & CommonProps;
```
veekeys, line 63: Does `title` prop work from `CommonProps` ?

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+    nodeCheckbox,
+    ...rest
+  } = props;
+  useTheme();
+
+  const context = React.useContext(TreeContext);
+  const nodeDepth = context?.nodeDepth ?? 0;
+  const subNodeIds = context?.subNodeIds ?? [];
+
+  const styleLevel = React.useMemo(
```
veekeys, line 106: Maybe rename, so it could be clear that this is setting style for depth?

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+   */
+  isSelected?: boolean;
+  /**
+   * Callback fired when expanding or closing a TreeNode.
+   */
+  onNodeExpanded?: (nodeId: string, expanded: boolean) => void;
+  /**
+   * Callback fired when selecting a TreeNode.
+   */
+  onNodeSelected?: (nodeId: string, selected: boolean) => void;
```
veekeys, line 52: variable could be `isSelected` same with expanded above

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+   * Is TreeNode selected.
+   * @default false
+   */
+  isSelected?: boolean;
+  /**
+   * Callback fired when expanding or closing a TreeNode.
+   */
+  onNodeExpanded?: (nodeId: string, expanded: boolean) => void;
+  /**
+   * Callback fired when selecting a TreeNode.
```
veekeys, line 50: Could add that it gives node id and boolean if node is selected

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+            className={cx(
+              'iui-tree-node',
+              {
+                'iui-active': isSelected,
+                'iui-disabled': isDisabled,
+              },
+              className,
+            )}
+            style={styleLevel}
+            onClick={() => onNodeClick()}
```
veekeys, line 135: no need inline func, can have `onClick={onNodeClick}`

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+            tabIndex={isSelected ? 0 : -1}
+          >
+            {nodeCheckbox && React.isValidElement(nodeCheckbox)
+              ? React.cloneElement(nodeCheckbox, {
+                  className: cx(
+                    'iui-tree-node-checkbox',
+                    nodeCheckbox.props.className,
+                  ),
+                })
+              : nodeCheckbox}
```
veekeys, line 145: What is the case, when `nodeCheckbox` is not a valid element, but we still want to display it?

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <div className='iui-tree-node-content'>
+              {context?.hasSubNodes && (
+                <IconButton
+                  styleType='borderless'
+                  size='small'
+                  onClick={(e) => {
+                    onNodeExpanded?.(nodeId, !isExpanded);
+                    e.stopPropagation();
```
veekeys, line 153: what it fixes?

src/core/Tree/index.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+export { Tree } from './Tree';
+export type { TreeProps } from './Tree';
+export default './Tree';
```
veekeys, line 7: Not sure if default export is needed here.. Something useful it gives in this case?

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { CommonProps, useTheme } from '../utils';
+import '@itwin/itwinui-css/css/tree.css';
+import cx from 'classnames';
+
+export const TreeContext = React.createContext<
```
veekeys, line 10: In `ColorPicker` you had it in separate file.. I see MenuItem has it inside file, but its not exported anywhere. So need to align.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+type FlatNode<T> = {
+  nodeData: NodeData<T>;
+  depth: number;
+  subNodeIds?: Array<string>;
+};
+
+export type TreeProps<T> = {
+  /**
+   * Custom renderer for tree nodes, recommended to use TreeNode component.
```
veekeys, line 37: If you wrap `TreeNode` with ` would be better readability

src/core/Tree/Tree.tsx
saaaaaally: Added hasSubnodes

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+   */
+  nodeRenderer: (props: NodeData<T>) => JSX.Element;
+  /**
+   * Items inside tree.
+   */
+  data: T[];
+  /**
+   * Get the NodeData.
```
veekeys, line 45: I guess need better explanation here too. Users do now know what is that `NodeData` (wrap it with ` too). So could expand that they are going to get their passed data instance and need to map into custom interface object, which will be used to render a tree node in `nodeRenderer`. Just all of that in better english :D

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+   */
+  getNode: (node: T) => NodeData<T>;
+} & CommonProps;
+
+/**
+ * Tree
```
veekeys, line 51: We can do better than a single word, cant we

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+  nodeRenderer: (props: NodeData<T>) => JSX.Element;
+  /**
+   * Items inside tree.
+   */
+  data: T[];
+  /**
+   * Get the NodeData.
+   */
+  getNode: (node: T) => NodeData<T>;
+} & CommonProps;
```
veekeys, line 48: and here might need to omit that `title` from props..

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const flatNodes = (nodes: T[] = [], depth = 0) => {
+      const nodeIdList = Array<string>();
+      nodes.forEach((element) => {
+        const flatNode: FlatNode<T> = {
+          nodeData: getNode(element),
+          depth: depth,
+        };
+        nodeIdList.push(flatNode.nodeData.nodeId);
+        flatList.push(flatNode);
+        flatNode.depth = depth;
```
veekeys, line 180: you are assigning depth twice

src/core/Tree/TreeNode.tsx
saaaaaally: Yes I think so, should it be omitted?

src/core/Tree/TreeNode.tsx
saaaaaally: It prevents calling `onNodeClick` and selecting a node when expanding/collapsing a node,

src/core/Tree/index.ts
saaaaaally: Not sure, its what is added by default when running the `CreateComponent` script

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+export const Tree = <T,>(props: TreeProps<T>) => {
+  const { data, className, nodeRenderer, getNode, ...rest } = props;
+  useTheme();
+
+  const treeRef = React.useRef<HTMLUListElement>(null);
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLUListElement>) => {
+    const nodes: Array<HTMLElement> = Array.from(
+      treeRef.current?.querySelectorAll('.iui-tree-node:not(.iui-disabled)') ||
```
veekeys, line 83: Have you checked key nav in `Menu` component for example? Could you try to rewrite something similar to there to avoid css selectors as much as possible?

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+    [generateItem],
+  );
+
+  const getNode = React.useCallback(
+    (node: StoryData): NodeData<StoryData> => {
+      return {
+        subNodes: node.subItems,
+        nodeId: node.id,
+        node: node,
+        isExpanded: expandedNodes.findIndex((id) => id === node.id) != -1,
```
veekeys, line 108: another option would be `expandedNotes.some((id) => id === node.id)`

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  return (
+    <Tree<StoryData>
+      data={data}
+      getNode={getNode}
+      nodeRenderer={React.useCallback(
+        ({ node, ...rest }) => (
+          <TreeNode
+            label={node.label}
+            sublabel={node.subLabel}
```
veekeys, line 125: so `sublabel` or `subLabel`. I guess first one?

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+    subItems: StoryData[];
+  };
+
+  const [selectedNodes, setSelectedNodes] = React.useState([
+    'Node 0',
+    'Node 3.2',
+    'Node 22',
+  ]);
+  const onSelectedNodeChange = React.useCallback(
+    (nodeId: string, selected: boolean) => {
```
veekeys, line 35: I guess its css bug, but I really dont like that when you click it gets focus style and released it dissapears... It is really weird and not the first component I noticed.
Also onclick it stays focused so if you have 2 selected nodes near, it gets this double border or smth..

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+  />
+ */
+
+export const Tree = <T,>(props: TreeProps<T>) => {
+  const { data, className, nodeRenderer, getNode, ...rest } = props;
+  useTheme();
+
+  const treeRef = React.useRef<HTMLUListElement>(null);
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLUListElement>) => {
```
veekeys, line 81: so if I started to navigate with keyboard in tree and then press tab, it gets me to the neares next checkbox. But how to get back to the tree navigation? Need to review keyboard..

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+              'iui-tree-node',
+              {
+                'iui-active': isSelected,
+                'iui-disabled': isDisabled,
+              },
+              className,
+            )}
+            style={styleLevel}
+            onClick={() => onNodeClick()}
+            tabIndex={isSelected ? 0 : -1}
```
veekeys, line 136: dbl click should expand row? :D same as table

src/core/Tree/TreeNode.tsx
veekeys: If it shows something, I guess we can leave it

src/core/Tree/TreeNode.tsx
saaaaaally: Renamed to `styleDepth`

src/core/Tree/TreeNode.tsx
saaaaaally: Renamed to `isSelected` and `isExpanded`

src/core/Tree/TreeNode.tsx
saaaaaally: Changed

src/core/Tree/TreeNode.tsx
saaaaaally: I guess there is no case for that, changed to `undefined` instead

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,192 @@
...
+              {icon &&
+                React.cloneElement(icon, {
+                  className: cx(
+                    'iui-tree-node-content-icon',
+                    icon.props.className,
+                  ),
+                })}
+              <span className='iui-tree-node-content-label'>
+                <div className='iui-tree-node-content-title'>{label}</div>
+                <div className='iui-tree-node-content-caption'>{sublabel}</div>
```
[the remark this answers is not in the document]
saaaaaally: Changed

src/core/Tree/Tree.tsx
saaaaaally: Not sure why `ColorPicker` has it in a separate file but I think that was a special circumstance. But `TreeContext` does need to be exported so that it can be used in `TreeNode`

src/core/Tree/Tree.tsx
saaaaaally: Done

src/core/Tree/Tree.tsx
saaaaaally: Removed

stories/core/Tree.stories.tsx
saaaaaally: changed

stories/core/Tree.stories.tsx
saaaaaally: Changed to sublabel

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,215 @@
...
+export const TreeContext = React.createContext<
+  | {
+      nodeDepth: number;
+      subNodeIds?: string[];
+      hasSubNodes: boolean;
+    }
+  | undefined
+>(undefined);
+
+export type NodeData<T> = {
```
bentleyvk, line 19: Add comments to each property so users would know what each of them means.

src/core/Tree/TreeNode.tsx
bentleyvk: I think it is not a valid element when text is passed.

src/core/Tree/TreeNode.tsx
saaaaaally: Updated both `Tree` and `TreeNode` descriptions

src/core/Tree/TreeNode.tsx
saaaaaally: Added

src/core/Tree/TreeNode.tsx
saaaaaally: So should I change it back so it will show text if `nodeCheckbox` is not a valid element?

src/core/Tree/Tree.tsx
saaaaaally: Done

src/core/Tree/Tree.tsx
saaaaaally: Updated these comments

src/core/Tree/Tree.tsx
saaaaaally: Added comment

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Get the NodeData.
+   */
+  getNode: (node: T) => NodeData<T>;
+} & CommonProps;
+
+/**
+ * Tree
+ * @example
+  <Tree<TreeData>
```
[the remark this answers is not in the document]
saaaaaally: Updated code example

src/core/Tree/Tree.tsx
saaaaaally: Added comments

src/core/Tree/Tree.test.tsx
```diff
@@ -0,0 +1,399 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+
+import { Tree } from './Tree';
+import TreeNode from './TreeNode';
+
```
bentleyvk, line 10: Need to update and fix tests

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const onNodeClick = () => {
+    if (isDisabled) {
+      return;
+    }
+    onNodeSelected?.(nodeId, !isSelected);
+  };
+
+  return (
+    <>
+      <li role='treeitem' id={nodeId} aria-expanded={isExpanded} {...rest}>
```
bentleyvk, line 126: `aria-disabled` missing

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,139 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import { Tree, TreeProps, TreeNode, Checkbox } from '../../src/core';
+import { action } from '@storybook/addon-actions';
+import { SvgPlaceholder } from '@itwin/itwinui-icons-react';
+import { NodeData } from '../../src/core/Tree/Tree';
+
+export default {
```
bentleyvk, line 12: Add 
image
This hides warning message in story.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,269 @@
...
+      className={cx('iui-tree', className)}
+      role='tree'
+      onKeyDown={handleKeyDown}
+      ref={treeRef}
+      {...rest}
+    >
+      {flatNodesList.map((flatNode) => (
+        <TreeContext.Provider
+          key={flatNode.nodeData.nodeId}
+          value={{
```
bentleyvk, line 256: There is errors in console:
image
We are passing these using context but they are still in `nodeRenderer` props and in story we use `...rest` to pass remaining props.
@veekeys  @mayank99 Should we just enlist all props in our stories or should we do a hack like in Table:
image

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,269 @@
...
+import React from 'react';
+import { CommonProps, useTheme } from '../utils';
+import '@itwin/itwinui-css/css/tree.css';
+import cx from 'classnames';
+
+export const TreeContext = React.createContext<
+  | {
+      nodeDepth: number;
+      subNodeIds?: string[];
+      hasSubNodes: boolean;
```
bentleyvk, line 14: I don't think this one is needed in context as it is in `NodeData`

src/core/Tree/Tree.tsx
veekeys: wait, why rest gives errors? We pass rest on TreeNode component, which has those props, no?

src/core/Tree/TreeNode.tsx
veekeys: I dont think passing text could be taken as node checkbox... Opinions?

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,199 @@
...
+   */
+  nodeCheckbox?: React.ReactNode;
+  /**
+   * Content shown after TreeNode.
+   */
+  children?: React.ReactNode;
+} & CommonProps;
+
+/**
+ * TreeNode component to display node content within a Tree.
```
veekeys, line 68: COuld you wrap components again into ` ?

src/core/Tree/Tree.tsx
bentleyvk: It doesn't have these 2 currently.

src/core/Tree/TreeNode.tsx
bentleyvk: Should we care what user passes?

veekeys: amm.. I guess yes, if we want Tree to work as expected :D or we can render that string I guess if nothing will break

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,199 @@
...
+            <div className='iui-tree-node-content'>
+              {context?.hasSubNodes && (
+                <IconButton
+                  styleType='borderless'
+                  size='small'
+                  onClick={(e) => {
+                    onNodeExpanded(nodeId, !isExpanded);
+                    e.stopPropagation();
+                  }}
+                  disabled={isDisabled}
```
bentleyvk, line 158: People are asking from time to time how can they enable expander or checkbox in Table when row is disabled. I see it might be the same situation. Solution could be by making TreeNode subcomponents for expander, icon and content. What do you think @mayank99 @veekeys

veekeys: I guess we need to think about it but not necessarily implement it now. Just have an option to extend if needed..
Or you think we need it now cause later might be considered as breaking change?

bentleyvk: I don't think we could do that in the future easily as we already show expander when `hasSubrows` is true and it also will be confusing that there is props and some subcomponents.

src/core/Tree/Tree.test.tsx
saaaaaally: Updated tests

src/core/Tree/TreeNode.tsx
mayank99: > TreeNode subcomponents for expander, icon and content

Just expander and content should be enough because icon is really part of content

src/core/Tree/Tree.tsx
mayank99: Why are they in `nodeRenderer` if they're passed using context?

I don't think we can expect users to manually list those props.

src/core/Tree/TreeNode.tsx
saaaaaally: Added

stories/core/Tree.stories.tsx
saaaaaally: Added

src/core/Tree/Tree.tsx
saaaaaally: Removed `hasSubNodes` from context but I had to add it to `TreeNode` props

src/core/Tree/TreeNode.tsx
saaaaaally: Done

stories/core/Tree.stories.tsx
saaaaaally: Focus bug should be fixed with this pr https://github.com/iTwin/iTwinUI/pull/456

src/core/Tree/Tree.tsx
saaaaaally: I added `hasSubNodes` as prop to `TreeNode` so that warning is gone now. 
But I'm not sure how to fix the warning with `subNodes` since it is not needed as a prop in `TreeNode`

src/core/Tree/Tree.tsx
saaaaaally: Updated keyboard navigation to use `getFocusableElements` instead of css selectors

src/core/Tree/Tree.tsx
saaaaaally: Fixed keyboard to work with checkbox and button

src/core/Tree/TreeNode.tsx
saaaaaally: Ok I reverted it to `nodeCheckbox` so it will render if text was passed in

src/core/Tree/TreeNode.tsx
saaaaaally: Added expand node on double click

stories/core/Tree.stories.tsx
veekeys: I guess it is fixed, but we still have double border..
image

We can fix that later, I assume

src/core/Tree/Tree.tsx
veekeys: I dont know.. experience is weird, now there is no way to quickly go through the nodes.. It hits checkbox and expansion button on every node, when I press arrow keys or tab

src/core/Tree/TreeNode.tsx
veekeys: I just noticed now, we do not have that on table.
@bentleyvk  didnt we have it????????

veekeys: @saaaaaally There is prop called `onDoubleClick`, that did not work for expansion?

src/core/Tree/Tree.tsx
veekeys: Left/Right and Up/Down does same. Just Left/Right also expand/collapse.
What did you want to do with this 
image

src/core/Tree/TreeNode.tsx
bentleyvk: > I just noticed now, we do not have that on table. @bentleyvk didnt we have it????????

I think we never had it.

src/core/Tree/Tree.tsx
mayank99: Cant we make it intuitive so it matches the visual layout? Something like this:

- Up/Down will always traverse between nodes
- Left/Right will traverse between focusable elements inside the nodes.
  - Once you reach the leftmost focusable element (e.g. checkbox), pressing Left again will collapse the node
  - Once you reach the rightmost focusable element, pressing Right again will expand the node

W3C does not have recommendations for extra interactable elements inside the nodes, so we have to improvise while still providing a sensible experience.

src/core/Tree/TreeNode.tsx
saaaaaally: Yes `onDoubleClick` works the same, changed

src/core/Tree/Tree.tsx
saaaaaally: Updated keyboard navigation

src/core/Tree/TreeNode.tsx
veekeys: Lets talk about this next week, what we could do.

src/core/Tree/index.ts
saaaaaally: Removed default

src/core/Tree/Tree.tsx
saaaaaally: Fixed console error

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,215 @@
...
+            </div>
+          </div>
+        }
+        {hasSubNodes && (
+          <ul
+            className='iui-sub-tree'
+            role='group'
+            aria-owns={subNodeIds.join(', ')}
+          />
+        )}
```
roluk, line 209: Are you sure that this `aria-owns` property does something useful for screen readers? The flat tree structure does not follow the accessibility guidelines, so my guess is that the addition of this group does not produce the desired effect.

bentleyvk: I hope it should 😀
image

I also added some additional `aria-` attributes that are automatically added by the browser on semantically correct tree.

src/core/Tree/TreeNode.tsx
bentleyvk: Added a prop for expander

src/core/Tree/Tree.tsx
bentleyvk: It is too much work to handle inside node navigation with arrow keys, so I just made focused node inner items focusable with Tab key.

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,404 @@
...
+            {...rest}
+          />
+        ),
+        [onNodeExpanded, onSelectedNodeChange],
+      )}
+    />
+  );
+};
+
+export const AsyncLoading: Story<TreeProps<unknown>> = () => {
```
bentleyvk, line 147: I am thinking to remove this story and bring it back when we will have proper skeleton or other example.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,268 @@
...
+            'iui-tree-node',
+            {
+              'iui-active': isSelected,
+              'iui-disabled': isDisabled,
+            },
+            className,
+          )}
+          style={styleDepth}
+          onClick={() => !isDisabled && onSelected?.(nodeId, !isSelected)}
+          onDoubleClick={() =>
```
bentleyvk, line 214: I am thinking to remove double click, because it firstly selects the node, then deselects it:
tree-node-dbclick

And to fix this, we would need some hacky solutions to check whether it is one click or double click.

src/core/utils/functions/focusable.test.ts
```diff
@@ -26,6 +26,11 @@ describe('getTabbableElements', () => {
     disabledDiv.className = 'iui-disabled';
     container.append(disabledDiv);
 
+    const focusableDisabledDiv = document.createElement('div');
+    focusableDisabledDiv.setAttribute('tabindex', '-1');
```
veekeys, line 30: mb better `0` as before

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,265 @@
...
+    if (isExpanded) {
+      setExpandedNodes((oldExpanded) => [...oldExpanded, nodeId]);
+    } else {
+      setExpandedNodes((oldExpanded) =>
+        oldExpanded.filter((item) => item != nodeId),
+      );
+    }
+  };
+
+  const getNode = (node: DemoData): NodeData<DemoData> => {
```
veekeys, line 112: need to review again this code sample.. I think here node type is wrong

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,268 @@
...
+          <div className='iui-tree-node-content'>
+            {hasSubNodes && expander}
+            {hasSubNodes && !expander && (
+              <TreeNodeExpander
+                isExpanded={isExpanded}
+                disabled={isDisabled}
+                onClick={(e) => {
+                  onExpanded(nodeId, !isExpanded);
+                  e.stopPropagation();
+                }}
```
veekeys, line 236: wrap in `useCallback`

src/core/Tree/TreeNode.tsx
veekeys: I brought it up initially, but I also tohught we have that in table, which we dont. So I am ok to remove it.

src/core/Tree/TreeNode.tsx
bentleyvk: Moved

src/core/Tree/Tree.tsx
bentleyvk: Changed a bit

src/core/utils/functions/focusable.test.ts
bentleyvk: Changed

src/core/Tree/TreeNode.tsx
bentleyvk: Removed double click handling

veekeys (approved):

src/core/Tree/Tree.tsx
veekeys: nodeRenderer I guess also can wrap in useCallback as in story.. and should be ok

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
...
+      className={cx('iui-tree', className)}
+      role='tree'
+      onKeyDown={handleKeyDown}
+      ref={treeRef}
+      tabIndex={0}
+      style={{ outline: 0 }}
+      onFocus={() => {
+        const items = getFocusableNodes();
+        if (items.length > 0) {
+          items[focusedIndex.current]?.focus();
```
mayank99, line 227: This is also weird. How does it work? Because right now I'm seeing default browser outlines on the nodes.

mayank99 (commented):
Overall looks pretty ok, other than the confusing API.

Mainly need to fix focus/keyboard issues.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,270 @@
...
+          onExpanded(nodeId, true);
+        } else if (subNodeIds.length > 0) {
+          const subNodes = Array.from(
+            nodeRef.current?.ownerDocument.querySelectorAll(
+              `${subNodeIds.map((id) => `#${id}`).join(',')}`,
+            ) || [],
+          ).filter(
+            (el) => el.getAttribute('aria-disabled') !== 'true',
+          ) as HTMLElement[];
+          subNodes[0]?.focus();
```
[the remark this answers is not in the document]
bentleyvk: The code for that is ugly because you need to get `onExpanded`, sub-nodes IDs. I don't think `getFocusableNodes` would be beneficial here as it will get all focusable elements when I need only sub-nodes.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,270 @@
...
+        }
+        break;
+      case ' ':
+      case 'Spacebar':
+      case 'Enter':
+        // Ignore if it is called on the element inside, e.g. checkbox or expander
+        if (event.target !== nodeRef.current) {
+          break;
+        }
+        event.preventDefault();
```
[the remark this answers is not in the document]
bentleyvk: Yeah, it is the issue with the checkbox, I hope your PR will fix it, because it is working with expander button.
As you said it yourself there is no standard for navigating inside node, so I don't want to write this ugly handler and check whether it is first or last item and do actions accordingly, so I made actionable items tabbable and that really simplifies everything.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,270 @@
...
+          style={styleDepth}
+          onClick={() => !isDisabled && onSelected?.(nodeId, !isSelected)}
+        >
+          {checkbox && React.isValidElement(checkbox)
+            ? React.cloneElement(checkbox, {
+                className: cx(
+                  'iui-tree-node-checkbox',
+                  checkbox.props.className,
+                ),
+                tabIndex: isFocused ? 0 : -1,
```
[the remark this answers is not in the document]
bentleyvk: To prevent going trough each checkbox and expander when tabbing.

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
mayank99: By that logic, everything should be tabbable. Just because the code is simpler doesn't mean it's the best experience. https://twitter.com/annaecook/status/1494340020144001027

stories/core/Tree.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+  component: Tree,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  parameters: { controls: { hideNoControlsWarning: true } },
+  title: 'Core/Tree',
+} as Meta<TreeProps<unknown>>;
+
+export const Basic: Story<TreeProps<unknown>> = () => {
```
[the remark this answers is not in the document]
mayank99: selection might be ok, but no need for checkbox and disabled

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { CommonProps, useTheme, getFocusableElements } from '../utils';
+import '@itwin/itwinui-css/css/tree.css';
+import cx from 'classnames';
+
+export type TreeContextProps = {
```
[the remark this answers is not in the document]
bentleyvk: I am reusing it in tests and is not exported to the barrel. I also added some comments.

src/core/Tree/TreeNode.tsx
```diff
@@ -0,0 +1,270 @@
...
+    isSelected = false,
+    onSelected,
+    onExpanded,
+    checkbox,
+    expander,
+    ...rest
+  } = props;
+  useTheme();
+
+  const context = React.useContext(TreeContext);
```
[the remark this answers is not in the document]
bentleyvk: Created `useTreeContext`

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
...
+  indexInGroup: number;
+};
+
+export type NodeRenderProps<T> = Omit<NodeData<T>, 'subNodes'>;
+
+export type TreeProps<T> = {
+  /**
+   * Custom renderer for items inside tree, recommended to use `TreeNode` component.
+   */
+  nodeRenderer: (props: NodeRenderProps<T>) => JSX.Element;
```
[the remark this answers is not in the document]
bentleyvk: Changed comment a bit and added an example.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Array of custom data used for `TreeNodes` inside `Tree`.
+   */
+  data: T[];
+  /**
+   * Function to map custom type `T` to `NodeData`, which will be used to render a tree node in `nodeRenderer`.
+   */
+  getNode: (node: T) => NodeData<T>;
```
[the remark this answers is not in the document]
bentleyvk: Changed comment a bit and added an example. We use flatten user passed `data` and `getNode` maps to according state.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
...
+  }, [data, getNode]);
+
+  return (
+    <ul
+      className={cx('iui-tree', className)}
+      role='tree'
+      onKeyDown={handleKeyDown}
+      ref={treeRef}
+      tabIndex={0}
+      style={{ outline: 0 }}
```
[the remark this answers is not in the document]
bentleyvk: I am not sure whether it should be in CSS because it is kind of hack for easier focus handling.

src/core/Tree/Tree.tsx
bentleyvk: Dumb me, forgot that I need to add classes to `TreeNode`... Fixed.

stories/core/Tree.stories.tsx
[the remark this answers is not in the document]
bentleyvk: Made `Basic` only with expansion.

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
bentleyvk: I didn't go with the easiest path as it was handling arrow keys navigation but nvm... Changed, check it, I can't say it is better than it was previously.

src/core/Tree/Tree.tsx
```diff
@@ -0,0 +1,253 @@
...
+export const Tree = <T,>(props: TreeProps<T>) => {
+  const { data, className, nodeRenderer, getNode, ...rest } = props;
+  useTheme();
+
+  const treeRef = React.useRef<HTMLUListElement>(null);
+
+  const focusedIndex = React.useRef<number>(0);
+  React.useEffect(() => {
+    focusedIndex.current = 0;
+  }, [data, getNode]);
```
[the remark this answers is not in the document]
bentleyvk: Left only `data`

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
mayank99: It is much better now!

I think there is one small issue: currently if a node is collapsed and I press right arrow, then it traverses through the checkbox, expander, etc. I think it should first expand the node, and only after that traverse through the contents.

See this example: https://www.w3.org/TR/wai-aria-practices-1.2/examples/treegrid/treegrid-1.html

src/core/Tree/Tree.tsx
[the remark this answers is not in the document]
mayank99: small typo 😄 
image

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
mayank99: So we don't expect the TreeNode to be used outside Tree?

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
bentleyvk: This example is great! Now I prefer this keyboard navigation over my old one 😀 Btw, changed.

src/core/Tree/Tree.tsx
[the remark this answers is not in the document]
bentleyvk: No no, it is intentional but changed 😀

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
bentleyvk: Idk, I have a feeling that we still will not handle everything iTwin.js needs so they might just reuse our `TreeNode` alone.

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
mayank99: There seems to be a bug 🤔 After reaching the rightmost element in the node, pressing Right arrow goes into the next node's checkbox and stops.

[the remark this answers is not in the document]
mayank99: But it's really confusing when combined with the new behavior of allowing traversing contents. See gif.

In this gif, I'm only using left and right keys. I start on Node 2, press Right to focus checkbox and then expander, and then again to focus Node 2-0, and then again to focus checkbox in Node 2-0. But then it stops! I can press Left to go back but I can't press Right to go to Node 2-1.
tree

The suggestions from your screenshot were more relevant before we made this whole set of changes for allowing content traversal. What we want is something more like this:
image
https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-25

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
bentleyvk: Ok, won't expose context for now

src/core/Tree/TreeNode.tsx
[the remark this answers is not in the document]
bentleyvk: Fixed.

mayank99 (approved):

PR #470, opened 2021-12-09
feat(Checkbox): Add visibility checkbox

### LINKED ISSUE #476: Add visibility checkbox

saaaaaally:
Add `variant` prop to `Checkbox` to allow choice of default or eyeball checkbox

image

## Checklist


veekeys:
I was thinking instead of having a prop, where user can provide his own SVGs for every state of checkbox rather than having a specific variant in the checkbox component itself.

src/core/Checkbox/Checkbox.tsx
```diff
@@ -95,6 +102,7 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
             'iui-disabled': disabled,
             [`iui-${status}`]: !!status,
             'iui-loading': isLoading,
+            'iui-checkbox-visibility': variant === 'eyeball',
```
saaaaaally, line 105: > I was thinking having a prop instead, where user can provide his own SVGs for every state of checkbox rather than having a specific variant in the checkbox component itself.

@veekeys Not sure how to do this, I think we do need to know that it is going to be a visibility checkbox to set this class name

bentleyvk: I think we discussed this several times before. You can't simply pass any SVGs as we are applying classes to the inside elements of SVG. Maybe we can discuss this during our Friday meeting.

src/core/Checkbox/Checkbox.tsx
```diff
@@ -111,12 +119,36 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
         >
           {isLoading && <ProgressRadial indeterminate />}
-          {!isLoading && (
-            <svg viewBox='0 0 16 16' aria-hidden='true' focusable='false'>
-              <path className='iui-check' d='M6,14L0,8l2-2l4,4l8-8l2,2L6,14z' />
-              <path className='iui-check-partial' d='m1 6.5h14v3h-14z' />
-            </svg>
-          )}
+          {!isLoading &&
+            (variant === 'default' ? (
```
bentleyvk, line 123: Maybe move checkboxes html to variables because it is hard to distinct each of them.

gretanausedaite (commented):
No additional comments. Looks good.

bentleyvk (approved):

PR #473, opened 2021-12-09
feat(ComboBox): Add `status` prop

mayank99:
Exposed `status` and `message` props from InputContainer directly into ComboBox props. These should be used to avoid situations like https://github.com/iTwin/iTwinUI/issues/408 where the ComboBox is wrapped in another input-container.

Didn't expose the icon prop because we are already using it for the dropdown caret, and it is trivial for the user to style with just two lines of CSS (added example in story).

image

**Update**: removed `message` to avoid prop clutter. Only `status` is exposed now.

## Checklist


src/core/ComboBox/ComboBox.tsx
```diff
@@ -53,7 +54,8 @@ export type ComboBoxProps<T> = {
    * @default 'No options found'
    */
   emptyStateMessage?: string;
-} & Omit<CommonProps, 'title'>;
+} & Pick<InputContainerProps, 'status' | 'message'> &
```
veekeys, line 57: What about all other props? required, disabled, etc... Or those can work through `inputProps` ?

stories/core/ComboBox.stories.tsx
```diff
@@ -387,3 +389,30 @@ export const WithLabel: Story<Partial<ComboBoxProps<string>>> = (args) => {
...
+  const options = React.useMemo(() => countriesList, []);
+
+  return (
+    <ComboBox
+      options={options}
+      inputProps={{ placeholder: 'Select a country' }}
+      onChange={(value: string) => action(value ?? '')()}
+      status='negative'
+      message={
+        <div style={{ display: 'flex', gap: 4, alignItems: 'center' }}>
```
veekeys, line 403: So you still need to add icon and style it yourself??
Naah, I prefer `InputGroup` then..

src/core/ComboBox/ComboBox.tsx
mayank99: Yes, those should use `inputProps`

stories/core/ComboBox.stories.tsx
mayank99: It's on you if you want to use the InputGroup hack to avoid writing two lines of CSS, but the `status` styling is useful even without `message`

veekeys: Hack is wrong word. I think it is nice universal approach, just our css is buggy sometimes..

src/core/ComboBox/ComboBox.tsx
veekeys: And what about label? Why not this one?

stories/core/ComboBox.stories.tsx
mayank99: I would still call it a hack because you are forcing it in situations it was not designed for. And it's also bad for accessibility as there is no `label` associated with the input.
image

src/core/ComboBox/ComboBox.tsx
veekeys: I would say expose ability to work with InpuContainer fully as you would have it in  your code.. Not sure this mixture is not confusing..

mayank99: I would like to maintain a separation between label and input for this component.

The "mixture" exists in other components like LabeledInput. Here I'm just selectively exposing props, which is less confusing imo.

If you want, I can remove the `message` prop to reduce prop clutter even more. The main benefit really is the `status` which is hard to achieve from outside.

stories/core/ComboBox.stories.tsx
bentleyvk: I would agree with Vykintas that `InputGroup` looks better. And it is not jus 2 lines of CSS, it is all 5 CSS rules. We are the ones that should define standards and not leave them for the user. We might also change something in the inputs, this message styling will differ from the other inputs.

veekeys: But we do need the status styling on the input, which you do not get when using InputGroup... so what to do? Applying twice status also not  that nice :(

veekeys (approved):

PR #477, opened 2021-12-10
chore: Add a11y Storybook addon

FlyersPh9:
- Add a11y Storybook addon.
- Update Storybook.


mayank99:
`creevey` is throwing after updating sb to 6.4

This is the issue to track: https://github.com/wKich/creevey/issues/195

.eslintrc.js
```diff
@@ -1,39 +1,42 @@
...
     }
   },
-  extends: [
-    "plugin:react/recommended", // Uses the recommended rules from @eslint-plugin-react
-    "plugin:@typescript-eslint/recommended", // Uses the recommended rules from @typescript-eslint/eslint-plugin
-    "plugin:react-hooks/recommended",
-    "prettier",
-    "plugin:prettier/recommended" // Enables eslint-plugin-prettier and eslint-config-prettier. This will display prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
-  ],
+  extends: ["plugin:react/recommended", "plugin:@typescript-eslint/recommended", "plugin:react-hooks/recommended", "prettier", "plugin:prettier/recommended", "plugin:storybook/recommended"],
```
veekeys, line 24: would be nice to fix formatting back...
Also return comments as prettier plugin needs to be the last one, so move storybook up.

bentleyvk:
Can we merge it now after Storybook and Creevey update?

.eslintrc.js
FlyersPh9: Fixed.

FlyersPh9:
> Can we merge it now after Storybook and Creevey update?

Should be good to go now! 👍

veekeys (approved):
It works..
The only thing I am concerned is, that this is exposed to our users too, which might decrease the value of our product. Would be nice to hide it and maybe enable it only on local development? Or somehow else toggle it. We dont have authentication, so cannot do it by user :/

Update:
Another concern is how actually its helpful.. We do have some more custom demos, like our `ColorPicker` which just has button to open it, but it tests that button and looks for some a11y stuff, which is not the point there :/

gretanausedaite (approved):
Works for me.
1. It is useful for us.
2. Our users will have iTwinUI documentation site.
3. Once we have that documentation site, we could rework storybook to show only components, without interactions to open them (eg. ColorPicker always displayed)

mayank99:
I never got an answer on https://github.com/storybookjs/storybook/discussions/15642 so I don't think we'll be able to disable these for users unfortunately

bentleyvk:
I expect there won't be a lot of people checking this.
Btw, it is really cool addon as I already found a bug in a `Tree`

PR #478, opened 2021-12-13
feat(Table): Added zebra stripes to rows

### LINKED ISSUE #424: ErrorPage: CommonProps are missing
Need to add `CommonProps` to `ErrorPage` component.
Not related, but is the image not too small?

### LINKED ISSUE #439: Table: zebra rows
image

veekeys:
Added zebra stripes class. Surprisingly, works with sub rows too.
Visual tests did not fail though.. So just manually updated the image.
Added prop `isStriped` so up for discussion for the name and description.

Closes #439 

image


bentleyvk:
Have you tested properly? 😁 Because it needs some CSS fixes:
zebra-table

veekeys:
> Have you tested properly? 😁 Because it needs some CSS fixes: zebra-table

Fixed through css.

src/core/Table/Table.tsx
```diff
@@ -213,6 +213,11 @@ export type TableProps<
    * @default false
    */
   isResizable?: boolean;
+  /**
+   * Use to have zebra-striped table rows.
+   * @default false
+   */
+  isStriped?: boolean;
```
gretanausedaite, line 220: Maybe something like `styleType` : default or striped would be better? This way we could add more styles later.

gretanausedaite (commented):
I think separate story for striped table would be better. I know we have like a million stories for table but I'd never look into paginator story for style (striped vs default).

src/core/Table/Table.tsx
veekeys: Changed prop and added separate story

mayank99: Might be better to do `zebra-rows`. we could add zebra columns in the future

mayank99 (approved):

PR #481, opened 2021-12-14
chore(deps): Bump itwinui-css to 0.42.0

mayank99:
Updated itwinui-css to 0.42.0 and made required changes in ButtonGroup and SplitButton (wrapping children with div).

Also updated browserslist db to get rid of console warning.

## Checklist


src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -54,9 +54,10 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
   (props, ref) => {
     const { children, className, style, overflowButton, ...rest } = props;
 
-    const items = React.useMemo(() => React.Children.toArray(children), [
-      children,
-    ]);
+    const items = React.useMemo(
+      () => React.Children.map(children, (child) => <div>{child}</div>) ?? [],
```
veekeys, line 58: when children can be undefined? or when map can return undefined?

veekeys: And I guess would be nice to have some tests just to ensure it works correctly

mayank99: `React.Children.map` has `| undefined` as its type so I needed the `?? []`.

And I'm not sure about testing implementation details like a wrapper div. https://kentcdodds.com/blog/testing-implementation-details

veekeys: I dont know if div is implementation detail.. Cause then classes are also details for sure. But we know, that if we do not add certain classes, stuff will not work. Same is here, if div does not appear for whatever reasons, it will not work with current css implementation. Same if div changes to something else, I would expect tests to fail.
But as you wish. Maybe it really does not make sense to test. 
Just if it was written html, I wouldnt have thought, just maybe specific map function made me think about it :/

veekeys (approved):
Would be nice to have new tests images lowered in height for small elements, but minor.

PR #482, opened 2021-12-14
chore(Combobox): Update status story with usage of InputGroup

veekeys:
As requested to give full example to the users, added InputGroup with label and message to show full statused combobox.

image


stories/core/ComboBox.stories.tsx
```diff
@@ -392,13 +398,19 @@ export const WithStatus: Story<Partial<ComboBoxProps<string>>> = (args) => {
...
-      options={options}
-      inputProps={{ placeholder: 'Select a country' }}
-      onChange={(value: string) => action(value ?? '')()}
-      status='negative'
-      {...args}
-    />
+    <InputGroup
+      status={args.status ?? 'negative'}
+      message='Country is required.'
+    >
```
mayank99, line 404: Really don't like this hack. Can't we make a new `StatusMessage` component or something?

stories/core/ComboBox.stories.tsx
```diff
@@ -392,13 +398,19 @@ export const WithStatus: Story<Partial<ComboBoxProps<string>>> = (args) => {
...
+      <Label>Country</Label>
```
mayank99, line 405: Add `htmlFor` here and `id` to inputProps. See the label story for example

veekeys: yeah, good catch. added

stories/core/ComboBox.stories.tsx
veekeys: hmm...I dont know. I guess we can, but then what to use when...? InputGroup it not needed anymore..?

mayank99: InputGroup would still be valid for these use cases
image

veekeys: But I think you could have same without it too. Just put `Label`, few `Checkbox` and this new `StatusMessage` component at the end.

mayank99: InputGroup adds a bit of gap between the checkboxes by setting min-height

veekeys: and its only helpful if you have checkbox without label...
idk, tricky situation and not sure what would be the best here..

stories/core/ComboBox.stories.tsx
```diff
@@ -392,16 +398,22 @@ export const WithStatus: Story<Partial<ComboBoxProps<string>>> = (args) => {
...
+    <InputGroup
+      status={args.status ?? 'negative'}
+      message='Country is required.'
+    >
+      <Label htmlFor='country-input'>Country</Label>
+      <ComboBox
+        options={options}
+        inputProps={{ id: 'country-input', placeholder: 'Select a country' }}
+        onChange={(value: string) => action(value ?? '')()}
+        status='negative'
```
gretanausedaite, line 410: I think this hardcoded `status` is not needed anymore.

veekeys: What do you mean by "not needed"? Is it because storybook args will override it? Then I left it on purpose for better readability.
Or is it because input group now has it? If thats, then it is needed, because without it Combobox will not be styled with the red bottom border.

gretanausedaite: Better readability is good purpose.
Combobox is styled with red bottom border even without this `status: 'negative'`.

gretanausedaite (approved):
Looks good 🌵

stories/core/ComboBox.stories.tsx
veekeys: At first I picked only `message` prop from `InputContainer`, but.. It does not handle icon and we are already using icon, so that means we cannot have 2 icons (inline and below) at the same time with current implementation. So I guess this is also not an option, cause just message prop mostly used on status messages, IMO.
So I guess we need to rethink whole thing and decide what we wanna do with this situation.

veekeys:
I am closing this PR as it seems it is going nowhere. Need to rethink how we want to handle message. 
We have here possibly 2 icons to put (inline and below), where our input components just accept icon and its position.
Maybe `StatusMessage` as a component could be an option, just it brings another inconsistency all over the APIs...

bentleyvk:
> I am closing this PR as it seems it is going nowhere. Need to rethink how we want to handle message. We have here possibly 2 icons to put (inline and below), where our input components just accept icon and its position. Maybe `StatusMessage` as a component could be an option, just it brings another inconsistency all over the APIs...

I think this PR https://github.com/iTwin/iTwinUI/pull/444 should fix having status icon.

PR #487, opened 2021-12-16
fix(Table): Prevents from triggering sort when filtering

### LINKED ISSUE #483: Table: clicking on filter toggle triggers sorting
### Describe the bug
Clicking on the filter icon triggers sorting each time.

#### Reproduction
https://itwin.github.io/iTwinUI-react/?path=/story/core-table--filters&args=isSortable:true

#### Additional information
Package version: `@itwin/itwinui-react@1.27.0`

bentleyvk:
Closes #483
I was trying to think of better solution instead of adding `e.stopPropagation()` but every time I would bump into edge case that breaks the fix.
I thought that this should do the trick:
```tsx
<div
  onMouseDown={e => {
    if (e.target === e.currentTarget) {
      onSortClick(e);
    }
  }}
/>  
```
But clicking on sort icon would not sort, also if users have some elements inside header, clicking on them would also not sort.

I am using `userEvent` in `Table` tests because its click simulates real user click: `pointerDown`, `mouseDown`, `focus`, `pointerUp`, `mouseUp`, and `click`.

## Checklist

### DIFF
--- a/src/core/Table/Table.test.tsx
+++ b/src/core/Table/Table.test.tsx
@@ -19,6 +19,7 @@ import { EditableCell } from './cells';
 import { TablePaginator } from './TablePaginator';
 import * as UseOverflow from '../utils/hooks/useOverflow';
 import * as UseResizeObserver from '../utils/hooks/useResizeObserver';
+import userEvent from '@testing-library/user-event';
 
 const intersectionCallbacks = new Map<Element, () => void>();
 jest
@@ -148,7 +149,7 @@ function assertRowsData(rows: NodeListOf<Element>, data = mockedData()) {
     expect(cells[0].textContent).toEqual(name);
     expect(cells[1].textContent).toEqual(description);
     expect(cells[2].textContent).toEqual('View');
-    fireEvent.click(cells[2].firstElementChild as HTMLElement);
+    userEvent.click(cells[2].firstElementChild as HTMLElement);
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/core/Table/Table.test.tsx
```diff
@@ -806,6 +807,7 @@ it('should not trigger sorting when filter is clicked', () => {
     columns: mockedColumns,
     onFilter,
     onSort,
+    isSortable: true,
```
bentleyvk, line 810: This was missing therefore test doesn't check anything.

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -56,10 +56,10 @@ export const FilterToggle = <T extends Record<string, unknown>>(
             styleType='borderless'
             isActive={isVisible || column.filterValue}
             className={cx('iui-filter-button', className)}
-            onClick={(e: React.MouseEvent) => {
-              e.stopPropagation();
+            onClick={() => {
               setIsVisible((v) => !v);
             }}
+            onMouseDown={(e) => e.stopPropagation()}
```
mayank99, line 62: Maybe add a comment here? Because it is a bit unintuitive

bentleyvk: Added.

mayank99 (approved):

src/core/Table/filters/FilterToggle.tsx
```diff
@@ -56,10 +56,11 @@ export const FilterToggle = <T extends Record<string, unknown>>(
             styleType='borderless'
             isActive={isVisible || column.filterValue}
             className={cx('iui-filter-button', className)}
-            onClick={(e: React.MouseEvent) => {
-              e.stopPropagation();
+            onClick={() => {
```
veekeys, line 59: have you checked if keyboard clicks work fine?

mayank99: Yes, it does work but we have bigger problems there like not being able to tab into the filter icon (it's hidden without hover						)

veekeys: it should be visible with focus though...?

bentleyvk: I was just to write the same. I think I never have tested keyboard navigation for opening filter menu or triggering sort...

veekeys (approved):

src/core/Table/filters/FilterToggle.tsx
bentleyvk: Created https://github.com/iTwin/iTwinUI-react/issues/488

PR #491, opened 2021-12-17
feat(useTheme): Add event listener to respond to theme changes

### LINKED ISSUE #485: Using "os" as theme does not update when OS theme changes
If you use ThemeProvider with them set to "os" or useTheme("os"), it correctly detects the color theme at the time of component creation. However, it does not detect on-the-fly changes in the OS color theme.

#### Expected behavior
If a user uses "os" as the theme, ThemeProvider and useTheme() should both detect when the OS theme changes and update accordingly.

#### Reproduction
https://codesandbox.io/s/itwinui-react-example-forked-7fvtc?file=/src/App.tsx

##### Steps to reproduce
Load the sandbox. Note that its theme matches the OS theme.
Change your OS theme.
Notice that the theme does not change in the sandbox.

Note: I tested the above sandbox on macOS Big Sur. My actual code is running on iOS, where the same problem occurs.

#### Additional information
I have code in another location that detects changes to the OS color them. The code looks something like this:

```ts
    try {
      window.matchMedia("(prefers-color-scheme: dark)").addEventListener("change", MobileUi._colorSchemeListener);
    } catch (e) {
      // Safari didn't support the above BASIC functionality until version 14.
      // eslint-disable-next-line deprecation/deprecation
      window.matchMedia("(prefers-color-scheme: dark)").addListener(MobileUi._colorSchemeListener);
    }
```

The source file that the above code came from can be found here:

https://github.com/iTwin/mobile-ui-react/blob/main/src/mobile-ui-react/MobileUi.tsx

mayank99:
Closes #485

## Checklist


src/core/utils/hooks/useTheme.ts
```diff
@@ -40,16 +40,33 @@ export const useTheme = (
...
       case 'dark':
+        prefersDarkMediaQuery?.removeEventListener?.('change', addOSTheme);
         addDarkTheme(ownerDocument);
         break;
       case 'os':
-        if (getWindow()?.matchMedia?.('(prefers-color-scheme: dark)').matches) {
-          addDarkTheme(ownerDocument);
+        if (prefersDarkMediaQuery != undefined) {
+          addOSTheme(prefersDarkMediaQuery);
+          prefersDarkMediaQuery.addEventListener?.('change', addOSTheme);
```
bentleyvk, line 69: What about Safari? Does it support this method?

src/core/utils/hooks/useTheme.ts
```diff
@@ -40,16 +40,33 @@ export const useTheme = (
       return;
     }
 
+    const prefersDarkMediaQuery = getWindow()?.matchMedia?.(
+      '(prefers-color-scheme: dark)',
+    );
+
+    const addOSTheme = ({
```
bentleyvk, line 47: You need to memoize this function because every rerender it will have different reference and it won't be removed from listeners.

src/core/utils/hooks/useTheme.ts
mayank99: Safari 14 and later do. Earlier versions will still get the initial preferred theme so it should be fine

veekeys, line 69: works fine :O
some tests would be useful

veekeys: ok.. I cannot test in storybook, cause it works even without your changes...console log in listener never works :DD

mayank99: For tests I need to mock add/removeEventListener and figure out how to fire a change event

saaaaaally (approved):

src/core/utils/hooks/useTheme.ts
bentleyvk: You can try mocking `matchMedia` return object that has some trigger function.

mayank99: See [`be4e696`](https://github.com/iTwin/iTwinUI-react/pull/491/commits/be4e69635972e5f71f13fcf32d50ba2d09395c6f)

src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -25,19 +25,69 @@ afterEach(() => {
...
+
+  window.matchMedia = jest.fn().mockReturnValueOnce({
+    matches: matches,
+    addEventListener: (
+      _: 'change',
+      listener: (e: { matches: boolean }) => void,
+    ) => {
+      listeners.push(listener);
+    },
+    removeEventListener: (
```
bentleyvk, line 64: Maybe also check that `removeEventListener` was called.

mayank99: Added

bentleyvk (approved):

PR #503, opened 2022-01-10
chore: Updated CSS and icons libs

### LINKED ISSUE #502: New classnames and html from itwinui-css 0.44
iTwinUI-css 0.44.0 includes some breaking changes such as updated class names and restructured html.
- https://github.com/iTwin/iTwinUI/pull/426
- https://github.com/iTwin/iTwinUI/pull/429
- https://github.com/iTwin/iTwinUI/pull/427

Also some new icons were added, specifically `swap.svg` which is currently being inlined in `ColorPicker`
- https://github.com/iTwin/iTwinUI/pull/438

bentleyvk:
Some of the Table images were different because of Checkbox corners (red means different):
checkbox-change

Closes #502

## Checklist


src/core/Table/TablePaginator.tsx
```diff
@@ -158,18 +157,20 @@ export const TablePaginator = (props: TablePaginatorProps) => {
...
-        tabIndex={tabIndex}
-        size={buttonSize}
-      >
-        {index + 1}
-      </Button>
+      <div key={index}>
+        <button
+          className={cx('iui-paginator-page-button', {
+            'iui-active': index === currentPage,
+            'iui-paginator-page-button-small': buttonSize === 'small',
```
bentleyvk, line 164: Why modifier class has whole class name in the beginning? DatePicker classes were refactored a bit but modifiers remained the same.

mayank99: This happens because of `&-`. I don't think it's a bad thing

bentleyvk: Yeah, I understand why this class is like this. What I don't understand is the reason why it was done like this and only here. It looks ugly to me or then it should be enough to have only `iui-paginator-page-button-small` without `iui-paginator-page-button`.

mayank99: then you won't be able to target base button with one class.

in fact this is a popular BEM convention
https://en.bem.info/methodology/naming-convention/#element-modifier-name

bentleyvk: But then we are inconsistent here. Everywhere we use short modifiers and only here (for whatever reason) we decided to go with BEM convention.

mayank99: How would you solve this? Inconsistencies are everywhere in the CSS repo, but this solution keeps the specificity low while avoiding conflicts with other components.

Also maybe not the right place for this discussion as we can't change the CSS in this PR.

bentleyvk: What is the problem with `.iui-paginator-page-button.iui-small`? If we don't have modifier classes as global class, we shouldn't have a problem with that.

mayank99 (approved):
Looks good but some of the image changes (table/checkboxes) seem unnecessary

PR #512, opened 2022-01-18
fix(Button): Improve polymorphic handling

mayank99:
Made two changes in `PolymorphicForwardRefComponent`:
1. `as` prop is now optional (because it doesn't seem to get default value in itwinjs-core)
2. changed type narrowing to check `JSX.IntrinsicElements` first.

## Checklist

### DIFF
--- a/src/core/utils/props.ts
+++ b/src/core/utils/props.ts
@@ -60,12 +60,10 @@ export interface PolymorphicForwardRefComponent<
     >
   > {
   <As = T>(
-    props: As extends ''
-      ? { as: keyof JSX.IntrinsicElements }
+    props: As extends keyof JSX.IntrinsicElements
+      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as?: As }>
       : As extends React.ComponentType<infer P>
-      ? Merge<P, OwnProps & { as: As }>
-      : As extends keyof JSX.IntrinsicElements
-      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as: As }>
+      ? Merge<P, OwnProps & { as?: As }>
       : never,
   ): React.ReactElement | null;
 }


mayank99:
@bsteinbk Are you able to build this branch and test your itwinjs-core branch against it? Or if that's not possible, maybe copy the changes into your node_modules?

The goal here is that `Button` should work by default (pretty much the same way as before) and not require using the `as` prop.

bsteinbk:
@mayank99 - I built the branch and copied the files from lib directory into the node_modules\@itwin\itwinui-react folder on two separate packages in our mono repo and I was able to compile without any changes.

mayank99:
> I was able to compile without any changes.

Just to be clear, does this mean it fixes the typescript errors that you were getting re: `as` prop being required and `className` not being found, and that you don't need to add any workarounds from [`itwinjs-core@50fafdf`](https://github.com/iTwin/itwinjs-core/commit/50fafdf088f77d0fca23f37e0721922b804c183a)?

bsteinbk:
I tested the maplayers and core-react packages and they did not need any of the `as="button"` or ButtonProps<"button"> changes from  itwinjs-core@50fafdf. I expect all the other packages in the mono repo to be fine once this PR completes and a new itwin-react package is published.  I did not take the time to do all the overwritting to node_modules to every package since the two that I tried worked.

src/core/utils/props.ts
```diff
@@ -60,12 +60,10 @@ export interface PolymorphicForwardRefComponent<
     >
   > {
   <As = T>(
-    props: As extends ''
-      ? { as: keyof JSX.IntrinsicElements }
+    props: As extends keyof JSX.IntrinsicElements
+      ? Merge<JSX.IntrinsicElements[As], OwnProps & { as?: As }>
```
bentleyvk, line 64: Could you explain to us mortals what it does? 😀

mayank99: It's basically the same as before, the difference is that previously it was checking against empty string first to make sure that `as` prop was actually used, whereas now it's _always_ checking against `JSX.IntrinsicElements` first. I'm guessing the empty string was causing problems in itwinjs-core

bentleyvk (approved):

mayank99:
@bsteinbk This fix is now available in v1.29.1 🚀

PR #519, opened 2022-01-20
fix(Button): Fix event typings for inline handlers

mayank99:
Making `as` optional in the type narrowing breaks automatic inferring of function props (like event handlers and ref). Modified one of the tests to include an inline handler so that we don't get this bug again.

Before this PR:
image

After this PR:
image

## Checklist

### DIFF
--- a/src/core/Buttons/Button/Button.test.tsx
+++ b/src/core/Buttons/Button/Button.test.tsx
@@ -10,7 +10,13 @@ import { Button } from './Button';
 it('renders default button correctly', () => {
   const onClickMock = jest.fn();
   const { container } = render(
-    <Button onClick={onClickMock}>Click me!</Button>,
+    <Button
+      onClick={(e) => {
+        onClickMock(e);
+      }}
+    >
+      Click me!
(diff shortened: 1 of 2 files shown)


src/core/Buttons/Button/Button.test.tsx
```diff
@@ -10,7 +10,13 @@ import { Button } from './Button';
 it('renders default button correctly', () => {
   const onClickMock = jest.fn();
   const { container } = render(
-    <Button onClick={onClickMock}>Click me!</Button>,
+    <Button
+      onClick={(e) => {
```
mayank99, line 14: @veekeys I created a branch for you to show how Ben's approach doesn't have correct types for events. All the code is in [`Button.tsx`](https://github.com/iTwin/iTwinUI-react/blob/mayank/polymorphic-types-ben/src/core/Buttons/Button/Button.tsx#L40). Check out that branch and open `Button.test.tsx:14` to see the `any` type.

veekeys (approved):

PR #524, opened 2022-01-25
feat(ButtonGroup): Add `ButtonGroupItem` subcomponent

mayank99:
See https://github.com/iTwin/iTwinUI/pull/448

Closes # 

## Checklist

### DIFF
--- a/src/core/ButtonGroup/ButtonGroup.tsx
+++ b/src/core/ButtonGroup/ButtonGroup.tsx
@@ -59,8 +59,10 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
       () =>
         React.Children.map(children, (child) => {
           return React.isValidElement(child) ? (
-            child.type === ButtonGroupItem || child.type === Tooltip ? (
+            child.type === ButtonGroupItem ? (
               child
+            ) : child.type === Tooltip ? (
+              <div>{child}</div>
             ) : (
               <ButtonGroupItem>{child}</ButtonGroupItem>
             )
@@ -92,11 +94,9 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
       </div>
     );
   },
-) as React.ForwardRefExoticComponent<ButtonGroupProps> & {
-  Item: typeof ButtonGroupItem;
-};
+);
 
-const ButtonGroupItem = React.forwardRef(
+export const ButtonGroupItem = React.forwardRef(
   (
     { className, children, ...rest }: React.ComponentPropsWithoutRef<'div'>,
     ref: React.Ref<HTMLDivElement>,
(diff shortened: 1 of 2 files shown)


src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -55,7 +57,18 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
...
 
     const items = React.useMemo(
-      () => React.Children.map(children, (child) => <div>{child}</div>) ?? [],
+      () =>
+        React.Children.map(children, (child) => {
+          return React.isValidElement(child) ? (
+            child.type === ButtonGroupItem ? (
+              child
+            ) : child.type === Tooltip || child.type === DropdownMenu ? (
+              <div>{child}</div>
```
mayank99, line 66: I am not sure about this part. This div would not be needed if `appendTo={document.body}` is used (which we recommend doing for z-index issues anyway).

Also I need to test `child.type` in production React. I think development environment has extra info in `.type`

veekeys: I didnt get, why we need all this logic? Why we just cannot render all children without wrapping div? Our `ButtonGroupItem` should have div... no?

mayank99: This is how we can expect it to be used:
image

The `ButtonGroupItem` is used for getting access to the child's `disabled` attribute via `> *`. The tooltip still needs to go around `ButtonGroupItem`, so if tippy is appended to parent, it still needs a wrapping div to avoid breaking `:last-child` selector.

veekeys: damn... Thats not good and no chance to explain that to users...

veekeys: And what happens, if we wrap `IconButton` with `Tooltip` ?

mayank99: Would it make sense to offer a built-in way to add tooltips directly in `ButtonGroupItem`?

```tsx
<ButtonGroupItem tooltip={<Tooltip .../>}>
  <IconButton ... />
</ButtonGroupItem>
```

veekeys: That is always a possibility I guess.. But we were avoiding that with other components.

bentleyvk: So, is `.type` available in prod? From what I read, it should be there as JSX is transformed into JS objects that has `type` prop.

mayank99: > So, is .type available in prod?

I checked the built `.js` file and it was not there. I also tried console logging in codesandbox and was getting undefined in all our components.

PR #527, opened 2022-01-27
feat(ButtonGroup): Add new `overflowPlacement` prop

### LINKED ISSUE #531: ButtonGroup: Allow placing overflow button at the start
overflow button needs to be placed at the "start" if the button-group hugs the right corner of the window
image

mayank99:
New `overflowPlacement` prop allows placing overflowButton at the start.
image

Closes #531

## Checklist

### DIFF
--- a/src/core/ButtonGroup/ButtonGroup.tsx
+++ b/src/core/ButtonGroup/ButtonGroup.tsx
@@ -20,6 +20,11 @@ export type ButtonGroupProps = {
    * and returns the `ReactNode` to render.
    */
   overflowButton?: (firstOverflowingIndex: number) => React.ReactNode;
+  /**
+   * If `overflowButton` is specified, should it placed at the start or the end?
+   * @default 'end'
+   */
+  overflowPosition?: 'start' | 'end';
 } & React.ComponentPropsWithRef<'div'>;
 
 /**
@@ -52,7 +57,14 @@ export type ButtonGroupProps = {
  */
 export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
   (props, ref) => {
-    const { children, className, style, overflowButton, ...rest } = props;
+    const {
+      children,
+      className,
+      style,
+      overflowButton,
+      overflowPosition = 'end',
+      ...rest
+    } = props;
 
     const items = React.useMemo(
       () => React.Children.map(children, (child) => <div>{child}</div>) ?? [],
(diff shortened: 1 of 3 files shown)


adamhock (approved):
This looks good to me

src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -73,8 +85,15 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
       >
         {!!overflowButton && visibleCount < items.length ? (
           <>
+            {overflowButton && overflowPosition === 'start' && (
+              <div>{overflowButton(visibleCount)}</div>
```
bentleyvk, line 89: Why div is needed?

mayank99: it was forgotten before. If you go to the live version and try to hover/focus on the overflow button, you will see messed up borders

src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -73,8 +85,15 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
...
+            )}
+
             {items.slice(0, visibleCount - 1)}
-            {overflowButton(visibleCount)}
+
+            {overflowButton && overflowPosition === 'end' && (
```
bentleyvk, line 94: I just thought that instead of `overflowButton`, we could have children as a render function so user could place  overflow button whethever they want. But I guess it will be a breaking change.

src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -20,6 +20,11 @@ export type ButtonGroupProps = {
    * and returns the `ReactNode` to render.
    */
   overflowButton?: (firstOverflowingIndex: number) => React.ReactNode;
```
bentleyvk, line 22: Maybe change `overflowButton` description a bit

mayank99: Changed and added info about the new prop, which made me realize maybe the prop should be called `overflowPlacement`

bentleyvk: Ok, lets wait for others opinion about the naming 😀

veekeys: I would have gone with `position`, but only because we have tippy in multiple components and its using placement, `placement` feels like more consistent one...

mayank99: image

bentleyvk (approved):

PR #528, opened 2022-01-27
fix(Slider): Handle pointermove events only when thumb is active

### LINKED ISSUE #511: Pointer actions not working with slider
We encountered a specific bug in the iTwin-React Slider component during development. This component overrides the default behavior of pointer move event which means that we are not able to perform default pointer actions such e.g. selecting text when the slider is in the DOM.I am unable to select text after adjusting slider value. As seen in the gif below:
chrome-capture (1)

#### Expected behavior
Adjusting the slider should not effect pointer actions in other elements. 

#### Reproduction
[Link to a minimal repro: ](https://codesandbox.io/s/itwinui-react-example-forked-kb7oj?file=/src/App.tsx)

##### Steps to reproduce
1. Select Text (works fine)
1. Adjust Slider (triggers bug)
1. Select Text Again (doesn't work)

mayank99:
Returning early when thumb is undefined so that it doesn't break other pointer event listeners on document.

Fixes #511

## Checklist

### DIFF
--- a/src/core/Slider/Slider.test.tsx
+++ b/src/core/Slider/Slider.test.tsx
@@ -714,3 +714,18 @@ it('should activate thumb on pointerDown and move to closest step on move/ no up
   expect(thumbs[0].getAttribute('aria-valuenow')).toEqual('40');
   expect(thumbs[1].getAttribute('aria-valuenow')).toEqual('80');
 });
+
+it('should not break pointer events when thumb is released', () => {
+  const { container } = render(
+    <Slider min={0} max={100} values={[20, 80]} step={1} />,
+  );
+
+  assertBaseElement(container);
+  const thumb = container.querySelector('.iui-slider-thumb') as HTMLDivElement;
+  expect(thumb).not.toHaveClass('iui-active');
+
+  act(() => {
+    // userEvent.pointer is not released yet 😕
+  });
+  // expect(thumb).toHaveClass('iui-active');
+});
--- a/src/core/Slider/Slider.test.tsx
+++ b/src/core/Slider/Slider.test.tsx
@@ -714,18 +714,3 @@ it('should activate thumb on pointerDown and move to closest step on move/ no up
   expect(thumbs[0].getAttribute('aria-valuenow')).toEqual('40');
   expect(thumbs[1].getAttribute('aria-valuenow')).toEqual('80');
 });
-
-it('should not break pointer events when thumb is released', () => {
-  const { container } = render(
-    <Slider min={0} max={100} values={[20, 80]} step={1} />,
-  );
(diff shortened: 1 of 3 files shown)


src/core/Slider/Slider.test.tsx
```diff
@@ -714,3 +714,18 @@ it('should activate thumb on pointerDown and move to closest step on move/ no up
...
+  const { container } = render(
+    <Slider min={0} max={100} values={[20, 80]} step={1} />,
+  );
+
+  assertBaseElement(container);
+  const thumb = container.querySelector('.iui-slider-thumb') as HTMLDivElement;
+  expect(thumb).not.toHaveClass('iui-active');
+
+  act(() => {
+    // userEvent.pointer is not released yet 😕
```
mayank99, line 728: Thoughts on how to test this? I can't get the test to replicate the bug. I've tried adding pointer events on a sibling element, container, document, and on window

src/core/Slider/Slider.tsx
```diff
@@ -331,12 +334,15 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
 
     const handlePointerUp = React.useCallback(
       (event: PointerEvent) => {
+        if (activeThumbIndex === undefined) {
```
bentleyvk, line 337: When it is set to `undefined`?

mayank99: in this same function 😆 
image

src/core/Slider/Slider.test.tsx
bentleyvk: I have tried few cases myself but nothing worked. Maybe it is impossible to reproduce in jsdom.

mayank99: Maybe we should complete this PR without a test? 😄

bentleyvk: We might skip tests for this one then. I am just thinking whether it is a good solution because it is always listening for the document events, I kind of like the way you did in Info Panel:
image

mayank99: I tried doing that and it got too complicated and was breaking things, so I reverted it

elephantcatdog (commented):
Why didn't you include a check for the active color dot in handleSquarePointerLeave alongside the other two pointer handling functions?

veekeys (approved):

mayank99:
> Why didn't you include a check for the active color dot in handleSquarePointerLeave alongside the other two pointer handling functions?

Good point, I've added it to handleSquarePointerLeave now.

bentleyvk (approved):

PR #534, opened 2022-02-03
fix(Popover): Prevents from adding 5px padding

bentleyvk:
Fixes the issue when `Select` or any other component with `Popover` is close to the side of container, it gets 5px padding which appear isn't random.
Popper documentation: https://popper.js.org/docs/v2/modifiers/prevent-overflow/
Issue in Popper that helped to figure it out: https://github.com/floating-ui/floating-ui/issues/704
image

## Checklist

### DIFF
--- a/src/core/utils/components/Popover.tsx
+++ b/src/core/utils/components/Popover.tsx
@@ -63,13 +63,6 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
     arrow: false,
     duration: 0,
     interactive: true,
-    popperOptions: {
-      strategy: 'fixed',
-      modifiers: [
-        { name: 'flip' },
-        { name: 'preventOverflow', options: { padding: 0 } },
-      ],
-    },
     role: undefined,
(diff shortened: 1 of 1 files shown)


src/core/utils/components/Popover.tsx
```diff
@@ -65,7 +65,10 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
     interactive: true,
     popperOptions: {
       strategy: 'fixed',
-      modifiers: [{ name: 'flip' }],
+      modifiers: [
+        { name: 'flip' },
+        { name: 'preventOverflow', options: { padding: 0 } },
+      ],
     },
```
mayank99, line 72: I just realized that if user adds `popperOptions`, these will all get overwritten. Do we want to merge user's popperOptions here?

bentleyvk: I think so, we are doing the same with `plugins`. So I added merging.

mayank99 (approved):

PR #536, opened 2022-02-03
chore(deps): Bump itwinui-css to 0.45.0

mayank99:
Updated Tile classnames according to https://github.com/iTwin/iTwinUI/pull/431 and added `iui-popover` class to Popover.

## Checklist

- [x] ~~Add meaningful unit tests for your component (verify that all lines are covered)~~

### DIFF
--- a/src/core/Tile/Tile.tsx
+++ b/src/core/Tile/Tile.tsx
@@ -147,15 +147,15 @@ export const Tile = (props: TileProps) => {
       {...rest}
     >
       {thumbnail && (
-        <div className='iui-thumbnail'>
+        <div className='iui-tile-thumbnail'>
           {typeof thumbnail === 'string' ? (
             <div
-              className='iui-picture'
+              className='iui-tile-thumbnail-picture'
               style={{ backgroundImage: `url(${thumbnail})` }}
             />
           ) : thumbnail && (thumbnail as JSX.Element).type === 'img' ? (
             React.cloneElement(thumbnail as JSX.Element, {
-              className: 'iui-picture',
+              className: 'iui-tile-thumbnail-picture',
             })
           ) : React.isValidElement(thumbnail) ? (
             React.cloneElement(thumbnail, {
@@ -167,20 +167,22 @@ export const Tile = (props: TileProps) => {
 
           {leftIcon &&
             React.cloneElement(leftIcon as React.ReactElement, {
-              className: 'iui-small iui-type-indicator',
+              className: 'iui-small iui-tile-thumbnail-type-indicator',
             })}
 
           {rightIcon &&
             React.cloneElement(rightIcon as React.ReactElement, {
-              className: 'iui-small iui-quick-action',
+              className: 'iui-small iui-tile-thumbnail-quick-action',
             })}
 
-          {badge && <div className='iui-badge-container'>{badge}</div>}
+          {badge && (
+            <div className='iui-tile-thumbnail-badge-container'>{badge}</div>
+          )}
         </div>
       )}
 
-      <div className='iui-content'>
-        <div className='iui-name'>
+      <div className='iui-tile-content'>
+        <div className='iui-tile-name'>
           {isSelected && (
             <SvgCheckmark
               className={cx('iui-tile-status-icon', 'iui-informational')}
@@ -193,15 +195,15 @@ export const Tile = (props: TileProps) => {
               aria-hidden
             />
           )}
-          <span className='iui-name-label'>{name}</span>
+          <span className='iui-tile-name-label'>{name}</span>
         </div>
 
         {description != undefined && (
-          <div className='iui-description'>{description}</div>
+          <div className='iui-tile-description'>{description}</div>
         )}
 
         {metadata != undefined && (
-          <div className='iui-metadata'>{metadata}</div>
+          <div className='iui-tile-metadata'>{metadata}</div>
         )}
 
         {moreOptions && (
@@ -223,7 +225,7 @@ export const Tile = (props: TileProps) => {
             <IconButton
               styleType='borderless'
               size='small'
-              className={cx('iui-more-options', {
+              className={cx('iui-tile-more-options', {
                 'iui-visible': isMenuVisible,
               })}
               aria-label='More options'
--- a/src/core/Tile/Tile.tsx
+++ b/src/core/Tile/Tile.tsx
@@ -167,12 +167,12 @@ export const Tile = (props: TileProps) => {
 
           {leftIcon &&
             React.cloneElement(leftIcon as React.ReactElement, {
-              className: 'iui-small iui-tile-thumbnail-type-indicator',
+              className: 'iui-tile-thumbnail-type-indicator',
             })}
 
           {rightIcon &&
             React.cloneElement(rightIcon as React.ReactElement, {
-              className: 'iui-small iui-tile-thumbnail-quick-action',
+              className: 'iui-tile-thumbnail-quick-action',
             })}
 
           {badge && (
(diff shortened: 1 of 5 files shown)


src/core/Tile/Tile.tsx
```diff
@@ -193,15 +195,15 @@ export const Tile = (props: TileProps) => {
...
         </div>
 
         {description != undefined && (
-          <div className='iui-description'>{description}</div>
+          <div className='iui-tile-description'>{description}</div>
         )}
 
         {metadata != undefined && (
-          <div className='iui-metadata'>{metadata}</div>
+          <div className='iui-tile-metadata'>{metadata}</div>
```
veekeys, line 206: Just noticed, we dont user `iui-tile-metadata-icon` anywhere :O

src/core/Tile/Tile.tsx
```diff
@@ -223,7 +225,7 @@ export const Tile = (props: TileProps) => {
             <IconButton
               styleType='borderless'
               size='small'
-              className={cx('iui-more-options', {
+              className={cx('iui-tile-more-options', {
                 'iui-visible': isMenuVisible,
               })}
               aria-label='More options'
```
veekeys, line 231: Was not allowed to comment there, but there is no `iui-tile-button` class

src/core/Tile/Tile.tsx
```diff
@@ -167,20 +167,22 @@ export const Tile = (props: TileProps) => {
 
           {leftIcon &&
             React.cloneElement(leftIcon as React.ReactElement, {
-              className: 'iui-small iui-type-indicator',
+              className: 'iui-small iui-tile-thumbnail-type-indicator',
```
veekeys, line 170: actually, this is more interesting.. we do include buttons and small buttons mixin for `iui-tile-thumbnail-type-indicator`, so there is no need of `iui-small`.. This was forgotten in html though,..There is no need of any button classes, to be honest.
Same with quick action below

src/core/Tile/Tile.tsx
```diff
@@ -147,15 +147,15 @@ export const Tile = (props: TileProps) => {
       {...rest}
     >
       {thumbnail && (
-        <div className='iui-thumbnail'>
+        <div className='iui-tile-thumbnail'>
           {typeof thumbnail === 'string' ? (
```
veekeys, line 151: Order of thumbnail elements are different in html and css, but I guess that does not influence anything

src/core/Tile/Tile.tsx
```diff
@@ -147,15 +147,15 @@ export const Tile = (props: TileProps) => {
...
             <div
-              className='iui-picture'
+              className='iui-tile-thumbnail-picture'
               style={{ backgroundImage: `url(${thumbnail})` }}
             />
           ) : thumbnail && (thumbnail as JSX.Element).type === 'img' ? (
             React.cloneElement(thumbnail as JSX.Element, {
-              className: 'iui-picture',
+              className: 'iui-tile-thumbnail-picture',
```
veekeys, line 158: This part is also interesting.. we have example with ifram as picture and there we need to pass class `iui-tile-thumbnail-picture`.. in react it will get `iui-thumbnail-icon` instead...
Rules actually for `iui-thumbnail-icon` are also interesting in css... We have its size set if its children of thumbnail and we have some more rules, if its next to `iui-tile-thumbnail-picture` <- not even possible in react right now :/

src/core/Tile/Tile.tsx
mayank99: I could not find it in the CSS either

src/core/Tile/Tile.tsx
mayank99: Huh. I see the CSS PR didn't even have it before. Why did we add it to React then?

src/core/Tile/Tile.tsx
veekeys: image

src/core/Tile/Tile.tsx
veekeys: I think that is a really good question

src/core/Tile/Tile.tsx
mayank99: So what do you suggest doing? iframes are kinda an edge case so the user could just add their own CSS

src/core/Tile/Tile.tsx
mayank99: Removed

src/core/Tile/Tile.tsx
veekeys: Found it :)

https://github.com/iTwin/iTwinUI/commit/ebcc9b28df136ab6c66e29b388b252ae3d924370#diff-fde6de9322de8edfd8f91b47b9dc6e00ca10ba514a532e2d73a61c978e3f99b3

src/core/Tile/Tile.tsx
veekeys: I am more wondering, why that explicit img check was added...

src/core/Tile/Tile.tsx
mayank99: I think it may have been added for webcomponents? Not sure. But I see the CSS also targets the `> svg` so there is no class needed in React

src/core/Tile/Tile.tsx
mayank99: wait, it's breaking visuals because the button is coming from user who is likely using our `Button`

veekeys: what?
Whats breaking

mayank99: image

src/core/Tile/Tile.tsx
veekeys: weird.. As long as svg is there, its fine, but it might be also img, then its gonna be broken

src/core/Tile/Tile.tsx
veekeys: because we put users classes after ours :/

mayank99: order of classnames doesnt matter. Order of css imports does, which we cant control

src/core/utils/components/Popover.tsx
```diff
@@ -56,6 +57,15 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
     }),
   };
 
+  // Plugin to revert default tippy styles by applying iui-popover class to tippy-box
+  const revertTippyStyles = {
+    fn: () => ({
+      onCreate: (instance: Instance) => {
+        instance.popper.firstElementChild?.classList.add('iui-popover');
```
mayank99, line 64: This class is being added correctly our CSS needs a fix as we are reverting our own styles
image

bentleyvk: Created PR to fix https://github.com/iTwin/iTwinUI/pull/469

src/core/utils/components/Popover.tsx
```diff
@@ -56,6 +57,15 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
     }),
   };
 
+  // Plugin to revert default tippy styles by applying iui-popover class to tippy-box
+  const revertTippyStyles = {
```
bentleyvk, line 61: You can just add `className` to `<Tippy />`

src/core/Tile/Tile.tsx
veekeys: we get such a weird situation then.. All the mixins we are reusing in our components might mean nothing in react.. if user is reusing our existing components and the css gets overriden :/

src/core/utils/components/Popover.tsx
mayank99: Done

veekeys (approved):

PR #539, opened 2022-02-07
feat(Combobox): Virtualization

### LINKED ISSUE #469: ComboBox Virtualization
#### Feature
For performance reasons in the data-viz service, we need virtualization (ie only rendering rows that are "visible" in the dropdown) for combobox dropdowns.

#### Examples
See the react-virtualized-select implementation of a combobox dropdown that virtualizes everything in the dropdown.
React-virtualized-select:
https://github.com/bvaughn/react-virtualized-select

Demo:
http://bvaughn.github.io/react-virtualized-select/

veekeys:
- Added new `enableVirtualization` flag to use `VirtualScroll` component for the menu items
- Changed `menuItems` return type to be always array instead of `JSX.Element | JSX.Element[]`
- Changed `scrollIntoView` behavior. It was always aligning bottom element, which resulted in weird scrolling up with keyboard behavior, where focus always stays on the bottom element. Changed now to `nearest` so walking through the menu items with keyboard looks more intuitive
- Added new prop `scrollToIndex` to have initial render according to the selected element and it recalculates whenever `scrollToIndex` changes and element is not visible.
This needed to add some additional calculations like first and last child + include margins in height calculation, because of negative margins in combobox. Also, `VirtualScroll` is responsible for the first scroll (full control), so for that I needed to disable `scrollIntoView` in `Combobox`
- Changed keyboard nav in combobox, so it works when focused element is not visible/rendered. Now it is using `SelectOption` instead of actual dom elements.
- Added new prop `as` to `VirtualScroll` so it is rendered as `ul` in Combobox.. Just it is bad now, because of styling and applying `iui-menu` twice..
- Changed height of select and comobobox, so there is no empty space

Issues
~~- Semantics
image~~
~~- on first scroll down with keyboard
image~~
~~- if you firstly scroll with mouse so first element is not rendered and then try keyboard nav, its broken :/~~

Closes #469


src/core/ComboBox/ComboBox.tsx
```diff
@@ -356,7 +385,15 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             role='listbox'
             ref={menuRef}
           >
-            {menuItems}
+            {enableVirtualization ? (
+              <VirtualScroll
+                itemsLength={menuItems.length}
+                itemRenderer={(index) => menuItems[index] ?? <></>}
+                scrollToIndex={focusedIndex}
```
mayank99, line 392: There is a state update happening after the menu is closed
image

I haven't investigated but I'm guessing it has to do with this focusedIndex passed into VirtualScroll

src/core/ComboBox/ComboBox.tsx
```diff
@@ -128,6 +137,14 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
   const menuRef = React.useRef<HTMLUListElement>(null);
   const toggleButtonRef = React.useRef<HTMLSpanElement>(null);
   const [isOpen, setIsOpen] = React.useState(false);
+  // used to dismiss scrolling into view on first menu open
+  // if virtualization is enabled
+  const isFirstRender = React.useRef<boolean>(false);
```
mayank99, line 142: So the menu is opened and this becomes true? But it's still true on subsequent renders so the name is a bit confusing.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -41,8 +52,16 @@ const getNumberOfNodesInHeight = (childHeight: number, totalHeight: number) => {
...
+const getTranslateValue = (
+  childHeight: number,
+  firstChildHeight: number,
+  startIndex: number,
+) => {
+  if (startIndex > 0) {
+    return childHeight * (startIndex - 1) + firstChildHeight;
+  }
+
+  return 0;
```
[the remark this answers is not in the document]
veekeys: hm.. I see the same visuals in both virtualized and basic stories
Basic
image

Virtualized
image

src/core/ComboBox/ComboBox.tsx
veekeys: its false, when after 1st menu open focused element is reached in `menuItems` :D

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +124,188 @@ export type VirtualScrollProps = {
...
+      if (scrollToIndex > 0 && scrollContainer.current) {
+        const scrollTop =
+          (scrollToIndex - 1) * childHeight.current.child +
+          childHeight.current.firstChild;
+        scrollContainer.current.scrollTo({ top: scrollTop });
+      } else {
+        updateVirtualScroll();
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [scrollContainerHeight]);
```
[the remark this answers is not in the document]
veekeys: well.. I know, but this is supposed to be run only on first render and on resize, to recalculate things.. I dont want to run it if input of `scrollToIndex` is changed (support for user controlled virtual scroll).. Maybe I could add `updateVirtualScroll` as dep.. not sure, need to think :D

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +124,188 @@ export type VirtualScrollProps = {
...
+      parentRef.current.style.transform = `translateY(${getTranslateValue(
+        childHeight.current.child,
+        childHeight.current.firstChild,
+        startIndex,
+      )}px)`;
+    }, [bufferSize, itemsLength]);
+
+    const onScroll = React.useCallback(() => {
+      updateVirtualScroll();
+    }, [updateVirtualScroll]);
```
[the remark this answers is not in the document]
veekeys: This function is for scroll event listener only, wanted to make clear difference between first render run and then every run on scroll

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +124,188 @@ export type VirtualScrollProps = {
...
+          childHeight.current.firstChild;
+        scrollContainer.current.scrollTo({ top: scrollTop });
+      } else {
+        updateVirtualScroll();
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [scrollContainerHeight]);
+
+    return (
       <div
```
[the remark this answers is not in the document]
veekeys: So this is the biggest issue.. you know how it ended up for us with polymorphism in certain client apps...

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +124,188 @@ export type VirtualScrollProps = {
...
         style={{
-          willChange: 'transform',
+          overflow: 'hidden',
+          minHeight:
+            Math.max(itemsLength - 2, 0) * childHeight.current.child +
+            childHeight.current.firstChild +
+            childHeight.current.lastChild,
```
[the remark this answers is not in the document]
veekeys: fixed

src/core/ComboBox/ComboBox.tsx
veekeys: As for now, cant find anyhow, where is the issue :/
Tried commenting all the useEffect and setStates, but could not get the result without error..Went thourgh VS, Combobox and Menu components.. Unless I still missed something

veekeys: Fixed 🎉

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
mayank99: react-window has the same
image
https://react-window.vercel.app/#/api/FixedSizeList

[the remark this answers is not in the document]
mayank99: What if we expose a hook as an alternative to the `VirtualScroll` component?

```tsx
const { outerProps, innerProps } = useVirtualization();

return (
  <div {...outerProps}>
    <ul {...innerProps}>{...}</ul>
  </div>
);
```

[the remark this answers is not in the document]
veekeys: Added prop `as`..
Looks really bad on combobox, situation with menu and styles..
Can you help me define the correct typing for `parentProps` .. ?

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
veekeys: Now there is no disabling!

src/core/ComboBox/ComboBox.tsx
bentleyvk: Is it still a problem?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -241,39 +256,50 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-        ({ id }) =>
-          id === inputRef.current?.getAttribute('aria-activedescendant'),
-      );
       switch (event.key) {
         case 'ArrowDown':
           if (isOpen) {
-            const nextIndex = Math.min(
-              focusedIndexInFilteredList + 1,
-              focusableOptions.length - 1,
+            const filteredFocusedIndex =
```
bentleyvk, line 262: Why so many finds?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
+      if (!scrollContainerHeight) {
+        return;
+      }
+
+      const scrollableContainer =
+        scrollContainer.current ??
+        (parentRef.current?.ownerDocument.scrollingElement as HTMLElement);
+
+      // if `scrollToIndex` is not visible, scroll to it
+      if (
```
bentleyvk, line 294: I think `scrollToIndex` doesn't need default value, it should be `undefined` if not set. Because it is scrolling to 0 if nothing is passed. Open virtualized Table story, scroll a bit down and resize the window -> table will scroll to the first row.

bentleyvk: Now I am thinking maybe there is a problem with passed `scrollToIndex` after you scroll down and resize, it should scroll to the previous index.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
+      updateVirtualScroll();
+    }, [scrollContainerHeight, updateVirtualScroll]);
+
+    return (
       <div
         style={{
-          willChange: 'transform',
+          overflow: 'hidden',
+          minHeight:
+            itemsLength > 1
```
bentleyvk, line 329: Is this check needed? If there are no items, length === 0, that means 0 * height is 0.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
-      );
-    } else {
-      scrollContainer.current.addEventListener('scroll', updateVirtualScroll);
-    }
-    return removeScrollListener;
-  }, [updateVirtualScroll, removeScrollListener]);
-
-  React.useLayoutEffect(() => {
-    updateVirtualScroll();
-  }, [scrollContainerHeight, itemsLength, updateVirtualScroll]);
```
bentleyvk, line 219: Was it intentional to remove `itemsLength`?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
+      }
+      return removeScrollListener;
+    }, [onScroll, removeScrollListener]);
+
+    React.useLayoutEffect(() => {
+      if (!scrollContainerHeight) {
+        return;
+      }
+
+      const scrollableContainer =
```
bentleyvk, line 289: Maybe have a getter function for `scrollableContainer` as you have it in several places.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
+          scrollToIndex < visibleIndex.current.start)
+      ) {
+        const indexDiff =
+          scrollToIndex > visibleIndex.current.end
+            ? scrollToIndex - visibleIndex.current.end
+            : scrollToIndex - visibleIndex.current.start;
+
+        // smallest scroll to the index
+        const scrollTop =
+          scrollableContainer.scrollTop + indexDiff * childHeight.current.child;
```
bentleyvk, line 306: Idk, but something seems wrong here as you are multiplying by the child height and then dividing and multiplying again 😀

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -74,6 +92,20 @@ export type VirtualScrollProps = {
...
+  scrollToIndex?: number;
+  /**
+   * Tag of the parent element. In most cases default `div` should be good enough.
+   * @default 'div'
+   */
+  as?: React.ElementType;
+  /**
+   * Props object to set on parent element.
+   */
+  parentProps?: React.ComponentPropsWithRef<T>;
```
bentleyvk, line 108: It can be a bit misleading as there are 2 parents. But I am not sure is this really a problem 😀

src/core/utils/components/VirtualScroll.tsx
mayank99: See https://github.com/iTwin/iTwinUI-react/pull/539#discussion_r800798768

src/core/ComboBox/ComboBox.tsx
```diff
@@ -322,8 +350,19 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      const focusScrollRef = (el: HTMLElement) => {
+        if (
+          enableVirtualization &&
+          isFirstRender.current &&
+          focusedIndex === index
+        ) {
+          isFirstRender.current = false;
+          return;
+        }
+        !enableVirtualization &&
```
bentleyvk, line 362: Use normal `if`

src/core/Select/Select.tsx
```diff
@@ -289,7 +289,7 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
         style={{
           minWidth,
           maxWidth: `min(${minWidth * 2}px, 90vw)`,
-          maxHeight: `300px`,
+          maxHeight: 297,
```
bentleyvk, line 292: I think height should really be that half of item would be visible 😀 I know there were user raised issues in Android or iOS with share menu or somewhere else that it is not clear that there are more items:
image
But maybe it is ok when scroll is visible.

mayank99: i agree
image
image

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +132,224 @@ export type VirtualScrollProps = {
...
+      parentProps,
+      ...rest
+    },
+    ref,
+  ) => {
+    const [startNode, setStartNode] = React.useState(0);
+    const [visibleNodeCount, setVisibleNodeCount] = React.useState(0);
+    const scrollContainer = React.useRef<HTMLElement>();
+    const parentRef = React.useRef<HTMLDivElement>(null);
+    const childHeight = React.useRef({ firstChild: 0, child: 0, lastChild: 0 });
```
bentleyvk, line 153: Maybe rename to `first`, `middle`, `last`

src/core/Select/Select.tsx
veekeys: Then you are both free to take over this PR and fix this 😄

veekeys: my calculations were messed up when taking container height and calculating amount of items to fit in..
I guess I can try to change that now and see, maybe this one also magically starts working just fine 😟

src/core/ComboBox/ComboBox.tsx
```diff
@@ -354,8 +393,14 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     focusedIndex,
     itemRenderer,
     memoizedItems,
+    enableVirtualization,
   ]);
 
+  const virtualizedItemRenderer = React.useCallback(
+    (index) => menuItems[index] ?? <></>,
```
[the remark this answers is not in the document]
veekeys: removed

src/core/ComboBox/ComboBox.tsx
```diff
@@ -241,39 +256,50 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
           break;
         case 'ArrowUp':
-          if (isOpen) {
-            const previousIndex = Math.max(focusedIndexInFilteredList - 1, 0);
-            setFocusedIndex(
-              options.findIndex(
-                (_, index) =>
-                  getOptionId(index) === focusableOptions[previousIndex].id,
-              ),
+          if (isOpen && focusedIndex > 0) {
```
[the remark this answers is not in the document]
veekeys: amm, i dont know. For down findIndex works fine cause it will stop with the first find, and for up need reverse cycle, so used simple for..

src/core/utils/components/VirtualScroll.tsx
veekeys: `itemsLength` is in deps of `updateVirtualScroll`, so should work same.

src/core/utils/components/VirtualScroll.tsx
veekeys: left as undefined, if nothing is passed. It fixes table bug with resize...
2nd part will need to see, how to fix

src/core/utils/components/VirtualScroll.tsx
veekeys: updated description a little

src/core/utils/components/VirtualScroll.tsx
veekeys: Added

src/core/utils/components/VirtualScroll.tsx
veekeys: changed a little

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
veekeys: now pushed :D

src/core/Select/Select.tsx
veekeys: Ok, changed max height but didnt hack with the scroll down..
Check it up

src/core/utils/components/VirtualScroll.tsx
veekeys: Fixed 2nd part...

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+    }, [itemsLength, itemRenderer, bufferSize, startNode, visibleNodeCount]);
+
+    // Get child height when children available
+    React.useLayoutEffect(() => {
+      if (!parentRef.current || !visibleChildren.length) {
+        return;
+      }
+
+      const firstChild = parentRef.current.children.item(0) as HTMLElement;
+      const child = parentRef.current.children.item(1) as HTMLElement;
```
mayank99, line 210: nit: maybe rename this to `secondChild`. I got confused with all the different children in here

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+    const scrollContainer = React.useRef<HTMLElement>();
+    const parentRef = React.useRef<HTMLDivElement>(null);
+    const childHeight = React.useRef({ first: 0, middle: 0, last: 0 });
+    const onScrollRef = React.useRef<(e: Event) => void>();
+    // Used only to recalculate on resize
+    const [scrollContainerHeight, setScrollContainerHeight] = React.useState(0);
+    const visibleIndex = React.useRef({ start: 0, end: 0 });
+    // Used to mark when scroll container has height (updated by resize observer)
+    // because before that calculations are not right
+    const [isMounted, setIsMounted] = React.useState(false);
```
mayank99, line 159: Shouldn't this be a ref? Or does it take too long to update and you need it to re-render later?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+      ) {
+        const indexDiff =
+          scrollToIndex > visibleIndex.current.end
+            ? scrollToIndex - visibleIndex.current.end
+            : scrollToIndex - visibleIndex.current.start;
+
+        // scrollTop value to display full item height
+        const adjustedScrollTop =
+          scrollableContainer.scrollTop -
+          (scrollableContainer.scrollTop % childHeight.current.middle);
```
mayank99, line 314: What does this calculation do? Not sure I understand.

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+
+        // smallest scroll to the index
+        scrollableContainer.scrollTo({
+          top: adjustedScrollTop + indexDiff * childHeight.current.middle,
```
mayank99, line 318: I'm noticing a weird issue If you scroll down past like 20 items and then start scrolling back up, then the first item doesn't come back into view but scroll up and the second item comes back into view, and it keeps alternating.
virtual-scroll-top

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
mayank99: but how is it different from before?
image

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
mayank99: it's a weird approach you've taken where `as` prop stays in the main props but the rest of the props are in `parentProps`. Now you can't use all the utility types we made in #409

Hard to say how to get correct typings but you can get default div typings if you do this:
image

I still think the hook approach will be a lot simpler. And it's not too much work. See [`24e21df`](https://github.com/iTwin/iTwinUI-react/commit/24e21df3f30354978e2161479f3b4b191b7e165a?w=1).

src/core/utils/components/VirtualScroll.tsx
veekeys: Sure you rebuilt with newest changes? I know I had this issue and I think I fixed it..
Tried doing same as in GIF, but could not reproduce.

src/core/utils/components/VirtualScroll.tsx
veekeys: I need my useEffects to run, when this changes, so cannot be a ref..

src/core/ComboBox/ComboBox.tsx
```diff
@@ -249,39 +256,50 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
 
   const onKeyDown = React.useCallback(
     (event: React.KeyboardEvent) => {
-      const focusableOptions = getFocusableElements(menuRef.current);
-      const focusedIndexInFilteredList = focusableOptions.findIndex(
-        ({ id }) =>
-          id === inputRef.current?.getAttribute('aria-activedescendant'),
```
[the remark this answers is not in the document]
veekeys: How it can be kept, if virtual scroll does not render active element? (i.e. if you open menu and scroll down with mouse)

src/core/ComboBox/ComboBox.tsx
```diff
@@ -392,20 +419,44 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-            </Menu>
+            <>
+              {enableVirtualization ? (
+                <div
+                  style={{
+                    minWidth,
+                    maxWidth: `min(${minWidth * 2}px, 90vw)`,
+                    maxHeight: 315,
+                  }}
+                  className='iui-menu iui-scroll'
```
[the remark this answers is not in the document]
veekeys: It is intentional, for styling and scroll... If I remove this, I could add scroll through style, but i.e. box-shadow is also needed to be added here :/

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -89,5 +109,83 @@ it('should not crash with empty list items', () => {
...
     </div>,
   );
+  act(() => triggerResize({ height: 400 } as DOMRectReadOnly));
+
   expect(container.querySelectorAll('.element').length).toBe(20);
 });
+
+it('should show provided index on first render', () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  heightsMock.mockImplementation(function (this: Record<string, any>) {
```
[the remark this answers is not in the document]
veekeys: image

src/core/utils/components/VirtualScroll.test.tsx
```diff
@@ -89,5 +109,83 @@ it('should not crash with empty list items', () => {
...
+          <li key={index} className='element' style={{ height: 40 }}>{`Element${
+            index + 1
+          }`}</li>
+        )}
+        as='ul'
+        parentProps={{ className: 'customClass' }}
+      />
+    </div>,
+  );
+  act(() => triggerResize({ height: 400 } as DOMRectReadOnly));
```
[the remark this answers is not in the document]
veekeys: It is needed to fully imitate VS working.. So I wanted this to be in ever test to ensure test is valid (functionality and rendering). 
So Id better leave it..

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -33,6 +34,17 @@ const getElementHeight = (element: HTMLElement | undefined) => {
...
 
+const getElementHeightWithMargins = (element: HTMLElement | undefined) => {
+  if (!element) {
+    return 0;
+  }
+
+  const margin =
+    parseFloat(getElementStyle(element, 'margin-top')) +
+    parseFloat(getElementStyle(element, 'margin-bottom'));
+  return getElementHeight(element) + (isNaN(margin) ? 0 : margin);
```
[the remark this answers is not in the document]
veekeys: Where did you find this recommendation?
I would think opposite

image

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+      );
+      scrollContainer.current = scrollableParent;
+
+      resizeRef(scrollableParent);
+    }, [resizeRef]);
+
+    // Stop watching resize, when virtual scroll is unmounted
+    React.useLayoutEffect(() => {
+      return () => resizeObserver?.disconnect();
+    }, [resizeObserver]);
```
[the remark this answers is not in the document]
veekeys: so good, wont trigger this effect many times.
Or you suggest removing it from deps?

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
+
+    const getScrollableContainer = () =>
       scrollContainer.current ??
       (parentRef.current?.ownerDocument.scrollingElement as HTMLElement);
```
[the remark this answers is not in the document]
veekeys: I dont know, what to say

https://github.com/iTwin/iTwinUI-react/pull/539#r816825203

Something was not working if I dont use scrollingELement.. But it is a long time I have tested anything on body scroll :/

src/core/utils/components/VirtualScroll.tsx
veekeys: If at that time, scroll is somehow in the middle (displaying not full height element), I want to adjust the `scrollTop` value so I scroll the amount of pixels, that first element is displayed fully (not in the middle or just a part of it)

image

src/core/ComboBox/ComboBox.tsx
```diff
@@ -392,20 +419,44 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+                >
+                  <VirtualScroll
+                    itemsLength={menuItems.length}
+                    itemRenderer={virtualizedItemRenderer}
+                    scrollToIndex={focusedIndex}
+                    as='ul'
```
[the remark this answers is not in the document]
veekeys: I dont know.. If I put like that, something gets broken and scroll doesnt work ( I think something connected with `setFocus` done in `Menu`)

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
veekeys: Default I know, but I need this to react also to any custom element...
I did this on purpose, keeping all the props going to the root element and having a separate prop for the parent.. So much simpler and hopefully avoiding bugs (same as had with `Button`)..
Just could not figure the typings :/

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,145 +131,232 @@ export type VirtualScrollProps = {
...
-    resizeRef(scrollableParent);
-  }, [resizeRef]);
+    const onResize = React.useCallback(({ height }) => {
+      if (height > 0) {
+        setIsMounted(true);
+      }
+      setScrollContainerHeight(height);
+    }, []);
+    const [resizeRef, resizeObserver] = useResizeObserver(onResize);
+    const Component = as;
```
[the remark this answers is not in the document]
veekeys: moved

src/core/utils/components/VirtualScroll.tsx
veekeys: renamed

src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
veekeys: I tried the hook without any additional changes, but it does not work. Because VS gets mounted before even Menu is opened. And then nothing recalculates, when menu is opened..
Can I do conditional hook somehow ( I think not ) or I will need to add some additional logic to VS component...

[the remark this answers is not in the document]
mayank99: conditional hooks can be emulated by created a wrapper component which is conditionally rendered
image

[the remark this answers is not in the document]
veekeys: added

src/core/utils/components/VirtualScroll.tsx
veekeys: Ok, pushed something...
Thats the best I can do this month, check up current situation :/

src/core/utils/components/VirtualScroll.tsx
veekeys: Have a little different now.. Read, if its readable :D

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
veekeys: `as` was removed 😢

stories/core/ComboBox.stories.tsx
```diff
@@ -497,3 +497,24 @@ WithCustomMessageIcon.args = {
     <StatusMessage startIcon={<SvgCamera />}>This is a message</StatusMessage>
   ),
 };
+
+export const Virtualized: Story<Partial<ComboBoxProps<string>>> = (args) => {
```
adamhock, line 501: I missed the explanation of why the `WithMessage` story was removed? I think there should be a note about it in the PR

src/core/utils/components/VirtualScroll.tsx
```diff
@@ -42,7 +53,11 @@ const getNumberOfNodesInHeight = (childHeight: number, totalHeight: number) => {
 };
 
 const getTranslateValue = (childHeight: number, startIndex: number) => {
-  return childHeight * startIndex;
+  if (startIndex > 0) {
```
adamhock, line 56: I couldn't find a case where startIndex isn't greater than 0. Was this `if` statement added just for clarification?

veekeys:
@bentleyvk @mayank99 @adamhock I added virtualization to new Combobox. Please check it up again.

src/core/utils/components/VirtualScroll.tsx
veekeys: Wont be able to remember, to be honest.. but might have been some edge case or so :/

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,107 @@
...
+  VirtualScrollProps,
+  mergeRefs,
+  getWindow,
+} from '../utils';
 import { ComboBoxStateContext, ComboBoxRefsContext } from './helpers';
 
 type ComboBoxMenuProps = Omit<MenuProps, 'onClick'> &
   React.ComponentPropsWithoutRef<'ul'>;
 
+const VirtualComboboxMenu = ({ children, ...props }: VirtualScrollProps) => {
```
bentleyvk, line 22: It has no shadow

bentleyvk, line 22: Why it doesn't have shadow?

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,107 @@
...
+    const styles = React.useMemo(
+      () => ({
+        minWidth,
+        maxWidth: `min(${minWidth * 2}px, 90vw)`,
+        maxHeight: 315,
+      }),
+      [minWidth],
+    );
+
+    const overflowY = getWindow()?.CSS?.supports?.('overflow-x: overlay')
```
bentleyvk, line 69: Why needed?

bentleyvk: Can't you reuse `iui-scroll`?

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,107 @@
...
+            role='listbox'
+            ref={refs}
+            className={cx('iui-scroll', className)}
+            {...rest}
+          />
+        ) : (
+          <Surface style={{ ...styles, ...(overflowY as React.CSSProperties) }}>
+            <VirtualComboboxMenu
+              itemsLength={filteredOptions.length}
+              itemRenderer={(index) => getMenuItem(filteredOptions[index])}
```
bentleyvk, line 89: Don't you need to wrap with `useCallback`?

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,20 +120,65 @@ export type VirtualScrollProps = {
...
   const onScrollRef = React.useRef<(e: Event) => void>();
   // Used only to recalculate on resize
   const [scrollContainerHeight, setScrollContainerHeight] = React.useState(0);
+  const visibleIndex = React.useRef({ start: 0, end: 0 });
+  // Used to mark when scroll container has height (updated by resize observer)
+  // because before that calculations are not right
+  const [isMounted, setIsMounted] = React.useState(false);
 
   const onResize = React.useCallback(({ height }) => {
+    if (height > 0) {
```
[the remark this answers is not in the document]
veekeys: Want to share, what you remember? :D

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: Surface was supposed to provide one...

mayank99: Surface doesn't have elevation by default, you need to pass it as a prop

veekeys: oh..good catch! somehow I forgot we reset the css default.. 
Added

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: wrapped

veekeys:
> 1. Select any item from virtualized list
> 2. Press backspace
>    Result:
>    image

Woow, nice one!!!
I think fixed and added test...

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,121 @@
...
+            id={`${id}-list`}
+            style={{ ...styles, ...style }}
+            setFocus={false}
+            role='listbox'
+            ref={refs}
+            className={cx('iui-scroll', className)}
+            {...rest}
+          />
+        ) : (
+          <Surface
```
bentleyvk, line 97: Maybe just move all virtual scroll related stuff to `VirtualComboboxMenu`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -314,13 +339,9 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
             <ComboBoxMenu>
-              {filteredOptions.length > 0 ? (
-                filteredOptions.map(getMenuItem)
-              ) : (
-                <MenuExtraContent>
-                  <Text isMuted>{emptyStateMessage}</Text>
-                </MenuExtraContent>
-              )}
+              {filteredOptions.length > 0
+                ? filteredOptions.map(getMenuItem)
```
bentleyvk, line 343: Won't this end up generating react elements for all options no matter whether you have virtualisation enabled or not?

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,20 +120,65 @@ export type VirtualScrollProps = {
...
+      <div {...innerProps}>{visibleChildren}</div>
+    </div>
+  );
+});
+
+/**
+ * `useVirtualization` hook is using `VirtualScroll` component to render a huge amount of items.
+ * It returns `outerProps` and `innerProps`, which need to be applied on 2 container elements and `visibleChildren` which is a list of virtualized items.
+ * @example
+ * const itemRenderer = React.useCallback(() => (
```
bentleyvk, line 137: `index: number` is missing

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -35,7 +36,7 @@ const getElementHeight = (element: HTMLElement | undefined) => {
 
 const getElementHeightWithMargins = (element: HTMLElement | undefined) => {
   if (!element) {
-    return 0;
+    return undefined;
```
mayank99, line 39: why undefined?

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -396,8 +401,8 @@ export const useVirtualization = (props: VirtualScrollProps) => {
     } as React.HTMLAttributes<HTMLElement>,
     innerProps: {
       style: { willChange: 'transform' },
-      ref: parentRef,
-    } as React.HTMLAttributes<HTMLElement>,
+      ref: mergeRefs(parentRef),
```
mayank99, line 404: Maybe add a comment here because `mergeRefs` looks unnecessary at first glance

something like this:
```suggestion
      ref: mergeRefs(parentRef), // converts object ref into callback ref for better types
```

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,121 @@
...
+        minWidth,
+        maxWidth: `min(${minWidth * 2}px, 90vw)`,
+        maxHeight: 315,
+      }),
+      [minWidth],
+    );
+
+    const overflowY = getWindow()?.CSS?.supports?.('overflow-x: overlay')
+      ? { overflowY: 'overlay' }
+      : { overflowY: 'auto' };
```
mayank99, line 72: Do you plan to move this to CSS later?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -302,7 +318,16 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-          value={{ id, minWidth, isOpen, focusedIndex }}
+          value={{
+            id,
+            minWidth,
+            isOpen,
+            focusedIndex,
+            enableVirtualization,
+            filteredOptions,
+            getMenuItem,
+            emptyStateContent,
```
mayank99, line 329: Now it looks like too much stuff in context. `emptyStateContent` at least could be gotten rid of because we already pass it as `children` of `<ComboBoxMenu>`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -314,13 +339,9 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
           <ComboBoxDropdown {...dropdownMenuProps}>
             <ComboBoxMenu>
-              {filteredOptions.length > 0 ? (
-                filteredOptions.map(getMenuItem)
-              ) : (
-                <MenuExtraContent>
-                  <Text isMuted>{emptyStateMessage}</Text>
-                </MenuExtraContent>
-              )}
+              {filteredOptions.length > 0
```
mayank99, line 342: Maybe need to check for virtualization here as it doesn't need all those extra menu items in `children`

```suggestion
              {filteredOptions.length > 0 && !enableVirtualization
```

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,121 @@
...
+  VirtualScrollProps,
+  mergeRefs,
+  getWindow,
+} from '../utils';
 import { ComboBoxStateContext, ComboBoxRefsContext } from './helpers';
 
 type ComboBoxMenuProps = Omit<MenuProps, 'onClick'> &
   React.ComponentPropsWithoutRef<'ul'>;
 
+const VirtualComboboxMenu = ({ children, ...props }: VirtualScrollProps) => {
```
mayank99, line 22: nit: can you move this to the bottom of the file? I think the exported component should come first

My preferred order is this:
```
imports
constants
types
named exports
helpers // <-- VirtualComboBoxMenu goes here
default export
```

But it's a nit, so upto you if you have a different preference.

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,121 @@
...
+            ...(children as JSX.Element).props.style,
+          },
+        },
+        visibleChildren,
+      )}
+    </div>
+  );
+};
+
 export const ComboBoxMenu = React.forwardRef(
```
mayank99, line 45: With the current API, virtualization will not be possible to use when we export subcomponents for composition.

I think we might want to move some of the things from context into props so that the user can easily do something like this:

```tsx
<ComboBox>
  <Combobox.Input />

  <ComboBox.Menu
     enableVirtualization
     length={myOwnOptions.length}
     itemRenderer={(index) => (
       <ComboBox.MenuItem>{myOwnOptions[index]?.label}</ComboBox.MenuItem>
     )}
     emptyStateMessage='Nothing to see here'
  />
</ComboBox>
```

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: also why is it called `VirtualComboboxMenu` and not `VirtualizedComboBoxMenu`?

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -1,38 +1,121 @@
...
+    } = useSafeContext(ComboBoxStateContext);
     const { menuRef } = useSafeContext(ComboBoxRefsContext);
 
     const refs = useMergedRefs(menuRef, forwardedRef);
 
+    const styles = React.useMemo(
+      () => ({
+        minWidth,
+        maxWidth: `min(${minWidth * 2}px, 90vw)`,
+        maxHeight: 315,
```
mayank99, line 65: Maybe this height change from `300` to `315` should go in a different PR?

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -100,20 +120,65 @@ export type VirtualScrollProps = {
...
+
+  return (
+    <div {...outerProps} ref={ref}>
+      <div {...innerProps}>{visibleChildren}</div>
+    </div>
+  );
+});
+
+/**
+ * `useVirtualization` hook is using `VirtualScroll` component to render a huge amount of items.
```
mayank99, line 134: This wording looks wrong. `useVirtualization` is not using `VirtualScroll`, it's the other way around.

You could skip any mention of `VirtualScroll` and just explain what it does.
```suggestion
 * `useVirtualization` hook is used for efficiently rendering only the visible rows from a large list.
```

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -84,6 +84,13 @@ export type ComboBoxProps<T> = {
       index: number;
     },
   ) => JSX.Element;
+  /**
+   * Virtualization is used for the scrollable dropdown list.
```
mayank99, line 88:
```suggestion
   * If enabled, virtualization is used for the scrollable dropdown list.
```

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: No

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: I left it at the top as I usually put helpers before the usage. We have multiple files with this order, but it is really preference, not a set rule.
I have renamed the component.

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
veekeys: I thought it makes more sense. If element is not existing (i.e. there is single child and I am checking for the 2nd childs height) I should not get height value at all. And then check I had for `childHeights` is easier, to assign first childs height to all of the children.
I better return height `0` when element is existing but it is empty, for example.

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: amm... But this can be done later, right? I dont think we have already done everything for Combobox to work with composition approach. You have even removed the story as an example.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Changed

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
veekeys: Changed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Added

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Removed

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
veekeys: Added

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
veekeys: Added

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: I added `enableVirtualization` check, so it only passes empty content component

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: What exactly you want me to do here? Move `Surface` and `Menu` to `VirtualizedComboboxMenu`? But then I would need to pass all the props through in there (for example `style` prop needs to be passed different for `Menu` and for `Surface`)

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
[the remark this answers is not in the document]
veekeys: Added

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
veekeys: Ok, I have moved everything into `VirtualizedComboBoxMenu`.
2 thins I dont like
- forwarded ref goes onto `Menu`, whereas I think it should go on `Surface`.. But different types, so not sure if I wanna hack it i.e. `mergeRefs`)
- Some repeatance of code (i.e. styles)

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: Yeah but it would be nice to avoid having to come back here and figure out any virtualization-specific logic that might need to be moved.

I removed the story because we haven't finalized and exported the subcomponents yet. I think only keyboard navigation doesn't fully work with composition. For everything else I was thinking about how to reduce future work.

veekeys: amm... Well I think current situation is same as with any other combobox component.
If you use it with Combobox and it provides the context, everything should work fine.
And if not, you need to expose needed props. But that is applicable to other parts too :/
So not sure, what is the problem or I misunderstood

bentleyvk (approved):
I really hope that I reviewed virtualization changes previously because I don't want to dive into that again 😀

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: > If you use it with Combobox and it provides the context, everything should work fine.

Not true. With this approach you need to always pass the `options` prop into `ComboBox` root, which doesn't make sense when used with subcomponents.

I guess lets revisit this when we expose subcomponents.

mayank99 (approved):

PR #540, opened 2022-02-07
feat(Select): End icon is svg

### LINKED ISSUE #465: ComboBox: icon is different than select
ComboBox and Select looks too different. So one or another should be changed to have same arrow icon.

ComboBox:
image

Select:
image

gretanausedaite:
- Updated `Select` and `ComboBox` to have new className.
- Added end icon for caret down.

Select:
image
image

Combobox:
image


gretanausedaite:
Should I add `svgIcon` prop to ComboBox (as in `LabeledSelect`) for custom message icon?

src/core/Select/Select.tsx
```diff
@@ -273,36 +276,38 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+        aria-haspopup='listbox'
+        style={style}
+        {...rest}
       >
         <div
           ref={selectRef}
           className={cx('iui-select-button', {
             'iui-placeholder': !selectedItem && !!placeholder,
             'iui-disabled': disabled,
             'iui-active': isOpen,
```
mayank99, line 309: I think `iui-active` is not used

src/core/Select/Select.tsx
```diff
@@ -196,6 +197,8 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
   const toggle = () => setIsOpen((open) => !open);
 
   const selectRef = React.useRef<HTMLDivElement>(null);
+  const toggleButtonRef = React.useRef<HTMLSpanElement>(null);
```
mayank99, line 200: Seems you are not doing anything with `toggleButtonRef`

src/core/Select/Select.tsx
```diff
@@ -196,6 +197,8 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+  const inputWithIconRef = React.useRef<HTMLDivElement>(null);
```
mayank99, line 201: I see you are using this ref for focusing select-button when icon is clicked, but that doesn't do anything. In ComboBox, the menu automatically opens when the input is focused, but that's not the case here

src/core/Select/Select.tsx
```diff
@@ -322,8 +327,25 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+            'iui-disabled': disabled,
+            'iui-open': isOpen,
+          })}
+          onClick={() => {
+            if (isOpen) {
+              setIsOpen(false);
+            } else {
+              inputWithIconRef.current?.focus();
+            }
+          }}
```
mayank99, line 343: You can replace this whole thing with just `toggle`. I had to use `onPointerDown` because `onClick` was not working for some reason (even with stopPropagation).
```suggestion
        onPointerDown={() => toggle()}
```

src/core/Select/Select.tsx
```diff
@@ -273,36 +276,38 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
   }, [options, value]);
 
   return (
-    <div
-      className={cx('iui-select', { [`iui-${size}`]: !!size }, className)}
-      aria-expanded={isOpen}
-      aria-haspopup='listbox'
-      style={style}
-      {...rest}
+    <DropdownMenu
```
mayank99, line 279: You will need to move this DropdownMenu inside the wrapper if you want the `focus-within` logic to work. Right now there is no focus ring when the menu is open.

Something like this:
image

package.json
```diff
@@ -40,7 +40,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
-    "@itwin/itwinui-css": "^0.45.0",
+    "@itwin/itwinui-css": "^0.46.0",
```
mayank99, line 43: 0.46.1 is available now

src/core/Select/Select.tsx
gretanausedaite: Removed this ref

src/core/Select/Select.tsx
gretanausedaite: removed

src/core/Select/Select.tsx
gretanausedaite: changed to toggle

src/core/Select/Select.tsx
gretanausedaite: changed structure

src/core/Select/Select.tsx
mayank99: You might want to try investigating why onClick doesn't work, because onPointerDown is not an ideal solution (it should allow canceling the action if you hold your mouse down, then move mouse away from icon, then release).

src/core/Select/Select.tsx
mayank99: This works for now, but you should start thinking about how to tackle this once we add `appendTo={document.body}` to Popover (#514)

src/core/Select/Select.tsx
bentleyvk: I think you can just do the same hack as in `Combobox`:
image
If I remember correctly, the issue here is that when you click on the icon, it is outside click for the `DropdownMenu` and then is like `outside click => close => icon click => open`. Possible fix could be moving `DropdownMenu` to make it as the root element but we moved it in the current place to make it easier to query dropdown in tests. But if we plan to do `appendTo={document.body}` change, then it shouldn't matter as dropdown will be completely separate from `Select`.

src/core/Select/Select.tsx
bentleyvk: > This works for now, but you should start thinking about how to tackle this once we add `appendTo={document.body}` to Popover (#514)

Should we add some class to make it look like focused?

src/core/Select/Select.tsx
```diff
@@ -323,6 +324,18 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
         </div>
       </DropdownMenu>
+      <span
+        className={cx('iui-end-icon', {
+          'iui-actionable': !disabled,
+          'iui-disabled': disabled,
+          'iui-open': isOpen,
+        })}
+        onPointerDown={() => {
+          toggle();
```
bentleyvk, line 334: If you leave this, don't you need to check whether it is disabled?

src/core/Select/Select.test.tsx
```diff
@@ -69,14 +69,14 @@ beforeEach(() => {
 it('should render empty select', () => {
   const { container } = renderComponent();
 
-  const select = container.querySelector('.iui-select') as HTMLElement;
+  const select = container.querySelector('.iui-input-with-icon') as HTMLElement;
```
bentleyvk, line 72: Just thought that we should have constants in the begining of the test file for common selectors, that way we won't need to go trough each line after renames. But it is just an idea for the future, no need to implement 😀

bentleyvk:
> Should I add `svgIcon` prop to ComboBox (as in `LabeledSelect`) for custom message icon?

Don't we have an item for that?

src/core/Select/Select.tsx
mayank99: Currently we moved it down one level to make `focus-within` work.

@gretanausedaite Did you try adding that onClickOutside listener?

If it doesn't work, we should add that focused class in CSS and use the wrapper div as the trigger

mayank99 (approved):
Looks good 🍟 lets do status icon in a different PR

bentleyvk (approved):

PR #541, opened 2022-02-08
docs(Table): update onSelect description

### LINKED ISSUE #522: Table: onSelect is not called when row selection is reset when data is updated
The table doesn't call onSelect when selected rows are reset when data is updated and autoResetSelectedRows is set to true.

#### Expected behavior
onSelect should be called when selected rows are reset when data is updated and autoResetSelectedRows is set to true.

#### Additional information
It seems that the action that resets the selected rows, when table data is updated, is not being handled inside the tableStateReducer in https://github.com/iTwin/iTwinUI-react/blob/449dc4229c15b040e62395a4ab49592217e3bf13/src/core/Table/Table.tsx#L327

veekeys:
Updated description after #522

### DIFF
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -109,7 +109,7 @@ export type TableProps<
   isSelectable?: boolean;
   /**
    * Handler for rows selection. Must be memoized.
-   * This is triggered only by user initiated actions (i.e. data change will not call it ).
+   * This is triggered only by user initiated actions (i.e. data change will not call it).
    */
   onSelect?: (
     selectedData: T[] | undefined,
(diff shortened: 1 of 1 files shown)


src/core/Table/Table.tsx
```diff
@@ -109,6 +109,7 @@ export type TableProps<
   isSelectable?: boolean;
   /**
    * Handler for rows selection. Must be memoized.
+   * This is triggered only by user initiated actions (i.e. data change will not call it ).
```
mayank99, line 112: nit: whitespace
```suggestion
   * This is triggered only by user initiated actions (i.e. data change will not call it).
```

mayank99 (approved):

PR #544, opened 2022-02-10
fix(Footer): Override default elements

### LINKED ISSUE #223: Footer component : we can't override default elements

Currently we can't show custom footer only, it will always shows default footer with custom elements.

#### Expected behavior

As our components are public, there should a way to override default footer.

#### Reproduction
Link to a minimal repro: 

##### Steps to reproduce
1.  Try to show only custom footer elements.

elephantcatdog:
Added prop "showOnlyCustomElements" that, when true, hides all of the default elements - copyright and links.

image

If showOnlyCustomElements is true but no custom elements are provided, the footer will be empty.

Closes #223

## Checklist


src/core/Footer/Footer.test.tsx
```diff
@@ -78,6 +78,35 @@ it('should show all default footer elements', () => {
...
+  allLi.forEach((element, i) => {
+    expect(element).toBeTruthy();
+    expect(element.textContent).toBe(customUrls[i].title);
+    expect(element.href).toBe(customUrls[i].url);
+    if (i > 0) {
+      expect((element.previousSibling as HTMLSpanElement).classList).toContain(
+        'iui-separator',
+      );
+    } else {
+      expect(element.previousSibling as HTMLSpanElement).not.toBeDefined;
```
veekeys, line 105: isnt `toBeDefined` a function?

stories/core/Footer.stories.tsx
```diff
@@ -46,6 +46,39 @@ Custom.args = {
...
 } as FooterProps;
 
+export const OnlyCustomElements: Story<FooterProps> = ({
+  customElements,
+  showOnlyCustomElements,
+  ...rest
+}: FooterProps) => {
+  return (
+    <Footer
+      customElements={customElements}
```
veekeys, line 56: For better code readability in Storybook, having a definition of `customElements` in the story would be helpful.
`const customElements = [...]`

stories/core/Footer.stories.tsx
```diff
@@ -46,6 +46,39 @@ Custom.args = {
...
+      showOnlyCustomElements={showOnlyCustomElements}
```
veekeys, line 57: Just enough having `showOnlyCustomElements`, for the same readability

src/core/Footer/Footer.tsx
```diff
@@ -74,18 +84,27 @@ export const Footer = (props: FooterProps) => {
...
   ];
-  const elements = customElements
-    ? [...defaultElements, ...customElements]
-    : defaultElements;
+
+  let elements: FooterElement[] = [];
+  if (!showOnlyCustomElements) {
+    elements = elements.concat(defaultElements);
+  }
+  if (customElements) {
```
veekeys, line 92: Maybe something like that

```jsx
const elements = showOnlyCustomElements
    ? [...(customElements ?? [])]
    : [...defaultElements, ...(customElements ?? [])];

**Review by veekeys — COMMENTED** [review]
It should do the job, but not a fan of additional prop.
Ideally, we should provide our default list, which users could reuse + add custom.
But for that, we will need to have some breaking changes, so we can plan only for V2.

File: src/core/Footer/Footer.tsx (hunk 4 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx]: Fixed

File: stories/core/Footer.stories.tsx (hunk 3 above)
Comment (elephantcatdog) [reply stories/core/Footer.stories.tsx]: Fixed

File: stories/core/Footer.stories.tsx (hunk 2 above)
Comment (elephantcatdog) [reply stories/core/Footer.stories.tsx]: Fixed

File: src/core/Footer/Footer.test.tsx (hunk 1 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.test.tsx]: Fixed

File: stories/core/Footer.stories.tsx (hunk 2 above)
Comment (veekeys) [reply stories/core/Footer.stories.tsx]: I dont think you did it.
I just mean copying into the story code too like in this story for example
https://github.com/iTwin/iTwinUI-react/blob/c0ad0921f859130f4cc866999ba2d31187b192ac/stories/core/DropdownMenu.stories.tsx#L95
```js
const customElements: [
    {
      title: 'Custom Element 1',
      url: 'https://www.bentley.com/',
    },
    {
      title: 'Custom Element 2',
    },
    {
      title: 'Custom Element 3',
    },
    {
      title: 'Custom Element 4',
    },
  ],

File: src/core/Footer/Footer.tsx [hunk 5]
```diff
@@ -24,6 +24,10 @@ export type FooterProps = {
    * Provide localized strings.
    */
   translatedTitles?: TitleTranslations;
+  /**
+   * If true, only custom elements will be shown.
+   */
+  showOnlyCustomElements?: boolean;
```
veekeys, line 30: So I was thinking about this prop and it bothers me, cause its really specific and not extendable..
I was talking with @mayank99 and he suggested something like that
```js
customElements?: FooterElement[] | (defaultElements: FooterElement[]) => FooterElement[];
```
So basically, we would accept now prop `customElements` as function, where we would pass our default ones (for first implementation might not) and expect user to give us a list of `FooterElement` which we would render straight away as is (without any checks). That way user could pass the list of needed elements and we would not include the default ones.
The only issue for `defaultElements` as function prop is the Bentley copyright part which now is toggled by the new flag. Then it would be removed, I guess?

Do you think you could try to play around a little with the API as see what you could do without a new prop, but instead extending the current API?

mayank99: Maybe we should include the Bentley copyright part in `defaultElements`

elephantcatdog: My thoughts are:

1. This is potentially more confusing and difficult to use and needs more explanation than a true/false prop.
2. Another way to implement this might be to have an `options` prop that includes `showOnlyCustomElements` and could include more options in the future, which would make it less specific and more extendable.
3. I understand that doing it your suggested way would allow users more control over which default elements are in the footer, and in what order. However, I think that there are two scenarios for the Footer: it is being used in a Bentley product or it's not. In the case of it being used in a Bentley product, the default elements are required and I think it's better for the order and placement of the default elements to be the same across all Bentley products. In the case of the Footer not being in a Bentley product, users won't need any of the default elements. So while this idea provides more flexibility, I think that flexibility is uneccessary and could potentially lead to less cohesion across Bentley products.

But I'm new to this team, so if you think that your way is better, I'll do it that way.

veekeys: 1. I do understand this is not something used widely, but actually it is not something fancy. Users either pass array or they get to implement function, where they could fully customize footer content whereas we ensure the correct styling. Explanation is needed anyway, so be sure to have a good prop description and story example will be helpful too.
Whenever you use a component, you will need to get familiar with its API and to achieve some custom/advanced case it might always need not regular implementation.
2. This is something better and worth thinking.. Just the issue with it is the usage, which is not nice. As the prop gonna be object, if user wants to change something, he will need to create a new object reference everytime to make footer react to it. If you want, try to apply this suggestion and see, how it looks. We could talk about it more.
3. That is a really logical thinking, but few points to mention from experience.
- In our default links there are some, which need authentication to be opened (check the first one, for example). We might have products, which do not want to have such legal links, cause users can fully use the product without authentication (developer portal as an example). So such products might want to customize the links, by changing URL, for example
- Bentley also has a lot of acquired products, which not necessarily are fully under Bentley and considered as Bentley product. They might still work under their own environment including legal rights and we could expect them having different links to certain aspects of legal
- And as usual, there might be some weird and not thought through use cases, where usually we just try to have as customizable solution as possible, just to be safe, because we already had some decisions made which now we already need to change. You can check our [v2 wiki page](https://github.com/iTwin/iTwinUI-react/wiki/v2-release), to see the collection of needed breaking changes.

elephantcatdog: I am writing a story for a use case where default and custom elements are being intermixed. In doing so I realized that a problem with this implementation is that it's going to be difficult for users to know what is incoming in the `defaultElements` parameter.

I initially used the array position, but for this to be intuitive for the users there will need to be documentation that is kept up to date. I know that we have documentation; does it include information to this level of detail? I feel like outlining what is in each position is too much information to be added to the prop description.

I also experimented with using `.find()` on the `defaultElements` parameter to get the element by the `title` property. For this to be simple, users will have to use a non-null assertion. Otherwise, they'll have to null check every time they use a value from `defaultElements`. There would also need to be documentation for the titles of each default element, although that might be succinct enuogh to include in the prop description.

Do you have an alternative method of using the `defaultElements` parameter?

mayank99: Honestly, I don't find the below to be *too* difficult or complicated or unreadable. Array operations are pretty common.

Only custom:
```tsx
const newFooterParts = [ /* populate this with new stuff */ ];
<Footer customElements={() => newFooterParts} />
```

Appending custom unchanged:
```tsx
const newFooterParts = [ /* populate this with new stuff */ ];
<Footer customElements={(defaultParts) => [...defaultParts, ...newFooterParts]} />
```

Filtering out:
```tsx
const translatedTiles = { /* localized */ };
<Footer
  translatedTitles={translatedTitles}
  customElements={(defaultParts) => defaultParts.filter(({ title }) => title !== translatedTitles.termsOfService)}
/>
```

Changing one url:
```tsx
const translatedTitles = { /* localized */ };
const customPrivacyUrl = '...';
<Footer
  translatedTitles={translatedTitles}
  customElements={(defaultParts) => defaultParts.map(d => ({ ...d, url: d.title === translatedTitles.privacy ? customPrivacyUrl : d.url }))}
/>
```

Changing all urls:
```tsx
const translatedTitles = { /* localized */ };
const customUrls = { [translatedTitles.privacy]: '...', /* other urls using values of titles as keys */  };
<Footer
  translatedTitles={translatedTitles}
  customElements={(defaultParts) => defaultParts.map(({ title, url }) => ({ title: title, url: customUrls[title] ?? url }))}
/>
```

However, I agree that using `title` is weird so maybe we could expose a `key` in each of our defaultElements. It would simplify the last example like this (notice how we don't even need the `translatedTitles` prop anymore):
```tsx
const translatedTitles = { /* localized */ };
const customUrls = { privacy: '...', /* urls using actual keys */  };
<Footer
  customElements={(defaultParts) => defaultParts.map(({ key }) => ({ key, title: translatedTitles[key], url: customUrls[key] }))}
/>
```

src/core/Footer/Footer.test.tsx
```diff
@@ -45,6 +45,10 @@ const customUrls: FooterElement[] = [
   },
 ];
 
+const customUrlsFunction = (defaultElements: FooterElement[]) => {
```
veekeys, line 48: How did you manage to push with linter warnings? I thought we have a hook for that :/

src/core/Footer/Footer.tsx
```diff
@@ -19,15 +20,13 @@ export type FooterProps = {
   /**
    * Customize footer elements.
```
veekeys, line 21: Some additional description would be useful. Can have even example in here (or in `Footer` description below

src/core/Footer/Footer.test.tsx
```diff
@@ -54,15 +68,6 @@ it('should show all default footer elements', () => {
 });
 
 it('should show all default footer elements', () => {
```
veekeys, line 70: Same test case name? :D

src/core/Footer/Footer.tsx
veekeys: This is a very nice list of examples!
need to store it somewhere :D
Title looks weird to me too and story has null assertions all over the place now.

stories/core/Footer.stories.tsx
```diff
@@ -30,21 +30,91 @@ Basic.args = {
...
+      {
+        title: 'Custom Element 4',
+      },
+    ];
+  };
+
+  return <Footer customElements={customElements} {...rest} />;
+};
+
+export const CustomDefaultElements: Story<FooterProps> = ({
```
veekeys, line 69: Doest work for me
image

Maybe rename also to something like `DefaultAndCustomElements`, so its clear you are merging those

src/core/Footer/Footer.tsx
elephantcatdog: These are pretty good solutions for most cases, and the key is an excellent idea. However, this doesn't solve the difficulty of reordering elements. I suppose that's not a big concern though.

src/core/Footer/Footer.test.tsx
elephantcatdog: Fixed

src/core/Footer/Footer.tsx
elephantcatdog: Fixed

src/core/Footer/Footer.test.tsx
elephantcatdog: Fixed

stories/core/Footer.stories.tsx
elephantcatdog: I changed it to Customized Default And Custom Elements to differentiate it from Appended Custom Elements.

stories/core/Footer.stories.tsx
elephantcatdog: Ah, I fixed it this time.

src/core/Footer/Footer.tsx
veekeys: Hover over the prop to see, how it is displayed now:
image
Examples are not visible.

I suggest in description wrapping `FooterElements` with char ` for better readability that it is type.

Also, examples do not require anything additional, just:

```
* @example Returning only custom elements
   * <Footer customElements={() => newFooterElements)} />
   * @example Filtering out a specific element
   * <Footer customElements={(defaultElements: FooterElement[]) => defaultElements.filter(({ key }) => key !== 'privacy' )} />
   * @example Changing a url
   * <Footer customElements={(defaultElements: FooterElement[]) => defaultElements.map(element => ({ ...element, url: element.key === 'privacy' ? customPrivacyUrl : element.url }))} />
```

And I suggest move examples to whole component description, to have a list of those there.

src/core/Footer/Footer.tsx
veekeys: But `key` type now is weird.. What if user passes custom elements? key will not be from translations if he wants to add it..
Then I guess something like
```js
key?: keyof TitleTranslations | string;

**Comment by veekeys** [comment]
Need to update visual tests too, they are failing

File: src/core/Footer/Footer.tsx (hunk 7 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx:21]: Huh, the examples show up for me. You're looking at it in VS Code?

I wrapped `FooterElement[]` and moved examples to component description.

Wrapping the examples with ``` does makes a difference, though. Here the first example has the description included as part of the code, while the second example does not. The word `default` is highlighted because TSDoc is treating that line as code.
image

File: src/core/Footer/Footer.tsx (hunk 5 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx]: I thought about this, but thought that the (intended) use for keys is in accessing the default elements. I can imagine that someone might want to utilize them though.

File: src/core/Footer/Footer.tsx (hunk 7 above)
Comment (veekeys) [reply src/core/Footer/Footer.tsx:21]: Yes, vs code

File: src/core/Footer/Footer.tsx (hunk 5 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx]: A negative to adding `| string` to the key definition is that it removes the ability to see what the defined keys, and thus the default elements, are. Which is important for ease of use for users.

Is it really terrible for it to stay as is?

**Review by elephantcatdog — COMMENTED** [review]
What is a long footer that goes onto the second row supposed to look like? One of the test images has an example of this. Is that correct, with a `|` starting on the second row?

File: src/core/Footer/Footer.tsx (hunk 7 above)
Comment (veekeys) [reply src/core/Footer/Footer.tsx:21]: You current syntax is just wrong.
https://jsdoc.app/tags-example.html

This is official docs. For description you can use `<caption>`, so please change it, because current one shows nothing. Maybe you have some addition tools installed in vscode, so its shown like that.

**Comment by veekeys** [comment]
> What is a long footer that goes onto the second row supposed to look like? One of the test images has an example of this. Is that correct, with a `|` starting on the second row?

I think it is fine
@FlyersPh9

File: src/core/Footer/Footer.tsx (hunk 5 above)
Comment (veekeys) [reply src/core/Footer/Footer.tsx]: well, we export this type for users to use for custom element.
And we do not allow to set any other key to them, so if they have custom element, they cannot set key (cant use the default one).
As I mentioned, either allow to pass in key anything or change the type for external use without key.

Do you have other suggestions?

Comment (mayank99) [reply src/core/Footer/Footer.tsx]: > A negative to adding `| string` to the key definition is that it removes the ability to see what the defined keys, and thus the default elements, are.

This is a bug in typescript, can be worked around by adding `| {}` or `| Record<never, never>`. See example in https://github.com/iTwin/iTwinUI-react/blob/main/src/core/Badge/Badge.tsx#L53

File: src/core/Footer/Footer.tsx (hunk 7 above)
Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx:21]: Oh, I thought we used TS Doc.

Comment (mayank99) [reply src/core/Footer/Footer.tsx:21]: TSDoc is a library that needs to be manually used. JSDoc is built into the language and IDEs.

Comment (veekeys) [reply src/core/Footer/Footer.tsx:21]: Never thought about tsdoc (and I have never heard of it, to be honest :D), but I dont think it would work for us, cause we deliver js code, so component should be described in jsdoc (thinking logically)

Comment (veekeys) [reply src/core/Footer/Footer.tsx:21]: @elephantcatdog as this was confusion, which we didnt think might happen, could you create a PR for updating our docs, that we use jsdoc? Maybe could add some more info, that every prop should have description and component should have examples...

Comment (elephantcatdog) [reply src/core/Footer/Footer.tsx:21]: As far as I can tell, I don't seem to have TSDoc manually added, though it might be a part of an extension I have. But it still works for me. But yes, I will add that to the documentation.

File: src/core/Footer/Footer.tsx [hunk 10]
```diff
@@ -35,9 +40,15 @@ export type FooterElement = {
    * URL of the footer element.
    */
   url?: string;
+  /**
+   * Key of the footer element.
+   */
+  // eslint-disable-next-line @typescript-eslint/ban-types -- This allows custom strings and keeps intellisense. See https://github.com/Microsoft/TypeScript/issues/29729
+  key?: keyof TitleTranslations | (string & {});
```
mayank99, line 47: Can avoid eslint error by using `Record<never, never>`
```suggestion
  // This allows custom strings and keeps intellisense. See https://github.com/Microsoft/TypeScript/issues/29729
  key?: keyof TitleTranslations | (string & Record<never, never>);
```

src/core/Footer/Footer.tsx
```diff
@@ -8,6 +8,7 @@ import { useTheme, StylingProps } from '../utils';
 import '@itwin/itwinui-css/css/footer.css';
 
 export type TitleTranslations = {
+  copyright: string;
```
mayank99, line 11: I don't think this `copyright` should be in TitleTranslations. Or at least it shouldn't be required.

We can still keep it in footerElements by changing the type of key to `keyof TitleTranslations | 'copyright' | (string & Record<never, never>)`

@veekeys What do you think?

src/core/Footer/Footer.tsx
```diff
@@ -49,43 +60,70 @@ const footerTranslations: TitleTranslations = {
...
+        : [...defaultElements, ...customElements];
+  }
 
   return (
     <footer className={cx('iui-legal-footer', className)} {...rest}>
       <ul>
-        <li>© {today.getFullYear()} Bentley Systems, Incorporated</li>
         {elements.map((element, index) => {
           return (
             <li key={`${element.title}-${index}`}>
```
mayank99, line 125: the key from elements can be used now
```suggestion
            <li key={key || `${element.title}-${index}`}>
```

src/core/Footer/Footer.tsx
```diff
@@ -49,43 +60,70 @@ const footerTranslations: TitleTranslations = {
...
  * You can use position 'absolute' with relative body or set the height of the content and place footer at the end.
- * @example
+ * @example <caption>Appending custom element after default elements</caption>
  * <Footer customElements={[{title: 'Bentley', url: 'https://www.bentley.com/'}]} />
+ * @example <caption>Returning only custom elements</caption>
+ * <Footer customElements={() => newFooterElements)} />
+ * @example <caption>Filtering out a specific element</caption>
+ * <Footer customElements={(defaultElements: FooterElement[]) => defaultElements.filter(({ key }) => key !== 'privacy' )} />
+ * @example <caption>Changing a url</caption>
+ * <Footer customElements={(defaultElements: FooterElement[]) => defaultElements.map(element => ({ ...element, url: element.key === 'privacy' ? customPrivacyUrl : element.url }))} />
```
mayank99, line 70: types are automatically inferred and don't need to be specified
```suggestion
 * <Footer customElements={(defaultElements) => defaultElements.filter(({ key }) => key !== 'privacy' )} />
 * @example <caption>Changing a url</caption>
 * <Footer customElements={(defaultElements) => defaultElements.map(element => ({ ...element, url: element.key === 'privacy' ? customPrivacyUrl : element.url }))} />
```

src/core/Footer/Footer.test.tsx
```diff
@@ -35,6 +35,20 @@ const urls: FooterElement[] = [
...
+    url: 'https://www.bentley.com/',
+  },
+  {
+    title: 'Products link',
+  },
+];
+
+const customUrlsFunction = () => {
+  return customUrls;
+};
```
mayank99, line 50: Is this helper needed? Could just inline it

src/core/Footer/Footer.tsx
veekeys: oh wait, right... Now users who use `TitleTranslations` will get this as breaking change!
Good catch, somehow missed it..
If they want to "translate" copyright, they most probably will use that through custom elements, cause not much to translate in copyright.
Lets remove that from here.

src/core/Footer/Footer.tsx
veekeys: Now perfect.

veekeys:
Everything looks good to me except that translation type issue. After it is fixed, feel free to merge it after Mayank approves, no need to wait for me.

src/core/Footer/Footer.tsx
elephantcatdog: These are going to be different values than before. "termsOfService" instead of "Terms of service - 1" Is that ok?

src/core/Footer/Footer.tsx
elephantcatdog: Fixed? I think

src/core/Footer/Footer.tsx
elephantcatdog: Fixed

src/core/Footer/Footer.test.tsx
elephantcatdog: Fixed

src/core/Footer/Footer.tsx
elephantcatdog: Fixed

src/core/Footer/Footer.tsx
elephantcatdog: I did it anyway.

veekeys (approved):

mayank99 (approved):

PR #547, opened 2022-02-10
feat(ComboBox): Add `itemRenderer` prop

### LINKED ISSUE #517: Improvements for iTwin Select/Combobox component
#### Feature
Hello!

We need to have select field with the search functionality in a few places in **Synchro Control**. So iTwin Combobox meets our requirements to replace our component with iTwin but there are some issues that are described further.
We have some Selects with search functionality and possibility of custom rendering for selected fields and menu items (see **Examples**).
All these requirements can be followed by iTwin Select component because it contains **selectedItemRenderer** and **itemRenderer** props to override rendering methods. But Select doesn’t support search functionality.
Search functionality is supported by iTwin Combobox component but it doesn’t support custom rendering on selected field and menu items.
Could supporting of custom renderings be added to Combobox (like in Select component) or could supporting of search functionality be added to Select (like in Combobox component)?

#### Examples
 Selected field rendering:
MicrosoftTeams-image (1)

Menu items rendering (show tooltips only on truncated text):
MicrosoftTeams-image

mayank99:
Added `itemRenderer` prop in which the user passes a function that returns custom JSX using two arguments:
1. first arg is the actual option that the user passes into `option`. This can be used to access `value` and `label`.
2. second arg is a set of states that can be put in the JSX. `isSelected` and `id` are the main ones here.
    - `isFocused` is also available in case the user doesn't want to use `MenuItem`.

Closes #517 

image

## Checklist


src/core/ComboBox/ComboBox.tsx
```diff
@@ -54,6 +55,20 @@ export type ComboBoxProps<T> = {
...
    */
   emptyStateMessage?: string;
+  /**
+   * A custom item renderer can be specified to control the rendering.
+   * This function should ideally return a customized version of `MenuItem`,
+   * otherwise you will need to make sure to provide styling for the `isFocused` state.
+   */
+  itemRenderer?: (
+    option: SelectOption<T>,
+    states: {
```
bentleyvk, line 65: Move it to separate type because I think `itemRenderer` should be memoized and therefore users won't have IDE for the types as in your story:
image

bentleyvk: I also think it should be renamed because `id` and `index` is not really a state 😀

src/core/ComboBox/ComboBox.tsx
```diff
@@ -103,25 +119,44 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        };
+        if (itemRenderer) {
+          return React.cloneElement(
+            itemRenderer(option, {
+              id: getOptionId(index),
+              index,
+              isSelected: false,
+              isFocused: false,
+            }),
+            additionalProps,
```
bentleyvk, line 145: How it will work?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -103,25 +119,44 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-        <MenuItem
-          id={getOptionId(index)}
-          key={getOptionId(index)}
-          value={value}
-          role='option'
-          onClick={(value: T) => {
+      options.map((option, index) => {
+        const { label, value, ...rest } = option;
+        const additionalProps = {
+          value: value,
```
bentleyvk, line 129: Is `value` a valid prop/attribute to pass to user rendered element?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -103,25 +119,44 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     [options, id],
   );
 
+  const userOnChange = React.useRef(onChange);
```
bentleyvk, line 122: What was the part that you said caused too much rerenders? Was it this one?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -54,6 +55,20 @@ export type ComboBoxProps<T> = {
    * @default 'No options found'
    */
   emptyStateMessage?: string;
+  /**
+   * A custom item renderer can be specified to control the rendering.
+   * This function should ideally return a customized version of `MenuItem`,
+   * otherwise you will need to make sure to provide styling for the `isFocused` state.
+   */
+  itemRenderer?: (
```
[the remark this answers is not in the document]
mayank99: What do you mean "not show"? The menu items are not even in the DOM.

src/core/ComboBox/ComboBox.tsx
mayank99: Yeah, this combined with the actual returned menu items, which I'm now memoizing right below this (previously I wasn't).

src/core/ComboBox/ComboBox.tsx
mayank99: I think it might be ok. If not used it will just appear in the DOM 🤷‍♀️

src/core/ComboBox/ComboBox.tsx
mayank99: This seems to work
image

The name of the parameter doesn't matter because it will not be used directly. I was originally going to call it `options` but that got really confusing because of our menu options. Do you have a better suggestion?

src/core/ComboBox/ComboBox.tsx
mayank99: Not sure what you're asking. These additionalProps just get applied on top of the user's returned JSX

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
bentleyvk: Yeah, bad comment. I wanted to say is it ok that we are not providing a way to have custom renderer here: 
image

src/core/ComboBox/ComboBox.tsx
bentleyvk: Doesn't look nice 😀 In `Select` I just named it `ItemRendererProps`

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: That is a separate topic which will require different implementation if we even do it. See https://github.com/iTwin/iTwinUI-react/issues/517#issuecomment-1017821332.

src/core/ComboBox/ComboBox.tsx
bentleyvk: Nvm 😀

src/core/ComboBox/ComboBox.tsx
mayank99: so this one cannot be called `ItemRendererProps` if we want it in the barrel, to avoid conflicts.

Idk, all those extra props feel unnecessary to me when typescript has all these utilities

bentleyvk: Yeah and I hate all those utilities. It just makes simple things hard to understand and it is not that comfortable to use. Name it `ComboboxItemRendererProps`, idk.

bentleyvk (approved):
Everything looks good, nice! Besides missing type 😀 but it is holding this PR for too long so maybe it can be added later (or never).

Issue #542, opened 2022-02-08
Submenu 


S-Bra:
#### Describe the bug (actual behavior)
When a dropdown with submenu is located close to the right edge of the browser, the submenu does not display.

#### Expected behavior
Submenu is displayed left from its parent menu when insufficient space to the right.

#### Reproduction
Link to a minimal repro: 
https://codesandbox.io/s/itwinui-react-submenu-display-issue-at-browser-edge-xsxte?file=/src/App.tsx

##### Steps to reproduce
1. Click the "..." button and open the "Font Size" submenu (working). 
2. Click "Toggle Justification Start/End" - The more button now displays close to the right edge
3. In this state the "Font Size" submenu does not display anymore

#### Additional information
Saw in the story-book the menu floating around and properly displaying and could not figure out what is different here.
In the story book I modified the style of the surrounding div of the "more" button to 'display: flex; justify-content: "flex-end";'

veekeys:
As till now, the workaround I found to make it work is disabling `gpuAcceleration` on `DropdownMenu`.

```js
popperOptions={{modifiers: [{ name: 'computeStyles', options: { gpuAcceleration: false }}]}}
```

Would be nice to figure out the reason.

PR #552, opened 2022-02-15
feat(Popover): Append to body by default and z-index increase

### LINKED ISSUE #553: Popover: Increase default z-index and append to Body
Need to change some default values to Popover.
Increase z-index so no other elements in app would have higher.
Also, by default append to body, instead of parent. This should help us with issues in apps, where they have high z-index stack context and elements are not visible

veekeys:
- Now by default Popover is always appended to `ownerDocument.body`
- default z-index is increased so hopefully it is always bigger than iTwin.js elements
- append to body is causing issues on 2nd+ level elements. So I added back parent to the sub menus,, table date filter
- Test images updated, previously select behavior was weird with the new changes, which I noticed now thanks to Popover. We will need to follow up on that with removing the focus from menu in select.

Plan:
release dev version and test it out on itwin.js apps


mayank99:
> Tests of select are failing weirdly.. Seems something is changed with focus..

This was mentioned in https://github.com/iTwin/iTwinUI-react/pull/540#discussion_r801893631. We'll need to make a CSS change because it currently uses `focus-within` which relies on the popover being inside the container.

src/core/utils/components/Popover.tsx
```diff
@@ -61,13 +61,14 @@ export const Popover = React.forwardRef((props: PopoverProps, ref) => {
   const computedProps: Partial<TippyProps> = {
     allowHTML: true,
     animation: false,
-    appendTo: 'parent',
+    appendTo: (ref) => ref.ownerDocument.body,
```
mayank99, line 64: Maybe call it `el` instead of `ref`?

`ref` implies that it's a ref object, which it isn't. This is more like `ref.current`.

src/core/utils/components/Popover.tsx
```diff
@@ -38,7 +38,7 @@ export type PopoverProps = {
 export const Popover = React.forwardRef((props: PopoverProps, ref) => {
   const [mounted, setMounted] = React.useState(false);
 
-  const tippyRef = React.useRef(null);
+  const tippyRef = React.useRef<Element>(null);
```
mayank99, line 41: Why did we avoid specifying the type before? 🤔

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
```diff
@@ -66,6 +66,7 @@ const DatePickerInput = (props: DatePickerInputProps) => {
           close();
         }
       }}
+      appendTo='parent'
```
mayank99, line 69: Why do some of these have `appendTo='parent'` now?

The submenu one is a known issue #542 but I'm surprised this date picker input breaks.

src/core/Wizard/Wizard.test.tsx
```diff
@@ -7,6 +7,14 @@ import React from 'react';
 import { Wizard } from './Wizard';
 
 describe('<Wizard />', () => {
+  afterEach(() => {
+    // cleanup tippy after every test, so it does not stay in the dom
+    const tippy = document.querySelector('[data-tippy-root]') as HTMLElement;
+    if (!!tippy) {
+      tippy.remove();
+    }
```
mayank99, line 15: nit:

```suggestion
    document.querySelector('[data-tippy-root]')?.remove();
```

although, I'm not sure why it even stays in the dom

src/core/utils/components/Popover.tsx
veekeys: yeah, took ref from their definition

image

changed

src/core/utils/components/Popover.tsx
veekeys: I dont know, less writing?

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
veekeys: Would need to catch events ourselves and check the target. There is prop `[hideOnClick](https://atomiks.github.io/tippyjs/v5/all-props/)`, which we could change to `toggle` and then just listen to clicks and check :/

src/core/Wizard/Wizard.test.tsx
veekeys: changed

veekeys:
> > Tests of select are failing weirdly.. Seems something is changed with focus..
> 
> This was mentioned in [#540 (comment)](https://github.com/iTwin/iTwinUI-react/pull/540#discussion_r801893631). We'll need to make a CSS change because it currently uses `focus-within` which relies on the popover being inside the container.
> 
> Something like this:
> 
> ```diff
> --- a/src/inputs/input-with-icon.scss
> +++ b/src/inputs/input-with-icon.scss
> @@ -13,7 +13,8 @@
>      padding-right: $iui-icons-default + $iui-l;
>    }
> 
> -  &:focus-within {
> +  &:focus-within,
> +  &.iui-focused {
>      > :first-child {
> ```
> 
> /cc @gretanausedaite

This is wrong and should not be in Select, IMO. Having combobox with active state on both input and menu item makes sense, because you can type in input and at the same time navigate through items. And after popover changes, it still works, because we use classes there for 'focus' style in menu.
Now in select, after the refactors, we get such situation in visuals:
image

Which does not make sense, because its not that you can type in input or do anything with it. Native select also acts same with only single item being active. 
In that case, either leave css like that or better change to something like `:first-child:focus-visible`, so we actually apply the styling on `select-button`, when it is focused, not when menu items are focused too.

src/core/utils/components/Popover.tsx
mayank99: no I think it caused an error somewhere. Can't remember now 🤷‍♀️

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
mayank99: So wait, this `DatePickerInput` will always be used inside `ColumnFilter` which is already appended to body. Which means `appendTo='parent'` still places outside of the main component tree. Then I am ok with it actually

image

mayank99 (approved):
Looks good. Maybe we should merge it to a separate branch not to block main release?

src/core/Table/filters/DateRangeFilter/DatePickerInput.tsx
veekeys: Yes, exactly.
We will want to have a separate `DatePickerInput` one day, so then it should be important.

veekeys:
> Looks good. Maybe we should merge it to a separate branch not to block main release?

Does it make sense to merge it into another branch? I was thinking to make a release from this one maybe..?
Or we could make a specific dev release branch and there update package.json.. not sure

PR #557, opened 2022-02-22
feat(ExpandableBlock): Add borderless option to expandable block

### LINKED ISSUE #549: ExpandableBlock: Add borderless option
As the css is ready, lets add the support in react too. `Button` component could be a good example for the api part.

adamhock:
Support for borderless expandable block has been added using the `isBorderless` property (defaults to `false`)

New story and test have also been added

Closes #549


stories/core/ExpandableBlock.stories.tsx
```diff
@@ -140,3 +140,16 @@ Small.args = {
   title: 'Basic Block',
   size: 'small',
 };
+
+export const Borderless: Story<Partial<ExpandableBlockProps>> = (args) => {
```
adamhock, line 144: I noticed that there is no borderless story for `Button`, so I'm not sure if this story is needed either, but I figured it couldn't hurt

mayank99: That's because borderless button should be very careful used (almost discouraged). It is better for icons than for text buttons

adamhock:
I forgot to update the visual tests

stories/core/ExpandableBlock.stories.tsx
```diff
@@ -111,6 +111,7 @@ export const Accordion: Story<Partial<ExpandableBlockProps>> = (args) => {
 Accordion.argTypes = {
   title: { control: { disable: true } },
   isExpanded: { control: { disable: true } },
+  isBorderless: { control: { disable: true } },
```
adamhock, line 114: I disabled this for Accordion because toggling it doesn't change the blocks at all

adamhock, line 114: `isExpanded` is set to `false` in all of the other stories' args except for `Small`. Should `isBorderless` do the same? Also, should both `isBorderless` and `isExpanded` be set to false in the `Small.args` as well?

stories/core/ExpandableBlock.stories.tsx
adamhock: I assumed the expandable block was a type of text button more so than an icon.

The stories have been added back.

stories/core/ExpandableBlock.stories.tsx
mayank99: Did you try to make them work? I think the expandable block is just missing `{...args}` available from the story parameter.

Usually we only disable these if it's not possible to control, or if it's too complicated, or if it doesn't make sense in the story's context, or if it's adding too much noise (e.g. ReactNode props).

src/core/ExpandableBlock/ExpandableBlock.tsx
```diff
@@ -51,6 +51,11 @@ export type ExpandableBlockProps = {
    * @default 'default'
    */
   size?: 'default' | 'small';
+  /**
+   * Whether or not the block is borderless.
+   * @default false
+   */
+  isBorderless?: boolean;
```
mayank99, line 58: Not sure about this prop, vs something like `styleType: 'default' | 'borderless'`. Any particular reason you went with a boolean?

src/core/ExpandableBlock/ExpandableBlock.test.tsx
```diff
@@ -166,3 +166,16 @@ it('should render small size', () => {
...
+  const { container } = render(
+    <ExpandableBlock title='test title' isBorderless={true}>
+      content
+    </ExpandableBlock>,
+  );
+  const expandableBlock = container.querySelector(
+    '.iui-expandable-block',
+  ) as HTMLElement;
+  expect(expandableBlock).toBeTruthy();
+  expect(expandableBlock.classList).toContain('iui-borderless');
```
mayank99, line 180: This can be simplified down to one statement

```suggestion
  expect(container.querySelector('.iui-expandable-block').toHaveClass('iui-borderless');
```

src/core/ExpandableBlock/ExpandableBlock.tsx
adamhock: I wasn't sure about which one to go with, but I now remember you telling me use `styleType` for consistency. I'll change this

stories/core/ExpandableBlock.stories.tsx
adamhock: This has been updated

src/core/ExpandableBlock/ExpandableBlock.tsx
adamhock: This has been changed to styleType

src/core/ExpandableBlock/ExpandableBlock.test.tsx
adamhock: This statement has been simplified

mayank99 (approved):

PR #561, opened 2022-02-24
feat(Table): support for draggable columns

### LINKED ISSUE #538: Table improvements
#### Feature
Table needs to have these features:
- Column pinning (left and right side)
- Column rearranging
These features are blocking from using this table in our products
(Synchro cost, Synchro control, PW365, Inspections)

#### Examples
image
image

### LINKED ISSUE #571: Table: draggable columns
Separate issue to track from #538

Nmzik:
- Added an option `enableDraggableColumns` to enable draggable columns in Table component.
- Added property to column object `disableDragging` that prevents column from being draggable.
- I think I finally found proper fix to prevent triggering sort on other element clicks.

table-dragndrop

## Checklist


stories/core/Table.stories.tsx
```diff
@@ -2269,3 +2269,63 @@ VirtualizedSubRows.argTypes = {
   isLoading: { control: { disable: true } },
   data: { control: { disable: true } },
 };
+
+export const DraggableColumns: Story<Partial<TableProps>> = (args) => {
+  const onClickHandler = (
+    props: CellProps<{ name: string; description: string }>,
+  ) => action(props.row.original.name)();
+
+  const columns = useMemo(
```
bentleyvk, line 2278: It would be nice to have more columns, you could copy from `HorizontalScroll` story but without widths.

src/core/Table/TableRowMemoized.tsx
```diff
@@ -146,6 +146,7 @@ export const TableRowMemoized = React.memo(
     prevProp.rowProps === nextProp.rowProps &&
     prevProp.expanderCell === nextProp.expanderCell &&
     prevProp.tableHasSubRows === nextProp.tableHasSubRows &&
+    prevProp.state.columnOrder === nextProp.state.columnOrder &&
```
bentleyvk, line 149: I am afraid that there can be new `columnOrder` reference even it is the same. Maybe better check like:
```tsx
prevProp.state.columnOrder.every((column, index) => column === nextProp.state.columnOrder[index])
```

src/core/Table/Table.tsx
```diff
@@ -589,6 +599,48 @@ export const Table = <
     [getPreparedRow, page],
   );
```
bentleyvk, line 601: Tests are needed.

src/core/Table/Table.tsx
```diff
@@ -642,6 +694,21 @@ export const Table = <
                       {...columnProps}
                       key={columnProps.key}
                       title={undefined}
+                      draggable={props.enableDraggableColumns}
+                      onDragStart={(event) =>
+                        enableDraggableColumns &&
+                        dragStartHandler(event, column.id)
+                      }
+                      onDragOver={
```
bentleyvk, line 702: I think we should check position like whether it is first half of the column or the last half and to add column there accordingly:
tree-column-reorder
This is what I did:
image

bentleyvk: Also, you can use newly added CSS for columns borders: https://github.com/iTwin/iTwinUI/pull/498

src/core/Table/Table.tsx
Nmzik: > Tests are needed.

```
    columns,
    enableDraggableColumns: true,
  });

  const headerCells = container.querySelectorAll<HTMLDivElement>(
    '.iui-table-header .iui-cell',
  );
  headerCells.forEach((cell) =>
    expect(cell.getAttribute('draggable')).toBe('true'),
  );

  const nameColumn = headerCells[0];
  const viewColumn = headerCells[2];

  fireEvent.dragStart(nameColumn);
  fireEvent.dragEnter(viewColumn);
  fireEvent.dragOver(viewColumn, { dataTransfer: { text: 0 } });
  fireEvent.drop(viewColumn);
```
I spent some time trying to make it work without any success.

src/core/Table/Table.tsx
Nmzik: @bentleyvk Not sure how to modify cell classes: via event.currentTarget.classList or dedupe version from classNames. Also, would it be a performance issue to reassign same class multiple times in allowDrop event? I'm sorry but I think it would be better for you to look at it. "Allow edits by maintainers" is enabled. @itwin/itwinui-css is updated.

src/core/Table/Table.tsx
bentleyvk: Can you push what you have so I could play around with the tests?

src/core/Table/TableRowMemoized.tsx
Nmzik: @bentleyvk Columns' data are not being reordered after this change.

src/core/Table/Table.tsx
bentleyvk: Pushed fixed test. I had to mock `dataTransfer` object to make everything work.

src/core/Table/Table.test.tsx
```diff
@@ -2227,3 +2227,193 @@ it('should sync body horizontal scroll with header scroll', () => {
...
+      emptyTableContent='Empty table'
+      emptyFilteredTableContent='No results. Clear filter.'
+      enableDraggableColumns
+      isSelectable
+      subComponent={(row) => (
+        <div>{`Expanded component, name: ${row.original.name}`}</div>
+      )}
+    />,
+  );
+
```
bentleyvk, line 2410: Tried to mock drag n drop events and in jsdom they just work, so I guess wit is bug in jsdom because in real browser you can't drag if `draggable` isn't present.

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const onDragStart = (event: React.DragEvent<HTMLDivElement>) =>
+    event.dataTransfer.setData('text', header.id);
+
+  const setOnDragColumnStyle = (
+    event: React.DragEvent<HTMLDivElement>,
+    position?: 'left' | 'right',
+  ) => {
+    const columnElement = event.currentTarget as HTMLElement;
+    columnElement.classList.remove('iui-reorder-column-right');
```
gretanausedaite, line 47: I think there is a css bug where two borders don't overlap

gretanausedaite (commented):
We should consider adding cursor styling to draggable/non-draggable columns or maybe remove hover effects (leave hover for filter and sort buttons) on non-draggable columns. It's not clear which columns could be moved (eg. Cursor on Products and other columns in story are the same and it's confusing why I couldn't move Products column)

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const columnElement = event.target as HTMLElement;
+    const middlePoint =
+      columnElement.offsetLeft + columnElement.offsetWidth / 2;
+
+    instance.setColumnOrder(
+      reorderColumns(
+        columnIds,
+        srcIndex,
+        // When dropped on the right side of the column, need to increase the index by 1
+        event.clientX > middlePoint ? dstIndex + 1 : dstIndex,
```
mayank99, line 107: I'm not sure about this middle point logic. If you look at windows explorer for example, it reorders as soon as it enters the destination column.

src/core/Table/hooks/useColumnDragAndDrop.tsx
mayank99: So how would you suggest fixing this?

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+  makePropGetter,
+  TableInstance,
+  TableKeyedProps,
+  useGetLatest,
+} from 'react-table';
+
+export const useColumnDragAndDrop = <T extends Record<string, unknown>>(
+  isEnabled: boolean,
+) => (hooks: Hooks<T>) => {
+  hooks.getDragAndDropProps = [defaultGetDragAndDropProps(isEnabled)];
```
veekeys, line 19: `defaultGetDragAndDropProps` is generic. Dont you need to pass it here? Is it even needed to be generic?

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+    );
+    return newTableColumns;
+  };
+
+  const onDragOver = (event: React.DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    const columnElement = event.target as HTMLElement;
+    const middlePoint =
+      columnElement.offsetLeft + columnElement.offsetWidth / 2;
+    setOnDragColumnStyle(event, event.clientX > middlePoint ? 'right' : 'left');
```
veekeys, line 78: Maybe move this middle point logic into separate function? As you have 2 places which uses same.
`isDraggedToTheRight` or something like that.

src/core/Table/hooks/useColumnDragAndDrop.tsx
veekeys: I like that it it appears either on left or right of the destination column. And it shows where its gonna land.

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+      draggable: true,
+      onDragStart,
+      onDragOver,
+      onDragLeave,
+      onDrop,
+    },
+  ];
+};
+
+const useInstance = <T extends Record<string, unknown>>(
```
veekeys, line 124: What does this do?

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+  instance: TableInstance<T>,
+) => {
+  const { flatHeaders, getHooks } = instance;
+
+  const getInstance = useGetLatest(instance);
+
+  flatHeaders.forEach((header) => {
```
veekeys, line 131: Will it work in dialogs? Does it use `document` for any of those events, to be exact..

src/core/Table/Table.test.tsx
```diff
@@ -2227,3 +2227,193 @@ it('should sync body horizontal scroll with header scroll', () => {
   expect(header.scrollLeft).toBe(100);
   expect(body.scrollLeft).toBe(100);
 });
+
+it.each([
+  // Dragging Name to the left of View
```
veekeys, line 2232: Maybe instead of comments, add proper test case name? Beacuse now if it fails, not really clear which exactly
image

veekeys (commented):
I do agree with Greta that it is not clear at all, which column is draggable now... Could we see if we can add `drag` cursor at least?

src/core/Table/hooks/useColumnDragAndDrop.tsx
bentleyvk: I don't know why it works without passing explicitly but this where generic type is needed 
image

src/core/Table/hooks/useColumnDragAndDrop.tsx
bentleyvk: It accesses table instance and adds or takes required properties. This is how all `react-table` plugins are written.

src/core/Table/hooks/useColumnDragAndDrop.tsx
bentleyvk: It uses only events attached to the column itself so it really doesn't care where Table is placed.

veekeys (approved):
My comments were resolved, so its fine.
I approve, but need UX changes, so available columns for reorder are visible.

src/core/Table/hooks/useColumnDragAndDrop.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  const onDrop = (event: React.DragEvent<HTMLDivElement>) => {
+    event.preventDefault();
+    setOnDragColumnStyle(event);
+
+    const srcColumnId = event.dataTransfer.getData('text');
+    const columnIds = instance.flatHeaders.map((x) => x.id);
+    const srcIndex = columnIds.findIndex((x) => x === srcColumnId);
+    const dstIndex = columnIds.findIndex((x) => x === header.id);
```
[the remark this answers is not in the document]
bentleyvk: Added

src/core/Table/hooks/useColumnDragAndDrop.tsx
mayank99: Here's a gif demonstrating the problem with the current middle point approach. You have to move the mouse too far into the next column. The blue feedback makes it even more confusing why nothing is happening

dragging

Compared with Windows Explorer, where it performs the reorder as soon as the mouse enters the next column.
dragging2

mayank99 (approved):
UI (no marker, shifting border, no hover) and UX (middle point logic) are messed up but functionality and code looks good

src/core/Table/hooks/useColumnDragAndDrop.tsx
FlyersPh9: I know what is happening here.  The blue stripe is a 2px border that is applied on the left or right.  In this GIF we see the border switching from the left of the ratings column to the right of the quantity column.  We could explore using `::before` and absolute positioning similar to the column resizer.

src/core/Table/hooks/useColumnDragAndDrop.tsx
FlyersPh9: Here is how macOS handles it for comparison:
column-reorder

src/core/Table/hooks/useColumnDragAndDrop.tsx
bentleyvk: Will be fixed with this https://github.com/iTwin/iTwinUI/pull/518

src/core/Table/hooks/useSelectionCell.tsx
```diff
@@ -26,6 +26,7 @@ export const useSelectionCell = <T extends Record<string, unknown>>(
       maxWidth: 48,
       columnClassName: 'iui-slot',
       cellClassName: 'iui-slot',
+      disableReordering: true,
```
gretanausedaite, line 29: Is there are difference between `disableReordering` and `enableReordering` prop names? 
Why in table we use positive version `enableColumnReordering` and here negative `disable...`?

gretanausedaite (approved):
Code looks good. Will you merge or wait for CSS changes?

PR #564, opened 2022-02-24
feat(Table): Added ColumnManager for hiding columns

### LINKED ISSUE #462: Column manager for iTwin UI React Table
#### Feature
Column Manager: It would be good to have a way to hide columns, edit column names and rearrange column ordering.
Most important feature being hiding/showing columns for a user. Such that the settings persist for a user.

#### Examples

image

adamhock:
Added a new Column Manager feature which allows users to hide columns from their react table. 

This PR is incomplete because of the following

~~- Toggling column visibility only toggles the visibility of the table column headers and not entire column~~

Before hiding **Name**:
image

After hiding **Name**:
image

~~- The Column Manager story has not been finished~~

~~- No testing has been written for the Column Manager~~

- Keyboard Accessibility for the Column Manager menu still needs to be fixed

Partially closes #462 

EDIT: Toggling column visibility works on the entire column now


src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,53 @@
...
+      disableGroupBy: true,
+      minWidth: 48,
+      width: 48,
+      maxWidth: 48,
+      columnClassName: 'iui-slot',
+      cellClassName: 'iui-slot',
+      Header: ({ allColumns }: HeaderProps<T>) => {
+        const headerCheckBoxes = () =>
+          allColumns.map((column) =>
+            isString(column.Header) ? (
```
adamhock, line 35: `isString` was used because the non-data columns that are created by things such as having a Column Manager or Selectable Rows still have a `Header`, but their headers are objects instead of strings. I'm sure there's a better way to check for this

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,53 @@
...
+              <Checkbox
+                {...column.getToggleHiddenProps()}
+                label={column.Header}
+                key={column.id}
+              />
+            ) : (
+              <></>
```
adamhock, line 42: This is here because `headerCheckBoxes` cannot be undefined or it will not pass in the `menuItems` parameter of `<DropdownButton`>

veekeys:
So the issue you are having is connected with our memoized row, which has a check written manually, if it needs to rerendered.
I needed to add this, so it checks if something changed in hidden columns and then trigger the row rerender
```js
prevProp.state.hiddenColumns?.length === nextProp.state.hiddenColumns?.length &&
prevProp.state.hiddenColumns?.every(
    (column, index) => nextProp.state.hiddenColumns?.[index] === column,
)
```
in `TableRowMemoized`

Read here more about [React.memo](https://reactjs.org/docs/react-api.html#reactmemo)

adamhock:
@mayank99 Here is the draft PR and branch I have for the Column Manager. The screenshot in the first comment shows what adding a `MenuItem` list of Checkboxes results in. 

I was discussing with Jon about what CSS changes need to be made for the Column Manager, and he suggested the following

- Remove the caret at the end of the Column Manager button
- Have `padding` and `height` properties set similar to `iui-menu-item` for each of the columns listed in the menu
- Have the entire row be clickable instead of just the checkbox

I'm not sure the best way to go about removing the caret at the end of the Column Manager button. Maybe if an element contains the classes `iui-dropdown` and `iui-column-manager`, then remove the caret? 

I have committed a change which fixes the height issues by adding `iui-menu-item` to the class name for checkboxes.
image
I will push this to the PR now

For `padding` and making the entire row clickable, I was going to add CSS props for `iui-checkbox-wrapper iui-menu-item` to set these. If there's another way I should go about doing that, please let me know.

mayank99:
I think you ended up in this situation because you are using the wrong tools for the job and/or trying to recreate existing functionality.

The two main changes are:
- use `DropdownMenu` instead of `DropdownButton` to get rid of the caret
- use `MenuItem` component instead of trying to add class names to the checkbox

image

This gets you most of the way there. No padding or height issues.
image

The only two visual issues remaining are:
- width of clickable `<label>` area
- vertical alignment of checkbox with label

Those are the ones you need to fix with CSS.

stories/core/Table.stories.tsx
```diff
@@ -49,6 +49,7 @@ export default {
   argTypes: {
     columns: { control: { disable: true } },
     isSelectable: { control: { disable: true } },
+    hasColumnManager: { control: { disable: true } },
```
adamhock, line 52: I disabled this control because I heard it may be impossible to switch the table columns dynamically (the same reason why isSelectable is disabled)

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    <Checkbox
+                      id={`iui-column-${column.id}`}
+                      checked={checked}
+                      disabled={column.disableToggleVisibility}
+                    />
+                  }
+                  onClick={() => onChange({ target: { checked: !checked } })}
+                  disabled={column.disableToggleVisibility}
+                >
+                  {column.Header}
```
veekeys, line 53: hm.. I thought for semantics we wanted to have here `<label>` but I see clicking works fine without it now too.. So maybe no that important (accessibility question)

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+        const [isOpen, setIsOpen] = React.useState(false);
+        const headerCheckBoxes = () =>
+          allColumns
+            .filter(({ id }) => !id.includes('iui-table'))
+            .map((column) => {
+              const { onChange, checked } = column.getToggleHiddenProps();
+              return (
+                <MenuItem
+                  key={column.id}
+                  {...column.getToggleHiddenProps()}
```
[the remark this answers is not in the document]
veekeys: This also breaks disabled menu item styling, because it has `cursor: pointer` while disabled should have `not-allowed`

src/core/Table/Table.tsx
```diff
@@ -231,6 +232,11 @@ export type TableProps<
    * @beta
    */
   enableVirtualization?: boolean;
+  /**
+   * Flag whether table has a column manager.
+   * @default false
+   */
+  hasColumnManager?: boolean;
```
veekeys, line 239: Still not sure about this name.. Currently all it does is toggles visiblity of columns. DOnt know if we gonna add something more in there..

stories/core/Table.stories.tsx
```diff
@@ -2292,7 +2295,152 @@ VirtualizedSubRows.argTypes = {
...
+      data={data}
+      emptyTableContent='No data.'
+    />
+  );
+};
+
+ColumnManager.args = { hasColumnManager: true };
+
 export const DraggableColumns: Story<Partial<TableProps>> = (args) => {
+  const columns = useMemo(
```
veekeys, line 2401: Can you move this to previous place, so it does not show any changes in this story?
Your story could also go last, by time order..

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+  hasColumnManager: boolean,
+) => (hooks: Hooks<T>) => {
+  if (!hasColumnManager) {
+    return;
+  }
+
+  hooks.allColumns.push((columns: ColumnInstance<T>[]) => [
+    ...columns,
+    {
+      id: COLUMN_MANAGER_ID,
```
veekeys, line 25: After #599 is merged, you might want to move this and export as separate column.
Then it should be possible to check if such column already exist and add menu to that header OR just add whole new column

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+          allColumns
+            .filter(({ id }) => !id.includes('iui-table'))
+            .map((column) => {
+              const { onChange, checked } = column.getToggleHiddenProps();
+              return (
+                <MenuItem
+                  key={column.id}
+                  {...column.getToggleHiddenProps()}
+                  icon={
+                    <Checkbox
```
veekeys, line 44: There is also this error in console, which is not nice
image

Just need to think for the fix.. Maybe you could add `onChange` to checkbox too and then `stopPropagation` in Checkbox `onClick` event, so it does not go up to the `MenuItem`

stories/core/Table.stories.tsx
veekeys: Can someone explain more apart its `react-table` thing? :D

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { ColumnInstance, HeaderProps, Hooks } from 'react-table';
+import { Checkbox } from '../..';
```
veekeys, line 7: Wondering, if this path would not break any builds...
I do remember we had something with it.. @bentleyvk maybe you recall?

stories/core/Table.stories.tsx
```diff
@@ -2292,7 +2295,152 @@ VirtualizedSubRows.argTypes = {
   data: { control: { disable: true } },
 };
 
+export const ColumnManager: Story<Partial<TableProps>> = (args) => {
```
veekeys, line 2298: Currently, test image does not really make sense... Maybe better take a screenshot of opened menu too?
YOu can find some examples of custom actions in `Select` stories, for example.. Just of course, you need to add such test for individual story (yours), not all the table stories.

stories/core/Table.stories.tsx
```diff
@@ -1367,6 +1368,7 @@ export const Full: Story<Partial<TableProps>> = (args) => {
         isSelectable
         isSortable
         isResizable
+        hasColumnManager
```
veekeys, line 1371: hmm... should we also change width rules here, cause this does not look nice

image

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,71 @@
...
+                  onClick={() => onChange({ target: { checked: !checked } })}
+                  disabled={column.disableToggleVisibility}
+                >
+                  {column.Header}
+                </MenuItem>
+              );
+            });
+        return (
+          <DropdownMenu
+            menuItems={headerCheckBoxes}
```
veekeys, line 59: Another thing.. Would there be a case that users want a little different content there? Maybe add additional description or something..
Might be worth thinking, if we can provide also a way to override our default behavior

veekeys (commented):
Looks pretty good! Mostly concerned what should be the correct naming and of course, what to do with keyboard nav

src/core/Table/Table.tsx
adamhock: `enableColumnVisibility` or `enableColumnVisibilityToggle` ?

Does this mean we'd also change the name of the icon as well?

stories/core/Table.stories.tsx
adamhock: image

This is the current opened test image I have. Is it okay if it's cut off, or should I put less columns in the story so that they all fit in the screenshot?

src/core/Table/hooks/useColumnVisibility.tsx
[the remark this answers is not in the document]
adamhock: Removed

stories/core/Table.stories.tsx
adamhock: Story has been moved

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: Added an `onChange` event handler to `Checkbox`

stories/core/Table.stories.tsx
adamhock: margin left screenshot cropped

Can we add `margin-left: auto` to the column manager column

adamhock:
> Looks pretty good! Mostly concerned what should be the correct naming and of course, what to do with keyboard nav

Would changing the checkboxes to checkbox svg's work? All of the checkbox functionality can already be done with the menu item, and it would fix the keyboard navigation issues

src/core/Table/hooks/useColumnVisibility.tsx
veekeys: `DropdownMenu` not using labels is not really connected... `label` semantically should be used with any input element. DropdownMenu` just has a prop called label :D
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

So the only concern of mine is a11y and screen readers.

src/core/Table/Table.tsx
veekeys: Icon for the column manager you mean? No, I mean that icon opens menu which is supposed to be column manager (so not necessarily only visibility control, I think).
hmm.. Maybe Id vote `enableColumnVisibilityToggle` out from these 2.
But then if we find the way for the user to customize lets say that menu, then we might need a prop to just enable it to function :D

stories/core/Table.stories.tsx
veekeys: Good, but now could you move `columns` in line 2299 in story `DraggableColumns` to the place as it was before, so it is not shown as a change in your PR? :D

stories/core/Table.stories.tsx
veekeys: So 2 options here.
- Just add more rows
- Add wrapper element with some padding so everything is visible in screenshot (you can find examples in our stories with `decorators`

stories/core/Table.stories.tsx
veekeys: hmmm... But in thise case `Click` column is left in small width and that empty space is still for nothing, right?
I dont know.. Kinda fixes visually..
@FlyersPh9 opinion?
For context, my screenshot of table and column manager left aligned appears when other columns have certain width rules. Question: should we try fixing column manager placement or just change our examples and set width to auto adjust..?

src/core/Table/hooks/useColumnVisibility.tsx
veekeys: So few comments here:
- instead of creating mocked event now in `Checkbox` `onChange` handler, you can just pass the original one. (`onChange={onChange}`
- not sure if you have checked, but `stopPropagation` in `onChange` does not help in any way. Because `change` and `click` events are different. So you need to stop it on `onClick` event in `Checkbox`. Very old, but still good debugging way is using `console.log`. Add this in `MenuItem` on click function and see, if clicking checkbox makes it called. What we want to do here is not to make it called, so we dont handle same action twice.

veekeys:
> > Looks pretty good! Mostly concerned what should be the correct naming and of course, what to do with keyboard nav
> 
> Would changing the checkboxes to checkbox svg's work? All of the checkbox functionality can already be done with the menu item, and it would fix the keyboard navigation issues

oh wow!
Now thats thinking outside the box!
Initially I thought "it would be so bad", but.. it actually fixes the issue :D Just not sure, how that would be `liked` by same screen readers and anyone, who would inspect whats there :D
Oh wait, and you mean using just checkmark instead of checkbox in this case, right? So actually no one cares, its just an image we show as an additional marker of `selected`. 
Then I thought we could also apply `isSelected` on `MenuItem`, but this visually looks overcrowded...
image

Also, we could put checkmark on the right
image

@bentleyvk, @FlyersPh9 what do you think?

gretanausedaite (commented):
I have no additional comments on code.

I think adding checkmark icon as simple svg might work and I agree with @veekeys that isSelected on MenuItem makes it too much. Maybe we could color just icon?
image

stories/core/Table.stories.tsx
adamhock: Story has been moved again :D

stories/core/Table.stories.tsx
adamhock: I removed the width rules in the columns for the Column Manager story. If I do that for the Full story, when all of the columns are hidden, I still get the issue.

image

I think we should try fixing column manager placement

stories/core/Table.stories.tsx
veekeys: @bentleyvk @mayank99

stories/core/Table.stories.tsx
adamhock: Added extra rows to Column Manager story

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: This has been updated

stories/core/Table.stories.tsx
bentleyvk: I think because it is made through `react-table` hooks and therefore can't be undone.

src/core/Table/hooks/useColumnVisibility.tsx
bentleyvk: I think there were problems in rollup because of cyclic dependencies.

bentleyvk:
> Also, we could put checkmark on the right 
> 
> @bentleyvk, @FlyersPh9 what do you think?

I would vote for checkmark on the right as we also have this in css demo:
image

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,72 @@
...
+                  icon={
+                    <Checkbox
+                      id={`iui-column-${column.id}`}
+                      checked={checked}
+                      disabled={column.disableToggleVisibility}
+                      onClick={(e) => e.stopPropagation()}
+                      onChange={onChange}
+                    />
+                  }
+                  onClick={() => onChange({ target: { checked: !checked } })}
```
bentleyvk, line 51: Maybe change to `() => column.toggleHidden(!checked)` as in `react-table` https://github.com/TanStack/react-table/blob/v7/src/hooks/useColumnVisibility.js#L35

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,72 @@
...
+      width: 48,
+      maxWidth: 48,
+      disableReordering: true,
+      columnClassName: 'iui-slot',
+      cellClassName: 'iui-slot',
+      Header: ({ allColumns }: HeaderProps<T>) => {
+        const [isOpen, setIsOpen] = React.useState(false);
+        const headerCheckBoxes = () =>
+          allColumns
+            .filter(({ id }) => !id.includes('iui-table'))
```
bentleyvk, line 36: Maybe add comment that this filters out default columns like selection and expansion ones.

stories/core/Table.stories.tsx
```diff
@@ -2500,3 +2503,139 @@ DraggableColumns.args = {
...
+            Cell: (props: CellProps<TableStoryDataType>) => {
+              return props.row.original.endDate.toLocaleDateString('en-US');
+            },
+          },
+        ],
+      },
+    ],
+    [],
+  );
+  const data = useMemo(
```
bentleyvk, line 2564: You need to also add `data` to `args` or just disable changing `data` trough `Controls`.

src/core/Table/Table.test.tsx
```diff
@@ -2373,3 +2373,159 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+        },
+        {
+          id: 'view',
+          Header: 'View',
+          Cell: () => 'View',
+        },
+      ],
+    },
+  ];
+  const { container } = render(
```
bentleyvk, line 2450: Use `renderComponent`

src/core/Table/Table.test.tsx
```diff
@@ -2373,3 +2373,159 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
   expect(headerCells[3].getAttribute('draggable')).toBe('true'); // Description column
   expect(headerCells[4].getAttribute('draggable')).toBeFalsy(); // View column
 });
+
+it('should render column manager', () => {
```
bentleyvk, line 2377: I am not sure whether it is worth having this test when the next one also checks how column manager is renderer.

src/core/Table/Table.tsx
bentleyvk: If we plan to add column rearrange here for accessibility then I would say it can be called column manager.

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const COLUMN_MANAGER_ID = 'iui-table-column-manager';
+
+export const useColumnVisibility = <T extends Record<string, unknown>>(
+  hasColumnManager: boolean,
+) => (hooks: Hooks<T>) => {
+  if (!hasColumnManager) {
+    return;
+  }
+
+  hooks.allColumns.push((columns: ColumnInstance<T>[]) => [
```
bentleyvk, line 22: Shouldn't it go in the last user column instead of a separate column?

bentleyvk: nvm, found answer myself 😀

stories/core/Table.stories.tsx
bentleyvk: I have no idea how we should fix this since columns has static width. I guess in most cases there should be name/id column that can't be removed and it would take remaining space like in `Column Manager` story.

src/core/Table/Table.test.tsx
adamhock: I have this story to make sure the Column Manager renders with the correct menu items that exclude the Selection and Expandable columns. I don't have those columns in the next story so that it's easier to follow

Should I just combine the checks in this story in the `should hide column when selected in column manager` story?

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: Changed. I used `() => column.toggleHidden(checked` instead of `!checked` because the `react-table` story hides columns that are checked when we are trying to show columns that are checked

stories/core/Table.stories.tsx
adamhock: Added `data` to `args`

src/core/Table/Table.test.tsx
adamhock: Used `renderComponent` for all three Column Manager stories

stories/core/Table.stories.tsx
FlyersPh9: > there should be name/id column that can't be removed and it would take remaining space

That makes sense to me.  Named column can be disabled from hiding & take remaining width.  Otherwise with this solution there is a lot of wasted space.
Screen Shot 2022-03-29 at 3 17 21 PM

src/core/Table/hooks/useColumnVisibility.tsx
veekeys: @adamhock could you just import that directly as you did with other components in this file? Consistency is good

stories/core/Table.stories.tsx
veekeys: I think you still have one old image left from previous visuals (before you added logic to open and take screenshot).

stories/core/Table.stories.tsx
veekeys: So lets add then `Name` column as not configurable in our `Full` story to avoid weird case without any useful column

veekeys:
> > I think adding checkmark icon as simple svg might work and I agree with @veekeys that isSelected on MenuItem makes it too much. Maybe we could color just icon?
> > image
> 
> I like this option best.
> 
> > I would vote for checkmark on the right as we also have this in css demo:
> > image
> 
> We put the checkmark on the right so it wouldn't replace any left aligned icon like the flags in this image (checkmark is missing from this image though): Screen Shot 2022-03-29 at 3 23 10 PM The checkmark goes in the same place as the submenu arrow because you can't select an expandable menu item. Screen Shot 2022-03-29 at 3 24 48 PM So I think it might be ok to left align the checkmarks here, I don't anticipate anyone placing icons to the left.

So the weird part with the left icon is that we need to pass `Blank` icon for non selected menu items :D
But I guess thats expected...

src/core/Table/Table.tsx
veekeys: but do we plan? cause I dont think so :D

src/core/Table/hooks/useColumnVisibility.tsx
veekeys: @bentleyvk any ideas on this one?

gretanausedaite:
> So the weird part with the left icon is that we need to pass `Blank` icon for non selected menu items :D But I guess thats expected...

I say let's add this version with `Blank` icon. Do we have css class to colour checkbox? I hardcoded HEX in my test.

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: Fixed

stories/core/Table.stories.tsx
adamhock: Test image removed

stories/core/Table.stories.tsx
adamhock: `Full` story changed

src/core/Table/hooks/useColumnVisibility.tsx
veekeys: As we wont probably use Checkbox anymore, this is not a concern. Resolving

adamhock:
Animation

Found this issue when trying to use the column manager on the `Full` story after resizing a column

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -0,0 +1,32 @@
...
+    if (hasManagerColumn) {
+      return columns;
+    }
+    const lastColumn = columns[columns.length - 1];
+    const managerColumn = ManagerColumn();
+    // If the last column is an iui-slot that has it's own Cell props, don't generate a separate column for Column Manager
+    if (
+      lastColumn.cellClassName?.includes('iui-slot') &&
+      lastColumn.Cell !== undefined
+    ) {
```
adamhock, line 25: Is this the correct conditions for adding just the Column Manager header instead of creating a new column? Should there be another check for if `lastColumn` doesn't contain a header?

adamhock:
Column Manager Centering Issue

Found a centering issue when users add custom `iui-slot` column

src/core/Table/hooks/useColumnVisibility.tsx
```diff
@@ -2,72 +2,31 @@
...
+  hooks.allColumns.push((columns: ColumnInstance<T>[]) => {
+    const hasManagerColumn = columns.find((c) => c.id === COLUMN_MANAGER_ID);
+    if (hasManagerColumn) {
+      return columns;
+    }
+    const lastColumn = columns[columns.length - 1];
+    const managerColumn = ManagerColumn();
+    // If the last column is an iui-slot that has it's own Cell props, don't generate a separate column for Column Manager
+    if (
+      lastColumn.cellClassName?.includes('iui-slot') &&
```
veekeys, line 23: So I was thinking a little bit different... Users by themselves do not add `iui-slot` class.. They do not know about it and we do not document it anywhere.
The purpose of creating this column as a component was for users to actually use that for their actionable last column (to display menu, for example). And then what would you do is to check if such column exists and just update its header part, leaving everything else.

adamhock: This has been changed. Now an empty `ActionColumn` can be generated with a flag for the Column Manager. The `Cell` prop can be overridden by the user if they want to populate the column or not (see [here](https://github.com/iTwin/iTwinUI-react/pull/564/files#diff-7e8181f0c561b1370b6fadd31d714fcf9293c0fd45d8de3f75d055f553d922a8R2439) in Table test)

https://github.com/iTwin/iTwinUI-react/pull/564/files#diff-7e8181f0c561b1370b6fadd31d714fcf9293c0fd45d8de3f75d055f553d922a8R2439

src/core/Table/Table.test.tsx
```diff
@@ -2373,3 +2373,184 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+        },
+        {
+          id: 'description',
+          Header: 'Description',
+          accessor: 'description',
+        },
+        {
+          ...ActionColumn(true),
+          id: 'view',
+          Cell: () => 'View',
```
veekeys, line 2441: Not sure how useful this test is.. You do not even check if Cell content appears.

src/core/Table/columns/index.ts
```diff
@@ -0,0 +1,5 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+export { ACTION_CELL_ID, ActionColumn } from './actionColumn';
```
veekeys, line 5: hmm..Not sure we need to export ID.. You are not using it anywhere outside.
Is there a use case that users need that?

stories/core/Table.stories.tsx
```diff
@@ -1273,6 +1274,8 @@ export const Full: Story<Partial<TableProps>> = (args) => {
     [],
   );
 
+  const hasColumnManager = true;
```
veekeys, line 1277: looks weird... why you need this variable?
Just pass true directly on line 1318

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { HeaderProps } from 'react-table';
+import { Checkbox } from '../../Checkbox';
+import SvgColumnManager from '@itwin/itwinui-icons-react/cjs/icons/ColumnManager';
+import { DropdownMenu } from '../../DropdownMenu';
+import { IconButton } from '../../Buttons/IconButton';
+import { MenuItem } from '../../Menu';
+export const ACTION_CELL_ID = 'iui-table-action';
+
+export const ActionColumn = <T extends Record<string, unknown>>(
+  hasColumnManager: boolean,
```
veekeys, line 15: Just randomly thought... Would we expect here more props in future and would it make sense to have an object instead..
But maybe not worth it as barely we will add that many props here :/

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,70 @@
...
+) => {
+  return {
+    id: ACTION_CELL_ID,
+    minWidth: 48,
```
gretanausedaite, line 19: Just noticing: Column manager width is less than column.

src/core/Table/columns/index.ts
adamhock: Removed

src/core/Table/columns/actionColumn.tsx
adamhock: Decided to leave as is

adamhock:
> Column Manager Centering Issue
> 
> Found a centering issue when users add custom `iui-slot` column

@veekeys do we want to center this button in the column or leave this as is?

veekeys:
> > Column Manager Centering Issue
> > Found a centering issue when users add custom `iui-slot` column
> 
> @veekeys do we want to center this button in the column or leave this as is?

Id say center, but thats a css change, I guess.
Might not even hold this pr.

src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -392,7 +392,9 @@ const getPreviousResizableHeader = <T extends Record<string, unknown>>(
   headerColumn: ColumnInstance<T>,
   instance: TableInstance<T>,
 ) => {
-  const headersList = headerColumn.parent?.columns || instance.flatHeaders;
+  const headersList = (
+    headerColumn.parent?.columns || instance.flatHeaders
+  ).filter(({ isVisible }) => isVisible);
```
adamhock, line 397: ColumnManagerResizerShown

This is to fix the resizer being shown after a column is hidden. As you can see, the change works, but doesn't refresh until another action is done to the table such as clicking on a row

adamhock (commented):
ColumnManagerReorderBug

Column Manager bug where if a column to the left and right of a hidden column are switched, then the hidden column is placed at the end of the table, past the column manager column

adamhock:
> Animation
> 
> Found this issue when trying to use the column manager on the `Full` story after resizing a column

Resizing the Column Manager column has been partially fixed (see ), but there is still an issue where hiding columns after the resizing of ANY column results in the column manager being shifted

ColumnManagerResizingThenHidingIssue

src/core/Table/Table.tsx
```diff
@@ -550,7 +548,11 @@ export const Table = <
           ].getBoundingClientRect().width;
         }
       });
-      dispatch({ type: tableResizeEndAction, columnWidths: newColumnWidths });
+      console.log(newColumnWidths);
```
veekeys, line 551: ups

adamhock: removed

stories/core/Table.stories.tsx
adamhock: column manager is no longer a table prop, so this has been removed

src/core/Table/Table.tsx
adamhock: this table prop has been removed

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: updated

src/core/Table/hooks/useColumnVisibility.tsx
adamhock: Users can add their own custom `Header` to action column

src/core/Table/hooks/useResizeColumns.tsx
adamhock: Fixed

src/core/Table/Table.tsx
```diff
@@ -43,17 +43,15 @@ import {
...
   onTableResizeStart,
-} from './actionHandlers/resizeHandler';
+  TABLE_RESIZE_END_ACTION,
+  TABLE_RESIZE_START_ACTION,
+} from './actionHandlers';
 import VirtualScroll from '../utils/components/VirtualScroll';
 
 const singleRowSelectedAction = 'singleRowSelected';
-const tableResizeStartAction = 'tableResizeStart';
-const tableResizeEndAction = 'tableResizeEnd';
```
mayank99, line 56: Don't we want a convention for these?

src/core/Table/columns/actionColumn.tsx
mayank99: I agree this should be an object, even if we don't add anything in the future. A boolean parameter makes the usage really hard to read

```ts
ActionColumn(true)
```

vs

```ts
ActionColumn({ hasColumnManager: true })
```

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+    disableReordering: true,
+    Header: ({ allColumns, dispatch }: HeaderProps<T>) => {
+      const [isOpen, setIsOpen] = React.useState(false);
+      if (!hasColumnManager) {
+        return null;
+      } else {
+        const headerCheckBoxes = () =>
+          allColumns
+            //Filters out any default columns made such as selection and expansion
+            .filter(({ id }) => !id.includes('iui-table'))
```
mayank99, line 37: Can we use the real ids for those columns instead of this hardcoded string?

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+                      onClick={(e) => e.stopPropagation()}
+                      onChange={(e) => {
+                        onChange(e);
+                        dispatch({ type: TABLE_RESIZE_START_ACTION });
+                      }}
+                    />
+                  }
+                  onClick={() => {
+                    column.toggleHidden(checked);
+                    dispatch({ type: TABLE_RESIZE_START_ACTION });
```
mayank99, line 57: Can you explain why this dispatch is needed here? Looks unrelated

/cc @bentleyvk

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,79 @@
...
+          allColumns
+            //Filters out any default columns made such as selection and expansion
+            .filter(({ id }) => !id.includes('iui-table'))
+            .map((column) => {
+              const { onChange, checked } = column.getToggleHiddenProps();
+              return (
+                <MenuItem
+                  key={column.id}
+                  icon={
+                    <Checkbox
```
mayank99, line 44: Can you add an `aria-label` to this checkbox with the name of the column? Or use `aria-labelledby` to reference the same text that is inside the MenuItem

src/core/Table/columns/actionColumn.tsx
adamhock:
```suggestion
            .filter(
              (column) =>
                column.id !== SELECTION_CELL_ID &&
                column.id !== EXPANDER_CELL_ID &&
                column.id !== ACTION_CELL_ID,
            )

```

Would this work or is there a better way to check for this

mayank99: Yeah, or this:
```js
.filter(({ id }) => ![SELECTION_CELL_ID, EXPANDER_CELL_ID, ACTION_CELL_ID].includes(id))
```

src/core/Table/columns/actionColumn.tsx
adamhock: changed to an object

src/core/Table/columns/actionColumn.tsx
adamhock: `aria-label` added

src/core/Table/columns/actionColumn.tsx
adamhock: `filter` updated

src/core/Table/columns/actionColumn.tsx
adamhock: It's to update the state of the resizer once a column's visibility has been toggled. It fixes [this](https://github.com/iTwin/iTwinUI-react/pull/564#discussion_r841305771) issue

bentleyvk: When you resize any of the columns, their width becomes static. Therefore when you remove a column, you need to update the width of all visible columns so this dispatch triggers this update by making columns' width dynamic, saving that width and setting to static width again. It is kind of a hack that I used for Table resize with resized columns to avoid recalculating each column by myself.

src/core/Table/hooks/useColumnVisibility.tsx
bentleyvk: Maybe for now we can release as it is and if someone asks for such feature we could add some property to the column like `columnName` or smth

stories/core/Table.stories.tsx
```diff
@@ -1320,6 +1322,7 @@ export const Full: Story<Partial<TableProps>> = (args) => {
               </>
             ),
           },
+          ActionColumn({ hasColumnManager: true }),
```
bentleyvk, line 1325: I would say to marge this column with `click-me` column but better change content to triple dots dropdown menu with some mocked actions like edit, delete. Because it is the whole point of having `ActionColumn` as a column and not regular prop in Table.

src/core/Table/Table.test.tsx
```diff
@@ -2445,3 +2445,189 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+          Header: 'View',
+          Cell: () => 'View',
+        },
+        ActionColumn({ hasColumnManager: true }),
+      ],
+    },
+  ];
+  const { container } = renderComponent({
+    columns,
+    data: mockedData(),
```
bentleyvk, line 2475: Only columns is needed here, remove the remaining props.

src/core/Table/Table.test.tsx
```diff
@@ -2445,3 +2445,189 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+    isSelectable: true,
+    subComponent: (row) => (
+      <div>{`Expanded component, name: ${row.original.name}`}</div>
+    ),
+  });
+
+  const columnManager = container.querySelector('.iui-button') as HTMLElement;
```
bentleyvk, line 2482: Maybe search for the button in the Table header in the last cell. Because it looks like just any random button.

src/core/Table/Table.test.tsx
```diff
@@ -2445,3 +2445,189 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+    data: mockedData(),
+  });
+
+  expect(container.querySelectorAll('[role="columnheader"]').length).toBe(3);
+  const actionColumn = container.querySelectorAll<HTMLInputElement>(
+    '.iui-slot',
+  );
+  expect(
+    actionColumn[0].firstElementChild?.className.includes(
+      'iui-button iui-borderless',
```
bentleyvk, line 2529: Just move these class selectors to the query.

bentleyvk: Oh, ok I see what you did there.

src/core/Table/Table.test.tsx
```diff
@@ -2445,3 +2445,189 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+    ),
+  ).toBeTruthy();
+  expect(actionColumn[1].textContent).toBe('View');
+  expect(actionColumn[2].textContent).toBe('View');
+  expect(actionColumn[3].textContent).toBe('View');
+});
+
+it('should hide column when selected in column manager', () => {
```
bentleyvk, line 2537: Maybe `deselected`?

src/core/Table/Table.test.tsx
```diff
@@ -2445,3 +2445,189 @@ it('should not have `draggable` attribute on columns with `disableReordering` en
...
+          ...ActionColumn({ hasColumnManager: true }),
+          id: 'view',
+          Cell: () => 'View',
+        },
+      ],
+    },
+  ];
+  const { container } = renderComponent({
+    columns,
+    data: mockedData(),
```
bentleyvk, line 2520: Why are you passing `mockedData`? It is not needed anywhere in your tests.

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { DropdownMenu } from '../../DropdownMenu';
+import { IconButton } from '../../Buttons/IconButton';
+import { MenuItem } from '../../Menu';
+import { TABLE_RESIZE_START_ACTION } from '../actionHandlers';
+import { EXPANDER_CELL_ID, SELECTION_CELL_ID } from '../hooks';
+
+const ACTION_CELL_ID = 'iui-table-action';
+
+export const ActionColumn = <T extends Record<string, unknown>>({
+  hasColumnManager = true,
```
veekeys, line 18: So what would be the best name for this? Sounds generic even though now it gives only columns visibility

@bentleyvk @mayank99

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,86 @@
...
+      const [isOpen, setIsOpen] = React.useState(false);
+      if (!hasColumnManager) {
+        return null;
+      } else {
+        const headerCheckBoxes = () =>
+          allColumns
+            //Filters out any default columns made such as selection and expansion
+            .filter(
+              ({ id }) =>
+                ![SELECTION_CELL_ID, EXPANDER_CELL_ID, ACTION_CELL_ID].includes(
```
veekeys, line 40: minor, but could have generated this array as const above, instead of doing that on every column item

src/core/Table/columns/actionColumn.tsx
veekeys: Hopefully, its always a string...

src/core/Table/columns/actionColumn.tsx
bentleyvk: Is it even makes sense to have this prop? Or you think users might use several action columns?

bentleyvk: My suggestion `showColumnManager`. So yes, I would still prefer column manager name 😀

veekeys: I assume they could have as many as thy want.
Also, if they take action column, not necessarily they want column manager in there.
Just I thought if its column manager and later has more features (visiblity and order), how that would be set up

src/core/Table/columns/actionColumn.tsx
mayank99: Oh, looks like it won't always be a string.
image

So then we should use `aria-labelledby` on the rendered JSX

src/core/Table/Table.test.tsx
adamhock: Removed

src/core/Table/Table.test.tsx
adamhock: changed

src/core/Table/Table.test.tsx
adamhock: Changed

src/core/Table/Table.test.tsx
adamhock: removed

stories/core/Table.stories.tsx
adamhock: Changed this column to an Action Column whose cells are triple dot dropdown menus

src/core/Table/columns/actionColumn.tsx
mayank99: I suggest we rename the current flag to `columnManager`, so that it doesn't need to be confined to a boolean.

In the future, we could accept an object as an alternative to the boolean.
```ts
type ActionColumnParams = {
  columnManager: boolean | { enableVisibilityToggle?: boolean; enableReordering?: boolean };
};
```

Usage:
```ts
ActionColumn({ columnManager: true })
```
```ts
ActionColumn({ columnManager: false }) // equivalent to ActionColumn()
```
```ts
ActionColumn({ columnManager: { enableVisibilityToggle = true, enableReordering = false } })
```

adamhock (commented):
ColumnManagerResizesOnToggle

Columns sizes are changing when column resizing is on

src/core/Table/columns/actionColumn.tsx
mayank99: Maybe there should be a comment next to the code to prevent confusion

src/core/Table/columns/actionColumn.tsx
adamhock: `aria-labelledby` added. Also, The menu item's inner HTML is rendered using `column.render('Header')` instead of `column.Header` to work with Headers that are strings or return JSX. See [here](https://github.com/iTwin/iTwinUI-react/pull/564/files#diff-e9eae8b7f2f64123d6cf3fcdc63983c5293641819768d19dfaa47ec82399fe88R69)

stories/core/Table.stories.tsx
```diff
@@ -2574,3 +2584,183 @@ DraggableColumns.args = {
...
+        startDate: new Date('May 5, 2021'),
+        endDate: new Date('Jun 5, 2021'),
+      },
+    ],
+    [],
+  );
+
+  return (
+    <Table
+      isSelectable={true}
```
bentleyvk, line 2695: Just `isSelectable`, no need for `={true}`

stories/core/Table.stories.tsx
```diff
@@ -2574,3 +2584,183 @@ DraggableColumns.args = {
...
+      columns={columns}
+      data={data}
+      {...args}
```
bentleyvk, line 2698: Move up, before `{...args}`

adamhock: moved

stories/core/Table.stories.tsx
adamhock: changed

bentleyvk:
> ColumnManagerResizesOnToggle
> 
> Columns sizes are changing when column resizing is on

I kind of fixed it, not fully but it is not so bad now and if columns were not resized then there is no issue at all. Also, I see that this issue only happens when there is only few columns with bigger number of columns, everything seems to be ok.

adamhock:
> > ColumnManagerResizesOnToggle
> > Columns sizes are changing when column resizing is on
> 
> I kind of fixed it, not fully but it is not so bad now and if columns were not resized then there is no issue at all. Also, I see that this issue only happens when there is only few columns with bigger number of columns, everything seems to be ok.

Thanks! I think another column should be added to the Full story anyway. It seems kind of light

src/core/Table/Table.test.tsx
bentleyvk: What about your other tests?

src/core/Table/columns/actionColumn.tsx
adamhock: Comment added [here](https://github.com/iTwin/iTwinUI-react/pull/564/files#diff-e9eae8b7f2f64123d6cf3fcdc63983c5293641819768d19dfaa47ec82399fe88R49)

src/core/Table/Table.test.tsx
adamhock: Updated. They should be removed in all Column Manager tests

src/core/Table/Table.tsx
adamhock: Changed. `tableResizeStartAction` is now exported from Table. These three variable names have also been changed to UPPER_CASE

mayank99: This diff feels unnecessary now.
image

Maybe revert the order and convention back to what it was.

src/core/Table/columns/actionColumn.tsx
adamhock: Added another comment

`                // Triggers an update to resize the widths of all visible columns `

src/core/Table/Table.tsx
adamhock: Reverted order and naming convention

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { DropdownMenu } from '../../DropdownMenu';
+import { IconButton } from '../../Buttons/IconButton';
+import { MenuItem } from '../../Menu';
+import { tableResizeStartAction } from '../Table';
+import { EXPANDER_CELL_ID, SELECTION_CELL_ID } from '../hooks';
+
+const ACTION_CELL_ID = 'iui-table-action';
+
+export const ActionColumn = <T extends Record<string, unknown>>({
+  columnManager = false,
```
veekeys, line 18: Even though you set `columnManager` to false, if its not passed, but whole object is required to pass, so if i want to use `ActionColumn` without it, I need to write like that:
image

So can make object optonal and.. if you had test with such a case, you would have caught it :D so add it

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,93 @@
...
+    disableReordering: true,
+    Header: ({ allColumns, dispatch, state }: HeaderProps<T>) => {
+      const [isOpen, setIsOpen] = React.useState(false);
+      const defaultColumnIds = [
+        SELECTION_CELL_ID,
+        EXPANDER_CELL_ID,
+        ACTION_CELL_ID,
+      ];
+
+      if (!columnManager) {
```
veekeys, line 38: You can move this if up anb return early. No need to init variables above, if you not gonna use them

src/core/Table/columns/actionColumn.tsx
```diff
@@ -0,0 +1,93 @@
...
+import SvgColumnManager from '@itwin/itwinui-icons-react/cjs/icons/ColumnManager';
+import { DropdownMenu } from '../../DropdownMenu';
+import { IconButton } from '../../Buttons/IconButton';
+import { MenuItem } from '../../Menu';
+import { tableResizeStartAction } from '../Table';
+import { EXPANDER_CELL_ID, SELECTION_CELL_ID } from '../hooks';
+
+const ACTION_CELL_ID = 'iui-table-action';
+
+export const ActionColumn = <T extends Record<string, unknown>>({
```
veekeys, line 17: wait, so why file name is lower case? It is having a react component.
@bentleyvk

src/core/Table/Table.tsx
```diff
@@ -550,7 +548,10 @@ export const Table = <
           ].getBoundingClientRect().width;
         }
       });
-      dispatch({ type: tableResizeEndAction, columnWidths: newColumnWidths });
+      dispatch({
```
veekeys, line 551: minor, but as you gonna have commit, revert this change too... should fit in one line

adamhock: reverted

src/core/Table/columns/actionColumn.tsx
adamhock: moved

src/core/Table/columns/actionColumn.tsx
adamhock: Object is now optional. Test is written to for empty action column

veekeys (approved):
File name is the last comment I have, but apart for that looks good!

src/core/Table/actionHandlers/index.ts
```diff
@@ -5,9 +5,4 @@
 export { onExpandHandler } from './expandHandler';
 export { onFilterHandler } from './filterHandler';
 export { onSelectHandler, onSingleSelectHandler } from './selectHandler';
-export {
-  onTableResizeStart,
-  onTableResizeEnd,
-  TABLE_RESIZE_START_ACTION,
-  TABLE_RESIZE_END_ACTION,
```
bentleyvk: Why reset these?

adamhock: https://github.com/iTwin/iTwinUI-react/pull/564#discussion_r842031236

I didn't see the point of exporting them from resizeHandler if they weren't being used in there. Why not just export from Table.tsx?

bentleyvk: Ok, minor, still need to unify our actions.

bentleyvk (approved):

PR #569, opened 2022-02-28
feat: Create `Carousel` component

### LINKED ISSUE #555: Carousel in react
Follow up to https://github.com/iTwin/iTwinUI/issues/374 and https://github.com/iTwin/iTwinUI/pull/377

- Base carousel with dots navigation and proper a11y attributes
- Focus management and keyboard navigation
- Autoplay/pause support that takes into account user interactions and [page visibility](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API#:~:text=a%20site%20has%20an%20image%20carousel%20that%20shouldn't%20advance%20to%20the%20next%20slide%20unless%20the%20user%20is%20viewing%20the%20page)

mayank99:
Added `Carousel` component with subcomponents for different parts. `Carousel.DotsList` can be used outside `Carousel`.

Sliding is handled with a combination of native scrolling and `scrollTo` events.

Supports keyboard navigation for switching between slides. Uses intersection observer for updating active index when scrolled natively.

Autoplay is excluded on purpose as it's bad for accessibility.

https://user-images.githubusercontent.com/9084735/158457415-b0fc265d-1b55-423f-b026-5e9b1bd6c9b3.mp4

Closes #555 

## Checklist


mayank99:
@iTwin/itwinui-react-1 Tests and semantics are pending, but would be good to get initial review on this to finalize the API part first.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+    {},
+  );
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.altKey || event.ctrlKey || event.metaKey) {
+      return;
+    }
+
+    if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
+      setKeysPressed((old) => ({ ...old, [event.key]: true }));
```
bentleyvk, line 47: Can you explain this part?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    switch (event.key) {
+      case 'ArrowLeft': {
+        setCurrentIndex((old) => (slideCount + old - 1) % slideCount);
```
bentleyvk, line 52: Why not just `old + 1`?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+      type='button'
+      role='tab'
+      tabIndex={-1}
+      className={cx(
+        'iui-carousel-navigation-dot',
+        {
+          'iui-active': isActive,
+          'iui-first': isFirst,
+          'iui-second': isSecond,
+          'iui-invisible': isInvisible,
```
bentleyvk, line 283: Shouldn't we just don't show invisible dots?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+    >
+      {children ?? dots}
+    </ol>
+  );
+};
+
+const CarouselDot = (
+  props: {
+    isActive?: boolean;
+    isFirst?: boolean;
```
bentleyvk, line 258: `isFirst` and `isSecond` really looks strange. Maybe we should rename them to `small`, `medium` or `ellipsisStart`, `ellipsisEnd` or something better.

bentleyvk:
I like your approach of composing everything with components. It allows more customization and people could reuse just the navigation part.

src/core/Carousel/Carousel.tsx
mayank99: You mean `old - 1`?

I'm using `%` to loop over from one edge to another.

src/core/Carousel/Carousel.tsx
mayank99: I'm saving the keys pressed on key down (and removing on key up) so that I can apply `[data-pressed]` on the left/right navigation buttons. If you try to tab to the carousel and press left/right arrow keys, you will see a visual indicator around the buttons.

src/core/Carousel/Carousel.tsx
bentleyvk: Oh, ok. Then why `(old - 1) % slideCount` is not enough?

src/core/Carousel/Carousel.tsx
bentleyvk: Is it some standard or what?

src/core/Carousel/Carousel.tsx
mayank99: you tell me 😄 
image

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+import { useTheme } from '../utils';
+import { IconButton } from '../Buttons';
+import SvgChevronLeft from '@itwin/itwinui-icons-react/cjs/icons/ChevronLeft';
+import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
+import '@itwin/itwinui-css/css/carousel.css';
+
+export type CarouselProps = {
+  currentIndex?: number;
+  autoPlay?: boolean;
+} & React.ComponentPropsWithoutRef<'div'>;
```
veekeys, line 16: Think `ref` wont be needed?

src/core/Carousel/Carousel.tsx
mayank99: no, it's the design we decided on

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+  const [currentIndex, setCurrentIndex] = React.useState(userCurrentIndex);
+  const [slideCount, setSlideCount] = React.useState(0);
+
+  const [keysPressed, setKeysPressed] = React.useState<Record<string, boolean>>(
+    {},
+  );
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.altKey || event.ctrlKey || event.metaKey) {
```
veekeys, line 42: As you started...what about `shift` then? :D

veekeys: And seems like we should have some util for all our events to ignore those additional keys :/ but out of this scope

src/core/Carousel/Carousel.tsx
veekeys: It is expected to always loop over? Wont be situations, that you want to stop it at the edge?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+    }
+  | undefined
+>(undefined);
+
+export const Carousel = (props: CarouselProps) => {
+  const { currentIndex: userCurrentIndex = 0, children, ...rest } = props;
+
+  useTheme();
+
+  const [currentIndex, setCurrentIndex] = React.useState(userCurrentIndex);
```
veekeys, line 34: no `useEffect` for this for user to control?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+        transform: `translateX(-${currentIndex * 100}%)`,
+        ...props.style,
+      }}
+    >
+      {items}
+    </ol>
+  );
+};
+
+const CarouselSlide = (props: React.ComponentPropsWithoutRef<'li'>) => {
```
veekeys, line 130: Have you considered passing elements as prop straight to CarouselSlider as array of `ReactNode` and then map those straight to `li` elements in there? Cause this component by itself does nothing

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+    </li>
+  );
+};
+
+const CarouselNavigation = (props: React.ComponentPropsWithoutRef<'nav'>) => {
+  const { className, children, ...rest } = props;
+
+  const context = React.useContext(CarouselContext);
+  if (!context) {
+    throw new Error('TODO');
```
veekeys, line 150: I assume these could be used outside of Carousel..no ?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+  const dots = React.useMemo(
+    () =>
+      Array(slideCount)
+        .fill(null)
+        .map((_, index) => {
+          const firstDotIndex = Math.max(
+            0,
+            Math.min(
+              currentIndex - Math.ceil(maxCount / 2) + 1,
```
veekeys, line 212: lost me in maths :D

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,298 @@
...
+          'iui-invisible': isInvisible,
+        },
+        className,
+      )}
+      {...rest}
+    />
+  );
+};
+
+Carousel.Slider = CarouselSlider;
```
veekeys, line 292: Would you move then those compnents in separate files or leave all in here?

veekeys (commented):
Looks pretty good for the start.
Just animation on dots is really bothering me here same as in css :/

src/core/Carousel/Carousel.tsx
mayank99: Haven't decided yet. It's kinda neat that these are "internal" components only exposed through `Carousel`

src/core/Carousel/Carousel.tsx
mayank99: it's just trying to calculate how many dots to show on each side.
- `maxCount / 2` gives "half" of the dots
- `Math.ceil`/`floor` handles odd numbers
- `+1`/`-1` is accounting for current dot
- `Math.max`/`Math.min` is for clamping.

src/core/Carousel/Carousel.tsx
mayank99: I think it's ok for this component to not "do anything". It just represents a DOM element, which is expected in the composition approach.

I have considered accepting an array of slides in the parent though, as an alternative to the whole composition approach. But it requires too many props if we want it to be controlled.

```tsx
<Carousel slides={[<div>, <div>, <div>]} current={current} onSlideChange={() => {}} />
```

src/core/Carousel/Carousel.tsx
mayank99: I was just mirroring the names from CSS. I find your names to also be strange 😅 I already thought about `truncateStart`/`truncateEnd` and `truncateFirst`/`truncateSecond` but they are all strange

src/core/Carousel/Carousel.tsx
mayank99: I think these dots need to be accessible to screen readers.

Currently the CSS is just setting `display: none` but we should probably use visually hidden styles

src/core/Carousel/Carousel.tsx
mayank99: `ref` never comes from props anyway. I can forwardRef the component, but that still doesn't require changing the props here

src/core/Carousel/Carousel.tsx
mayank99: I think only `Carousel.Dots` would be useful outside `Carousel`, so I've added some extra props to that.

src/core/Carousel/Carousel.tsx
bentleyvk: I am just wondering whether it is worth having these all handlers.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,335 @@
...
+              id={`${idPrefix}--dot-${index}`}
+              aria-controls={`${idPrefix}--slide-${index}`}
+            />
+          );
+        }),
+    [slideCount, currentIndex, maxCount, idPrefix, handleSlideChange],
+  );
+
+  return (
+    <div
```
bentleyvk, line 280: Shouldn't this be `ol` or `ul` and items `li` elements?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,335 @@
...
+  const context = React.useContext(CarouselContext);
+  if (!context) {
+    throw new Error('CarouselNavigation should be used inside Carousel');
+  }
+
+  const { slideCount, setCurrentIndex, keysPressed } = context;
+
+  return (
+    <nav className={cx('iui-carousel-navigation', className)} {...rest}>
+      <div className='iui-carousel-navigation-left'>
```
veekeys, line 164: If you gonna add animation, hover part was interestin as this whole article too
https://www.smashingmagazine.com/2016/07/ten-requirements-for-making-home-page-carousels-work-for-end-users/#the-10-carousel-requirements

veekeys: But..
https://web.dev/carousel-best-practices/#don't-use-autoplay
:D

src/core/Carousel/Carousel.tsx
veekeys: yeah..on quick googling couldnt find anything on this, so.. I guess doesnt matter

src/core/Carousel/Carousel.tsx
mayank99: yeah, after spending some time on it, i removed autoplay because of all the requirements (hover, focus, reduced motion, page visibility)

src/core/Carousel/Carousel.tsx
mayank99: I changed it because the `tab` and `tablist` roles should be enough and it was adding an extra element for no reason. Current approach will reduce the chance of misuse (we are expecting dots to be usable outside carousel).

W3 has the same in their [example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/carousel/carousel-2-tablist.html) 
image

src/core/Carousel/Carousel.tsx
```diff
@@ -113,35 +114,64 @@ const CarouselSlider = (props: React.ComponentPropsWithoutRef<'ol'>) => {
...
     [props.children, idPrefix],
-  );
+  ) as React.ReactNode[];
+
+  React.useEffect(() => {
+    if (!sliderRef.current) {
+      return;
+    }
+
+    sliderRef.current.children.item(currentIndex)?.scrollIntoView();
```
bentleyvk, line 131: What happens if auto switching is on in carousel and user has scrolled down the page so the carousel is not in the viewport anymore? Will it scroll the page up?

src/core/Carousel/Carousel.tsx
mayank99: It's needed now because I've moved the state update into keyup to fix a bug when you hold down the key for too long.

src/core/Carousel/Carousel.tsx
bentleyvk: I think we can leave that `Carousel` has subcomponents as its properties but we should have each component in separate file - for me it is more readable.

src/core/index.ts
```diff
@@ -270,3 +270,6 @@ export {
   MiddleTextTruncation,
 } from './utils';
 export type { ThemeType, MiddleTextTruncationProps } from './utils';
+
+export { Carousel } from './Carousel';
```
bentleyvk, line 274: Move up

src/core/Carousel/Carousel.tsx
bentleyvk: Ok, still strange behavior 😕
carousel-scroll-top

bentleyvk: This seems to fix the issue: 
image

src/core/Carousel/Carousel.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+  // 0 -> 1
+  const keyState2 = userEvent.keyboard('{ArrowLeft>}');
+  expect(prev).toHaveAttribute('data-pressed', 'true');
+  userEvent.keyboard('{/ArrowLeft}', { keyboardState: keyState2 });
+  expect(prev).not.toHaveAttribute('data-pressed');
+  expect(dots[0]).toHaveClass('iui-active');
+});
+
+// TODO: Write more tests here!
```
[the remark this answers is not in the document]
veekeys: Good, cause I have suggestions:
- add test cases to use components outside context to test error and success accordingly
- would it be possible to test that case with intersection observer and keyboard, which you were trying to solve?
- maybe some other important code parts, which you noticed while developing..?

veekeys (commented):
Looks pretty good. I just really dont like the dots animation....

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+import SvgChevronRight from '@itwin/itwinui-icons-react/cjs/icons/ChevronRight';
+import '@itwin/itwinui-css/css/carousel.css';
+
+export type CarouselProps = {
+  /**
+   * Index of the currently shown slide.
+   * Can be used to set the default index or control the active slide programmatically.
+   * @default 0
+   */
+  activeSlideIndex?: number;
```
bentleyvk, line 24: Then we also need `onSlideChange`

mayank99: Is it worth it? `onSlideChange` is present in the dots

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+} & React.ComponentPropsWithoutRef<'div'>;
+
+export const CarouselContext = React.createContext<
```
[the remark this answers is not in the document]
mayank99: I thought could be useful to build custom stuff. Mainly would be needed to set value of `isManuallyUpdating` because that messes with slide updates

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  const { currentIndex, isManuallyUpdating, setCurrentIndex } = context;
+
+  const updateActiveIndex = React.useCallback(() => {
+    if (!isManuallyUpdating.current) {
+      setCurrentIndex((prev) => (prev !== index ? index : prev));
+    }
+    if (currentIndex === index) {
+      isManuallyUpdating.current = false;
```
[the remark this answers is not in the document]
mayank99: yeah, if by "revert" you mean the navigation is completed and ready for any future scrolling

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+      {children}
+    </li>
+  );
+});
+
+/**
+ * The CarouselNavigation component by default consists of the previous/next slide buttons
+ * shown on the left and right, and the CarouselDots component shown in the middle.
+ *
+ * `children` can be specified to override what is shown in this section.
```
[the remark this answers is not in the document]
mayank99: _this_ section. The navigation section.

src/core/Carousel/Carousel.tsx
mayank99: I forwarded in all subcomponents but left it here because it was messing with the dot notation. Typescript doesn't like if you try attaching subcomponents to a forwardRef component

image

src/core/Carousel/Carousel.tsx
mayank99: added shiftKey

src/core/Carousel/Carousel.tsx
mayank99: added

mayank99: added useEffect

src/core/Carousel/Carousel.tsx
mayank99: Would it make sense to have CarouselSlider and CarouselSlide in one file and CarouselDots and CarouselDot in one file? They are very tightly coupled

src/core/Carousel/Carousel.tsx
bentleyvk: For me it looks odd to set index in one place and have callback in other.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  React.useLayoutEffect(() => {
+    setSlideCount(React.Children.count(props.children));
+  }, [props.children, setSlideCount]);
+
+  const items = React.useMemo(
+    () =>
+      React.Children.map(props.children, (child, index) =>
+        React.isValidElement(child)
+          ? React.cloneElement(child, {
+              id: `${idPrefix}--slide-${index}`,
```
bentleyvk, line 168: So if user sets `id` on `CarouselSlide`, it will be overwritten. Because `CarouselSlide` props extends default `li` props.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  const context = React.useContext(CarouselContext);
+  if (!context) {
+    throw new Error('CarouselSlider must be used within Carousel');
+  }
+
+  const { currentIndex, isManuallyUpdating, setCurrentIndex } = context;
+
+  const updateActiveIndex = React.useCallback(() => {
+    if (!isManuallyUpdating.current) {
+      setCurrentIndex((prev) => (prev !== index ? index : prev));
```
bentleyvk, line 217: You still have `currentIndex` in deps list. So maybe just move this condition to the `if` condition above or have it in different `useEffect`.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  const { className, children, ...rest } = props;
+
+  const context = React.useContext(CarouselContext);
+  if (!context) {
+    throw new Error('CarouselSlider must be used within Carousel');
+  }
+
+  const { currentIndex, isManuallyUpdating, setCurrentIndex } = context;
+
+  const updateActiveIndex = React.useCallback(() => {
```
bentleyvk, line 215: Can you explain this function?

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+ *
+ * `children` can be specified to override what is shown in this section.
+ */
+const CarouselNavigation = React.forwardRef<
+  HTMLElement,
+  React.ComponentPropsWithoutRef<'nav'>
+>((props, ref) => {
+  const { className, children, ...rest } = props;
+
+  const context = React.useContext(CarouselContext);
```
bentleyvk, line 257: Seems like this could be moved to a hook.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+      /** Number of slides in the carousel. Gets set in CarouselSlider for reading in CarouselDots.  */
+      slideCount: number;
+      /** State updater for slideCount. */
+      setSlideCount: (length: number | ((old: number) => void)) => void;
+      /** Stateful value of the left/right arrow keys currently pressed. */
+      keysPressed: Record<string, boolean>;
+      /** Prefix used for setting id for internal carousel components. */
+      idPrefix: string;
+      /** Ref object used for preventing intersection observer from breaking manual slide updates. */
+      isManuallyUpdating: React.MutableRefObject<boolean>;
```
bentleyvk, line 42: I still don't understand this prop.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+
+  return (
+    <div
+      className={cx('iui-carousel-navigation-dots', className)}
+      role='tablist'
+      aria-label='Slides'
+      ref={ref}
+      {...rest}
+    >
+      {children ?? dots}
```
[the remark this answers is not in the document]
bentleyvk: Need to use `useOverflow`?

src/core/Carousel/Carousel.tsx
mayank99: Are these comments helpful?
image

src/core/Carousel/Carousel.tsx
mayank99: How about now?
image

src/core/Carousel/Carousel.tsx
mayank99: Yeah, but then you lose the more specific error messages where you can use the name of the component

src/core/Carousel/Carousel.tsx
mayank99: So if we take `id` from user, then we need to somehow set in the `aria-controls` of dots, which adds complexity. Do you think it's worth it?

src/core/Carousel/Carousel.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+    '.iui-carousel-navigation-right > button',
+  );
+
+  userEvent.tab();
+
+  // 0 -> 1
+  const keyState1 = userEvent.keyboard('{ArrowRight>}');
+  expect(next).toHaveAttribute('data-pressed', 'true');
+  userEvent.keyboard('{/ArrowRight}', { keyboardState: keyState1 });
+  expect(next).not.toHaveAttribute('data-pressed');
```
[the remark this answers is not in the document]
mayank99: I'm setting undefined in my code

src/core/Carousel/Carousel.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+      <Carousel.Slider>
+        <Carousel.Slide>1</Carousel.Slide>
+        <Carousel.Slide>2</Carousel.Slide>
+        <Carousel.Slide>3</Carousel.Slide>
+      </Carousel.Slider>
+      <Carousel.Navigation />
+    </Carousel>,
+  );
+
+  const root = container.querySelector('section') as HTMLElement;
```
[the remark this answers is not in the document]
mayank99: I prefer the explicit-looking check given by `toHaveClass`

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+ */
+const CarouselDot = React.forwardRef<
+  HTMLButtonElement,
+  {
+    /** Is this dot currently active? */
+    isActive?: boolean;
+    /** Should be set to true for dots that are not visible, i.e. truncated. */
+    isInvisible?: boolean;
+    /** Should be set to true for dots that are one spot from the edge of truncation. The dot size becomes small.  */
+    isSecond?: boolean;
```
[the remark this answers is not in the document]
mayank99: Renamed

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+              data-pressed={keysPressed['ArrowLeft'] || undefined}
+            >
+              <SvgChevronLeft />
+            </IconButton>
+          </div>
+
+          <CarouselDots />
+
+          <div className='iui-carousel-navigation-right'>
+            <IconButton
```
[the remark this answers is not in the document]
mayank99: Not sure how to go about fixing the multiple button clicks.

Can't do anything about animation here, but if you have a suggestion I can implement the CSS.

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+          </div>
+        </>
+      )}
+    </nav>
+  );
+});
+
+/**
+ * The CarouselDots component shows a list of CarouselDot components which can be used to
+ * choose a specific slide. If used as a descandant of Carousel, then this component does not need
```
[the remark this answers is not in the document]
mayank99: Changed

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+ * CarouselSlide is used for the actual slide content. The content can be specified through `children`.
+ *
+ * It is recommended that the slide content bring its own dimensions (esp. height) and that
+ * the dimensions should be the same for all slides.
+ */
+const CarouselSlide = React.forwardRef<
+  HTMLLIElement,
+  { index?: number } & React.ComponentPropsWithoutRef<'li'>
+>(({ index, ...props }, ref) => {
+  const { className, children, ...rest } = props;
```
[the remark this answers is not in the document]
mayank99: Changed

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+
+/**
+ * CarouselSlide is used for the actual slide content. The content can be specified through `children`.
+ *
+ * It is recommended that the slide content bring its own dimensions (esp. height) and that
+ * the dimensions should be the same for all slides.
+ */
+const CarouselSlide = React.forwardRef<
+  HTMLLIElement,
+  { index?: number } & React.ComponentPropsWithoutRef<'li'>
```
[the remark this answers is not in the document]
mayank99: Added

src/core/index.ts
mayank99: Moved

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: Now it's in a separate file so it's exported anyway

src/core/Carousel/Carousel.tsx
mayank99: ok, split into 7 files then.

Tests currently pending

src/core/Carousel/Carousel.tsx
mayank99: Added

src/core/Carousel/Carousel.tsx
veekeys: a maaaan...
Could have fixed everything and then just move those files, if so important... Now need to review everything in not very comfortable way :./

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: Good, wont we have issues that props are not exported...?

src/core/Carousel/Carousel.tsx
mayank99: what's not comfortable? 

I made only two commits, the first one is purely cosmetic: moving stuff around, adding comments, creating types for props, destructuring props, etc.
image

adamhock (commented):
https://user-images.githubusercontent.com/78564221/158849170-81a9e9e6-2325-4f5e-869d-516c45423b28.mp4

I came across this when trying to use the trackpad to scroll through the carousel horizontally with multiple swipes at a time. I'm not sure if this is the same behavior on mobile when swiping through the carousel or if there's even a fix for this, but this is the only issue I noticed outside of the other comments made

adamhock:
> @adamhock Do you have the latest css installed (`yarn install`)?

Yes

adamhock (approved):
I'll approve these changes assuming the other comments are resolved

src/core/Carousel/Carousel.tsx
mayank99: Done

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: what issues? props should almost never need to be exported imo, but I have `CarouselProps` in the barrel at least

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  );
+};
+
+/**
+ * CarouselSlider is the scrollable list that should consist of CarouselSlide components.
+ */
+const CarouselSlider = React.forwardRef<
+  HTMLOListElement,
+  React.ComponentPropsWithoutRef<'ol'>
+>((props, ref) => {
```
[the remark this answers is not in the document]
mayank99: Fixed

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,494 @@
...
+  if (!context) {
+    throw new Error('CarouselSlider must be used within Carousel');
+  }
+
+  const { currentIndex, setSlideCount, idPrefix } = context;
+  const sliderRef = React.useRef<HTMLOListElement>(null);
+  const refs = useMergedRefs(sliderRef, ref);
+
+  React.useLayoutEffect(() => {
+    setSlideCount(React.Children.count(props.children));
```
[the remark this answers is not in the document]
mayank99: Fixed

src/core/Carousel/Carousel.tsx
mayank99: Fixed and added Safari fallback
image

Would be good if @gretanausedaite can check

stories/core/Carousel.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    { from: '#e65c00', to: '#f9d423' },
+    { from: '#1488cc', to: '#2b32b2' },
+    { from: '#bbd2c5', to: '#536976' },
+    { from: '#9796f0', to: '#fbc7d4' },
+    { from: '#b79891', to: '#94716b' },
+    { from: '#acb6e5', to: '#86fde8' },
+  ];
+
+  return (
+    <Carousel style={{ width: 'min(800px, 90vw)' }} {...args}>
```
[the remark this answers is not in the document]
mayank99: Fixed in [`f1b45e6`](https://github.com/iTwin/iTwinUI-react/pull/569/commits/f1b45e69629d6f93f3823b124ce90e205e57e438)

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: This should be kinda fixed with https://github.com/iTwin/iTwinUI/pull/527. Still want me to add `useOverflow`?

src/core/Carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,94 @@
...
+    'not-mounted' | 'just-mounted' | 'post-mount'
+  >('not-mounted');
+
+  React.useLayoutEffect(() => {
+    if (mountState === 'not-mounted') {
+      setMountState('just-mounted');
+    } else if (mountState === 'just-mounted') {
+      setMountState('post-mount');
+    }
+  }, [mountState]);
```
mayank99, line 53: Need opinions on this prop. I wanted a triple-state value, so I went with a string union. Another option would be to use `boolean | undefined` or `boolean | 'not-mounted'`.

src/core/Carousel/Carousel.tsx
bentleyvk: Will it throw an error if not supported?

src/core/Carousel/CarouselDots.tsx
```diff
@@ -0,0 +1,138 @@
...
+ *
+ * @example
+ * <Carousel.Dots
+ *   length={10}
+ *   maxCount={3}
+ *   currentIndex={current}
+ *   onSlideChange={(i) => setCurrent(i)}
+ * />
+ */
+export const CarouselDots = React.forwardRef<HTMLDivElement, CarouselDotsProps>(
```
bentleyvk, line 45: Maybe rename to `CarouselDotsList` because for a few moments I thought that VS Code is showing me the same file in the list then I thought it might be some bug 😀

src/core/Carousel/Carousel.tsx
bentleyvk: Have component name as a parameter then? Or user could just look at call stack and see that Error was called from hook that was called from a component.

src/core/Carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,94 @@
...
+      return;
+    }
+
+    try {
+      slideToShow?.scrollIntoView({ block: 'nearest' });
+    } catch (_) {
+      // Safari fallback
+      sliderRef.current.scrollTo({ left: slideToShow?.offsetLeft });
+    }
+  }, [mountState, currentIndex]);
```
bentleyvk, line 82: If deps list is empty, then it will scroll once on init, no? Then no need for `mountState`

src/core/Carousel/CarouselDots.tsx
```diff
@@ -0,0 +1,138 @@
...
+  (props, ref) => {
+    const {
+      maxCount = 5,
```
bentleyvk, line 48: Shouldn't we take `maxCount` dynamically and not ask user to pass it? Because we kind of know one item width and container width.

src/core/Carousel/CarouselSlider.tsx
mayank99: It can't scroll if it's not mounted yet

src/core/Carousel/Carousel.tsx
mayank99: > Will it throw an error if not supported?

That's also something @gretanausedaite could check.

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
bentleyvk: Yes, I want 😀

src/core/Carousel/CarouselSlider.tsx
bentleyvk: But isn't it already mounted in first useEffect call?

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: so now I guess dep could also be `items.length` ?

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: well, remember we had no props expoted before at all, but then people needed those for different reasons...
Well, leave for now, I guess :/

stories/core/Carousel.stories.tsx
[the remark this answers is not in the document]
veekeys: In basic story ot works, but controlled story does not for me :/

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: I could not find the exact example I found months ago (and chat history in teams is of course cleaned), but if we had at least something like here, would be already much better
https://github.com/fulopdaniel/react-carousel-dots

And about that bug in GIF.. Have you tried with `requestAnimationFrame` and canceling it on nex press...? Maybe would help somehow.. If not, need to have a look and think, what else we can do :/

src/core/Carousel/CarouselDots.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useTheme } from '../utils';
+import { CarouselDot } from './CarouselDot';
+import '@itwin/itwinui-css/css/carousel.css';
+
+type CarouselDotsProps = {
+  /** Number of total dots/slides in the carousel. Will be inferred from Carousel context. */
+  length?: number;
+  /** Maximum number of dots to show in the viewport. @default 5 */
+  maxCount?: number;
+  /** Index of currently acitve dot. Will be inferred from Carousel context. */
```
veekeys, line 17: acitve

Can you install spell checker in vscode please? :D

src/core/Carousel/CarouselSlider.tsx
veekeys: Yeah, it should be mounted and ref should be available. Did you face some other issue so you added `mountState` ?

src/core/Carousel/CarouselSlider.tsx
adamhock: I like the string union. An alternative `boolean | 'just-mounted'` could work instead of `boolean | 'not-mounted'`.

src/core/Carousel/CarouselDots.tsx
mayank99: image

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: Changed

stories/core/Carousel.stories.tsx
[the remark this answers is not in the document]
mayank99: The Controlled story shows code for using the `activeSlideIndex` on `onSlideChange`

so I have disabled the `activeSlideIndex` control

src/core/Carousel/CarouselDots.tsx
veekeys: so it didnt mark or whats the deal?

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: Tried a few different ways with `requestAnimationFrame`. No effect.
image

So I've added a hack using `event.detail` where I'm scrolling instantly if clicked too many times. This is how it looks:

https://user-images.githubusercontent.com/9084735/159352793-9352a7e5-f0a9-4fe6-92da-79038b2ee51a.mp4

Also I've removed dots animation (temporarily using inline `style`) and added smooth scrolling to dots container. This is how it looks:

https://user-images.githubusercontent.com/9084735/159352966-50bb7cf9-c2a5-4e69-968d-51ee2abb501b.mp4

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
mayank99: `useOverflow` is going to be difficult because it relies on removing stuff from the DOM, whereas we want to keep it in the DOM for activating later

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,145 @@
...
+ * @example
+ * <Carousel>
+ *   <Carousel.Slider>
+ *     <Carousel.Slide>...</Carousel.Slide>
+ *     <Carousel.Slide>...</Carousel.Slide>
+ *     <Carousel.Slide>...</Carousel.Slide>
+ *   </Carousel.Slider>
+ *   <Carousel.Navigation />
+ * </Carousel>
+ */
```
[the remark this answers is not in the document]
mayank99: Fixed

src/core/Carousel/Carousel.tsx
mayank99: Oh ell, looks like this will not get tested on macos, so I've removed the try/catch and replaced `scrollIntoView` with _only_ `scrollTo`. It feels somewhat jittery on mobile emulation but works ok otherwise.

src/core/Carousel/CarouselDots.tsx
mayank99: Reinstalled and it's working now

src/core/Carousel/CarouselDots.tsx
mayank99: Renamed

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: Now the issue with multiple clicks is fixed, but if I use keyboard to navigate, it is same issue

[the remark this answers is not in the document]
veekeys: Dots animation also look much better!
Can you try to add back those small dots back and try to use `useOverflow` to cut overflow and just have those smaller dots at the sides? (if enough space)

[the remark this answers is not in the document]
veekeys: Oh, and also, now on that fast clicks, active dot seems lost :( all dots are marked as inactive

src/core/Carousel/CarouselDotsList.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { CarouselContext } from './CarouselContext';
+import { useMergedRefs, useResizeObserver, useTheme } from '../utils';
+import { CarouselDot } from './CarouselDot';
+import '@itwin/itwinui-css/css/carousel.css';
+
+type CarouselDotsListProps = {
+  /** Number of total dots/slides in the carousel. Will be inferred from Carousel context. */
+  length?: number;
+  /** Maximum number of dots to show in the viewport. */
+  maxCount?: number;
```
mayank99, line 16: This `maxCount `prop is in a weird situation now after adding the resizeObserver logic. Maybe I should just remove this prop? If not, we will have to do some hacks like this:

```tsx
style={{ maxWidth: maxCount * dotSize }}
```

src/core/Carousel/Carousel.tsx
FlyersPh9: Confirmed is working as expected in Safari on macOS.

veekeys:
Managed to get this weird state...
If you press on right arrow with mouse and then use keyboard and press left, you get this:
image

And whenever press left, you get focused style on both buttons

src/core/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,152 @@
...
+  React.forwardRef<HTMLElement, CarouselProps>((props, ref) => {
+    const {
+      activeSlideIndex: userActiveIndex = 0,
+      onSlideChange,
+      children,
+      ...rest
+    } = props;
+
+    // Generate a stateful random id if not specified
+    const [id] = React.useState(
```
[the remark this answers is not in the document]
mayank99: I need the value to stay the same between renders.

It's the same implementation as ComboBox. But I think it will not be stable between SSR and hydration, so we should come up with a better solution.

src/core/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,83 @@
...
+          });
+        }
+      }
+    },
+    [isActive],
+  );
+  const refs = useMergedRefs(ref, dotRef);
+
+  React.useEffect(() => {
+    if (justMounted.current) {
```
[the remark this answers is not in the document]
mayank99: Scroll down

src/core/Carousel/CarouselDot.tsx
```diff
@@ -0,0 +1,83 @@
...
+ * It should be used as a child of `CarouselDotsList`.
+ */
+export const CarouselDot = React.forwardRef<
+  HTMLButtonElement,
+  CarouselDotProps
+>((props, ref) => {
+  const { isActive, isSmaller, isSmall, className, ...rest } = props;
+
+  const justMounted = React.useRef(false);
+  const motionOk = React.useRef(
```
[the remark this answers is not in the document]
mayank99: I got this name from https://open-props.style/#media-queries and I've seen it in other places too. https://web.dev/building-split-text-animations/#preparing-motion-conditionals

src/core/Carousel/CarouselDot.tsx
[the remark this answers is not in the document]
veekeys: I did, cant see still
 
 
image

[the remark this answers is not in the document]
mayank99: Screen Shot 2022-03-24 at 7 32 04 AM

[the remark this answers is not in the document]
mayank99: oh ok I see, there's a typo 😆 it should be initialized with `true`

src/core/Carousel/CarouselDotsList.tsx
```diff
@@ -0,0 +1,157 @@
...
+    [context, onSlideChange],
+  );
+
+  if (
+    slideCount == undefined ||
+    currentIndex == undefined ||
+    (!idPrefix && !children)
+  ) {
+    throw new Error(
+      'CarouselDotsList needs to be used inside Carousel, or missing props need to be specified',
```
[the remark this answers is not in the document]
mayank99: Made error message longer

src/core/Carousel/CarouselDot.tsx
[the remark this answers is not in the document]
mayank99: Ok fixed now.

To answer your question, it is needed for initializing default state with instant scroll. e.g. if the default dot is not in view, then we don't want it to smooth scroll when the page first loads.

This same logic is in CarouselSlider too.

src/core/Carousel/CarouselDots.tsx
mayank99: Removed maxCount

src/core/Carousel/CarouselDotsList.tsx
mayank99: Removed maxCount 🤷

src/core/Carousel/CarouselDot.tsx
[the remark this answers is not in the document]
mayank99: this has been removed now bc smooth scrolling two elements doesn't work in chrome

src/core/Carousel/CarouselDot.tsx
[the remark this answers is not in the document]
mayank99: update: I've removed smooth scrolling from dots because it doesn't work in chrome using `scrollIntoView`, so all this code has been moved into `CarouselDotsList` and uses `scrollTo`. It's working in all browsers now

src/core/Carousel/Carousel.tsx
bentleyvk: carousel-resize-hide

mayank99: can you stop resizing everything? 🔪 😄

bentleyvk: Just doing basic QA stuff 😀

src/core/Carousel/CarouselSlide.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * the dimensions should be the same for all slides.
+ */
+export const CarouselSlide = React.forwardRef<
+  HTMLLIElement,
+  CarouselSlideProps
+>((props, ref) => {
+  const { index, className, children, ...rest } = props;
+
+  const context = React.useContext(CarouselContext);
+  if (!context || index == null) {
```
bentleyvk, line 31: `index == null` and this error message is completely unrelated

src/core/Carousel/Carousel.tsx
mayank99: should be good now [`8bef04f`](https://github.com/iTwin/iTwinUI-react/pull/569/commits/8bef04f229db88e5497f9ab6692002b66afa49ed)

veekeys:
So ok.
I think state improved, much better.
Last 2 things 
first comment in this https://github.com/iTwin/iTwinUI-react/pull/569#issuecomment-1077454351
and I guess minor (comparing to clipping), but resizing sometimes gives this state
image

Not sure if you checked 1st thing? 
Last one as I said, really minor.. I could approve.

src/core/Carousel/Carousel.tsx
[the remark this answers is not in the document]
veekeys: Hopefully, we will be able to get rid of it
https://reactjs.org/docs/hooks-reference.html#useid

mayank99:
> first comment in this https://github.com/iTwin/iTwinUI-react/pull/569#issuecomment-1077454351

That is happening because anything you click gets focus. Initially there is no ring shown because of `:focus-visible` styling but when you start using keyboard, it leaves focus on the last element you clicked and. uses `:focus` styling.

I've fixed it by manually focus to the carousel root.

> I guess minor (comparing to clipping), but resizing sometimes gives this state image

It's the same as clipping (there is padding being cut off on the leftmost dot).

> Oh, ofc, with visual tests passing :D

can't run visual tests on macOS so I guess I will boot up Windows before merging this PR 😄

stories/core/Carousel.stories.tsx
```diff
@@ -105,3 +110,99 @@ Controlled.args = {};
...
+    { from: '#ffe000', to: '#799f0c' },
+    { from: '#e65c00', to: '#f9d423' },
+    { from: '#1488cc', to: '#2b32b2' },
+    { from: '#bbd2c5', to: '#536976' },
+    { from: '#9796f0', to: '#fbc7d4' },
+    { from: '#b79891', to: '#94716b' },
+    { from: '#acb6e5', to: '#86fde8' },
+  ];
+
+  const id = useId();
```
veekeys, line 129: Why you are hiding that id :DDD

stories/core/Carousel.stories.tsx
```diff
@@ -105,3 +110,99 @@ Controlled.args = {};
...
+              style={{
+                background: `linear-gradient(to right, ${from}, ${to})`,
+                height: 'min(600px, 70vh)',
+                display: 'grid',
+                placeItems: 'center',
+                fontSize: 48,
+                color: 'hsl(0deg 0% 100% / 0.7)',
+              }}
+            >
+              {index}
```
veekeys, line 176: Maybe start from `1` as in other stories? :D

mayank99: done

stories/core/Carousel.stories.tsx
```diff
@@ -105,3 +110,99 @@ Controlled.args = {};
 Controlled.argTypes = {
   activeSlideIndex: { control: { disable: true } },
 };
+
+const useId = () => 'my-custom-carousel';
+export const Custom: Story<CarouselProps> = () => {
```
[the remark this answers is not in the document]
mayank99: done

stories/core/Carousel.stories.tsx
mayank99: thought we could just use the new `useId` when we upgrade to react 18

veekeys (approved):
Good job! 🥳

adamhock (approved):
Looks good to me!

PR #579, opened 2022-03-07
fix(ButtonGroup): Add vertical support to `overflowButton`

### LINKED ISSUE #573: ButtonGroup: vertical orientation
CSS support is there, need to add it to `ButtonGroup` now.

mayank99:
Added `orientation` prop to `useOverflow` and removed inline `width: 100%` as it's now handled in iTwinUI-css 0.50.1.

Added new story with hardcoded `height` (`maxHeight` doesn't work).

image

Closes #573 (for real this time)

## Checklist


src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -96,8 +98,9 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
           className,
         )}
         style={{
-          ...(!!overflowButton &&
-            orientation === 'horizontal' && { width: '100%' }),
+          ...(!!overflowButton && orientation === 'horizontal'
+            ? { width: '100%' }
+            : { height: '100%' }),
```
mayank99, line 103: Maybe we should move these to CSS?

veekeys (approved):
Would be good to have visual test for vertical too.
Do you know, why it works differently with height? (maxWIdth vs height)

mayank99:
> Would be good to have visual test for vertical too.

We already have one. This is just fixing overflow

> Do you know, why it works differently with height? (maxWIdth vs height)

I am not sure but I think `maxHeight` is not enough for `height: 100%` to work. As an alternative, the user could apply `maxHeight` on the ButtonGroup instead of the wrapper. That's why I'm thinking we should move these inline rules to CSS.

elephantcatdog (commented):
The overflow story looks wrong for me. 
image


Nevermind, something was wrong with my project. I did notice that if I shrink the screen the button bar shrinks, but it doesn't grow when I increase the screen again.

stories/core/Buttons/ButtonGroup.stories.tsx
```diff
@@ -77,6 +77,7 @@ export const Overflow: Story<ButtonGroupProps> = (args) => {
           maxWidth: 'clamp(300px, 50%, 100%)',
           border: '1px solid hotpink',
           padding: 8,
+          height: 'max(200px, 30vh)',
```
elephantcatdog, line 80: I think there's some problems with this story:

1. I can see why you wouldn't want to add another story for just vertical overflow, but it's not clear that the extra height is meant to be used to look at the overflow of a vertical button bar. I would add a note like the one already there ("Resize the viewport to see overflow behavior.") to say this explicitly.
2. Because of the note, I thought I would be able to resize the vertical box as well. I think this should be added in, or the vertical overflow should have its own story, perhaps included as part of the "Vertical" story.
3. The horizontal story resizes when the box becomes smaller, but does not resize when the box becomes bigger.

https://user-images.githubusercontent.com/37936169/158854082-8fb4dc14-24e4-4c08-9561-b80234a80769.mp4

elephantcatdog (approved):
The code looks fine to me, but I think the story could be better.

PR #582, opened 2022-03-09
chore: Add `once` param to `useIntersection` hook

mayank99:
For some reason, the `useIntersection` hook only triggers the `onIntersect` callback once, so I've added a param to be able to trigger it more than once.

This is needed for watching the slide scrolls in #569 

## Checklist


src/core/utils/hooks/useIntersection.test.tsx
```diff
@@ -135,3 +135,27 @@ it('should do nothing when element is missing', () => {
...
+    const onIntersect = jest.fn();
+    const { result } = renderHook(() => useIntersection(onIntersect, {}, once));
+
+    act(() => {
+      result.current(mockedElement);
+    });
+
+    expect(onIntersect).not.toHaveBeenCalled();
+    mockIntersection(mockedElement);
+    mockIntersection(mockedElement);
```
bentleyvk, line 155: Minor, but 2 times is enough 😀

bentleyvk (approved):

src/core/utils/hooks/useIntersection.tsx
```diff
@@ -40,15 +42,17 @@ export const useIntersection = (
       observer.current = new IntersectionObserver(
         ([entry], obs) => {
           if (entry.isIntersecting) {
-            obs.disconnect();
+            if (once) {
+              obs.disconnect();
```
bentleyvk, line 46: Hm, do we need to have `useEffect` that would disconnect on unmount?

src/core/utils/hooks/useIntersection.test.tsx
mayank99: how about 3?

PR #588, opened 2022-03-15
feat(Modal): Add full page modal

elephantcatdog:
Added full page modal
- Added optional parameter `modalSize`. Used a type to implement switching instead of a boolean in anticipation of a container sized modal in the future (or any other size).
- Class name `"iui-modal-full-page"` is added to the modal if it's full page.

Added animations for small and full page modal
- Animations/transitions match those in the CSS project.
- Uses CSSTransition to accomplish transitions. Placed outside of modal parent div to encompass transitions in both the dialog and the overlay.

Added class name `"iui-modal-content"` to story so that full page modal button bar is formatted correctly.

Related to https://github.com/iTwin/iTwinUI/pull/489. Must be linked with that version of @itwin-css for this PR to work.

https://user-images.githubusercontent.com/37936169/158485177-5776473f-0c37-4cde-aeb2-8666ed446d63.mp4

https://user-images.githubusercontent.com/37936169/158485184-2db2d905-a4cb-41c9-92ea-9da0f7481640.mp4

Closes #? - issue not created

## Checklist


stories/core/Modal.stories.tsx
```diff
@@ -90,7 +90,7 @@ export const Basic: Story<ModalProps> = ({
         onKeyDown={action('onKeyDown', { depth: 1 })}
         {...rest}
       >
-        <div>
+        <div className='iui-modal-content'>
```
gretanausedaite, line 93: Maybe content needs to have its own component? Like `ModalButtonBar`. Then users won't need to add `iui-modal-content` manually to their code.

elephantcatdog:
Full page modal test is failing because the CSS isn't a thing yet.

src/core/Modal/Modal.tsx
```diff
@@ -170,10 +176,20 @@ export const Modal = (props: ModalProps) => {
 
   return !!container ? (
     ReactDOM.createPortal(
-      isOpen && (
-        <FocusTrap>
+      <FocusTrap>
+        <CSSTransition
+          in={isOpen}
+          classNames='iui-modal-animation'
+          timeout={4000}
```
bentleyvk, line 183: Change to `400` as it is in CSS animations.

src/core/Modal/Modal.tsx
```diff
@@ -170,10 +176,20 @@ export const Modal = (props: ModalProps) => {
 
   return !!container ? (
     ReactDOM.createPortal(
-      isOpen && (
-        <FocusTrap>
+      <FocusTrap>
```
bentleyvk, line 179: Move it inside `CSSTransition`

bentleyvk:
For me full-modal exit animation is a bit too long, I would like it to be faster. What you think? @FlyersPh9 @gretanausedaite

bentleyvk:
What changed in old test images? Can't really see.

stories/core/Modal.stories.tsx
```diff
@@ -90,15 +91,17 @@ export const Basic: Story<ModalProps> = ({
...
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
-          eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad
-          minim veniam, quis nostrud exercitation ullamco laboris nisi ut
-          aliquip ex ea commodo consequat. Duis aute irure dolor in
-          reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
-          pariatur. Excepteur sint occaecat cupidatat non proident, sunt in
-          culpa qui officia deserunt mollit anim id est laborum.
-        </div>
+        <ModalContent>
+          <div>
```
bentleyvk, line 95: Is this `div` wrapper still needed?

src/core/Modal/ModalContent.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export type ModalContentProps = {
+  /**
+   * Main content in the modal.
+   */
+  children: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * Container for content in modal.
```
bentleyvk, line 18: Uppercase and wrap `Modal`

src/core/Modal/Modal.tsx
```diff
@@ -60,6 +61,10 @@ export type ModalProps = {
    * @default document
    */
   ownerDocument?: Document;
+  /**
+   * Size of the modal.
+   */
+  modalSize?: 'small' | 'fullPage'; //later add size container?
```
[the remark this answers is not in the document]
bentleyvk: When I see `size` prop, I expect modal to be small, medium and large. But in this case modal has different types. I would say to rename prop to `variant`/`styleType`/`type` and have values `default` and `full-page`

stories/core/Modal.stories.tsx
```diff
@@ -11,6 +11,7 @@ import { Modal, ModalButtonBar } from '../../src/core/Modal';
 import { ModalProps } from '../../src/core/Modal/Modal';
 import { useEffect, useState } from '@storybook/addons';
 import { CreeveyMeta } from 'creevey';
+import ModalContent from '../../src/core/Modal/ModalContent';
```
bentleyvk, line 14: Move this import to line 10. Or you can even import everything from `src/core`.

src/core/Modal/Modal.tsx
bentleyvk: Ah, you also need to add `focusRef` to `FocusTrap`. You can check other components, simplest example `Body` component.

FlyersPh9:
> For me full-modal exit animation is a bit too long, I would like it to be faster. What you think? @FlyersPh9 @gretanausedaite

Without inspecting the code I think the speed is probably correct, but inverting the ease might help.  So if it's `ease-out` swapping it with `ease-in` might help.

gretanausedaite:
> For me full-modal exit animation is a bit too long, I would like it to be faster. What you think? @FlyersPh9 @gretanausedaite

Speed looks good to me.

gretanausedaite (commented):
Dark to light background flicker on exit looks strange, or is it just for me?

https://user-images.githubusercontent.com/83585998/159871217-c4606ef8-c5e0-4572-a6bd-d9bc75b5494a.mov



However, I'm not sure how to solve this.

bentleyvk:
> Dark to light background flicker on exit looks strange, or is it just for me?

It bugs me too but also no idea what would be the best solution. Maybe here also need to adjust animation.

adamhock (commented):
This looks good to me. I have no additional input aside from the comments already made

stories/core/Modal.stories.tsx
elephantcatdog: Done

src/core/Modal/ModalContent.tsx
elephantcatdog: Done

stories/core/Modal.stories.tsx
elephantcatdog: No, removed

src/core/Modal/Modal.tsx
elephantcatdog: Whoops. Done

src/core/Modal/Modal.tsx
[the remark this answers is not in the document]
elephantcatdog: Done

src/core/Modal/Modal.tsx
```diff
@@ -100,6 +105,7 @@ export const Modal = (props: ModalProps) => {
     className,
     style,
     children,
+    styleType: type,
```
bentleyvk, line 108: Why casting to another name?

elephantcatdog: I think this happened when I changed the name. Fixed.

src/core/Modal/Modal.test.tsx
```diff
@@ -46,6 +46,15 @@ it('should render in basic form', () => {
   assertBaseElement(overlay);
 });
 
+it('should render in full page form', () => {
+  renderComponent({ modalSize: 'fullPage' });
```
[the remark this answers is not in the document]
elephantcatdog: Fixed

src/core/Modal/Modal.tsx
elephantcatdog: Done. Changes in CSS were needed but they are mostly complete now. I have a PR that fixes an issue I introduced with the regular modal animations.

elephantcatdog:
I have another PR in CSS (https://github.com/iTwin/iTwinUI/pull/564) to
1. fix an issue with selector names
2. fix a part of the problem with the full page modal exit animation flickering (see below)
3. increase speed of animations a bit (see below)

https://user-images.githubusercontent.com/37936169/160776537-b194e9ed-6c5e-4b82-a615-1e65b08ec090.mp4

src/core/Modal/Modal.tsx
```diff
@@ -60,6 +61,10 @@ export type ModalProps = {
    * @default document
    */
   ownerDocument?: Document;
+  /**
+   * Type of the modal.
+   */
```
gretanausedaite, line 66:
```suggestion
   * @default 'default'
   */
```

src/core/Modal/Modal.tsx
```diff
@@ -100,6 +105,7 @@ export const Modal = (props: ModalProps) => {
     className,
     style,
     children,
+    styleType,
```
gretanausedaite, line 108:
```suggestion
    styleType = 'default',
```

src/core/Modal/Modal.tsx
elephantcatdog: Fixed

src/core/Modal/Modal.tsx
elephantcatdog: Fixed

elephantcatdog:
Updated CSS package. Should be good to go barring any other comments.

veekeys:
All looks good!
Just one question, what has changed in simple modals images? Image comparer doesnt show any difference (or I havent noticed it). Maybe could try reverting those and see, if all passes?

gretanausedaite (commented):
I think storybook code should not look like this? 
image

I think this is because we are using Basic.Bind in Modal stories? Do we want to change that?

mayank99:
@gretanausedaite  this is (yet another) bug in storybook. It can be seen in some other components too.
https://itwin.github.io/iTwinUI-react/?path=/story/core-informationpanel--custom-actions

veekeys:
@elephantcatdog can you try removind `.bind` and see if thats the cause of bad code?
I think I have checked that though and it didnt make a difference.. So it must be something else..

gretanausedaite:
> @gretanausedaite this is (yet another) bug in storybook. It can be seen in some other components too. [itwin.github.io/iTwinUI-react/?path=/story/core-informationpanel--custom-actions](https://itwin.github.io/iTwinUI-react/?path=/story/core-informationpanel--custom-actions)

Ok, got it. So if we are leaving this as is, this PR looks good to me.
I agree with the comment that default modal images should pass without changing them.

elephantcatdog:
I'm not exactly sure why the images changed now. I am apprehensive about trying to revert those changes because I don't want to make a mistake and delete everything.

The problem with the code bug in storybook does have to do with `.bind({})`.
- Similar problem documented on Storybook's github: https://github.com/storybookjs/storybook/issues/12065
- The more binds in a story, the more times the code is repeated
- If you shorten the code to below the repeating line (50 in the image above), then the code doesn't repeat
- If you change the code, then the previous and new code are both shown as duplicated code until you clear the cache
- Removing the binds solves the issue. In this case, should I copy the same code nearly completely to fix the issue? I couldn't figure out another short way of doing it like `.bind({})` does.

veekeys:
Id say then yes, put the code separately into every story, if that actually fixes the bug.
Priority here are users and showing them the correct code.
Or we can complete this PR and do it with separate one maybe for all the broken stories.

stories/core/Modal.stories.tsx
```diff
@@ -110,15 +113,213 @@ export const Basic: Story<ModalProps> = ({
...
+  return (
+    <>
+      <Button styleType='high-visibility' onClick={() => setIsModalOpen(true)}>
+        Open Modal
+      </Button>
+      <Modal
+        isOpen={isModalOpen}
+        title={title}
+        onClose={onClose}
+        onKeyDown={action('onKeyDown', { depth: 1 })}
```
veekeys, line 159: Now its fixed, great finding!!
One minor thing...
Could you also repeat the props specific per story in here?
If you look at storybook, it is not clear, what kinda props were passed to have non dismissible modal, for example.
The ones we set in args, are not really visible in the code.
So in this story just add `isDismissible={false}` just before `{...rest}`

And in other same (in full page you would add `styleType` explicitly, for example).

elephantcatdog: Ah, of course. I wanted it to show up in the controls. Turns out you can do both.

veekeys:
So last thing, we need to try reverting images.
From what I think now, maybe the easiest would be to take images from main branch into yours, run tests. If they pass, just commit those and then it should not show those are changed.

So why dont you try it with command
`git checkout main path/to/image.png`

elephantcatdog:
Thanks, great tip! A lot simpler than what I found.

veekeys (approved):
Looks good!
Dont forget to add info about that `ModalContent` compnent is also added in our Changelog in release PR.

gretanausedaite (approved):
Looks good 🌵

adamhock (approved):

PR #597, opened 2022-03-22
fix(ComboBox): Combine input onChange with user's prop

mayank99:
Previously, user's `onChange` was overriding ours. This change will make sure it's combined properly.

Before:
c1

After:
c2

## Checklist

### DIFF
--- a/src/core/ComboBox/ComboBox.tsx
+++ b/src/core/ComboBox/ComboBox.tsx
@@ -422,7 +422,6 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             ref={inputRef}
             onKeyDown={onKeyDown}
             onFocus={() => setIsOpen(true)}
-            onChange={onInput}
             value={inputValue}
             aria-activedescendant={
               isOpen && focusedIndex > -1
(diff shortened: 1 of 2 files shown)


src/core/ComboBox/ComboBox.tsx
```diff
@@ -436,6 +435,7 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             autoCapitalize='none'
             autoCorrect='off'
             {...inputProps}
+            onChange={onInput}
```
elephantcatdog, line 438: You said that this change would make sure the two onChanges are combined properly, but it seems to me like this would just make our onChange override theirs?

bentleyvk: If you expand whole file, you will see that our `onInput` handler are calling user's `onChange`:
image

src/core/ComboBox/ComboBox.tsx
```diff
@@ -436,6 +435,7 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             autoCapitalize='none'
             autoCorrect='off'
             {...inputProps}
```
bentleyvk, line 437: Add test

elephantcatdog (approved):
Approved, assuming other comments are taken care of.

src/core/ComboBox/ComboBox.tsx
mayank99: Added

bentleyvk (approved):

PR #599, opened 2022-03-24
feat(Table): Default columns customization

### LINKED ISSUE #472: Allow tables to disable selection for specific rows
#### Feature
It would be nice to have the ability to disable only the checkbox in individual rows so that users can still use subcomponents but cannot select the given row. Currently, it's all or nothing with disabling rows.

#### Examples
When checkbox is disabled, selection does not work on it, but all other elements in the row should work.

bentleyvk:
image

Closes #472

## Checklist


bentleyvk:
@mayank99 @veekeys What do you think about this kind of an approach?

mayank99:
Looks good I think. Only thing I don't like is that cells need individual disabled state. Could we default it to the value from the row? And then user could override it with `isDisabled={false}` for example.

veekeys:
Looks good. WIth this solution now we can deliver slot column, which hopefully will help in column manager pr

bentleyvk:
> Looks good I think. Only thing I don't like is that cells need individual disabled state. Could we default it to the value from the row? And then user could override it with `isDisabled={false}` for example.

We are using `isRowDisabled` in selection handler, e.g. when you press select all, it will not select disabled row. And if user has custom column, I have no idea whether checkbox will be clickable or not because it is controlled by the user.

stories/core/Table.stories.tsx
```diff
@@ -2005,7 +2007,7 @@ export const HorizontalScroll: Story<Partial<TableProps>> = (args) => {
   );
 
   const columns = useMemo(
-    (): Column[] => [
+    (): Column<typeof data[number]>[] => [
```
bentleyvk, line 2010: Linter was showing an error here.

src/core/Table/cells/DefaultCell.tsx
```diff
@@ -4,10 +4,12 @@
...
 import React from 'react';
 import { CellRendererProps } from 'react-table';
+import cx from 'classnames';
 
-export type DefaultCellProps<
-  T extends Record<string, unknown>
-> = CellRendererProps<T> & React.ComponentPropsWithoutRef<'div'>;
+export type DefaultCellProps<T extends Record<string, unknown>> = {
+  disabled?: (rowData: T) => boolean;
+} & CellRendererProps<T> &
```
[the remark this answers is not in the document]
bentleyvk: If `isRowDisabled` is passed, it is being used in selection action handler, which makes active checkbox not selectable. So how can I reuse `isRowDisabled` if I can't pass it to the Table?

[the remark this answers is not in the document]
mayank99: > If isRowDisabled is passed, it is being used in selection action handler, which makes active checkbox not selectable.

Can this behavior be changed if user has custom columns? Would be nice to utilize the `isRowDisabled` prop instead of just saying "this prop should not be used if you have custom columns".

src/core/Table/hooks/useSelectionCell.tsx
```diff
@@ -16,53 +13,9 @@ export const useSelectionCell = <T extends Record<string, unknown>>(
...
-          {...row.getToggleRowSelectedProps()}
-          disabled={isRowDisabled?.(row.original)}
-          onClick={(e) => e.stopPropagation()} // Prevents triggering on row click
-        />
-      ),
-    },
-    ...columns,
-  ]);
-
-  hooks.useInstanceBeforeDimensions.push(({ headerGroups }) => {
```
gretanausedaite, line 63: We don't need this fix anymore?

bentleyvk: I really hope not 😀 Because it was working ok when tested

src/core/Table/cells/DefaultCell.tsx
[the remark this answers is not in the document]
bentleyvk: Taking state from the row now.

bentleyvk:
Need to wait for https://github.com/iTwin/iTwinUI-react/pull/622 to be merged

src/core/Table/columns/selectionColumn.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { CellProps, CellRendererProps, HeaderProps } from 'react-table';
+import { Checkbox } from '../../Checkbox';
+import { DefaultCell } from '../cells';
+
+export const SELECTION_CELL_ID = 'iui-table-checkbox-selector';
+
+export const SelectionColumn = <T extends Record<string, unknown>>(
```
mayank99, line 12: Maybe add a JSDoc comment above this because it is exposed to the user.

Same thing in ExpanderColumn and ActionColumn.

mayank99 (approved):
Looks good now.

Documentation could be added/improved.

src/core/Table/columns/selectionColumn.tsx
bentleyvk: Something like this? https://github.com/iTwin/iTwinUI-react/pull/599/commits/aa3f0d97909336142ecc5920e2b4fb97f6639d07

gretanausedaite (approved):
Looks good. Agree with more documentation (eg. hooks)

src/core/Table/columns/selectionColumn.tsx
mayank99: Yeah that's better. I think `SelectionColumn` and `ExpanderColumn` could also use `@example` because they accept parameters.

Also, you're going to hate me for saying this but, now I'm realizing that we could probably align the parameters so that these columns accept an object (similar to `ActionColumn`),

e.g. usage could go from this:
```ts
SelectionColumn(() => true)
```

to this:
```ts
SelectionColumn({ isDisabled: () => true })
```

bentleyvk: Good that you raised this question as I forgot to ask everyone's opinion whether to have regular arguments or props object. In order to align with ActionColumn, I changed to props object.

mayank99 (approved):

PR #601, opened 2022-03-24
fix(Table): Date filter makes range from the beginning to the end of the day

### LINKED ISSUE #600: DateRangeFilter for table working strange.
The date range filter not working as expected.
#### Expected behavior
The filter should show all the records From - To date

#### Reproduction
https://codesandbox.io/s/datefilterexample-wvh38u?file=/src/App.js

##### Steps to reproduce
1. filter from 3/5/2022 to 3/5/2022 shows empty records.
2. filter from 3/4/2022 to 3/5/2022 shows records for 3/4/2022 and 3/5/2022 but not all. I have added 3 records for 3/4/2022 and 6 for 3/5/2022. 
image
3. Filter only from 3/5/2022 not return all records.
image
4. same For only To

bentleyvk:
The problem is that `DatePicker` returns selected day with the current time, e.g. if current time is 17:00 and I set "From" filter to "March 24, 2022" and there is a row that has date with 'March 24, 2022, 12:00`, this row wouldn't be shown.
So my change makes date range from 00:00 to 23:59.

Closes #600

## Checklist

### DIFF
--- a/src/core/Table/filters/DateRangeFilter/DateRangeFilter.test.tsx
+++ b/src/core/Table/filters/DateRangeFilter/DateRangeFilter.test.tsx
@@ -70,8 +70,8 @@ it('should set filter when both values entered', () => {
   screen.getByText('Filter').click();
 
   expect(setFilter).toHaveBeenCalledWith([
-    new Date(2021, 4, 1),
-    new Date(2021, 4, 3),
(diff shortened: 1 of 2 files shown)


veekeys (approved):

elephantcatdog (approved):
Your tests don't cover lines 75 & 95 (`return date;`) but otherwise seems fine to me.

PR #604, opened 2022-03-28
feat(DatePicker): Added year selection

veekeys:
Added buttons for year selection.

image


gretanausedaite (approved):
Looks good. Should we add indicators to show which chevron icon button does what? Right now it's not clear what these buttons do.

src/core/DatePicker/DatePicker.tsx
```diff
@@ -129,6 +131,11 @@ export type DatePickerProps = {
    * @default false
    */
   showTime?: boolean;
+  /**
+   * Show additional buttons to select year.
+   * @default false
+   */
+  showYear?: boolean;
```
bentleyvk, line 138: I would say to rename maybe to `showYearNavigation` because `showYear` for me sounds like if it is not enabled, you won't see what year it is 😀
Or maybe we should always show it.

veekeys: mb then `showYearSelection` better?

veekeys: renamed to `showYearSelection`

gretanausedaite (approved):

PR #605, opened 2022-03-28
chore(deps): update itwinui-css to 0.53.0

### LINKED ISSUE #602: ButtonGroup: Overflow is not growing back when there is enough space
Related changes: https://github.com/iTwin/iTwinUI/pull/514 and https://github.com/iTwin/iTwinUI-react/pull/579/
buttongroup-overflow-bug

veekeys:
- updated `DatePicker` classes
- added class to `ButtonGroup` overflow
- Images change from theme color changes and user icon


veekeys:
@FlyersPh9 
While updating images, 2 things I noticed:
- we increased font size of sub label in header, but did not in menu item. Inconsistency not a problem?
- With increased font size of sub label in header, we risk having buttons wider (decsription can be longer than name). Should we think of setting certain characters numbers in there (as we do in date picker months) and show overflow? Or you think its not really our problem and user should handle buttons width and overflow there...?

mayank99 (approved):

adamhock (approved):

PR #607, opened 2022-03-28
fix(Table): `onBottomReached` was triggered on filtered Table

### LINKED ISSUE #523: Table: onBottomReached isn't getting called when the last data row is filtered out
onBottomReached is only called when the row that displays the last data entry appears in the viewport, which never happens if the last row is filtered out. (The last row is being determined by its index being equal to unfiltered data's length https://github.com/iTwin/iTwinUI-react/blob/449dc4229c15b040e62395a4ab49592217e3bf13/src/core/Table/Table.tsx#L658, which doesn't work when filters are applied)

#### Expected behavior
onBottomReached should be called when the bottom of the table is reached, even when filters are applied.

bentleyvk:
Previously it was triggering `onBottomReached` on wrong row because it was setting row as a last if its index is the same as last `data` index which is wrong with filtered data.
Also adjusted filter and sort icon showing logic because previously it was blinking with lazy-loading:
table-blinking-filter-sort

Closes #523

## Checklist


src/core/Table/Table.tsx
```diff
@@ -558,13 +558,13 @@ export const Table = <
...
   const getPreparedRow = React.useCallback(
-    (row: Row<T>) => {
+    (row: Row<T>, index: number) => {
       prepareRow(row);
       return (
         <TableRowMemoized
           row={row}
           rowProps={rowProps}
-          isLast={row.index === data.length - 1}
+          isLast={index === page.length - 1}
```
adamhock, line 567: I was looking at solutions for this yesterday, and I just changed `data.length` to `instance.rows.length` to solve the issue

bentleyvk: I am afraid that `instance.rows` might be not in sync with `page` and we are using `page` to construct Table body rows.

src/core/Table/Table.tsx
```diff
@@ -595,7 +595,7 @@ export const Table = <
   );
 
   const virtualizedItemRenderer = React.useCallback(
-    (index: number) => getPreparedRow(page[index]),
+    (index: number) => getPreparedRow(page[index], index),
```
veekeys, line 598: Looks weird, when you pass instance and its index.. Could just pass index then and get the instance in `getPreparedRow`, but I guess minor and less changes :D

src/core/Table/Table.test.tsx
```diff
@@ -406,13 +406,23 @@ it('should not show sorting icon if sorting is disabled', () => {
 
 it('should not show sort icon if data is loading', () => {
```
veekeys, line 407: Amm, test name now is wrong. Maybe say `if there is no data` isntead

src/core/Table/Table.test.tsx
```diff
@@ -704,6 +749,33 @@ it('should not show filter icon when filter component is not set', () => {
...
+  ];
+  const { container } = renderComponent({
+    columns: mockedColumns,
+    isLoading: true,
+  });
+
+  setFilter(container, '2');
+
+  const filterIcon = container.querySelector(
+    '.iui-filter-button .iui-button-icon',
```
veekeys, line 774: shouldnt query `iui-active` ?

stories/core/Table.stories.tsx
```diff
@@ -861,13 +862,21 @@ export const LazyLoading: Story<Partial<TableProps>> = (args) => {
       emptyTableContent='No data.'
       onBottomReached={onBottomReached}
       isLoading={isLoading}
+      isSortable
       {...args}
       style={{ height: 440, maxHeight: '90vh' }}
       data={data}
+      // Prevents from resetting filters and sorting when more data is loaded
+      autoResetFilters={false}
```
veekeys, line 870: oh wow, interesting :O Wondering, if users were aware of these flags..

src/core/Table/Table.test.tsx
bentleyvk: No, it is not 😀 `data` is required field therefore passing empty array and `isLoading: true`

src/core/Table/Table.test.tsx
bentleyvk: Good catch! Added

src/core/Table/Table.tsx
bentleyvk: Changed

stories/core/Table.stories.tsx
bentleyvk: We have a link to `react-table` documentation in Table description. But I doubt that a lot of users go there.

veekeys (approved):

stories/core/Table.stories.tsx
veekeys: Personally, I havent noticed these flags before.
Shouldnt be this a default value?

veekeys: Or maybe we could start adding these things to some wiki or doc.. 
Looks like a often used case and having these tips might be helpful :/

bentleyvk: I hope that our documentation site will have these examples.

PR #612, opened 2022-04-01
chore: Fixed stories code

veekeys:
Removed `bind` from all stores but typography, which seemed not to be broken.


mayank99 (approved):
Love storybook

PR #618, opened 2022-04-06
feat(SkipToContent): Add SkipToContent Component

### LINKED ISSUE #568: SkipNavLink react component

elephantcatdog:
image
Added new `SkipToContent` component in React.
- Used similar method to Mayank's suggestion [SkipNavContent in ReachUI](https://reach.tech/skip-nav/#skipnavcontent).
- There are two pieces of the whole: the stylized `SkipToContentLink` and invisible `SkipToContentMarker`
  - `SkipToContentLink` is the link/component that appears when tabbed to, as seen in the image above. Its `href` directs to the `SkipToContentMarker`. The `href` is not editable by users.
  - `SkipToContentMarker` is an invisible div whose purpose is to provide a 'marker' for where the main content is. The main content should follow the marker or be a child of it. The `id` is not editable by users, as that is how the focus is directed from the `SkipToContentLink`.
- Added Jest tests.
  - Initially tried to test that the focus moved from the link to the marker, but realized/read that I don't need to test that part as it's handled by the ...DOM (? or something...).
- Added two stories: one for regular usage of the link and marker, and one that demonstates using both with children.
  - The children story is named 'WithChildren' and there might be a better name for that...
  - I got kinda close but couldn't figure out the issue with storybook opening a new tab, so I just provided a disclaimer in the stories.
- The two visual test images are see-through currently but [I have a fix out for that in CSS](https://github.com/iTwin/iTwinUI/pull/572).

Closes #568

## Checklist


src/core/SkipToContent/SkipToContentMarker.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Marks the start of the main content of the page.
+ * Used in conjunction with `SkipToContentLink`.
+ * @example <caption>`SkipToContentMarker` is placed before the main content</caption>
+ * <SkipToContentMarker /> <MainContentMarker />
+ * @example <caption>The main content is placed as a child of the `SkipToContentMarker` element</caption>
+ * <SkipToContentMarker> <MainContent /> </SkipToContentMarker>
+ */
+export const SkipToContentMarker = (props: SkipToContentMarkerProps) => {
```
mayank99, line 23: Looks like all this component does is attach an id, and it's not very flexible. So maybe it should be a hook that just gives you an id.

Also the id should probably not be hardcoded because someone might want multiple of these markers.

Usage could be something like this:

```tsx
const contentId = useSkipToContentMarker();
...
<SkipToContentLink href={`#${contentId}`}>Skip to content</SkipToContentLink>
```

But it's less useful now that React 18 offers a `useId` hook.

src/core/SkipToContent/SkipToContentLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { CommonProps, useTheme } from '../utils';
+import '@itwin/itwinui-css/css/skip-to-content.css';
+
+export type SkipToContentLinkProps = {
+  /**
+   * Localize 'Skip to main content' label if needed.
+   * @default 'Skip to main content'
+   */
+  children?: React.ReactNode;
+} & Omit<CommonProps, 'href'>;
```
mayank99, line 16: This should support all anchor props, including `href`, to provide the most flexibility to user.

```suggestion
& React.ComponentPropsWithoutRef<'a'>;
```

stories/core/SkipToContentLink.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+          await this.expect({ opened }).to.matchImages();
+        },
+      },
+    } as CreeveyStoryParams,
+  },
+} as Meta<SkipToContentLinkProps>;
+
+export const Basic: Story<SkipToContentLinkProps> = () => {
+  React.useEffect(() => {
+    window.frameElement?.setAttribute('Name', 'storybook-iframe');
```
mayank99, line 46: Maybe add a comment explaining what it does / why it's needed

stories/core/SkipToContentLink.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+  }, []);
+
+  return (
+    <>
+      <SkipToContentLink />
+      <Header
```
mayank99, line 52: I don't think we should have 50 lines of code for header. Maybe you can create a simpler version that doesn't need all this code.

I was going to mention this in the first CSS PR but it's _much_ more important here to have readable code in storybook.

src/core/SkipToContent/SkipToContent.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    </SkipToContentMarker>,
+  );
+
+  const marker = container.querySelector('#skip-to-content-marker');
+  expect(marker).toBeTruthy();
+  expect(marker?.children.length === 1);
+  expect(marker?.querySelector('#main')).toBeTruthy();
+});
+
+it('should propagate misc props in link (except href)', () => {
```
mayank99, line 52: This doesn't really test the `except href` part so the name looks wrong. But as I mentioned in my other comment, we should allow overriding href.

src/core/SkipToContent/SkipToContent.test.tsx
```diff
@@ -0,0 +1,85 @@
+/*---------------------------------------------------------------------------------------------
```
mayank99, line 1: Maybe rename this file (and folder) to `SkipToContentLink`. It doesn't make sense to have different names.

src/core/SkipToContent/SkipToContentLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Link is for screen reader and keyboard users and will not be visible unless tabbed to.
+ * Provides a shortcut to the main content of the page without navigating through the header, etc.
+ * Should be placed just inside the body. Used in conjunction with `SkipToContentMarker`.
+ * @example
+ * <body><SkipToContentLink /> ... </body>
+ * <body><SkipToContentLink>{localizedLabel}</SkipToContentLink> ... </body>
+ */
+export const SkipToContentLink = (props: SkipToContentLinkProps) => {
```
mayank99, line 26: I think we might want to forward this component's `ref`

src/core/SkipToContent/SkipToContentLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { children = 'Skip to main content', className, ...rest } = props;
```
mayank99, line 27: While it's nice that the user doesn't have to specify children, I am not so sure.

Do we really want to set default text here? The "Skip to" part is fine, but the "main content" part will probably depend on the app's context, e.g. "Skip to search results", "Skip to blog post" or even something like "Skip iframe" 

Another concern is that the user might forget to localize this.

src/core/SkipToContent/SkipToContentLink.tsx
```diff
@@ -0,0 +1,42 @@
...
+export type SkipToContentLinkProps = {
+  /**
+   * Localize 'Skip to main content' label if needed.
+   * @default 'Skip to main content'
+   */
+  children?: React.ReactNode;
+} & Omit<CommonProps, 'href'>;
+
+/**
+ * Link is for screen reader and keyboard users and will not be visible unless tabbed to.
```
mayank99, line 19: Typo?
```suggestion
 * SkipToContentLink is for screen reader and keyboard users and will not be visible unless tabbed to.
```

stories/core/SkipToContentLink.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+export default {
+  component: SkipToContentLink,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+  },
+  title: 'Core/SkipToContentLink',
+  parameters: {
+    creevey: {
+      captureElement: null,
```
adamhock, line 31: It seems that all of the content is shown even when this parameter is removed

src/core/index.ts
```diff
@@ -277,3 +277,9 @@ export {
   MiddleTextTruncation,
 } from './utils';
 export type { ThemeType, MiddleTextTruncationProps } from './utils';
+
+export { SkipToContentLink, SkipToContentMarker } from './SkipToContent';
```
adamhock, line 281: This should be in alphabetical order see [here](https://github.com/iTwin/iTwinUI-react/blob/main/CONTRIBUTING.md?plain=1#L98)

src/core/SkipToContent/SkipToContentMarker.tsx
elephantcatdog: I based this off of the example you suggested. I was going to do something more like this, but decided to do two components instead because our React stuff is component based and I didn't want to require users to fiddle with the ids on their components. I don't think adding an id to their components works as you suggest here - what if the component already has an id? I am also not sure how having multiple markers would work. What kind of situation would call for that?

src/core/SkipToContent/SkipToContentLink.tsx
elephantcatdog: Done

stories/core/SkipToContentLink.stories.tsx
elephantcatdog: Not needed; artifact from previous work.

stories/core/SkipToContentLink.stories.tsx
elephantcatdog: Done

src/core/SkipToContent/SkipToContent.test.tsx
elephantcatdog: Fixed

src/core/SkipToContent/SkipToContent.test.tsx
elephantcatdog: They're different names because `SkipToContent` is currently made up of both `SkipToContentLink` and `SkipToContentMarker`.

src/core/SkipToContent/SkipToContentLink.tsx
elephantcatdog: Sure, done

src/core/SkipToContent/SkipToContentLink.tsx
elephantcatdog: Default text is set in other components, such as `IdeasButton`, `Footer`, and `DatePicker`. Also, this phrase "Skip to content" or "Skip to main content" is standard for this feature and is found on large sites (Google, Amazon, LinkedIn, etc.) without specialized text.

src/core/SkipToContent/SkipToContentLink.tsx
elephantcatdog: Fixed

stories/core/SkipToContentLink.stories.tsx
elephantcatdog: Removed

src/core/index.ts
elephantcatdog: Fixed

src/core/SkipToContent/SkipToContentLink.tsx
mayank99: > this phrase "Skip to content" or "Skip to main content" is standard for this feature 

Your own example provides two variations, which means it's not really standard.

Also don't discard my concern about multiple links which need different text.

```html
<body>
  <SkipToContentLink href="#main">Skip to content</SkipToContentLink>
  <header>Main page header with some links to other parts of site...</header>

  <main id="main">
    <nav style="position: relative">
      <SkipToContentLink href="#table-skip" style="position: absolute">Skip table of contents</SkipToContentLink>
      <ul>Table of contents for this page...</ul> 
    </nav>

    <p id="table-skip">Content before the iframe…</p>

    <div style="position: relative">
      <SkipToContentLink href="#iframe-skip" style="position: absolute">Skip iframe content<SkipToContentLink>
      <iframe src="embed-url-from-other-site"></iframe>
    </div>

    <p id="iframe-skip">Content after the iframe…</p>
  </main>
</body>
```

src/core/SkipToContent/SkipToContentMarker.tsx
mayank99: > I based this off of the example you suggested.

That example is not gospel, it was just the top one I found in my google search when creating that issue. You could do more research into other libraries to see how they do this.

> our React stuff is component based

This is no reason to limit any future implementations. If hooks can provide us more flexibility, then we shouldn't avoid it because we haven't done anything like it in the past.

> I don't think adding an id to their components works as you suggest here - what if the component already has an id?

Not sure what's the concern here. They are the ones adding the id (we are simply giving them the id). If there are multiple, then the last one will be used.

> I am also not sure how having multiple markers would work. What kind of situation would call for that?

Multiple markers also means multiple links. The ids need to be unique. I have a few examples in my other comment: https://github.com/iTwin/iTwinUI-react/pull/618#discussion_r844411643

src/core/SkipToContent/SkipToContentLink.tsx
mayank99: Ok, fair. If you are okay with having this inconsistent usage (`<SkipToContent />` vs `<SkipToContent>Skip table of contents</SkipToContent>` in the same page), then it's fine.

I will leave this comment open for visibility.

src/core/SkipToContent/SkipToContentMarker.tsx
mayank99: > Elements are supposed to only have one id.

That's true for HTML, but we are working in JSX, where props are just JS objects. Please see above where I said "the last one is used".

I'm still not sure what your concern here is. Nowhere am I suggesting having multiple ids on the same element. You can see a clearer example of what I'm suggesting in my other comment: https://github.com/iTwin/iTwinUI-react/pull/618#discussion_r849469581

mayank99: If you really want to keep this component, then it will need to be more flexible:
- it should allow overriding the id (for when there are multiple links/markers)
- it should be polymorphic to allow changing what element is rendered (`<div>` is not always desirable)
- it might need `tabindex=-1` for older browsers and for receiving focus if we progressively enhance the functionality using JS

src/core/SkipToContent/SkipToContent.test.tsx
elephantcatdog: Renamed

src/core/SkipToContent/SkipToContentMarker.tsx
elephantcatdog: Removed the marker. Didn't add a hook because it seems unnecessary. Added a prop so that users are required to make it work.

mayank99: Why not just make `href` required?

elephantcatdog: Didn't know you could do that

adamhock (approved):
Nice Job 🎉

src/core/SkipToContent/SkipToContentMarker.tsx
elephantcatdog: Ok, I did that I think. I mean, it requires href now but this might not have been the method you were thinking of? The internet wasn't super helpful.

src/core/SkipToContentLink/SkipToContentLink.tsx
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * The id of the main content that the link directs to. Don't forget the #!
+   */
+  href: string;
+  /**
+   * Localize 'Skip to main content' label.
+   * @default 'Skip to main content'
+   */
+  children?: React.ReactNode;
+} & React.ComponentPropsWithoutRef<'a'>;
```
mayank99, line 20: `href` should be omitted from the `a` props now that it's being manually included

elephantcatdog: FIxed

mayank99 (approved):

PR #619, opened 2022-04-07
fix(Popover): Stop propagation of click in Popover

veekeys:
We have an issue, that elements created with Popover (`DropdownMenu` for example) are triggering events on the elements behind (imagine menu opened over the table row, it gets selected when you choose an item).
This is weird behavior, because we append Popover to body, so DOM structure should not have that.
But apparently, it is behavior of React Portal.

https://reactjs.org/docs/portals.html#event-bubbling-through-portals

Instead of fixing only menu, I decided to `stopPropagation` in Popover instead. Could not think of a situation, that you would need click event to go out.
Added only click, but I assume we might want to add all other?? (mouse, pointer down...)

Could not write a working jest test, cause it seems does not recognize element being on top of other.


elephantcatdog (commented):
Apologies if you tried this and I didn't understand your comment about testing, but could you write a test that replicates the problem without the code fix and then check that doesn't happen? Like, open a dropdown menu over another element, simulate a click, then check that the other element isn't focused when it would've been previously?

elephantcatdog:
Is this the kind of situation that this PR solves? I didn't have luck with reproducing it.
image

veekeys:
> Is this the kind of situation that this PR solves? I didn't have luck with reproducing it. image

Its hard to understand from image, but hopefully these will explain better:
Before this fix
table-bug
After
table-bug-fix

Notice that table row was being selected while clicking menu item. Thats what I meant by element behind was getting the click.




> Apologies if you tried this and I didn't understand your comment about testing, but could you write a test that replicates the problem without the code fix and then check that doesn't happen? Like, open a dropdown menu over another element, simulate a click, then check that the other element isn't focused when it would've been previously?

Thats what I was trying. Just it passes even without my fix, so must be again jest thing, I guess.

elephantcatdog:
> Notice that table row was being selected while clicking menu item. Thats what I meant by element behind was getting the click.

Ah, I see now, thank you

bentleyvk (approved):

PR #620, opened 2022-04-08
chore: Automate issues and prs addition to project

veekeys:
Github released this new action
https://github.com/actions/add-to-project

The concern I have is the token, not sure, if its gonna work.

UPDATE: ok, I am sure it wont work, cause its limited per repo.. We need token with the access to the org. 

LAST UPDATE: for now, I have linked my personal access token. I have tested with this PR and it successfully added it to project! I assume it would work with issues too.
For future, we might want to replace my token with some general org one, so thats gonna happen just maybe not that fast.

### DIFF
--- a/.github/workflows/add-issues-and-prs-to-project.yml
+++ b/.github/workflows/add-issues-and-prs-to-project.yml
@@ -1,4 +1,4 @@
-name: Add issues and pull requests to project
+name: Project
 
 on:
   issues:
@@ -10,10 +10,11 @@ on:
 
 jobs:
   add-to-project:
+    name: Add to project
     runs-on: ubuntu-latest
     steps:
       - uses: actions/add-to-project@main
         with:
-          project-url: https://github.com/orgs/iTwin/projects/3/views/1
+          project-url: https://github.com/orgs/iTwin/projects/3
           github-token: ${{ secrets.PROJECT_BOARD_TOKEN }}
           
--- a/.github/workflows/add-issues-and-prs-to-project.yml
+++ b/.github/workflows/add-issues-and-prs-to-project.yml
@@ -7,7 +7,6 @@ on:
   pull_request:
     types:
       - opened
-      - labeled
 
 jobs:
   add-to-project:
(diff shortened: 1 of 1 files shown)


mayank99:
Maybe you can create a PAT?
https://github.com/actions/add-to-project#creating-a-pat-and-adding-it-to-your-repository

veekeys:
> Maybe you can create a PAT? [actions/add-to-project#creating-a-pat-and-adding-it-to-your-repository](https://github.com/actions/add-to-project#creating-a-pat-and-adding-it-to-your-repository)

I can, but not with org scope :)

mayank99:
> I can, but not with org scope :)

Are you sure?
image

.github/workflows/add-issues-and-prs-to-project.yml
```diff
@@ -0,0 +1,19 @@
+name: Add issues and pull requests to project
```
mayank99, line 1: Looks weird right now
image


Maybe make this name shorter and add a name to the job.

Other ones look cleaner for example
image

.github/workflows/add-issues-and-prs-to-project.yml
```diff
@@ -0,0 +1,19 @@
...
+    types:
+      - opened
+
+jobs:
+  add-to-project:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@main
+        with:
+          project-url: https://github.com/orgs/iTwin/projects/3/views/1
```
mayank99, line 17: I think there's no need for views here
```suggestion
          project-url: https://github.com/orgs/iTwin/projects/3
```

.github/workflows/add-issues-and-prs-to-project.yml
veekeys: added

.github/workflows/add-issues-and-prs-to-project.yml
veekeys: good catch, changed

mayank99 (approved):

PR #623, opened 2022-04-11
refactor(ComboBox): Use compound components and context

mayank99:
Rewrote the entire ComboBox by splitting into compound components and using reducer+context. This provides extra flexibility (e.g. composition) and improves performance (e.g. click lag in virtualization).

One of the important changes I made for performance is save the original index from the passed options into an object whose keys are derived from the option label. This index is also stored as `data-iui-index` in the DOM element. This avoids the need for repeatedly calling `find`/`findIndex`.

## Checklist


src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -319,26 +339,28 @@ it('should work with custom itemRenderer', () => {
...
       '.iui-menu-item.iui-active.iui-focused.my-custom-item',
     ),
   ).toHaveTextContent('CUSTOM Item 1');
 
-  fireEvent.keyDown(input, { key: 'ArrowDown' });
+  userEvent.keyboard('{ArrowDown}');
+
   expect(
     document.querySelector('.iui-menu-item.iui-focused.my-custom-item'),
   ).toHaveTextContent('CUSTOM Item 2');
```
mayank99, line 361: I don't understand what is up with these keyboard navigation tests, I can't get them to pass.

mayank99:
@veekeys I think this PR is ready to be hooked up to virtualization. I tried to test it with the current (released) version of `VirtualScroll`, and the performance is great! 👍 

There are two non-performance issues:
- some size miscalculation (which #539 should fix I think)
- in my current keyboard navigation, when pressing ⬇️  on the last item (or ⬆️  on the first), it should loop over. This is not compatible with virtualization so we could try to fix it, or remove this behavior.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +122,312 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      const customItem = itemRenderer
+        ? itemRenderer(option, {
+            isFocused: focusedIndex === __originalIndex,
+            isSelected: selectedIndex === __originalIndex,
+            index: __originalIndex,
+            id: option.id,
+          })
+        : null;
+
+      return selectedIndex === __originalIndex ||
```
bentleyvk, line 279: Can you change to normal `if` condition. For me it is hard to read several ternary conditionals in one place.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +122,312 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  ((x: [ComboBoxAction] | [ComboBoxAction, number]) => void) | undefined
+>(undefined);
+ComboBoxActionContext.displayName = 'ComboBoxActionContext';
+
+const useSafeContext = <T,>(context: React.Context<T>) => {
+  const value = React.useContext(context);
+  if (!value) {
+    throw new Error(`${context.displayName} is undefined`);
+  }
+  return value!; // eslint-disable-line @typescript-eslint/no-non-null-assertion -- we already checked for undefined
```
bentleyvk, line 420: Hm, I am not getting this linter error

bentleyvk: Oh getting error later. Strange...

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +122,312 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+};
+
+const ComboBoxInputContainer = (
```
bentleyvk, line 423: What if you merge it with Input? Is it needed to be as a separate component?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +122,312 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+              customItem.props.onClick?.(e);
+            },
+          })
+        ) : (
+          <ComboBoxMenuItem
+            key={option.id ?? __originalIndex}
+            {...rest}
+            isSelected={selectedIndex === __originalIndex}
+            isFocused={focusedIndex === __originalIndex}
+            onClick={() => dispatch(['select', __originalIndex])}
```
bentleyvk, line 296: Click is still laggy.

src/core/ComboBox/ComboBox.tsx
mayank99: That's how I had it before but this is more flexible. `ComboBox.Input` contains the `<input>` element and all the logic, so for customizing the styling (e.g. icon on the left) you could replace `ComboBox.InputContainer` with something else or even remove it.

src/core/ComboBox/ComboBox.tsx
mayank99: Doesn't lag when virtualized. I'm not sure how to fix it in regular mode when there's a large number of items. In my `dispatch(['select'])` I'm setting `isOpen: false` but that triggers a whole bunch of updates before the popover actually closes.

mayank99:
So there is an issue with `Disabled Items` story not loading in dev storybook. I ran prod version (`yarn build-storybook`) and also tried in a test app and it loads fine there. That's where I also noticed that keyboard navigation for disabled items seems broken, so I will need to fix that.

I'm thinking of reworking keyboard navigation because it also doesn't work in custom `Async` story and in tests.

src/core/ComboBox/ComboBox.tsx
mayank99: Removed nested ternary, take a look

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   const inputRef = React.useRef<HTMLInputElement>(null);
   const menuRef = React.useRef<HTMLUListElement>(null);
   const toggleButtonRef = React.useRef<HTMLSpanElement>(null);
-  const [isOpen, setIsOpen] = React.useState(false);
+  const optionsRef = React.useRef<Array<HTMLLIElement>>([]);
+
+  // Latest value of the onChange prop
+  const onChangeProp = React.useRef(onChange);
+  React.useEffect(() => {
+    onChangeProp.current = onChange;
```
veekeys, line 146: whats the reason to save it in ref, instead of using directly?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  }, [onChange]);
+
+  // Reducer where all the component-wide state is stored
+  const [{ isOpen, selectedIndex, focusedIndex }, dispatch] = React.useReducer(
+    comboBoxReducer,
+    {
+      isOpen: false,
+      selectedIndex: -1,
+      focusedIndex: -1,
```
veekeys, line 155: ammm... cant you save this info instead of looking for index everytime?
Wont now happen `select` event twice? Once from us (lets say keybaord nav) and then once from user (because we will call `onChange` callback and user would change `valueProp` then)

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      inputRef.current?.focus();
+    } else {
+      // Reset the focused index when the dropdown closes
+      dispatch(['focus']);
+    }
+  }, [isOpen]);
 
   // Set min-width of menu to be same as input
-  const [minWidth, setMinWidth] = React.useState(0);
+  const [minWidth, setMinWidth] = React.useState(300);
```
veekeys, line 170: remember why needed initial value?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
 
-  // Filter options and update focus when input value changes
+  // When the value prop changes, update the selectedIndex
   React.useEffect(() => {
-    if (!isOpen) {
-      return;
-    }
+    dispatch([
+      'select',
+      options.findIndex((option) => option.value === valueProp),
```
veekeys, line 210: yeah, so if you saved initially index info, wouldnt need to look for it here

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            isFocused: focusedIndex === __originalIndex,
+            isSelected: selectedIndex === __originalIndex,
+            index: __originalIndex,
+            id: option.id,
+          })
+        : null;
+
+      return customItem ? (
+        React.cloneElement(customItem, {
+          onClick: (e: unknown) => {
```
veekeys, line 290: It mus be `li` element though, no?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            dispatch(['select', __originalIndex as number]);
+            customItem.props.onClick?.(e);
+          },
+          // ComboBox.MenuItem handles data-iui-index and iui-focused through context
+          // but we still need to pass them here for backwards compatibility with MenuItem
+          className: cx(customItem.props.className, 'iui-focused'),
```
veekeys, line 296: you need to add `iui-focused` to all elements? doesnt make sense

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,317 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   );
 
-  // update inputValue and focusedIndex every time selected value changes
+  // Reset focused item when filter changes
   React.useEffect(() => {
-    const selectedOption = options.find(({ value }) => value === selectedValue);
-    setInputValue(selectedOption?.label ?? '');
-    setFocusedIndex(selectedOption ? options.indexOf(selectedOption) : -1);
-  }, [selectedValue, options]);
+    dispatch(['focus']);
```
bentleyvk, line 203: Can you move this call to `handleInput`?

veekeys (commented):
Submitting code review. It is pretty complex structure, Im still left in thinking, what and how could be improved :D
But its good as testing with VS went pretty successful.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        {children ?? <SvgCaretDownSmall aria-hidden />}
+      </span>
+    );
+  },
+);
+ComboBoxEndIcon.displayName = 'ComboBoxEndIcon';
+
+const ComboBoxInput = React.forwardRef(
+  (props: InputProps, forwardedRef: React.RefObject<HTMLInputElement>) => {
+    const { onKeyDown: onKeyDownProp, onFocus: onFocusProp, ...rest } = props;
```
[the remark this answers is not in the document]
mayank99: this is just to make it clear because i have my own handlers for keydown and focus in this component

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-          className={cx('iui-end-icon', {
-            'iui-actionable': !inputProps?.disabled,
-            'iui-disabled': inputProps?.disabled,
-            'iui-open': isOpen,
-          })}
-          onClick={() => {
+
+            while (true) {
+              const currentElement =
+                menuRef.current?.querySelector(
```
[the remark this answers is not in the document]
mayank99: this is outdated, please look at https://github.com/iTwin/iTwinUI-react/pull/623/commits/69056c9770d96b6bdc0233ded4ce3fca26e384da

but it's still querying dom. can you elaborate on the map part and how it would work with custom items?

src/core/ComboBox/ComboBox.tsx
mayank99: i have an early return above
image

src/core/ComboBox/ComboBox.tsx
mayank99: it should be `li` but it can be anything. we dont control it

src/core/ComboBox/ComboBox.tsx
mayank99: i think this part is the same as before, it's to avoid requiring user to memoize onChange

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+              if (
+                prevElement?.ariaDisabled === 'true' ||
+                prevIndex == undefined
+              ) {
+                continue;
+              }
+
+              return dispatch(['focus', Number(prevIndex)]);
+            }
+          } else if (event.key === 'Enter') {
```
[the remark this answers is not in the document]
mayank99: space is a typeable character in input

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        {...rest}
+      />
+    );
+  },
+);
+ComboBoxInput.displayName = 'ComboBoxInput';
+
+const ComboBoxPopover = React.forwardRef(
+  (
+    props: PopoverProps & { children: JSX.Element },
```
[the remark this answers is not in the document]
mayank99: what for? For me it's nicer if we use `children` for composing those subcomponents.

src/core/ComboBox/ComboBox.tsx
mayank99: This is leftover from testing, reverted.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      );
+    };
+
+    const handleKeyDown = React.useCallback(
+      (event: React.KeyboardEvent<HTMLInputElement>) => {
+        (() => {
+          const length =
+            menuRef.current?.querySelectorAll('[data-iui-index]').length ?? 0;
+
+          if (event.key === 'ArrowDown') {
```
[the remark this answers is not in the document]
mayank99: no reason. using `switch` now in latest version

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +446,309 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    const { className, children, ...rest } = props;
+    const dispatch = useSafeContext(ComboBoxActionContext);
+    const { toggleButtonRef } = useSafeContext(ComboBoxRefsContext);
+    const refs = useMergedRefs(toggleButtonRef, forwardedRef);
+
+    return (
+      <span
+        ref={refs}
+        className={cx('iui-end-icon', className)}
+        onClick={() => dispatch(['toggle'])}
```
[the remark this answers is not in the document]
mayank99: done

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,252 +123,312 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-          value: option.value,
-          role: 'option',
-          onClick: () => {
-            setSelectedValue(option.value);
-            userOnChange.current?.(option.value);
-            setIsOpen(false);
-          },
-        });
-      }
+const ComboBoxRefsContext = React.createContext<
```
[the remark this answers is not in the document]
mayank99: added displayName

src/core/ComboBox/ComboBox.tsx
mayank99: Why? `filteredOptions` could also change if `optionsProp` changes for some reason

src/core/ComboBox/ComboBox.tsx
mayank99: This comment might be in the wrong place, but even if it wasn't, I don't think I understand what you mean by "save this info". Maybe elaborate how I can avoid the lookup?

src/core/ComboBox/ComboBox.tsx
veekeys: but wouldnt just `React.MouseEvent<HTMLElement>` work here?

bentleyvk:
Why tests are still failing?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+          // but we still need to pass them here for backwards compatibility with MenuItem
+          className: cx(customItem.props.className, 'iui-focused'),
+          'data-iui-index': __originalIndex,
+        })
+      ) : (
+        <ComboBoxMenuItem
+          key={option.id ?? __originalIndex}
+          {...rest}
+          isSelected={selectedIndex === __originalIndex}
+          onClick={() => dispatch(['select', __originalIndex])}
```
veekeys, line 311: onclick might need merge here too

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  ((x: [ComboBoxAction] | [ComboBoxAction, number]) => void) | undefined
+>(undefined);
+ComboBoxActionContext.displayName = 'ComboBoxActionContext';
+
+const useSafeContext = <T,>(context: React.Context<T>) => {
+  const value = React.useContext(context);
+  if (!value) {
+    throw new Error(`${context.displayName} is undefined`);
+  }
+  return value!; // eslint-disable-line @typescript-eslint/no-non-null-assertion -- we already checked for undefined
```
veekeys, line 434: this is weird.. why it fails to see the check above? because no return?
Maybe this whole function can be moved to utils and reused, right?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +459,365 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    //   }
+    // });
+
+    React.useEffect(() => {
+      focusedIndexRef.current = {
+        original: focusedIndex ?? -1,
+        filtered: optionsRef.current.findIndex(
+          (el) => Number(el.dataset.iuiIndex) === focusedIndex,
+        ),
+      };
```
veekeys, line 537: `optionsRef` in deps will do nothing.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +459,365 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        ref={refs}
+        className={cx('iui-scroll', className)}
+        {...rest}
+      />
+    );
+  },
+);
+ComboBoxMenu.displayName = 'ComboBoxMenu';
+
+const ComboBoxMenuItem = React.memo(
```
veekeys, line 743: Why memo?

src/core/ComboBox/ComboBox.tsx
bentleyvk: I think the problem here is that you change `selectedIndex` and `value` is changed from outside that causes full rerender. But if with virtual scroll it is working fine, maybe no need to dig deeper.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  }, [isOpen, options, selectedIndex]);
+
+  // Call user-defined onChange when the value actually changes
+  React.useEffect(() => {
+    if (selectedIndex != undefined && selectedIndex >= 0) {
+      const value = options[selectedIndex]?.value;
+      if (value === valueProp) {
+        return;
+      }
+      onChangeProp.current?.(value);
```
bentleyvk, line 236: It will be also called on options change and on selection, not only on value change.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      const { __originalIndex, ...rest } = option;
+
+      const customItem = itemRenderer
+        ? itemRenderer(option, {
+            isFocused: false,
+            isSelected: false,
+            index: __originalIndex,
+            id: option.id,
+          })
+        : null;
```
bentleyvk, line 252: Here could be `<ComboboxMenuItem>`. That way no need another trinary on return.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
 
-  const menuItems = React.useMemo(() => {
-    if (filteredOptions.length === 0) {
-      return [
-        <MenuExtraContent key={0}>
-          <Text isMuted>{emptyStateMessage}</Text>
-        </MenuExtraContent>,
-      ];
+  return (
+    <ComboBoxRefsContext.Provider
```
bentleyvk, line 322: Why 3 providers instead of 1?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+          value={{ id, minWidth, isOpen, focusedIndex }}
+        >
+          {children ?? (
+            <>
+              <ComboBoxInputContainer
+                disabled={inputProps?.disabled}
+                endIcon={
+                  <ComboBoxEndIcon
+                    className={cx({
+                      'iui-actionable': !inputProps?.disabled,
```
bentleyvk, line 336: These could be props and classes could be moved inside icon component.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    >
+      <ComboBoxActionContext.Provider value={dispatch}>
+        <ComboBoxStateContext.Provider
+          value={{ id, minWidth, isOpen, focusedIndex }}
+        >
+          {children ?? (
+            <>
+              <ComboBoxInputContainer
+                disabled={inputProps?.disabled}
+                endIcon={
```
bentleyvk, line 333: Why not just place icon at the end, instead of having prop?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      return { ...state, isOpen: false };
+    }
+    case 'toggle': {
+      return { ...state, isOpen: !state.isOpen };
+    }
+    case 'select': {
+      return {
+        ...state,
+        isOpen: false,
+        selectedIndex: value ?? state.focusedIndex ?? state.selectedIndex,
```
bentleyvk, line 389: How can selected index be focused index?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +459,365 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-              }}
-              setFocus={false}
-              role='listbox'
-              ref={menuRef}
-            >
-              {menuItems}
-            </Menu>
+
+            let nextIndex = focusedIndexRef.current.filtered;
+            do {
```
bentleyvk, line 570: Can you just get all focusable items and increase index of focusable list?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +459,365 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    props: Omit<MenuProps, 'onClick'> & React.ComponentPropsWithoutRef<'ul'>,
+    forwardedRef: React.RefObject<HTMLUListElement>,
+  ) => {
+    const { className, style, ...rest } = props;
+    const { minWidth, id } = useSafeContext(ComboBoxStateContext);
+    const { menuRef, optionsRef } = useSafeContext(ComboBoxRefsContext);
+
+    React.useEffect(() => {
+      if (menuRef.current != null) {
+        const items = menuRef.current.querySelectorAll('[data-iui-index]');
```
bentleyvk, line 713: What won't work in custom rendered items that are missing this attribute?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      );
       inputProps?.onChange?.(event);
     },
-    [inputProps],
+    [filterFunction, inputProps, options],
   );
 
-  // update inputValue and focusedIndex every time selected value changes
+  // Reset focused item when filter changes
   React.useEffect(() => {
```
bentleyvk, line 202: Can it be that this reset is causing you a bug where initialy selected item is not focused?

stories/core/ComboBox.stories.tsx
```diff
@@ -497,3 +503,104 @@ WithCustomMessageIcon.args = {
...
+  };
+
+  const handleItemClick = (label: string) => {
+    setInputValue(label);
+    setResults([]);
+    setIsPopoverVisible(false);
+  };
+
+  return (
+    <ComboBox options={[]}>
```
bentleyvk, line 563: Isn't this demo too complex?

veekeys (commented):
I still need to play around with it and check up keyboard nav

src/core/ComboBox/ComboBox.tsx
mayank99: Yeah, but the linter needs it because it's coming from context.

src/core/ComboBox/ComboBox.tsx
mayank99: To avoid excessive rerenders. Is there a downside to not having it?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -14,10 +14,9 @@ import {
...
   PopoverProps,
   CommonProps,
-  getFocusableElements,
   getRandomValue,
   InputContainerProps,
-  mergeRefs,
+  useMergedRefs,
 } from '../utils';
 import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
 import 'tippy.js/animations/shift-away.css';
```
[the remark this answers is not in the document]
mayank99: It's the same as before

src/core/ComboBox/ComboBox.tsx
mayank99: I tried to remove it and it didn't fix the bug

src/core/ComboBox/ComboBox.tsx
mayank99: keyboard navigation

src/core/ComboBox/ComboBox.tsx
mayank99: getFocusableElements will not work with virtualized lists because the elements are not always in the DOM

src/core/ComboBox/ComboBox.tsx
mayank99: it selects the currently focused index if value is not passed in the dispatch

src/core/ComboBox/ComboBox.tsx
mayank99: when any value inside a context changes, it rerenders the whole component that uses the context, even if that component is not using the part that changed. So it is better to split into multiple contexts.

src/core/ComboBox/ComboBox.tsx
mayank99: `MenuItem` doesn't expose the mouse event

src/core/ComboBox/ComboBox.tsx
mayank99: with what?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            dispatch(['select', __originalIndex as number]);
+            customItem.props.onClick?.(e);
+          },
+          'data-iui-index': __originalIndex,
+        })
+      ) : (
+        <ComboBoxMenuItem
+          key={(__originalIndex as number) ?? index}
+          index={option.__originalIndex}
+          onClick={() => dispatch(['select', __originalIndex as number])}
```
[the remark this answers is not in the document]
mayank99: that's what the `rest` in next line does

src/core/ComboBox/ComboBox.tsx
```diff
@@ -117,60 +123,48 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-      }),
-    [options, getOptionId, itemRenderer],
-  );
+  // Add id and originalIndex to each option
+  const [options, setOptions] = React.useState<ComboBoxOption<T>[]>([]);
+  React.useLayoutEffect(() => {
+    setOptions(
+      optionsProp.map((option, index) => ({
+        ...option,
+        id: `${id}-option${index}`,
```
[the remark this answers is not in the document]
mayank99: 👍 https://github.com/iTwin/iTwinUI-react/pull/623/commits/666c0686bd6acce4c1c60dfb974301fa6813ed76

mayank99:
I've reworked keyboard navigation (again), kinda reverting my last rework. Now that we are ok with losing focused index when using virtual scroll + mouse, It seems to work pretty well in all existing cases (regular, controlled, custom renderer, custom composed).

While testing custom composed (`Async` story) example, I noticed that pressing <kbd>⏎</kbd> and <kbd>Esc</kbd> does not work, which is kinda expected because there are no `options` available in state. I'm thinking we might want to provide something like an `onHighlight` callback (which will be called every time a menu item is focused) so that the user can handle it on their end.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        React.cloneElement(customItem, {
+          onClick: (e: unknown) => {
+            dispatch(['select', __originalIndex as number]);
+            customItem.props.onClick?.(e);
+          },
+          'data-iui-index': __originalIndex,
+        })
+      ) : (
+        <ComboBoxMenuItem
+          key={(__originalIndex as number) ?? index}
```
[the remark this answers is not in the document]
mayank99: hmm. I think there was a reason I added this, but can't remember.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +174,274 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            isSelected: false,
+            index: __originalIndex,
+            id: option.id,
+          })
+        : null;
+
+      return customItem ? (
+        React.cloneElement(customItem, {
+          onClick: (e: unknown) => {
+            dispatch(['select', __originalIndex as number]);
```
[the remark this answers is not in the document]
mayank99: Apparently not. Removed in [`f9e42d5`](https://github.com/iTwin/iTwinUI-react/pull/623/commits/f9e42d55b02ce8a225fbbd0765c9830d61300276)

src/core/ComboBox/ComboBox.tsx
bentleyvk: I know that it causes rerenders? But is it a case here?

src/core/ComboBox/ComboBox.tsx
bentleyvk: Does it make sense to select focused item? I would say it should be the other way arround.

bentleyvk:
> While testing custom composed (`Async` story) example, I noticed that pressing ⏎ and Esc does not work, which is kinda expected because there are no `options` available in state. I'm thinking we might want to provide something like an `onHighlight` callback (which will be called every time a menu item is focused) so that the user can handle it on their end.

Can you use `itemRenderer` instead of custom popover?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +175,268 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      if (value === valueProp) {
+        return;
+      }
+      onChangeProp.current?.(value);
+    }
+  }, [options, selectedIndex, valueProp]);
+
+  // Memoize the default (not selected/focused) version of menu items
+  const memoizedItems = React.useMemo(() => {
+    return options.map((option, index) => {
```
bentleyvk, line 243: index is not used

bentleyvk:
Is it only for me, that after selecting a value, I can't open the combobox menu?

src/core/ComboBox/ComboBox.tsx
veekeys: `option.onClick`

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
veekeys: well, I meant the `key`, `onClick`, etc props, which you assign yourself.

stories/core/ComboBox.stories.tsx
```diff
@@ -424,8 +424,14 @@ export const CustomRenderer: Story<Partial<ComboBoxProps<string>>> = (args) => {
   }, []);
 
   const itemRenderer = React.useCallback(
-    ({ value, label }, { isSelected, id }) => (
-      <MenuItem key={value} id={id} isSelected={isSelected} value={value}>
+    ({ value, label }, { isSelected, id, index }) => (
+      <ComboBox.MenuItem
```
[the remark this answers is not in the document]
veekeys: Would be good to leave story as it was so we are sure nothing got broken using it "the old way"

src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,189 +175,268 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  }, [filteredOptions]);
 
-  // Filter options and update focus when input value changes
+  // When the value prop changes, update the selectedIndex
   React.useEffect(() => {
-    if (!isOpen) {
-      return;
-    }
+    dispatch([
+      'select',
```
bentleyvk, line 210: Don't you need to set new focus index too?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -301,7 +301,9 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
           },
           // ComboBox.MenuItem handles data-iui-index and iui-focused through context
           // but we still need to pass them here for backwards compatibility with MenuItem
-          className: cx(customItem.props.className, 'iui-focused'),
+          className: cx(customItem.props.className, {
+            'iui-focused': focusedIndex === __originalIndex,
```
veekeys, line 305: now this got weird :D you said yourself, that you have early return, if focusedIndex would equal to original one

src/core/ComboBox/ComboBox.tsx
```diff
@@ -545,62 +522,73 @@ const ComboBoxInput = React.forwardRef(
           case 'ArrowDown': {
             event.preventDefault();
             if (!isOpen) {
-              return dispatch(['open']);
+              dispatch(['open']);
```
veekeys, line 525: I'd say, if you open, just focus first item too.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +454,363 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-            </Menu>
+
+            let nextIndex = focusedIndexRef.current;
+            do {
+              const currentElement = menuRef.current?.querySelector(
+                `[data-iui-index="${nextIndex}"]`,
+              );
+              const nextElement =
+                currentElement?.nextElementSibling ??
+                menuRef.current?.querySelector('[data-iui-index]');
```
bentleyvk, line 546: When it can't have sibling?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +454,363 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+              nextIndex = Number(nextElement?.getAttribute('data-iui-index'));
+              const __originalIndex = Number(
+                nextElement?.getAttribute('data-iui-index') ?? -1,
+              );
+
+              if (nextElement?.ariaDisabled !== 'true') {
+                return dispatch(['focus', __originalIndex]);
+              }
+            } while (nextIndex !== focusedIndexRef.current);
```
bentleyvk, line 555: Won't it get into infinite loop with 1 item?

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +454,363 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            let nextIndex = focusedIndexRef.current;
+            do {
+              const currentElement = menuRef.current?.querySelector(
+                `[data-iui-index="${nextIndex}"]`,
+              );
+              const nextElement =
+                currentElement?.nextElementSibling ??
+                menuRef.current?.querySelector('[data-iui-index]');
+              nextIndex = Number(nextElement?.getAttribute('data-iui-index'));
+              const __originalIndex = Number(
```
veekeys, line 548: amm.. why you need `__originalIndex` ? Value same as `nextIndex` just with -1 if get undefined attribute.

src/core/ComboBox/ComboBox.tsx
```diff
@@ -380,84 +454,363 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            let prevIndex = focusedIndexRef.current;
+            do {
+              const currentElement = menuRef.current?.querySelector(
+                `[data-iui-index="${prevIndex}"]`,
+              );
+              const prevElement =
+                currentElement?.previousElementSibling ??
+                menuRef.current?.querySelector('[data-iui-index]:last-of-type');
+              prevIndex = Number(prevElement?.getAttribute('data-iui-index'));
+              const __originalIndex = Number(
```
bentleyvk, line 584: Move it under `if` where it is used.

src/core/ComboBox/ComboBox.tsx
mayank99: So you want 3 new props on this super simple component?

src/core/ComboBox/ComboBox.tsx
mayank99: I have both cases.
image

src/core/ComboBox/ComboBox.tsx
mayank99: and how would i get original index from children if this attribute is not present?

src/core/ComboBox/ComboBox.tsx
mayank99: removed

src/core/ComboBox/ComboBox.tsx
mayank99: there is no `return`, so it will perform the appropriate action after opening

src/core/ComboBox/ComboBox.tsx
mayank99: when at the first or last item

src/core/ComboBox/ComboBox.tsx
mayank99: i didn't find any infinite loop

src/core/ComboBox/ComboBox.tsx
mayank99: removed

src/core/ComboBox/ComboBox.tsx
mayank99: i've removed it

mayank99:
> Can you use `itemRenderer` instead of custom popover?

No, because there are no `options`. It's either full composition or nothing.

I really think we should have some useful props like `onSelect` and `onHighlight` so that users could build all combinations of things.

src/core/ComboBox/ComboBox.tsx
mayank99: i've read this comment about 10 different times in 4 days and i'm still not sure what it means 😄

mayank99:
> Is it only for me, that after selecting a value, I can't open the combobox menu?

This should be fixed now. All tests are passing too.

src/core/ComboBox/ComboBox.tsx
veekeys: Huh? I dont understand. Click even is gonna be either `React.MouseEvent<HTMLElement>` or then will be no event at all (if `MenuItem` does not expose event at all).

src/core/ComboBox/ComboBox.tsx
veekeys: Oh, I think previously there wasnt (or I didnt notice) check for selectedIndex too.
So when I commented, idea was:
You already check if focused index is not equal to original and early return. So on that particalar code line `focused === __originalIndex` could be replaced just with `true` as it was before.

src/core/ComboBox/ComboBox.tsx
veekeys: I dont think it fully performs the appropiate action now.
image

After selected value and opening again with keyboard, it does not focus selected value.

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
veekeys: well.. just if we have `Popover` and then we have `ComboBoxPopover` kinda implies we talk about same `Popover` and having consistent APIs somehow feels better.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -382,84 +457,349 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-            </Menu>
+
+            let nextIndex = focusedIndexRef.current;
+            do {
+              const currentElement = menuRef.current?.querySelector(
+                `[data-iui-index="${nextIndex}"]`,
+              );
+              const nextElement =
+                currentElement?.nextElementSibling ??
+                menuRef.current?.querySelector('[data-iui-index]');
```
veekeys, line 548: What this would do? Just get first element, which has `data-iui-index`.

veekeys: a ok.. I guess thats the part which stops cycle going infinity

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -382,84 +457,349 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      );
+    };
+
+    const handleKeyDown = React.useCallback(
+      (event: React.KeyboardEvent<HTMLInputElement>) => {
+        onKeyDownProp?.(event);
+        const length = optionsRef.current.length ?? 0;
+
+        switch (event.key) {
+          case 'ArrowDown': {
```
veekeys, line 524: hm...whats not nice now, is that if you hold arrow, nothing happens.. You dont see list going down anymore..

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
```diff
@@ -424,8 +424,14 @@ export const CustomRenderer: Story<Partial<ComboBoxProps<string>>> = (args) => {
   }, []);
 
   const itemRenderer = React.useCallback(
```
veekeys, line 426: story controls dissapeared now.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -382,84 +457,349 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+              className: cx(badge.props.className, 'iui-icon'),
+            })}
+        </li>
+      );
+    },
+  ),
+);
+ComboBoxMenuItem.displayName = 'ComboBoxMenuItem';
+
+ComboBox.InputContainer = ComboBoxInputContainer;
```
bentleyvk, line 798: Maybe have subcomponents in different files like in other our components. Because it is hard to view ~1k lines of code.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -382,84 +457,349 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+        autoCapitalize='none'
+        autoCorrect='off'
+        {...rest}
+      />
+    );
+  },
+);
+ComboBoxInput.displayName = 'ComboBoxInput';
+
+const ComboBoxPopover = React.forwardRef(
```
bentleyvk, line 643: Maybe worth calling it `ComboBoxMenu` or `ComboBoxDropdown`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
bentleyvk: It is going down for me on holding the key. But I for me it is lagging. Opened deployed original version and it is navigating smoothly.

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
```diff
@@ -9,12 +9,12 @@ import {
...
   ComboBoxProps,
   Label,
-  MenuItem,
   StatusMessage,
   SelectOption,
+  ProgressRadial,
 } from '../../src/core';
 import { CreeveyStoryParams } from 'creevey';
-import { SvgCamera } from '@itwin/itwinui-icons-react';
+import { SvgCamera, SvgSearch } from '@itwin/itwinui-icons-react';
```
bentleyvk, line 18: Disabled story takes a long time to load, whole window freezes and fans start to go wild.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: i think @veekeys will disagree

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -183,188 +175,271 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-  const [inputValue, setInputValue] = React.useState<string>('');
-  const onInput = React.useCallback(
+  // Filter options based on input value
+  const [inputValue, setInputValue] = React.useState(inputProps?.value ?? '');
+  const handleOnInput = React.useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
-      setInputValue(event.target.value);
+      const { value } = event.currentTarget;
+      setInputValue(value);
+      dispatch(['open']);
```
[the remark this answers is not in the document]
mayank99: if you select an option, it closes. so it should reopen if closed when typing

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Yeah, cause other one for me did not even load (disabled one)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Carousel we  did separate in files. So I guess lets be consistent..?

src/core/ComboBox/ComboBox.tsx
veekeys: oh, if you checked the performance gain, thats good.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
adamhock: Holding down arrow works for me

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
adamhock: I agree to put subcompontents in different files

packages/iTwinUI-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -209,77 +210,99 @@ it('should handle keyboard navigation', () => {
...
   const { container } = renderComponent({ id, onChange: mockOnChange });
 
+  userEvent.tab();
+
   const input = assertBaseElement(container);
-  input.focus();
   expect(input).toHaveAttribute('aria-controls', `${id}-list`);
 
-  const items = document.querySelectorAll('.iui-menu-item');
+  let items = document.querySelectorAll('.iui-menu-item');
```
adamhock, line 218: I don't understand this change and why `items` needs to be redefined every time

mayank99: the test was failing and when I tried debugging, I saw that `items` was referring to the old instance, so it had to be queried again. i have no clue how/why it was working before, but this change was just to get the test to pass.

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
mayank99: Any idea why they disappeared? I haven't changed anything in most of the stories.

Not that they were useful, but still.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: I've split everything into separate files 👍

stories/core/ComboBox.stories.tsx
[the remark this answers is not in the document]
mayank99: Reverted story back to use `MenuItem`. Found a bug with scrollIntoView for which I had to add a ref in the `cloneElement`

src/core/ComboBox/ComboBox.tsx
mayank99: Added

src/core/ComboBox/ComboBox.tsx
mayank99: no, `MenuItem` (confusingly) passes the `value` of the item.

https://github.com/iTwin/iTwinUI-react/blob/4c9b4a8ae9973303c1ea9314de6958773308b8be/packages/iTwinUI-react/src/core/Menu/MenuItem.tsx#L28-L35
https://github.com/iTwin/iTwinUI-react/blob/4c9b4a8ae9973303c1ea9314de6958773308b8be/packages/iTwinUI-react/src/core/Menu/MenuItem.tsx#L146

src/core/ComboBox/ComboBox.tsx
mayank99: Using `children` now 👍

src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: The option type doesn't have any of those props, so is it expected that the user will pass something in there? The arbitrary `[key]` is mainly for `aria-` and `data-` attributes iirc.
image

mayank99:
I've split everything into separate files and removed the subcomponents from barrel. I've also removed the `Async` story (we should add this in a future PR when we have a better API) and the `Many Items` story (we should add this in the virtualization PR).

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: It still does not work for me... :/

veekeys:
Can we switch back behavior, that when you are at the last item, you cannot appear on the first?
Just in case Ill get issues with VS to achieve this :D

packages/iTwinUI-react/src/core/ComboBox/ComboBoxDropdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+      if (props.visible != undefined) {
+        dispatch([props.visible ? 'open' : 'close']);
+      }
+    }, [dispatch, props.visible]);
+
+    return (
+      <Popover
+        placement='bottom-start'
+        visible={isOpen}
+        onClickOutside={React.useCallback(
```
veekeys, line 29: using inline hooks behaves same as if you had variable, right?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: But you do not export them. On purpose just as internal use?

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
veekeys: I think some of them returned now :/

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -0,0 +1,46 @@
...
+        optionsRef.current = Array.from(items) as HTMLLIElement[]; // need this cast to make TS happy :(
+      }
+    }, [menuRef, optionsRef, props.children]);
+
+    const refs = useMergedRefs(menuRef, forwardedRef);
+
+    return (
+      <Menu
+        id={`${id}-list`}
+        style={React.useMemo(
```
veekeys, line 28: Inline hooks act same as if you had them as variables?

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
bentleyvk: Any ideas what is the cause?

mayank99: idk. must be some storybook bug. I don't think it's related to this PR

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: well yeah

mayank99:
> Can we switch back behavior, that when you are at the last item, you cannot appear on the first? Just in case Ill get issues with VS to achieve this :D

But it's nice to have this feature. We can always disable this for VS if it causes issues.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: Renamed to `ComboBoxDropdown`.

There is already a `ComboBoxMenu` expected as a child of this.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: Yup, lets keep internal for now. We can expose a better API with a better custom example in a future PR

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -45,9 +52,9 @@ export type ComboBoxProps<T> = {
    * Function to customize the default filtering logic.
    */
   filterFunction?: (
-    options: SelectOption<T>[],
+    options: ComboBoxOption<T>[],
```
veekeys, line 55: Isnt this one a breaking change????

mayank99: Is it? This is a function we call. The user just takes whatever from the `options`.

veekeys: In my understanding it is the change of API (types).
Not only the coming argument type changed (so if users defined that explicitly, will get errors), but also the return type changed (if users do not do `..rest` they will end up with errors too ) :/

mayank99: Oh yeah, good point about return type. Hmm, I will revert this then

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
bentleyvk: But why it was working previously? 😀

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+    case 'select': {
+      return {
+        ...state,
+        isOpen: false,
+        selectedIndex: value ?? state.selectedIndex,
+        focusedIndex: value ?? state.focusedIndex,
+      };
+    }
+    case 'focus': {
+      return { ...state, focusedIndex: value ?? state.selectedIndex ?? -1 };
```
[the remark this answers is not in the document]
mayank99: no, we want to fall back to `selectedIndex` because that's what needs to be focused when `value` is not passed

packages/iTwinUI-react/src/core/ComboBox/ComboBoxEndIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+            'iui-actionable': !disabled,
+            'iui-disabled': disabled,
+            'iui-open': isOpen,
+          },
+          className,
+        )}
+        onClick={(e) => {
+          onClickProp?.(e);
+          if (!e.isDefaultPrevented()) {
+            dispatch(['toggle']);
```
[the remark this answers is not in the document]
mayank99: um ok, done

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
```diff
@@ -9,9 +9,9 @@ import {
   ComboBox,
   ComboBoxProps,
   Label,
-  MenuItem,
   StatusMessage,
   SelectOption,
+  MenuItem,
```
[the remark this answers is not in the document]
mayank99: done

packages/iTwinUI-react/stories/core/ComboBox.stories.tsx
mayank99: it lags for me even without this PR, so I don't think it was working previously

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: This is going to be tough because our `filteredOptions` state expects `ComboBoxOption` so that it is able to access `__originalIndex`

packages/iTwinUI-react/src/core/ComboBox/ComboBoxEndIcon.tsx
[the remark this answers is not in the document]
adamhock: Now the `"toggle"` case can be removed from the `ComboBoxAction` list in helper.ts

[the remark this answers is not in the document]
adamhock: One last spot [here](https://github.com/iTwin/iTwinUI-react/pull/623/files#diff-65491c8f7818bdc2149a18e25ea8664c7eb9f092ce4961187bc4f5159baeae9aR3)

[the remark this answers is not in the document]
mayank99: removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: What can we do?

veekeys: Lets fix this and I approve. Too long hanging.

adamhock (approved):

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: Fixed by saving `__originalIndex` in `optionsExtraInfoRef` and getting rid of `ComboBoxOption`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -155,6 +143,28 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    Record<string, { __originalIndex: number }>
+  >({});
+
+  // Clear the extra info when the options change so that it can be reinitialized below
+  React.useEffect(() => {
+    optionsExtraInfoRef.current = {};
+  }, [id, options]);
+
+  // Initialize the extra info only if it is not already initialized
+  if (
```
veekeys, line 153: Why this approach instead of `useEffect` ?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -70,8 +70,10 @@ export type ComboBoxProps<T> = {
   emptyStateMessage?: string;
   /**
    * A custom item renderer can be specified to control the rendering.
-   * This function should ideally return a customized version of `ComboBox.MenuItem`,
-   * otherwise you will need to make sure to provide styling for the `isFocused` state.
+   *
+   * For keyboard navigation to work, the returned element should use the `id` provided by this function.
+   * The `isFocused` state is calculated using this `id` and can be used for specifying the focus styling.
+   * If a `MenuItem` is returned, then focus sytling is automatically handled.
```
veekeys, line 76: `sytling` you say.
Where is vscode extension of spell checker????

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -82,10 +84,6 @@ export type ComboBoxProps<T> = {
       index: number;
     },
   ) => JSX.Element;
-  /**
-   * Children can be specified to customize individual parts.
-   */
-  children?: React.ReactNode;
 } & Pick<InputContainerProps, 'status'> &
```
veekeys, line 87: why children removed?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -10,6 +10,7 @@ import { SvgCaretDownSmall } from '@itwin/itwinui-icons-react';
 import { MenuItem } from '../Menu';
 import { StatusMessage } from '../StatusMessage';
 import userEvent from '@testing-library/user-event';
+import { act } from 'react-dom/test-utils';
```
bentleyvk, line 13: Why it is imported not from testing-library?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: somehow missed it
image

fixed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.test.tsx
mayank99: Ask vscode autoimport. Fixed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: if using `useEffect`, this ref is not available on first render so I can't access `__originalIndex` inside `getMenuItem`

veekeys: ref is available in useEffect, unless element is conditional.

veekeys: a wait, ref of menu items?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: so bad marking... :O

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: `optionsExtraInfoRef` doesn't get set until after the useEffect, so it's not available when `getMenuItem` is first called

veekeys (approved):
Great job!
Looking forward to bug-free use cases of composition.

PR #624, opened 2022-04-12
feat(Surface): Create Surface component

### LINKED ISSUE #615: Add Surface component

adamhock:
Added `Surface` component which allows content to appear elevated through the use of a drop shadow. Users can change the elevation of the surface

SurfaceAfter

[CSS PR](https://github.com/iTwin/iTwinUI/pull/533)

Closes #615


stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  },
+  parameters: {
+    creevey: {
+      captureElement: null,
+    },
+  },
+  title: 'Core/Surface',
+} as Meta<SurfaceProps> & CreeveyMeta;
+
+const cardStyle = {
```
adamhock, line 27: I added additional styling to the story to make it look similar to the tiles shown in the CSS demo

Before: 
SurfaceBefore

After:
SurfaceAfter

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+   */
+  elevation?: 2 | 4 | 8 | 16 | 24;
+  /**
+   * Content in the surface.
+   */
+  children: React.ReactNode;
+} & CommonProps;
+
+/**
+ * The surface container allows content to appear elevated through the use of a drop shadow
```
adamhock, line 45: If anyone has a better suggestion for how to describe this component, feel free to comment

stories/core/Surface.stories.tsx
veekeys: Then I would put that inline so users could see code too..

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  height: '100px',
+  display: 'flex',
+  'justify-content': 'center',
+  'align-items': 'center',
+};
+export const Basic: Story<SurfaceProps> = ({ elevation }) => {
+  return (
+    <Surface elevation={elevation} style={cardStyle}>
+      Sample Surface Component
```
veekeys, line 36: hmm.. maybe write something more? Can be even repeated as the jsdoc of the component.

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
+    elevation: {
+      options: [2, 4, 8, 16, 24],
+    },
+  },
+  parameters: {
+    creevey: {
+      captureElement: null,
```
veekeys, line 21: Why added this?

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * @example
+ * <Surface>Surface Content</Surface>
+ * <Surface elevation={2}>Surface Content</Surface>
+ */
+export const Surface = (props: SurfaceProps) => {
```
veekeys, line 50: We can forward ref too

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Story, Meta } from '@storybook/react';
+import { CreeveyMeta } from 'creevey';
+import React from 'react';
+import { Surface, SurfaceProps } from '../../src/core';
+
+export default {
+  component: Surface,
+  argTypes: {
+    className: { control: { disable: true } },
+    style: { control: { disable: true } },
```
veekeys, line 14: remove `id` too

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  title: 'Core/Surface',
+} as Meta<SurfaceProps> & CreeveyMeta;
+
+const cardStyle = {
+  height: '100px',
+  display: 'flex',
+  'justify-content': 'center',
+  'align-items': 'center',
+};
+export const Basic: Story<SurfaceProps> = ({ elevation }) => {
```
veekeys, line 33: if you could also give default value, code readability to the users might improve.
`elevation = 2`

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Surface elevation={elevation} style={cardStyle}>
+      Sample Surface Component
+    </Surface>
+  );
+};
+
+Basic.args = {};
```
veekeys, line 41: Need to add here `elevation` value, so its visible in controls

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
```
veekeys, line 5: Wondering, if we should consider having folder for these `helper` components...

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+} & CommonProps;
+
+/**
+ * The surface container allows content to appear elevated through the use of a drop shadow
+ * @example
+ * <Surface>Surface Content</Surface>
+ * <Surface elevation={2}>Surface Content</Surface>
+ */
+export const Surface = (props: SurfaceProps) => {
+  const { elevation, className, style, children, ...rest } = props;
```
veekeys, line 51: so assign default value to `elevation` there.
It will also be always defined so culd you remove null check in helper func

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+  useTheme();
+
+  const _style = {
+    '--iui-surface-elevation': getSurfaceElevationValue(elevation),
```
veekeys, line 55: if you check other compoonents like `Badge` or `ColorBuilder` we do check if css supports variables. Not sure, if we want to continue doing that as we move away from IE, though:/
Would need to gather more opinions on this one

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+    case 4:
+      return '0 1px 10px rgba(0, 0, 0, 0.25)';
+    case 8:
+      return '0 3px 14px rgba(0, 0, 0, 0.25)';
+    case 16:
+      return '0 6px 30px rgba(0, 0, 0, 0.25)';
+    case 24:
+      return '0 9px 46px rgba(0, 0, 0, 0.25)';
+    default:
+      return '';
```
veekeys, line 28: Would there be valid case to have Surface without elevation? Then `elevation` could be undefined and make sense too :/

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+};
+
+export type SurfaceProps = {
+  /**
+   * Sets the size of the box shadow.
+   * @default 2
+   */
+  elevation?: 2 | 4 | 8 | 16 | 24;
```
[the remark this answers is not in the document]
adamhock: The types of elevation are mentioned in the documentation [here](https://ux.bentley.com/itwin/styles/elevation.html). I talked with @FlyersPh9 yesterday and he said elevation for sure needs to be an option.

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * Helper function that returns one of the preset surface elevation values.
+ */
+const getSurfaceElevationValue = (elevation: SurfaceProps['elevation']) => {
+  if (!elevation) {
+    return '';
+  }
+
+  switch (elevation) {
+    case 4:
+      return '0 1px 10px rgba(0, 0, 0, 0.25)';
```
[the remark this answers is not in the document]
adamhock: I found those in the Surface CSS demo for each of the elevation depths

src/core/Surface/Surface.tsx
veekeys: Also, another idea :D
Would it make sense to have a hook, which would just deliver a class or style, so you could apply that on already existing components?
I would add that to popover, for example (instead of reusing mixing in css)

stories/core/Surface.stories.tsx
adamhock: Without it, the test image was only capturing the border and not the box shadow around it

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
veekeys: Then you are missing no elevation option.
And we dont have this info available anywhere to everyone, so I'd still argue about the values.
I read this prop and have no idea what random numbers mean, so huge change users would react same. We need to think how else we could deliver this

stories/core/Surface.stories.tsx
veekeys: Right.. Just now image captures whole iframe, which is also not needed.. But it is 1 image, so I guess its fine.

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
veekeys: hmm
image

I'd prefer us using variables (add if not available), so we dont need to support these in 2 places...

stories/core/Surface.stories.tsx
veekeys: btw, I dont remember, what we have decided and why want to show prop, but disable control.. Is it just so people are aware of these props are existing?
Cause otherwise we could fully hide the control of those and not bother `{ table: { disable: true } }`...

stories/core/Surface.stories.tsx
adamhock: Code has been moved

stories/core/Surface.stories.tsx
adamhock: Removed `id`

stories/core/Surface.stories.tsx
adamhock: Added `elevation` value

src/core/Surface/Surface.tsx
adamhock: assigned default elevation here and removed null check in helper function

stories/core/Surface.stories.tsx
adamhock: Added

src/core/Surface/Surface.tsx
FlyersPh9: Adam and I were trying to come up with a description yesterday but struggled.  The surface component gives the user a container to place their content within.  It can or can not include drop shadow.  The main place I see it being used is within the home page iTwinUI-layouts, but it can also be used within components via mixin like tile, menu, etc.

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
adamhock: I found that these numbers are actually in units of density-independent pixels (dp). They go from 0 to 24, so maybe that can be the range?

stories/core/Surface.stories.tsx
veekeys: Did you?

veekeys: oh, you did.. Somehow it was showing me old history of the file :/

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
veekeys: So I guess we need to have a final decision, how many elevation levels our design system will support and have this prop acordingly in number, I guess. Not sure we can have self-descriptive string names. Instead, we could have a detailed description of the prop and explanation of the values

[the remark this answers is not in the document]
bentleyvk: I thought we are removing all of these.

src/core/Surface/Surface.tsx
adamhock: Case has been added for elevation 0

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
adamhock: Updated description for elevations
`Sets the elevation of the surface in density-independent pixels (dp)`

stories/core/Surface.stories.tsx
adamhock: Updated to match the Surface component description

src/core/Surface/Surface.tsx
adamhock: Forward ref added

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
adamhock: I just have the hardcoded elevations listed here for now

src/core/Surface/Surface.tsx
adamhock: Added IE support just in case

src/core/Surface/Surface.tsx
```diff
@@ -40,12 +40,12 @@ export type SurfaceProps = {
...
 
 /**
- * The surface container allows content to appear elevated through the use of a drop shadow
+ * The Surface container allows content to appear elevated through the use of a drop shadow
  * @example
  * <Surface>Surface Content</Surface>
  * <Surface elevation={2}>Surface Content</Surface>
  */
-export const Surface = (props: SurfaceProps) => {
+export const Surface = React.forwardRef((props: SurfaceProps) => {
```
veekeys, line 48: You are not using `ref` here (it comes as 2nd parameter). Check other components, how `forwardRef` syntax looks like or read here:
https://reactjs.org/docs/forwarding-refs.html

src/core/Surface/Surface.tsx
```diff
@@ -12,6 +12,8 @@ import '@itwin/itwinui-css/css/surface.css';
  */
 const getSurfaceElevationValue = (elevation: SurfaceProps['elevation']) => {
   switch (elevation) {
+    case 0:
+      return '0';
```
veekeys, line 16: Not sure..Docs say initial value is `none`
https://developer.mozilla.org/en-US/docs/web/css/box-shadow

adamhock: value changed to `none`

src/core/Surface/Surface.tsx
adamhock: forwardRef updated. Should be good

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Sets the size of the box shadow.
+   * @default 2
+   */
+  elevation?: 2 | 4 | 8 | 16 | 24;
+  /**
+   * Content in the surface.
+   */
+  children: React.ReactNode;
+} & CommonProps;
```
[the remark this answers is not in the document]
adamhock: `title` omitted

src/core/Surface/Surface.tsx
```diff
@@ -37,33 +37,40 @@ export type SurfaceProps = {
...
  * The Surface container allows content to appear elevated through the use of a drop shadow
  * @example
  * <Surface>Surface Content</Surface>
  * <Surface elevation={2}>Surface Content</Surface>
  */
-export const Surface = React.forwardRef((props: SurfaceProps) => {
-  const { elevation = 2, className, style, children, ...rest } = props;
-  useTheme();
+export const Surface = React.forwardRef<HTMLDivElement, SurfaceProps>(
+  (props: SurfaceProps, ref) => {
```
veekeys, line 49: You shouldnt need to redefine `props` type.

src/core/Surface/Surface.tsx
```diff
@@ -37,33 +37,40 @@ export type SurfaceProps = {
...
-  const _style = getWindow()?.CSS?.supports?.('--iui-surface-elevation: 0')
-    ? {
-        '--iui-surface-elevation': getSurfaceElevationValue(elevation),
-        ...style,
-      }
-    : {
-        'box-shadow': getSurfaceElevationValue(elevation),
-        ...style,
-      };
+    const _style = getWindow()?.CSS?.supports?.('--iui-surface-elevation: 0')
```
veekeys, line 53: Ideally, would need to check with `getSurfaceElevationValue`, not 0, but I guess minor. Can replace with `none`, for consistency.

src/core/Surface/Surface.tsx
```diff
@@ -37,33 +37,40 @@ export type SurfaceProps = {
...
+      ? {
+          '--iui-surface-elevation': getSurfaceElevationValue(elevation),
+          ...style,
+        }
+      : {
+          'box-shadow': getSurfaceElevationValue(elevation),
```
veekeys, line 59: Have you checked if it works? Shouldnt need to be `boxShadow` instead?

src/core/Surface/Surface.tsx
adamhock: removed

src/core/Surface/Surface.tsx
adamhock: Changed to include `getSurfaceElevationValue`

`--iui-surface-elevation: ${getSurfaceElevationValue(elevation)},`

src/core/Surface/Surface.tsx
```diff
@@ -0,0 +1,76 @@
...
+ * @example
+ * <Surface>Surface Content</Surface>
+ * <Surface elevation={2}>Surface Content</Surface>
+ */
+export const Surface = React.forwardRef<HTMLDivElement, SurfaceProps>(
+  (props: SurfaceProps, ref) => {
+    const { elevation = 2, className, style, children, ...rest } = props;
+    useTheme();
+
+    const _style = getWindow()?.CSS?.supports?.('--iui-surface-elevation: 0')
```
elephantcatdog: Lines 53-54 aren't being tested according to `yarn test`.

adamhock: These lines are just for IE 11 which doesn't supports vars. I've seen other examples with this check that don't test for the IE case (such as `Badge`). I don't think this needs to be tested

src/core/Surface/Surface.tsx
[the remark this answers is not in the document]
adamhock: Elevations have changed to `0 | 1 | 2 | 3 | 4 | 5`

veekeys (approved):

src/core/Surface/Surface.tsx
elephantcatdog: Ok, sounds good

stories/core/Surface.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  },
+  parameters: {
+    creevey: {
+      captureElement: null,
+    },
+  },
+  title: 'Core/Surface',
+} as Meta<SurfaceProps> & CreeveyMeta;
+
+export const Basic: Story<SurfaceProps> = ({ elevation = 0 }) => {
```
elephantcatdog, line 28: Can you change the story elevation to 1 so that the initial view of the story has a shadow?

elephantcatdog: That's a good solution. This is pretty nit-picky and not necessary, but if you add padding to the cardStyle the text won't be right against the edge of the box.

elephantcatdog (approved):
I would add padding to the box so the text isn't against the edge, but otherwise it looks good to me.

PR #630, opened 2022-04-19
chore: Update css to 0.55.0

veekeys:
Consuming latest css package.
- Tag required change, because apparently children for basic tag should not be wrapped with `label` span. So now either.. i have this change in react (which kinda changed DOM structure), or I will need to have something in css, where label rules (which are now separated) should not be applied to `basic` tag.
- Tile image changed, because now active Tile does not zoom in thumbnail. Is this expected? I know we need to add `actionable` prop to `Tile`.
- Header images changed to fix the button spacing.


src/core/Tag/Tag.tsx
```diff
@@ -40,16 +40,27 @@ export const Tag = (props: TagProps) => {
...
+        },
+        className,
+      )}
       {...rest}
     >
-      <span className='iui-label'>{children}</span>
+      {variant === 'default' ? (
+        <span className='iui-tag-label'>{children}</span>
+      ) : (
+        children
```
mayank99, line 55: so we previously had `iui-label` even for basic tags? Then we should probably make it consistent again with a css patch release?

mayank99:
> Tile image changed, because now active Tile does not zoom in thumbnail. Is this expected? I know we need to add actionable prop to Tile.

This is a bit bizarre for me that we are changing hover behavior. But according to @FlyersPh9, this change is intentional.

src/core/Tag/Tag.tsx
veekeys: In css, we didnt have it. only non-basic tags had examples with `iui-label`.
Just in react we were adding additional span for all the tags and because css selector had `:not(iui-basic)` we didnt face any issues.
I have separated label class in my [css pr](https://github.com/iTwin/iTwinUI/commit/0d3608f54efe0b327a4224c9fc7295e1d9029fec#diff-e67fd10966ce7f3a92e622e71a8bc6cca0ba784a83821f468487e64cb86e6f85R50) and now if we add it, some styling gets messed up :/

mayank99: Ok, makes sense. Are you planning to make it consistent again in the future?

mayank99 (approved):

src/core/Tag/Tag.tsx
veekeys: I dont know. Not now, I guess :/

src/core/Header/HeaderButton.test.tsx
```diff
@@ -149,7 +149,7 @@ it('should render menuItems correctly', () => {
   );
 
   const button = container.querySelector(
-    '.iui-header-button.iui-dropdown.iui-borderless',
+    '.iui-header-button.iui-dropdown.iui-borderless.iui-header-dropdown-button',
```
adamhock, line 152: Can we just grab `iui-header-dropdown-button` here instead of all of these?

veekeys: done

package.json
```diff
@@ -40,7 +40,7 @@
     "build-storybook": "build-storybook"
   },
   "dependencies": {
-    "@itwin/itwinui-css": "^0.54.1",
+    "@itwin/itwinui-css": "^0.55.0",
```
bentleyvk, line 43: There is 0.55.1 already

PR #631, opened 2022-04-20
feat(Tile): Add actionable prop

veekeys:
Added new prop to consume `iui-actionable` class.
Prop name and description up for debate.
Added all div props, so users could add `onClick` handler for example for interactable tile.
New story shows weird new Tile behavior, when onClick it gets `selected` state, but when you unselect, thumbnail is still scaled, because we have also `focus-within` 
image

Better to wait for #630 to be merged (or ignore other than Tile changes)


stories/core/Tile.stories.tsx
```diff
@@ -131,6 +133,68 @@ AllProps.argTypes = { ...Basic.argTypes };
 AllProps.args = {
   ...Basic.args,
   isSelected: true,
+  isActionable: true,
```
bentleyvk, line 136: Why passing like that? It is not visible to the users how props look.

veekeys: What do you mean its not? I also destructure in the code itself and assign default value

src/core/Tile/Tile.test.tsx
```diff
@@ -221,3 +221,8 @@ it('should propagate misc props correctly', () => {
   expect(child).toBeTruthy();
   expect(child.textContent).toBe('test-content');
 });
+
+it('should render actionable tile', () => {
```
adamhock, line 225: Should we write a test for tabbing to an actionable tile?

veekeys: And what would you like me to test there?

stories/core/Tile.stories.tsx
```diff
@@ -131,6 +133,68 @@ AllProps.argTypes = { ...Basic.argTypes };
...
+      thumbnail={thumbnail}
+      buttons={buttons}
+      leftIcon={leftIcon}
+      rightIcon={rightIcon}
+      isSelected={selected}
+      isNew={isNew}
+      moreOptions={moreOptions}
+      isActionable={isActionable}
+      {...rest}
+      onClick={() => setSelected((prev) => !prev)}
```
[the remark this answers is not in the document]
veekeys: I dont know.. It is not a custom action we have added (which we mostly want to show off in demos), but more of a funcnioliaty which was not there before (so actionable and onclick handler was not possible to be used straight on Tile).
`onClick` here is just same native handler as anything else available on dom elements.

src/core/Tile/Tile.test.tsx
adamhock: I was thinking appending this test with this

`expect(actionTile.getAttribute('tabIndex')).toBe('0');`

I guess I don't see other components have a test for this, so I guess we're good here

adamhock (approved):

bentleyvk (approved):

PR #645, opened 2022-05-02
fix(Table): Fix direction of sort icons

### LINKED ISSUE #641: Table: Sort icon is backward
The sort icon does not match [iTwin UI specs](http://ux.bentley.com/itwin/patterns/column-sort.html). It should point up when ascending, and down when descending. Currently, it points down when ascending and up when descending.

This can be seen on the current [sortable table story in the docs](https://itwin.github.io/iTwinUI-react/?path=/story/core-table--sortable).

image

FlyersPh9:
Swap the orientation of the sort icon to align with Windows and macOS. Closes #641.


mayank99 (approved):
Icon change looks good. I do have 3 comments:
- PR title should be updated to say `fix`, not `chore`. This is a user-facing change that affects more than just demos.
- I think the current conditional logic will show the wrong icons when the user overrides sorting order using [`sortDescFirst`](https://react-table.tanstack.com/docs/api/useSortBy#:~:text=will%20be%20disabled-,sortDescFirst,-%3A%20Bool). We may want to check and fix that in this PR.
- The formatting changes are unexpected. We should run `prettier` against all files (in a separate PR).

gretanausedaite (approved):
> * I think the current conditional logic will show the wrong icons when the user overrides sorting order using [`sortDescFirst`](https://react-table.tanstack.com/docs/api/useSortBy#:~:text=will%20be%20disabled-,sortDescFirst,-%3A%20Bool). We may want to check and fix that in this PR.

Tested with `sortDescFirst: true` and we should be good.

mayank99:
> > I think the current conditional logic will show the wrong icons when the user overrides sorting order using [`sortDescFirst`](https://react-table.tanstack.com/docs/api/useSortBy#:~:text=will%20be%20disabled-,sortDescFirst,-%3A%20Bool). We may want to check and fix that in this PR.
> 
> Tested with `sortDescFirst: true` and we should be good.

Doesn't look right to me. Hovering over a DescFirst column should show `SvgArrowDown` because that's the _next_ state (the result of the interaction).

https://user-images.githubusercontent.com/9084735/166505237-865cab90-1dc3-4182-9491-5e50083cf13b.mov

mayank99:
Update: Created #659 to track the `sortDescFirst` issue as it's unrelated to this PR.

We can merge this PR after the random formatting changes are reverted (#651 got merged, so `prettier`+`lint-staged` should be fixed).

PR #648, opened 2022-05-03
fix(Table): Fixes headless Table as previously it started to throw an error

bentleyvk:
Closes #647

When user doesn't pass columns with root wrapping header, then error is thrown because it can't find any columns for `hasManualSelectionColumn` check.
Also fixed an error where it would print to console that `isDisabled` is not a valid prop in `EditableCell`.

## Checklist

- [x] Add meaningful unit tests for your component (verify that all lines are covered)
- [x] Verify that all existing tests pass
- [x] ~Add component features demo in Storybook (different stories)~
- [x] ~Approve test images for new stories~
- [x] ~Add screenshots of the key elements of the component~

PR #636, opened 2022-04-22
fix(ComboBox): Use latest value of `onChange` prop

mayank99:
Previously, the ref was locked to the very initial value of `onChange`. The useEffect ensures that we get the latest value.

Fixes #635.

## Checklist

- [x] Add meaningful unit tests for your component (verify that all lines are covered)
- [x] Verify that all existing tests pass
- [x] ~Add component features demo in Storybook (different stories)~
- [x] ~Approve test images for new stories~
- [x] ~Add screenshots of the key elements of the component~

PR #613, opened 2022-04-01
chore: Convert to monorepo

mayank99:
Converted this repo into a monorepo using [turborepo](https://turborepo.org/) on top of yarn workspaces. Directory structure is split into `apps/` and `packages/`.
  - Moved all iTwinUI-react files into `packages/`
  - Added a vite playground under `apps/` which uses the linked itwinui-react package just for testing components in a more reliable setting compared to storybook.
  - In the future, the documentation `website` folder will also live under `apps/`

Most of the changes in this PR are simply additions from turborepo/vite and moving files around (git should show those as "renamed").

The only real "changes" are listed below:
- Node 16+ / npm 7+ restriction is now set in root `package.json`.
- iTwinUI-react build was changed to emit files into `cjs`/`esm` folders instead of `lib` so that it works with symlinks. There is no need to copy `package.json` into `lib` anymore.
- Some commands in github workflows had to be updated to reference itwinui-react instead of running from root.
- A lot of jest tests which test for tippy.js visibility started failing so I made some changes to get them to pass.
- Storybook was giving me the following error so I had to add `webpack@5` in the root `package.json`:
  - storybookjs/storybook/issues/15336
- Eslint+prettier was failing so I upgraded some deps and made changes accordingly.
- `ReactNodeArray` is deprecated and was showing an error in `Breadcrumbs` so I changed it to `ReactNode[]`.

A lot of those issues have nothing to do with turborepo but I think they started appearing because dependencies were updated when lockfile was moved.

---

Best way to test these changes:
- Create a fresh clone of this repo and switch to this branch.
- Run `yarn` to install everything.
- Run `yarn dev` to start vite playground and storybook together. 

For running any bespoke commands, you can `cd` into `packages/itwinui-react` or just run it from the repo root by prefixing `yarn workspace @itwin/itwinui-react`.

PR #651, opened 2022-05-03
chore: Fix prettier inconsistencies from duplicate versions

mayank99:
After #613, we had two different versions of `prettier` in the lockfile. One of them was being used for IDE save and `prettier --write`, and a different one was being used for `lint-staged`. This led to random unintended formatting changes like in https://github.com/iTwin/iTwinUI-react/pull/636#discussion_r856321814 and https://github.com/iTwin/iTwinUI-react/pull/648#discussion_r863614306.

So I deduped the prettier versions in our lockfile and ran `yarn prettier` against the repo and also verified that `lint-staged` doesn't break anything on committing. The only resulting formatting change is in `ComboBox.tsx`, reverting the unintended one from #636.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11090,15 +11090,10 @@ preserve@^0.2.0:
   resolved "https://registry.yarnpkg.com/preserve/-/preserve-0.2.0.tgz#815ed1f6ebc65926f865b310c0713bcb3315ce4b"
   integrity sha1-gV7R9uvGWSb4ZbMQwHE7yzMVzks=
 
-prettier@2.2.1:
+prettier@2.2.1, "prettier@>=2.2.1 <=2.3.0":
   version "2.2.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -11090,16 +11090,11 @@ preserve@^0.2.0:
   resolved "https://registry.yarnpkg.com/preserve/-/preserve-0.2.0.tgz#815ed1f6ebc65926f865b310c0713bcb3315ce4b"
   integrity sha1-gV7R9uvGWSb4ZbMQwHE7yzMVzks=
 
-prettier@2.2.1:
+prettier@2.2.1, "prettier@>=2.2.1 <=2.3.0":
```
veekeys, line 11093: didnt want to update to the newest?

veekeys (approved):

yarn.lock
mayank99: If I did, then it would change all the files and I would also need to force resolve because this `>=2.2.1 <=2.3.0` comes from storybook.

PR #656, opened 2022-05-04
feat: Add support of react 18

### LINKED ISSUE #633: React 18 support
CRA and create-vite ship with react 18, which our package.json is not compatible with, so `npm install` fails.
image

### Action items

- [ ] Update peer deps react version in package.json ‼️
- [ ] Update react version in dev deps (will need to use sb 6.5 and testing lib 13)
- [ ] Update any components that are using old APIs (e.g. `ReactDOM.render` in `Toaster.tsx`)
- [ ] Check components in new `StrictMode` (e.g. in vite test app)

veekeys:
Updated to react 18.
Still need to do some testing (with Vite app). I still get warning in there, might be connected that the code has old `render` already, no matter if it is actually called... But on every call it is not shown using this solution, while without it, it shows on every toaster call.
Added kinda weird solution to toaster, just I think having it through `setSettings` might not work well, cause this is set globally and need to think of an option to reset it to undefined again.
Had this on pre commit hook, so went with `--no-verify` to commit
image

UPDATE:
As I was fighting different issues accidentally ended up upgrading few more libs (only jest and testing-library).

- `userEvent` ended up having all APIs returning a promise, so.. Needed to go through all and add await
https://github.com/testing-library/user-event/releases/tag/v14.0.0
- Some tests were failing with html element click, so replaced those with `userEvent`
- Got rid of `@testing-library/react-hooks`, cause now testing-library has all it provided. https://github.com/testing-library/react-hooks-testing-library#a-note-about-react-18-support
- Added some inline comments for interesting places.


packages/iTwinUI-react/src/types/toaster.ts
```diff
@@ -0,0 +1,4 @@
+export interface Root {
```
mayank99, line 1: Why not just keep this in the `Toaster` file? I think the `types/` directory is only for declarations (`.d.ts` files)

packages/iTwinUI-react/src/types/custom.d.ts
```diff
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 declare module '*.svg' {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const content: any;
```
mayank99, line 7: Might be possible to use `unknown` here. But minor because this is removed in #660

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -30,6 +31,11 @@ export type ToasterSettings = {
     | 'bottom'
     | 'bottom-start'
     | 'bottom-end';
+  /**
+   * If using React v18, use`ReactDom.createRoot`function to return root and enable v18.
+   * Otherwise, old `ReadDom.render` method will be used to render toasts.
```
mayank99, line 36: Typos
```suggestion
   * If using React v18, use `ReactDOM.createRoot` function to return root and enable v18.
   * Otherwise, old `ReactDOM.render` method will be used to render toasts.
```

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -30,6 +31,11 @@ export type ToasterSettings = {
...
+   */
+  getRoot?: (container: Element | undefined) => Root;
```
mayank99, line 38: Why is this a function instead of just `Root`?

Also, maybe this could be named better. `getRoot` doesn't make it obvious root for what. And the JSDoc also doesn't explain. But I don't have a better suggestion yet.

packages/iTwinUI-react/src/core/Toast/Toast.tsx
```diff
@@ -177,11 +177,11 @@ export const Toast = (props: ToastProps) => {
       in={visible}
       appear={true}
       unmountOnExit={true}
-      onEnter={(node) => {
+      onEnter={(node: HTMLElement) => {
```
mayank99, line 180: Why is this needed? Inline functions should infer types

packages/iTwinUI-react/stories/core/Toasts.stories.tsx
```diff
@@ -98,6 +98,17 @@ export default {
...
   },
   title: 'Core/Toasts',
+  decorators: [
+    (Story) => {
+      toaster.setSettings({
+        getRoot: (container) =>
+          ReactDOM.createRoot(
+            container ?? document.getElementsByTagName('div')[0],
+          ),
+      });
```
mayank99, line 108: Shouldn't we put this in the story itself instead of hiding in the decorator?

mayank99 (commented):
I think PR title would need to be `feat` because this is _adding_ react 18 support and also _adding_ a new prop to Toaster.

Also, I think there is something wrong with TS still, because build is failing.

packages/iTwinUI-react/package.json
```diff
@@ -89,26 +89,27 @@
     "fast-glob": "^3.2.5",
     "husky": "^4.0.0",
     "inquirer": "^6.2.2",
-    "jest": "^27.0.6",
-    "jest-cli": "^27.0.6",
-    "jest-junit": "^12.0.0",
+    "jest": "^28.1.0",
+    "jest-cli": "^28.1.0",
+    "jest-environment-jsdom": "^28.1.0",
```
veekeys, line 94: https://jestjs.io/docs/28.0/upgrading-to-jest28#jsdom

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -321,25 +326,27 @@ it('should handle row clicks', () => {
...
 
-  userEvent.click(getByText(mockedData()[2].name));
+  await userEvent.click(getByText(mockedData()[2].name));
   expect(rows[1].classList).not.toContain('iui-selected');
   expect(rows[2].classList).toContain('iui-selected');
   expect(onSelect).toHaveBeenCalledTimes(2);
   expect(onRowClick).toHaveBeenCalledTimes(2);
 
-  userEvent.click(getByText(mockedData()[1].name), { ctrlKey: true });
+  const user = userEvent.setup();
```
veekeys, line 340: https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -890,19 +897,16 @@ it('should not trigger sorting when filter is clicked', () => {
...
 
-it('should render filter dropdown in the correct document', () => {
-  const mockDocument = new DOMParser().parseFromString(
-    `<!DOCTYPE html><html><body><div></div></body></html>`,
-    'text/html',
-  );
-  const mockContainer = mockDocument.querySelector('div') as HTMLDivElement;
-
+it('should render filter dropdown in the correct document', async () => {
+  const mockDocument = document.implementation.createHTMLDocument();
```
veekeys, line 907: It was failing for me before... This helped. I have verified it by getting rid of `ownerDocument` in filter and this test started to fail.

packages/iTwinUI-react/src/core/Toast/Toast.tsx
veekeys: no idea, but I was getting errors and only on those 2 :D

packages/iTwinUI-react/src/core/Wizard/Wizard.test.tsx
```diff
@@ -234,15 +234,15 @@ describe('<Wizard />', () => {
 
     expect(document.querySelector('.iui-tooltip')).toBeNull();
     expect(screen.queryByText('Step one tooltip')).toBeNull();
-    userEvent.hover(screen.getByText('Step One'));
+    fireEvent.mouseEnter(screen.getByText('Step One'), { bubbles: true });
```
veekeys, line 237: We hover over span, but only parent `li` has the popover to trigger tooltip.
`userEvent` didnt have (or I didnt find) bubbling option, so used `fireEvent` here instead.

packages/iTwinUI-react/src/types/toaster.ts
veekeys: got rid of it

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: removed

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: removed

packages/iTwinUI-react/stories/core/Toasts.stories.tsx
veekeys: no more story

veekeys:
@mayank99 @elephantcatdog @bentleyvk I have changed toaster a little to have support for react 18 included.
I'm not planning to make any more changes, so it's ready for review.

yarn.lock
```diff
@@ -10876,77 +10991,92 @@ postcss@8, postcss@^7.0.14, postcss@^7.0.26, postcss@^7.0.32, postcss@^7.0.36, p
...
 
-prettier@2.2.1, "prettier@>=2.2.1 <=2.3.0":
+prettier@2.2.1:
   version "2.2.1"
-  resolved "https://registry.yarnpkg.com/prettier/-/prettier-2.2.1.tgz#795a1a78dd52f073da0cd42b21f9c91381923ff5"
+  resolved "https://registry.npmjs.org/prettier/-/prettier-2.2.1.tgz#795a1a78dd52f073da0cd42b21f9c91381923ff5"
   integrity sha512-PqyhM2yCjg/oKkFPtTGUojv7gnZAoG80ttl45O6x2Ug/rMJw4wcc9k6aaf2hibP7BGVCCM33gZoGjyvt9mm16Q==
 
+"prettier@>=2.2.1 <=2.3.0":
+  version "2.3.0"
```
mayank99, line 11008: This got messed up again so I'm seeing random formatting changes when I save in vscode

veekeys: I had a bunch of conflicts in lock file, so I have regenerated it from scratch.
What is the expected version here to be used?

mayank99: needs to be 2.2.1. see #651

veekeys: Why the heck it takes this version if we explicitly said `2.2.1` ?? Is something not working?

mayank99: you regenerated lockfile so it makes sense that it would try to get the latest compatible version. The `>=2.2.1 <=2.3.0` is coming from storybook

mayank99:
I'm getting this error in browser console when I open storybook.

```console
The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type"
```

It doesn't really affect us because we are not server rendering storybook, but it is polluting the entire browser console. Any way to suppress it?
image

yarn.lock
veekeys: This is something I am not aware about versions then
I thought, if you say `prettier: 2.2.1` the only version it is going to get for you is `2.2.1`. Now how the heck it decides, that `2.3.0` is compatible? Then why we use caret symbols at all.

mayank99: well, it has both `2.2.1` and `2.3.0` now. See the PR I linked above for more explanation and fix

veekeys:
> I'm getting this error in browser console when I open storybook.
> 
> ```
> The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type"
> ```
> 
> It doesn't really affect us because we are not server rendering storybook, but it is polluting the entire browser console. Any way to suppress it? image

Yeah, I was checking it and somehow tohught prod build should not have it, but forgot to test.. I will see, if I find anything.. 
I think I saw somewhere, that this might be coming from storybook itself (it dedected first-child selector in its own css)

yarn.lock
veekeys: well ok..
to be honest, we should have a resolution for it instead, but reverted..

mayank99: yeah that makes sense to me. You could even sneak it into this PR

packages/iTwinUI-react/src/core/Select/Select.test.tsx
```diff
@@ -230,12 +231,14 @@ it.each(['Enter', ' ', 'Spacebar'])(
     let menu = document.querySelector('.iui-menu') as HTMLUListElement;
     expect(menu).toBeFalsy();
 
-    fireEvent.keyDown(
-      select.querySelector('.iui-select-button') as HTMLElement,
-      {
-        key,
-      },
-    );
+    act(() => {
```
bentleyvk, line 234: Still don't get when `act` is needed and when it is not 😀

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -220,28 +220,33 @@ const BooleanFilter = (
...
+const expandAll = async (
+  container: HTMLElement,
+  oldExpanders: Element[] = [],
+) => {
   const allExpanders = Array.from(
     container.querySelectorAll('.iui-row-expander'),
   );
   const newExpanders = allExpanders.filter((e) => !oldExpanders.includes(e));
-  newExpanders.forEach((button) => userEvent.click(button));
+  for (const button of newExpanders) {
```
bentleyvk, line 231: Was it you or prettier?

packages/iTwinUI-react/src/core/Table/filters/DateRangeFilter/DateRangeFilter.test.tsx
```diff
@@ -43,7 +43,8 @@ it('should render correctly with set filter value', () => {
   const { container } = renderComponent({
     column: {
       filterValue: [new Date(2021, 4, 1), new Date(2021, 4, 3)],
-    } as HeaderGroup,
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as HeaderGroup<any>,
```
bentleyvk, line 47: `unknown` wouldn't work?

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,37 @@ export default class Toaster {
...
+  private async createContainer() {
+    const container = getContainer(TOASTS_CONTAINER_ID) ?? getDocument()?.body;
+    if (!container) {
+      return;
+    }
+
+    const toastWrapper = <ToastWrapper ref={this.toastsRef} />;
+
+    try {
+      if (Number(React.version.split('.')[0]) > 17) {
```
bentleyvk, line 119: Have you checked whether this is working in prod builds in 17 and 18?

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,37 @@ export default class Toaster {
     this.updateView();
   }
 
+  private async createContainer() {
```
bentleyvk, line 110: Maybe add a comment that this is done for SSR/SSG that we wouldn't make the same mistake again.

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,37 @@ export default class Toaster {
...
+    const container = getContainer(TOASTS_CONTAINER_ID) ?? getDocument()?.body;
+    if (!container) {
+      return;
+    }
+
+    const toastWrapper = <ToastWrapper ref={this.toastsRef} />;
+
+    try {
```
bentleyvk, line 118: Add a comment that this is for 17 and 18 support.

packages/iTwinUI-react/src/core/Select/Select.test.tsx
veekeys: when you get warning, it is needed :D

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -1066,7 +1072,7 @@ it('should rerender table when columns change', () => {
             {
               id: 'name',
               Header: 'Name',
-              Cell: () => 'test1',
+              Cell: () => <>test1</>,
```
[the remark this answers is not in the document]
veekeys: It is not really a `ReactNode` in there.. It was not valid before, just somehow TS was not catching it back then :/

packages/iTwinUI-react/src/core/Table/Table.test.tsx
veekeys: It was me, because I could not write an async foreach

packages/iTwinUI-react/src/core/Table/filters/DateRangeFilter/DateRangeFilter.test.tsx
veekeys: image

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: I have checked only dev builds

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: added

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: added

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: ok, great catch!
Does not work neither in vite, nor in cra apps.. :(

mayank99: will `try`/`catch` work instead? like if you `await import` from `react-dom/client` on react 17, does it throw an error?

veekeys: It fails on build time...doesnt matters this code..

mayank99: What's the error? I think there's a way to hack around it at build time.

veekeys: image

aruniverse: found out last week dynamic imports are resolved at build time. with webpack you can ignore it with magic comments, not sure about vite

veekeys: Rollup also has additional config option, which might help, but it is not a solution for us, imo..
We would break current setup of the apps, so that technically should come with a major release.

mayank99: looks like esbuild is the only one that handles this in a sensible way
https://github.com/evanw/esbuild/commit/6be8e44beba8e765c2b99683608540ce40d6a58f

veekeys: Ok, I have tested the solution we have in:
- Rollup build with Vite app
- Webpacke build CRA app
- next.js app

I have tried both dev and prod builds and stuff seemed to work fine. Also no warnings were displayed in v18 (that old render is being used).
If someone is still up to, I would appreciate if you could run one more time through Vite or CRA app (separate one, not the playground we have, just to be sure), so that my local builds were not messed up..
But it looks for now we can go with this change and hopefully we will refactor fully toasts to get rid of render in future major releases, before react decided to get rid of their barrel export of `createRoot`.

Thank you @mayank99 for the hacky inspiration, but I think looks much better than antd one 😄

mayank99 (commented):
> > > There is this Carousel controlled story, which is failing... It does not keep the manually set slide. I might need @mayank99 insights on certain part of the code..
> 
> I spent a few hours trying to play around with different permutations of intersection observers, scroll events, effects, callbacks, etc, but could _not_ get all three cases to work with the same code:
> 
> * uncontrolled with manual slide change (e.g. clicking next/prev or pressing ←/→ )
> * controlled
> * automatic slide change (e.g. dragging on mobile, or focusing an element inside a slide)

Update: I was able to fix the Carousel after playing around some more. It works with all three cases now.

Not sure if we want to make it part of this PR so I pushed it to a separate branch: https://github.com/iTwin/iTwinUI-react/commit/bd7c8a81747292141200552d852fc13ea45a804e

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,40 @@ export default class Toaster {
...
+    } else {
+      // v17 and before
+      ReactDOM.render(toastWrapper, container);
+    }
+  }
+
   private updateView() {
-    this.toastsRef.current?.setToasts(this.toasts);
+    if (!this.isInitialized) {
+      this.createContainer().then(() => {
```
bentleyvk, line 137: Why `.then` if there is no awaits in `createContainer`?

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
bentleyvk: I tested on CRA that has v17 and on Vite with v18 and no errors in dev or in prod builds.

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,40 @@ export default class Toaster {
...
+    }
+
+    const toastWrapper = <ToastWrapper ref={this.toastsRef} />;
+
+    const _ReactDOM = ReactDOM as any; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+    // v18 mode
+    if (_ReactDOM.createRoot) {
+      // suppress warning about importing createRoot from react-dom/client
+      _ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.usingClientEntryPoint = true;
```
bentleyvk, line 125: Don't we need to bring back original value after the import?

bentleyvk: Also, will they you be fired for this? 😯

apps/playground/src/main.tsx
```diff
@@ -65,9 +65,8 @@ const ThemeButton = styled.button`
   }
 `;
 
-ReactDOM.render(
+createRoot(document.getElementById('root')!).render(
```
bentleyvk, line 68: eslint didn't say anything about using `!`?

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: This is very interesting... Need to leave `createContainer` as async, otherwise toasts do not work on initial render (on first call).
My guess is it is connected with react life cycles and initial render of component. Without wrapping it as Promise (putting the execution as the last one in queue), it was setting the list of the toasts before it even got mounted and as a result, first call on `toaster` would render nothing, but then second one would show you 2 toasts instantly.

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: yes, thats the exact thing I was thinking whole night after making this change..
I think we need to set it to false, cause otherwise, all apps using iTwinUI will never see any warnings :D

apps/playground/src/main.tsx
veekeys: seems like not..

packages/iTwinUI-react/package.json
```diff
@@ -52,56 +52,56 @@
...
     "ts-loader": "^9.2.8",
     "ts-node": "^8.0.2",
     "typescript": "^4.4.3",
     "webpack": "5"
   },
   "peerDependencies": {
-    "react": "^16.8.6 || ^17.0.0",
-    "react-dom": "^16.8.6 || ^17.0.0"
+    "react": ">=16.8.6",
+    "react-dom": ">=16.8.6"
```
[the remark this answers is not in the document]
veekeys: hmm..changed to a range of versions

packages/iTwinUI-react/package.json
```diff
@@ -52,56 +52,56 @@
...
-    "@storybook/api": "~6.4.19",
-    "@storybook/builder-webpack5": "~6.4.19",
-    "@storybook/components": "~6.4.19",
-    "@storybook/core-events": "~6.4.19",
-    "@storybook/manager-webpack5": "~6.4.19",
+    "@storybook/addon-a11y": "^6.5.0-beta.1",
+    "@storybook/addon-essentials": "~6.5.0-beta.1",
+    "@storybook/addon-storysource": "~6.5.0-beta.1",
+    "@storybook/addons": "~6.5.0-beta.1",
+    "@storybook/api": "~6.5.0-beta.1",
```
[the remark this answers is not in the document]
veekeys: uu, nice. updated

yarn.lock
```diff
@@ -3,27 +3,28 @@
 
 
 "@ampproject/remapping@^2.1.0":
-  version "2.1.2"
-  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.1.2.tgz#4edca94973ded9630d20101cd8559cedb8d8bd34"
-  integrity sha512-hoyByceqwKirw7w3Z7gnIIZC3Wx3J484Y3L/cMpXFbr7d9ZQj2mODrirNzcJa+SM3UlpWXYvKV4RlRpFXlWgXg==
+  version "2.2.0"
+  resolved "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
```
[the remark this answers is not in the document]
veekeys: fixed

packages/iTwinUI-react/package.json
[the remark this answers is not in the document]
veekeys: I just dont like this
image

Tried with `||` but got same.. So I guess syntax is fine, just something is not liked there :/

veekeys:
> > > > There is this Carousel controlled story, which is failing... It does not keep the manually set slide. I might need @mayank99 insights on certain part of the code..
> > 
> > 
> > I spent a few hours trying to play around with different permutations of intersection observers, scroll events, effects, callbacks, etc, but could _not_ get all three cases to work with the same code:
> > 
> > * uncontrolled with manual slide change (e.g. clicking next/prev or pressing ←/→ )
> > * controlled
> > * automatic slide change (e.g. dragging on mobile, or focusing an element inside a slide)
> 
> Update: I was able to fix the Carousel after playing around some more. It works with all three cases now.
> 
> Not sure if we want to make it part of this PR so I pushed it to a separate branch: [bd7c8a8](https://github.com/iTwin/iTwinUI-react/commit/bd7c8a81747292141200552d852fc13ea45a804e)

Would you mind creating a separate PR? I wanted to, but I saw you have all react 18 changes in there..
Would be nice to merge that before react 18, just to be sure.

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: I reverted suppression just after we use `createRoot`...
Hopefully, that does not influence users app now

veekeys: Tested in Vite app by using `createRoot` in there and it showed warning

yarn.lock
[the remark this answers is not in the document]
veekeys: yup, fixed

packages/iTwinUI-react/package.json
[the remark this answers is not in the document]
mayank99: where is this coming from? I wonder if it's because of deps vs dev deps.

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
mayank99: Maybe we should leave the warning and ask users to suppress it if they complain?

packages/iTwinUI-react/package.json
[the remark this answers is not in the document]
veekeys: I was just runing yarn after all the changes.. At first I tohught its because of wrong syntax, I had..But it had same with the `||`, so I am not sure..
When message does not show full path, kinda hard to guess..

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
mayank99: > Tested in Vite app by using `createRoot` in there and it showed warning

Maybe I misunderstood this one. Just want to double confirm, it shows a warning when `createRoot` from `react-dom` in client app and not when importing it from `react-dom/client`, correct?

And have you verified that the behavior remains the same with/without this new revert?

veekeys: > Maybe I misunderstood this one. Just want to double confirm, it shows a warning when `createRoot` from `react-dom` in client app and not when importing it from `react-dom/client`, correct?
> 
Yes, it just showed for importing in from wrong path in CLIENT APP (so not itwinui code)

> And have you verified that the behavior changes with/without this new revert? Without this revert, there was no warning in the user's app when importing `createRoot` from `react-dom`?
>

I have now and it shows no error :)
I mean.. if `createRoot` is called AFTER our toaster (cause only then we disable warning). So the fix there helps now.

mayank99 (commented):
> > Update: I was able to fix the Carousel after playing around some more. It works with all three cases now.
> > Not sure if we want to make it part of this PR so I pushed it to a separate branch: [bd7c8a8](https://github.com/iTwin/iTwinUI-react/commit/bd7c8a81747292141200552d852fc13ea45a804e)
> 
> Would you mind creating a separate PR? I wanted to, but I saw you have all react 18 changes in there.. Would be nice to merge that before react 18, just to be sure.

Created https://github.com/iTwin/iTwinUI-react/pull/683.

veekeys:
> > I'm getting this error in browser console when I open storybook.
> > ```
> > The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It doesn't really affect us because we are not server rendering storybook, but it is polluting the entire browser console. Any way to suppress it? image
> 
> Yeah, I was checking it and somehow tohught prod build should not have it, but forgot to test.. I will see, if I find anything.. I think I saw somewhere, that this might be coming from storybook itself (it dedected first-child selector in its own css)

Issue to follow: https://github.com/storybookjs/storybook/issues/18103

apps/playground/src/main.tsx
bentleyvk: Oh, for some reason demo app doesn't have eslint config...

packages/iTwinUI-react/package.json
```diff
@@ -52,56 +52,56 @@
...
+    "ts-jest": "^28.0.2",
     "ts-loader": "^9.2.8",
     "ts-node": "^8.0.2",
     "typescript": "^4.4.3",
     "webpack": "5"
   },
   "peerDependencies": {
-    "react": "^16.8.6 || ^17.0.0",
-    "react-dom": "^16.8.6 || ^17.0.0"
+    "react": ">=16.8.6 < 19.0.0",
```
bentleyvk, line 103: Looks odd then there is no v19 yet

mayank99: see https://github.com/iTwin/iTwinUI-react/pull/656#discussion_r878144263

packages/iTwinUI-react/stories/core/Toasts.stories.tsx
```diff
@@ -15,7 +15,6 @@ export default {
     content: {
       description:
         'Content of the Toast message. Can be passed in as a string or a jsx element.',
-      type: { required: true },
```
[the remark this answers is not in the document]
veekeys: yeah, now if you specify that its required, you also need to add the name and value
image

I added it back and said its string.. should be fine

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -115,8 +107,42 @@ export default class Toaster {
...
+
+      const root = _ReactDOM.createRoot(container);
+      root.render(toastWrapper);
+      // revert suppression, not to influence users app
+      _ReactDOM.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.usingClientEntryPoint = false;
```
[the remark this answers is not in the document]
veekeys: done

packages/iTwinUI-react/src/types/custom.d.ts
mayank99: These can be removed now

mayank99 (approved):
Looks good 👍

packages/iTwinUI-react/src/types/custom.d.ts
veekeys: you mean whole file?

veekeys: removed

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -49,22 +49,42 @@ export default class Toaster {
 
   // Create container on demand.
   // Cannot do it in constructor, because SSG/SSR apps would fail.
-  private asyncInit = new Promise<void>((resolve) => {
+  private asyncInit = async () => {
     if (this.isInitialized) {
-      resolve();
```
bentleyvk, line 54: Are those all `resolve` not needed anymore? I thought there was a bug and you needed them.

veekeys: I had Promise before, but with that did not work.. So came back to async function.

bentleyvk (approved):

yarn.lock
mayank99: @veekeys This got messed up again

veekeys: Added resolution...
Dare to mess up now

packages/iTwinUI-react/src/core/Table/TablePaginator.tsx
```diff
@@ -246,7 +246,8 @@ export const TablePaginator = (props: TablePaginatorProps) => {
 
   const hasNoRows = totalPagesCount === 0;
   const showPagesList = totalPagesCount > 1 || isLoading;
-  const showPageSizeList = pageSizeList && onPageSizeChange && !!totalRowsCount;
+  const showPageSizeList =
+    pageSizeList && !!onPageSizeChange && !!totalRowsCount;
```
mayank99, line 250: this formatting might need to be reverted

veekeys: How? Save without formatting?

veekeys: And it puts it same way...

mayank99 (approved):

PR #667, opened 2022-05-11
fix(toaster): Do not call `ReactDOM` render multiple times

veekeys:
I have changed approach a little in toaster not to call `render` multiple times (as this is not the recommended way to do ).
I changed that mostly because of my v18 work, where tests with new root render were not passing for some reason.
With this change toasts will use react change detection and rerender as state of the component will change.
Using fancy [useImperativeHandle](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) hook.

### DIFF
--- a/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
+++ b/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
@@ -3,11 +3,10 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
-import { render } from '@testing-library/react';
+import { act, render } from '@testing-library/react';
 
 import { ToastWrapper, ToastWrapperHandle } from './ToastWrapper';
 import { ToastProps } from './Toast';
-import { act } from 'react-dom/test-utils';
 
 const mockToastObject1 = {
   category: 'informational',
--- a/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
+++ b/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
@@ -30,7 +30,7 @@ beforeEach(() => {
   ref = React.createRef<ToastWrapperHandle>();
 });
 
-it.only('should render toasts', () => {
+it('should render toasts', () => {
   const { container } = render(<ToastWrapper ref={ref} />);
   expect(container.querySelector('.iui-toast-wrapper')).toBeTruthy();
 
--- a/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
+++ b/packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
@@ -5,7 +5,7 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 
-import { ToastWrapper } from './ToastWrapper';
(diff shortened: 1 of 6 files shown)


packages/iTwinUI-react/src/core/Toast/Toaster.tsx
```diff
@@ -55,6 +65,7 @@ export default class Toaster {
       ? 'ascending'
       : 'descending';
     this.settings = newSettings;
+    this.toastsRef?.current?.setPlacement(this.settings.placement ?? 'top');
```
bentleyvk, line 68: DO you need first `?`?

packages/iTwinUI-react/src/core/Toast/Toaster.test.tsx
```diff
@@ -2,6 +2,7 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+import { act } from 'react-dom/test-utils';
```
bentleyvk, line 5: Why imported not from testing-library?

packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
```diff
@@ -5,7 +5,7 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 
-import { ToastWrapper } from './ToastWrapper';
+import { ToastWrapper, ToastWrapperHandle } from './ToastWrapper';
 import { ToastProps } from './Toast';
 import { act } from 'react-dom/test-utils';
```
bentleyvk, line 10: Why act is imported from here???

packages/iTwinUI-react/src/core/Toast/Toaster.tsx
veekeys: no

packages/iTwinUI-react/src/core/Toast/Toaster.test.tsx
veekeys: I'm just gonna quote this
https://github.com/iTwin/iTwinUI-react/pull/623#discussion_r871357029

But removed all use cases of that import..

packages/iTwinUI-react/src/core/Toast/ToastWrapper.tsx
```diff
@@ -8,14 +8,31 @@ import cx from 'classnames';
...
+};
+
+export const ToastWrapper = React.forwardRef<ToastWrapperHandle>((_, ref) => {
+  const [toasts, setToasts] = React.useState<ToastProps[]>([]);
+  const [placement, setPlacement] = React.useState<ToastPlacement>('top');
   const placementPosition = placement.startsWith('top') ? 'top' : 'bottom';
 
+  React.useImperativeHandle(
+    ref,
+    () => ({
```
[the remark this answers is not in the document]
veekeys: changed

packages/iTwinUI-react/src/core/Toast/ToastWrapper.test.tsx
veekeys: But removed all use cases of that import..

adamhock (approved):
Looks good to me :tada:

bentleyvk (approved):

PR #668, opened 2022-05-11
chore(deps): Update deps and move resolutions to root

mayank99:
- Replaced `cpx` with `cpy-cli`. Installed in root package.json because it might be used by multiple packages.
- Also moved `rimraf` to root for the same reason
- Removed `mkdirp` as it's not used anymore
- Moved `resolutions` to root as previously it had no effect
- Went through all `resolutions` and removed unneeded ones

This should help us get rid of the dependabot warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "lint-staged": "turbo run lint-staged"
   },
   "devDependencies": {
-    "cpy-cli": "^4.1.0",
+    "cpy-cli": "^3.1.1",
     "rimraf": "^3.0.2",
     "turbo": "~1.2.4",
     "webpack": "5"
--- a/package.json
+++ b/package.json
@@ -15,8 +15,18 @@
     "lint-staged": "turbo run lint-staged"
   },
   "devDependencies": {
+    "cpy-cli": "^4.1.0",
+    "rimraf": "^3.0.2",
     "turbo": "~1.2.4",
     "webpack": "5"
+  },
+  "resolutions": {
+    "trim": "^0.0.3",
+    "glob-parent": "^5.1.2",
+    "set-value": "^4.1.0",
+    "ansi-regex": "^5.0.1",
+    "postcss": "8",
+    "nanoid": "^3.1.31"
   },
   "engines": {
     "npm": ">=7.0.0",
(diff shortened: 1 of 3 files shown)


mayank99:
Looks like `cpy` is not working, even though I thought I tested it. Will move it to draft to investigate tomorrow.

packages/iTwinUI-react/package.json
```diff
@@ -40,7 +40,7 @@
     "lint": "eslint \"**/*.{js,ts,tsx}\" --max-warnings=0",
     "lint:fix": "yarn lint --fix && yarn lint:copyright --fix */**/*.{js,ts,tsx}",
     "lint:copyright": "node scripts/copyrightLinter.js",
-    "copy-files": "cpx \"../../{README,LICENSE}.md\" .",
+    "copy-files": "cpy \"../../{README,LICENSE}.md\" .",
```
bentleyvk, line 43: Try `cpy ../../{README,LICENSE}.md .`

package.json
```diff
@@ -15,9 +15,19 @@
     "lint-staged": "turbo run lint-staged"
   },
   "devDependencies": {
+    "cpy-cli": "^3.1.1",
+    "rimraf": "^3.0.2",
     "turbo": "~1.2.4",
     "webpack": "5"
   },
+  "resolutions": {
```
veekeys, line 23: List is shorter.
You have removed those which are not needed?

packages/iTwinUI-react/package.json
```diff
@@ -94,10 +93,8 @@
     "jest-junit": "^12.0.0",
     "lint-staged": "^12.1.2",
     "markdown-to-jsx": "6.11.4",
-    "mkdirp": "^1.0.4",
```
veekeys, line 97: not used anywhere?

package.json
mayank99: yup

packages/iTwinUI-react/package.json
mayank99: nope

veekeys (approved):
hope it works

PR #669, opened 2022-05-12
chore: Move pre-commit hook to root

mayank99:
- Moved `husky` and `lint-staged` dependencies to root package.json.
  - Updated `husky` to latest version, so config lives in `.husky` folder instead of in `package.json`.
- Moved `copyrightLinter.js` into `configs` package so that it can be shared across workspaces.
  - Moved `lint:copyright` command to root and updated `build.yml` as well.
  - Also hardcoded fallback pattern in the script itself to fix CI not linting all files.
- Root `lint-staged` config runs prettier and copyright linter because we need those everywhere.
- itwinui-react workspace overrides root `lint-staged` config to add `eslint` to the hook.

Tested working on macOS. Would be good to have multiple people confirm it works on Windows, as latest `husky` might have some differences across platforms.


.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - run: yarn audit
 
       - run: yarn lint
-      - run: yarn workspace @itwin/itwinui-react lint:copyright ${{ github.workspace }}/*/**.{js,ts,tsx}
+      - run: yarn lint:copyright ${{ github.workspace }}/*/**.{js,ts,tsx}
```
mayank99, line 29: Looks like this is broken (even before this PR). It's not checking all files
image

mayank99: Adding quotes doesn't help
image

bentleyvk: So is it working?

bentleyvk:
Seems that pre-commit is working again!

.github/workflows/build.yml
mayank99: Works after most recent change. I'm hardcoding the pattern inside `copyrightLinter.js`

Tested failure and success cases in CI 👍
https://github.com/iTwin/iTwinUI-react/runs/6476456067
https://github.com/iTwin/iTwinUI-react/runs/6476485106

bentleyvk (approved):

Issue #614, opened 2022-04-04
Resizable and Movable Modal Dialog


jason-crow:
#### Feature
Requesting the Modal Dialog support configuration to enable it to be movable and resizable from user interaction

#### Examples
Example dialog:
![image](https://user-images.githubusercontent.com/25822334/161553336-13f78d52-6e73-47f4-8809-612635d7efdb.png)
 Would want such a dialog to be draggable from the header to move and from the corner drag be resized. This would be an optional configuration  that would also prevent the background overlay from showing

bentleyvk:
It is finally implemented in [v1.46.0](https://github.com/iTwin/iTwinUI-react/releases/tag/v1.46.0) 🚀

PR #670, opened 2022-05-13
chore(Dialog): Created `Dialog` subcomponents and reused in `Modal`

bentleyvk:
Preparation for #614

Created `Dialog` and its subcomponents that are reused inside `Modal` so no API changes.

Visual tests are failing because there are breaking CSS changes in other components that needs to be fixed in other PR.

## Checklist

- [X] Add meaningful unit tests for your component (verify that all lines are covered)


packages/iTwinUI-react/src/core/Dialog/DialogButtonBar.test.tsx
```diff
@@ -6,29 +6,31 @@ import { render, screen } from '@testing-library/react';
 import React from 'react';
 import { Button } from '../Buttons';
```
bentleyvk, line 8: For some reason Content and ButtonBar are switched and showing strange diffs

gretanausedaite (commented):
Are we planning to add `Dialog` story?

bentleyvk:
> Are we planning to add `Dialog` story?

`Dialog` is even exported in `index.ts` for now

packages/iTwinUI-react/src/core/ColorPicker/ColorSwatch.tsx
```diff
@@ -17,7 +17,7 @@ export type ColorSwatchProps = {
    * Is color selected.
    */
   isActive?: boolean;
-} & Omit<React.ComponentPropsWithRef<'div'>, 'color'>;
+} & Omit<React.ComponentPropsWithoutRef<'div'>, 'color'>;
```
bentleyvk, line 20: Otherwise it takes `css` from somewhere (I guess storybook). Checked whether it still allows to pass `ref`

packages/iTwinUI-react/src/core/Modal/index.ts
```diff
@@ -4,8 +4,11 @@
...
-export { ModalButtonBar } from './ModalButtonBar';
-export type { ModalButtonBarProps } from './ModalButtonBar';
-export { ModalContent } from './ModalContent';
-export type { ModalContentProps } from './ModalContent';
+
+export { DialogButtonBar as ModalButtonBar } from '../Dialog/DialogButtonBar';
+export type { DialogButtonBarProps as ModalButtonBarProps } from '../Dialog/DialogButtonBar';
+
+export { DialogContent as ModalContent } from '../Dialog/DialogContent';
+export type { DialogContentProps as ModalContentProps } from '../Dialog/DialogContent';
```
mayank99, line 12: This is weird. Shouldn't we mark the old ones as deprecated?

Like this:
```tsx
/** @deprecated - Use `DialogButtonBar` instead. */
export const ModalButtonBar = DialogButtonBar;
```

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { DialogBackdrop } from './DialogBackdrop';
+import { DialogTitleBar } from './DialogTitleBar';
+import { DialogContent } from './DialogContent';
+
+export type DialogProps = {
+  /**
+   * Content of the dialog.
+   */
+  children: React.ReactNode;
+} & Omit<CommonProps, 'title'>;
```
mayank99, line 18: I think we should expose all native element props in every component, i.e. `& React.ComponentProps<'div'>` instead of `& Omit<CommonProps, 'title'>`

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
```diff
@@ -0,0 +1,209 @@
...
+   * Document where the dialog will be rendered.
+   * Can be specified to render in a different document (e.g. a popup window).
+   * @default document
+   */
+  ownerDocument?: Document;
+  /**
+   * Type of the dialog.
+   * @default 'default'
+   */
+  styleType?: 'default' | 'fullPage';
```
mayank99, line 62: I know it's not exposed yet but we might want to rename this prop. I'm not sure why this prop exists on backdrop (should be on `Dialog` only) but it will cause confusion when we add draggable dialog (viewport and relative cases).

packages/iTwinUI-react/src/core/Modal/Modal.tsx
```diff
@@ -95,137 +49,30 @@ export type ModalProps = {
...
-    )
-  ) : (
-    <></>
+  return (
+    <Dialog.Backdrop
+      isDismissible={isDismissible}
+      dialogRootId={modalRootId}
+      {...rest}
+    >
+      <Dialog style={style} id={id} aria-modal>
```
mayank99, line 69: api looks kinda weird. `Dialog.Backdrop` is a parent of the main `Dialog` component?

gretanausedaite: It's the same in css `iui-dialog-backdrop`-> `iui-dialog`

mayank99: The user won't care about our CSS classnames. My concern here is just about the potentially confusing.

I was looking for other examples and found these:
- [`radix-ui`](https://www.radix-ui.com/docs/primitives/components/dialog#anatomy) uses `Dialog.Overlay` and `Dialog.Content` with an additional `Dialog.Root` as the main wrapper.
- [`reach-ui`](https://reach.tech/dialog#dialogoverlay) uses `DialogOverlay` and `DialogContent`.
- [`ariakit`](https://reakit.io/docs/dialog/#backdrop) uses `Dialog` and `DialogBackdrop` (similar to ours but without `.`)
- [`mui`](https://mui.com/material-ui/api/dialog/#props) has `Dialog` and `Backdrop` as separate components and `Dialog` has a `BackdropComponent` prop.

packages/iTwinUI-react/src/core/ColorPicker/ColorSwatch.tsx
bentleyvk: https://github.com/iTwin/iTwinUI-react/runs/6424346603?check_suite_focus=true#step:8:17

packages/iTwinUI-react/src/core/Modal/index.ts
bentleyvk: This PR is just for replacement, I plan to add deprecation tags with another PR where Dialog will be exposed.

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
bentleyvk: The problem is that everything is done trough backdrop in CSS.

packages/iTwinUI-react/src/core/Modal/Modal.tsx
bentleyvk: Yeah, I also don't really like how it is right now but I don't know how to name `Dialog`. Maybe `Dialog.Dialog` but also looks weird, `Dialog.Content` is also taken.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: Exposing

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+ *     <Button styleType='high-visibility'>
+ *       Primary button
+ *     </Button>
+ *     <Button>
+ *       Secondary button
+ *     </Button>
+ *   </Dialog.ButtonBar>
+ * </Dialog>
+ */
+export const Dialog = (props: DialogProps) => {
```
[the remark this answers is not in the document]
bentleyvk: Added

packages/iTwinUI-react/src/core/Modal/index.ts
mayank99: Then don't add deprecation but still `export const ModalButtonBar = DialogButtonBar` would be better. And I think you'd want to preserve the old prop types.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -37,20 +37,28 @@ export type DialogProps = {
...
+        role='dialog'
+        ref={ref}
+        {...rest}
+      >
+        {children}
+      </div>
+    );
+  },
+);
+export const Dialog = Object.assign(DialogComponent, {
```
mayank99, line 58: Creating a separate `const` will lose the jsdoc. You could either get rid of the intermediate `DialogComponent` (see `Carousel` for example) or move the jsdoc down to above `Dialog`.

packages/iTwinUI-react/src/core/Modal/Modal.tsx
mayank99: `Dialog.Main`?

packages/iTwinUI-react/src/core/Modal/index.ts
bentleyvk: What is the difference?

packages/iTwinUI-react/src/core/Modal/Modal.tsx
bentleyvk: But if we intend for the users to use just plain `Dialog` then using some subcomponent kind of feels odd to me.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: I've done that previously but didn't like that there is so many wrappers. But I also don't like moving jsdocs so just wrapped 😀

packages/iTwinUI-react/src/core/Modal/index.ts
mayank99: Well, for props you need to bring back the old types anyway because the new ones use `ComponentPropsWithRef`

And for components, it's still better because It would make them distinct and allow using different jsdoc (so adding deprecation in the future is a single line change). And it's also less confusing for me as there is less "magic" happening in barrel file.

bentleyvk: Changed

packages/iTwinUI-react/src/core/Modal/Modal.tsx
bentleyvk: Yeah, you can use just the `Dialog` but then you need to handle everything by yourself.

bentleyvk: What if we make `Dialog` root component that has almost the same props as `Modal`, but it would have a prop for toggling backdrop or additional type. Because I checked other libs and most of them has a boolean prop for backdrop or if they have a subcomponent, it is placed as a sibling for dialog content.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * </Dialog>
+ */
+export const Dialog = Object.assign(
+  React.forwardRef<HTMLDivElement, DialogProps>((props, ref) => {
+    const { className, children, styleType, isOpen = false, ...rest } = props;
+
+    useTheme();
+
+    return (
+      <CSSTransition
```
bentleyvk, line 57: I am not sure whether to have `CSSTransition` here or not.
Pros: animations come straight out of box.
Cons: maybe it makes `Dialog` not so dumb and reusable component.

mayank99: Didn't we talk about adding a wrapper component? Maybe lets move all the extra logic there?
image

packages/iTwinUI-react/package.json
```diff
@@ -46,7 +46,7 @@
     "build-storybook": "yarn copy-files && build-storybook"
   },
   "dependencies": {
-    "@itwin/itwinui-css": "^0.55.0",
+    "@itwin/itwinui-css": "^0.58.0",
```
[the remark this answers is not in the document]
bentleyvk: That PR was merged long time ago.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: Added a wrapper with a context. I kind of copied from radix-ui. But I am still not sure about that.

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  ownerDocument?: Document;
+} & DialogContextProps &
+  BackdropProps;
+
+export const DialogBackdrop = React.forwardRef<
+  HTMLDivElement,
+  DialogBackdropProps
+>((props, ref) => {
+  const dialogContent = useDialogContext();
```
gretanausedaite, line 24: should this be `dialogContext`?

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
mayank99: It looks like backdrop still has a lot of logic. Would be nice to somehow move all that out of it.

What if we add a `backdrop` prop which accepts the generic `<Backdrop />` component? Then we can handle all the logic in Dialog?

bentleyvk: But then we would have some kind of mix between composition and passing components trough props. Also, what it gives moving logic from `DialogBackdrop` to 'Dialog`?

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
```diff
@@ -0,0 +1,108 @@
...
+    style,
+    ...rest
+  } = props;
+
+  const overlayRef = React.useRef<HTMLDivElement>(null);
+  const refs = useMergedRefs(overlayRef, ref);
+  const previousFocusedElement = React.useRef<HTMLElement | null>();
+
+  // Give focus to overlay for key handling to work.
+  React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
bentleyvk: Idk, for me it feels like it belongs to the `Backdrop` as it covers everything and takes the focus.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
mayank99: I think passing components through props is still somewhat like composition, as it gives the user control over the actual component that gets rendered.

The way I see it, backdrop needs to be a super simple component and we should try to handle as much as logic as we can in the dialog itself. This way, we could even avoid creating a special `Dialog.Backdrop` component and just reuse `Backdrop` (or the user could pass their own).

Plus, I think some of the existing things like keydown handlers and focus should already be moved to the dialog body anyway. I find it really odd that our focus gets "lost" on the backdrop (really hard to see visually). The only thing that where we might need to mess with backdrop is the "outside click", but I think we could even handle that by adding events on the `document`.

Another benefit of this approach would be that it would make it more compatible with the native `<dialog>` element when we decide to support it.

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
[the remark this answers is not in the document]
bentleyvk: If they don't need backdrop, they don't need to include it at all. Or you have something different in mind?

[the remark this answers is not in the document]
gretanausedaite: If users don't include backdrop, they loose keyboard and mouse actions (eg. esc or click outside dialog close)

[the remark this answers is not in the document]
bentleyvk: Fixed that by having props for focus, overflow handling, keyboard and mouse actions in the Dialog.

packages/iTwinUI-react/src/core/Modal/Modal.tsx
bentleyvk: Dropped `Dialog.Main`, now the `Dialog` takes care for that and having `backdrop` prop where users can pass custom backdrop if they wish.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Traps the focus inside the dialog. This is useful when the dialog is modal.
+   * @default false
+   */
+  trapFocus?: boolean;
+  /**
+   * Prevents body from being scrollable. This is useful when the dialog is modal.
+   * @default false
+   */
+  preventBodyScroll?: boolean;
```
bentleyvk, line 56: Not sure about naming this, whether it is clear. I also thought maybe better rename to `preventDocumentScroll` so that `body` wouldn't mistaken with inner dialog body.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: https://github.com/iTwin/iTwinUI-react/pull/670#discussion_r920835897

packages/iTwinUI-react/src/core/ColorPicker/ColorSwatch.tsx
bentleyvk: Somehow it fixed itself 🤷‍♂️

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
```diff
@@ -0,0 +1,47 @@
...
+  DialogContextProps | undefined
+>(undefined);
+
+export const useDialogContext = () => {
+  const context = React.useContext(DialogContext);
+  return {
+    isOpen: false,
+    isDismissible: true,
+    closeOnEsc: true,
+    closeOnExternalClick: false,
```
mayank99, line 44: why are defaults set here?

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -0,0 +1,56 @@
...
+      {isDismissible && (
+        <IconButton
+          size='small'
+          styleType='borderless'
+          onClick={onClose}
+          aria-label='Close'
+        >
+          <SvgClose />
+        </IconButton>
+      )}
```
mayank99, line 51: Do we really need to include this button as part of this component? What happens if the user needs to customize this button (e.g. change styles or the aria-label)?

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Document where the dialog will be rendered.
+   * @default document
+   */
+  ownerDocument?: Document;
```
mayank99, line 61: So we are not creating a portal in this component, which means this doesn't need to be a prop. We can access and save the `ownerDocument` directly from the ref.

packages/iTwinUI-react/src/core/index.ts
```diff
@@ -5,6 +5,9 @@
 export { Alert } from './Alert';
 export type { AlertProps } from './Alert';
 
+export { Backdrop } from './Backdrop';
+export type { BackdropProps } from './Backdrop';
+
```
mayank99, line 10: so we don't want to expose `<Dialog>` component just yet?

packages/iTwinUI-react/src/core/Modal/Modal.tsx
```diff
@@ -62,16 +47,12 @@ export type ModalProps = {
...
   /**
    * Content of the modal.
    */
   children: React.ReactNode;
-} & Omit<CommonProps, 'title'>;
+} & Pick<DialogProps, 'isOpen' | 'styleType'> &
+  Omit<CommonProps, 'title'>;
 
 /**
  * Modal component which can wrap any content.
```
mayank99, line 58: Why did modal test images change?

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { render } from '@testing-library/react';
+import { DialogBackdrop } from './DialogBackdrop';
+
+it('should render in its most basic state', () => {
+  const { container } = render(<DialogBackdrop isVisible />);
+  const backdrop = container.querySelector('.iui-backdrop');
+  expect(backdrop).toBeTruthy();
+});
+
+it('should render with misc props', () => {
```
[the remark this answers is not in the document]
bentleyvk: I kind of did that in `Dialog` tests where testing context.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
bentleyvk: Now I am more concerned about customizing title, like someone wanted to have an icon in it. And this icon should be standartized, so I am pretty ok with it.

packages/iTwinUI-react/src/core/Backdrop/Backdrop.tsx
```diff
@@ -0,0 +1,34 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import { useTheme } from '../utils';
+import '@itwin/itwinui-css/css/backdrop.css';
+
+export type BackdropProps = {
+  /**
+   * Flag whether backdrop should be shown.
+   * @default false
```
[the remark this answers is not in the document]
veekeys: Feels like we better fix css to be able to animate not existing backdrop and get rid of this prop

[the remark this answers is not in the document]
mayank99: ? what do you mean "fix css"? you would still need the prop whether it's true or false by default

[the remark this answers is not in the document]
veekeys: I meant if we could achieve animation with backdrop being not present in the dom.
But if thats too complex, I guess lets leave it. 
I think changing default value might make sense if we think about using backdrop outside of `Dialog`

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: Changed

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+   *   <Dialog.Content>
+   *     Here is my dialog content
+   *   </Dialog.Content>
+   *   <Dialog.ButtonBar>
+   *     <Button styleType='high-visibility'>Confirm</Button>
+   *     <Button>Close</Button>
+   *   </Dialog.ButtonBar>
+   * </Dialog>
+   */
+  children: React.ReactNode;
```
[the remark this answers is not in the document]
bentleyvk: Changed

packages/iTwinUI-react/src/core/Backdrop/Backdrop.tsx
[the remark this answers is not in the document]
bentleyvk: To true

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+  preventBodyScroll?: boolean;
+  /**
+   * Document where the dialog will be rendered.
+   * @default document
+   */
+  ownerDocument?: Document;
+} & DialogContextProps &
+  React.ComponentPropsWithRef<'div'>;
+
+export const Dialog = Object.assign(
```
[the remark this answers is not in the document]
bentleyvk: Added

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+      onClose,
+      onKeyDown,
+      style,
+      className,
+      ...rest
+    } = props;
+
+    // Focuses dialog when opened and brings back focus to the previously focused element when closed.
+    const previousFocusedElement = React.useRef<HTMLElement | null>();
+    React.useLayoutEffect(() => {
```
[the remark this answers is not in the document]
bentleyvk: Changed, haven't seen any blinking so I hope it is ok.

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+      if (isOpen) {
+        previousFocusedElement.current = document.activeElement as HTMLElement;
```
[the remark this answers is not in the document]
bentleyvk: Using `ownerDocument` from `dialogRef`

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import { Backdrop, BackdropProps } from '../Backdrop';
+import { useDialogContext } from './DialogContext';
+
+export type DialogBackdropProps = BackdropProps;
+
+/**
+ * Backdrop component for dialog. Recommended to be used as a child of `Dialog`.
```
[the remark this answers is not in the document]
bentleyvk: Added

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
bentleyvk: Removed

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+              'iui-dialog-full-page': styleType === 'fullPage',
+              'iui-dialog-visible': isOpen,
+            },
+            className,
+          )}
+          role='dialog'
+          ref={refs}
+          onKeyDown={handleKeyDown}
+          tabIndex={-1}
+          style={{ outline: 0, ...style }}
```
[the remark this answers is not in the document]
bentleyvk: Removed

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
bentleyvk: Removeed

packages/iTwinUI-react/src/core/Modal/Modal.tsx
bentleyvk: No idea, tried to move imgs from main branch but then tests failed.

packages/iTwinUI-react/src/core/index.ts
bentleyvk: Maybe the main reason if we see something while implementing draggable dialog.

gretanausedaite (approved):
Looks good to me.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
bentleyvk: Added subcomponent `Dialog.TitleBar.Title` and if `children` passed, then other content is ignored.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -9,48 +9,78 @@ import { useTheme } from '@storybook/theming';
...
   /**
-   * Dialog title bar content.
+   * Dialog title bar content. If passed, then `title` prop is ignored.
    */
-  children: React.ReactNode;
+  children?: React.ReactNode;
+  /**
+   * Dialog title.
+   */
+  title?: React.ReactNode;
```
mayank99, line 22: Should we call this `titleText` or something? To avoid confusion with the native `title` attribute.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -9,48 +9,78 @@ import { useTheme } from '@storybook/theming';
...
-  );
-});
+export const DialogTitleBar = Object.assign(
+  React.forwardRef<HTMLDivElement, DialogTitleBarProps>((props, ref) => {
+    const dialogContext = useDialogContext();
+    const {
+      children,
+      title,
+      isDismissible = dialogContext.isDismissible,
+      onClose = dialogContext.onClose,
```
mayank99, line 50: Does it really need `onClose`? The Dialog should already have it.

I would say, if there is no context then don't even show the close icon.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBarTitle.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * <Dialog.TitleBar>My dialog title</Dialog.TitleBar>
+ */
+export const DialogTitleBarTitle = React.forwardRef<
+  HTMLDivElement,
+  DialogTitleBarTitleProps
+>((props, ref) => {
+  const { children, className, ...rest } = props;
+  useTheme();
+  return (
+    <div className={cx('iui-dialog-title', className)} ref={ref} {...rest}>
```
mayank99, line 29: For supporting multiple elements, we'll still need to improve the CSS to use `margin-right: auto` on this title element instead of `space-between`.

The API won't change but just mentioning this here in case we forget.

bentleyvk: https://github.com/iTwin/iTwinUI-react/pull/670#discussion_r926312491

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
bentleyvk: It is using it for the close icon click handler so I would say it is needed if someone for some reason decides to use it outside context.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
bentleyvk: Renamed

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
mayank99: If someone uses it outside the context, then they should provide their own icon imo.

bentleyvk: idk, it doesn't hurt to have these props if they are supported trough context.

veekeys (approved):

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -0,0 +1,149 @@
...
+        {...rest}
+      >
+        {children}
+      </div>
+    );
+
+    return (
+      <CSSTransition
+        in={isOpen}
+        classNames='iui-dialog-animation'
```
mayank99, line 136: Maybe I'm missing something but this animation doesn't work anymore because it is trying to apply the class to first child it finds (which is the dummy `<div>` placed by FocusTrap). I think it will work if you move the CSSTransition to wrap the `iui-dialog` element.

mayank99: Actually that will also not work because FocusTrap needs a ref from the child, and CSSTransition will not provide that.

mayank99: Maybe we should convert the FocusTrap component into a hook that returns a callback ref which we can pass to `<div class='iui-dialog'>`

bentleyvk: Have you tested it live? Because everything seems to work for me. �
dialog-focus-animation

mayank99: I did a fresh clone just for you and recorded a video showing the broken animation and the class being applied on the wrong element.

https://user-images.githubusercontent.com/9084735/181819183-4d5dec0f-60a7-45a5-8e79-6e9a56df1280.mov

bentleyvk: Oh, on full page it is more obvious. I found a prop on CssTransitionGroup that helped here - `nodeRef`. Now it applies class to correct div:
image

mayank99 (approved):

PR #679, opened 2022-05-18
chore: Publish visual tests result artifact

gretanausedaite:
Added action to publish tests results.


bentleyvk:
No no no 😀 just do that in that PR you are working on

mayank99:
This will become out of date real quick as #640 is near completion. So I can just add it in there instead.

veekeys:
> This will become out of date real quick as #660 is near completion. So I can just add it in there instead.

Wouldnt you then just need to change the path..?

gretanausedaite:
This will be added with #660

PR #681, opened 2022-05-19
chore: Move gitattributes to root

mayank99:
This was leftover/missing from the monorepo PR.


veekeys (approved):
Have you checked if all files in server are `lf` ? Or you just got lucky, because you are using MAC? 😄

PR #689, opened 2022-05-24
feat(Table): Scroll table to selected item

### LINKED ISSUE #558: Table: add ScrollToRow / ScrollToIndex
Add a method, which scrolls the item at the specified index into view.
Optional parameters: 
- Smooth scrolling
- Control the alignment of scrolled rows (e.x. 'start' | 'center' | 'end')

HaveSpacesuit:
With the completion of #539, we have the ability to use the table virtualization to scroll to an item. This can be useful for situations such as sharing a link to an item in a table, or refreshing a table with an item selected. The user would want to be automatically scrolled to a relevant item.

This initial commit is only supporting virtualized tables. For non-virtualized tables, there is a workaround the consumer may use that @veekeys documented on #558. I'm also not supporting paginated tables on this initial commit, as I'm not familiar enough with the code.

I can add some unit tests as well; I wanted to get some feedback on the initial concept first.


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -237,6 +237,11 @@ export type TableProps<
    * @beta
    */
   enableVirtualization?: boolean;
+  /**
+   * When using virtualized table, initialize the table scrolled to this item
+   * Does not support non-virtualized tables or tables with pagination
+   */
+  scrollToItem?: T;
```
veekeys, line 244: hm... Would we add support for all table types later on? Would it make sense?

Cause otherwise, not sure I am a fan of prop, which works only with virtualization is enabled :/

veekeys, line 244: hm... Would we add support for all table types later on? Would it make sense?

Cause otherwise, not sure I am a fan of prop, which works only with virtualization is enabled :/

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -568,6 +574,17 @@ export const Table = <
     ],
   );
 
+  const scrollToIndex = React.useMemo((): number | undefined => {
+    if (!enableVirtualization || paginatorRenderer) {
+      return undefined;
+    }
+    const index = rows.findIndex((row) => row.original == scrollToItem);
```
veekeys, line 581: Hmm.. Could you add an example on storybook? Currently I am thinking of passing an item versus passing an index... 

Of course, need to think about filtered out list of items too, so not sure :/

veekeys, line 581: Hmm.. Could you add an example on storybook? Currently I am thinking of passing an item versus passing an index... 

Of course, need to think about filtered out list of items too, so not sure :/

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: For non-virtualized tables, we could try an approach like [your workaround](https://codesandbox.io/s/itwinui-react-table-scrollintoview-ol1z9x?file=/src/App.tsx:1270-1319); we'd need to add the ref to a row instead of a cell, I think. Or there may be a simpler solution.

We could probably hack together a solution for paginated tables, though I'm not too familiar with that part of the code. I suppose you can use pagination with and without virtualization?

_Alternatively_, we could update the virtualization prop to take an object, and include the scroll to item as an optional property. That would solve the problem of the prop only working with virtualization, though it isn't as logical to me.

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: Added a simple story https://github.com/iTwin/iTwinUI-react/pull/689/commits/3ef398f1618d1e010c190fdee72ba8add0239186

veekeys: so when I look at your story.. why not make prop accept index instead? We could save all this `findIndex` and custom type too.
I was just not sure the exact flow and if user has index of the item he wants to scroll. Thats why I asked for story as I thought maybe you have an exact use case and would know the situation

packages/iTwinUI-react/src/core/Table/Table.tsx
veekeys: yeah, so pagination can be used with virtualization, tehcnically :D (@bentleyvk)
Maybe then we can have a separate prop, just mark it `@beta` as virtulization.. If we plan to add this scrolling to index with all the cases in future...
Need to think a little here.
@bentleyvk @mayank99 opinions?

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: The consumer isn't necessarily going to know the index. They have the data, but the table may be sorted or filtered. The user generally wants to have a certain _item_ highlighted, not necessarily an _index_.

I have two use cases in my app.
1) A user gets a link to an item. When loading the page, we want the table scrolled so the linked item is visible.
2) A user selects an item, performs an action, and we reload the table with new data. After reloading, we want the table scrolled so the recent item is still visible.

Each of these cases is easy for me, as the items are always selected. So I just pass in `scrollToItem={selectedItems[0]}`. Maybe I could update the story along those lines.

Another consideration is that object equality may not always work if people are doing something tricky with their data. We could query by the the data index by forcing consumers to pass in `getRowId`, and then take the row id instead of the item. I think, however, it works as expect as is.

packages/iTwinUI-react/src/core/Table/Table.tsx
mayank99: I think it makes sense to fallback to native `scrollIntoView` for non-virtualized tables. Adding paginated support will get complicated so I think we should leave that to the user because they can easily find out what page the row is in.

veekeys: Well ok.. Lets leave the prop then.

HaveSpacesuit: I added support for non-virtualized tables, but left off support for pagination for now. Since it adds a bit more code, I wrapped up all the functionality into its own `useScrollToRow` hook.

The useMergedRefs hook allows us to add an extra ref on the row div, which can then be imperatively triggered to scroll to the specified row. It still allows users to add their own refs to the row, and the onRowInViewport callback is not affected.

I did not add support for paginated tables.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -237,6 +238,14 @@ export type TableProps<
...
    */
   enableVirtualization?: boolean;
+  /**
+   * Initialize the table scrolled to this item.
+   *
+   * Must be used in conjunction with `getRowId` prop.
+   *
+   * Does not support tables with pagination.
+   */
+  scrollToItem: T;
```
bentleyvk, line 248: This should be an optional prop.

bentleyvk: If we require users to pass `getRowId` maybe instead of passing data item, they should pass id? I think by default `getRowId` returns row index so in theory users don't even need to pass `getRowId` if they are ok to pass index.

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // For non-virtualized tables, we need to add a ref to each row
+  // and scroll to the element
+  React.useEffect(() => {
+    if (enableVirtualization || index < 0) {
+      return;
+    }
+
+    rowRefs.current[index]?.scrollIntoView();
+  }, []);
```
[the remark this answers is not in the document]
HaveSpacesuit: Tested out the change and added it in.

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: Yes, it should be optional 😁. Updated that.

The reason I'm using both `getRowId` and `item: T` is that we can run `getRowId(item)` and compare that to the `getRowId` on each of the rows to find the exact row being referenced. We could ask the user to pass in the row Id for the row, but it seems easier not to.

I had originally implemented by checking `row.original === item`, but I had some cases in the app I'm using where the behind the scenes data-loading mutated some info on the item, and the equality check didn't work. I figure a way to get an unmutable id for the item is safer.

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,68 @@
...
+  scrollToIndex: number | undefined;
+  tableRowRef: (row: Row<T>) => (element: HTMLDivElement) => HTMLDivElement;
+};
+
+type ScrollToRowProps<T extends Record<string, unknown>> = Pick<
+  TableProps<T>,
+  'data' | 'enableVirtualization' | 'getRowId' | 'paginatorRenderer'
+> & { scrollToItem?: T };
+
+export function useScrollToRow<T extends Record<string, unknown>>({
```
bentleyvk, line 19: Need tests for virtualized and non-virtualized cases.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -237,6 +238,14 @@ export type TableProps<
    * @beta
    */
   enableVirtualization?: boolean;
+  /**
+   * Initialize the table scrolled to this item.
+   *
```
veekeys, line 243: No need empty lines

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,68 @@
...
+      return;
+    }
+
+    rowRefs.current[index]?.scrollIntoView();
+  }, [enableVirtualization, index]);
+
+  const tableRowRef = React.useCallback((row: Row<T>) => {
+    function addReference(element: HTMLDivElement): HTMLDivElement {
+      rowRefs.current[row.index] = element;
+      return element;
```
veekeys, line 61: do you need to return element?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -635,6 +645,7 @@ export const Table = <
           tableHasSubRows={hasAnySubRows}
           tableInstance={instance}
           expanderCell={expanderCell}
+          tableRowRef={tableRowRef(row)}
```
veekeys, line 648: Mybe lets not pass it if virtualization is not enabled? We dont want to collect all the refs of elements, if we not gonna use those..

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,68 @@
...
+  }, [enableVirtualization, index]);
+
+  // For non-virtualized tables, we need to add a ref to each row
+  // and scroll to the element
+  React.useEffect(() => {
+    if (enableVirtualization || index < 0) {
+      return;
+    }
+
+    rowRefs.current[index]?.scrollIntoView();
```
veekeys, line 55: Will those indexes work correctly if table has filtered data?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -237,6 +238,14 @@ export type TableProps<
    * @beta
    */
   enableVirtualization?: boolean;
+  /**
+   * Initialize the table scrolled to this item.
+   *
+   * Must be used in conjunction with `getRowId` prop.
```
veekeys, line 244: We might need to add example in here.. cause it is not obvious at all with this :/
Maybe we could have single prop instead, which would be object for item and comparer?
`scrollToItem: {item: T, comparer: row => string}` ?
We could even have default comparer with ids or so.. not sure

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: The empty lines provide a new line when shown in intellisense
image

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: Interesting improvement! I think it is the other way - if we have virtualization, we don't need the refs. We only need them on non-virtualized tables to imperatively scroll to the row.

I tested this change on virtualized and non-virtualized tables, and it works great.

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
HaveSpacesuit: Tested with sorted and filtered data on virtualized and non-virtualized table without any problems.

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: I spent a very long time trying to get the TypeScript props to require `getRowId` if `scrollToItem` was passed, but it always fell short. There is an example in the code sandbox which shows with the extra needed prop.

I wouldn't want to pass another comparer, as the react-table already has a getRowId prop that can be used. But I'm open to whatever the maintainers want to do.

veekeys: hmm... Have you tried conditional types?
Something like this, maybe...
```ts
| {
      scrollToItem: undefined;
      getRowId?: (originalRow: T, relativeIndex: number, parent?: Row<T>) => string;
  }
| {
      scrollToItem: T;
      getRowId: (originalRow: T, relativeIndex: number, parent?: Row<T>) => string;
  }
```
https://www.benmvp.com/blog/conditional-react-props-typescript/

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
veekeys: I dont see returning element there anywhere.. Can you try removing it and verify, if something is broken?
I am pretty sure it is not needed

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3021,3 +3021,55 @@ it('should not throw on headless table', () => {
...
+  renderComponent({
+    data,
+    enableVirtualization: false,
+    getRowId: (item) => item.name,
+    scrollToItem: data[19],
+    style: { overflow: 'auto', height: 400 },
+  });
+
+  screen.getByText('Name20');
+  screen.debug(undefined, 100000000);
```
veekeys, line 3074: I guess it is not needed

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3021,3 +3021,55 @@ it('should not throw on headless table', () => {
...
+
+  renderComponent({
+    data,
+    enableVirtualization: false,
+    getRowId: (item) => item.name,
+    scrollToItem: data[19],
+    style: { overflow: 'auto', height: 400 },
+  });
+
+  screen.getByText('Name20');
```
veekeys, line 3073: I think this test would still pass, even if you remove `scrollToItem`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -16,7 +16,7 @@ import {
   BaseFilter,
   FilterButtonBar,
   TableFilterProps,
-  tableFilters
+  tableFilters,
```
adamhock, line 19: Can you remove this comma so it does not show up as a change in the PR?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3021,3 +3021,55 @@ it('should not throw on headless table', () => {
...
+      ];
+    });
+
+  const data = mockedData(20);
+
+  renderComponent({
+    data,
+    enableVirtualization: true,
+    getRowId: (item) => item.name,
+    scrollToItem: data[19],
```
adamhock, line 3053: Minor, but I feel like we should use a different test index than the last index. If this test breaks in the future and scrolls to the last element instead of a specific index, this test would still pass. Maybe `data[9]`?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -50,6 +50,7 @@ import {
 } from './actionHandlers';
 import VirtualScroll from '../utils/components/VirtualScroll';
 import { SELECTION_CELL_ID } from './columns';
+import { useScrollToRow } from './hooks/useScrollToRow';
```
adamhock, line 53: Can you move this import up to wear the other hooks are imported? This also means editing index.ts in `src/core/Table/hooks`.

packages/iTwinUI-react/stories/core/Table.stories.tsx
```diff
@@ -2349,6 +2349,73 @@ Virtualized.argTypes = {
...
+  );
+
+  const data = useMemo(() => {
+    const size = 100000;
+    const arr = new Array(size);
+    for (let i = 0; i < size; ++i) {
+      arr[i] = {
+        id: i,
+        name: `Name${i}`,
+        description: `Description${i}`,
```
adamhock, line 2394: Minor, but maybe change "Description12345" to "Item scrolled to" or something so that users better understand what's going on in the story?

packages/iTwinUI-react/stories/core/Table.stories.tsx
```diff
@@ -2349,6 +2349,73 @@ Virtualized.argTypes = {
   data: { control: { disable: true } },
 };
 
+export const ScrollToItem: Story<Partial<TableProps>> = (args) => {
```
adamhock, line 2352: Would the name "Scroll To Row" make more sense here since it only applies to rows?

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: Ah, you [nerd-sniped](https://xkcd.com/356/) me! I still can't get it to play nice on the actual table. Playing around with dummy types, it looks like it should work, but it doesn't play nicely in the actual table code. Maybe you can crack it?

HaveSpacesuit: This implementation actually does work how I want in my app. If I leave off `scrollToItem`, I get an error, and I can use or not use `getRowId` without any issues.

```ts
type ScrollToItemProps<T extends Record<string, unknown>> =
  | {
      scrollToItem: T;
      getRowId: TableOptions['getRowId'];
    }
  | {
      scrollToItem?: never;
      getRowId?: TableOptions['getRowId'];
    };
```

One annoyance is that the typings in react-table aren't great, and I have to do a work around to prevent a compile error

```ts
getRowId={(row: PWData | object) => (row as PWData).instanceId}
// shouldn't need to use `| object`, but otherwise it gives 
// Type '(row: PWData) => string' is not assignable to type ((originalRow: PWData ...) => string) & ((originalRow: {} ...) => string)'
```

veekeys: Yeah.. I think we have issues in storybook after monorepo change...
I dont see the code pushed, maybe you forgot? Would like to check in Vite app too.

HaveSpacesuit: Didn't push it, but I will now.

veekeys: hm... How did you verify that it works? It doesnt at my end, I think :/

HaveSpacesuit: I did a build and moved the esm and cjs files into my app to replace what's in the iTwin package. I have a lot of trouble running the iTwin-react stuff on my computer; it complains about the version of Node I'm using, and when I change it, then my other projects are grumpy. I'm happy not to have this extra complication around the props if we can't get it to play nicely without a bunch of effort.

packages/iTwinUI-react/src/core/Table/Table.tsx
veekeys: hm.. I think it works fine.
I tried in our playground (just dropped the story code) and removed `getRowId` and I got the error (just needed to explicitly add types to data).
So it might be fine.

Update the tests and lets merge it, if there is nothing else?

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: Ok, I'll be able to get to the tests today or tomorrow. I'm having trouble getting them to work nicely with virtualization, though. There is a lot of mocking going on around all the virtualization code in the tests, and I can't follow it all...

veekeys: Let me know how it goes. I could try to have a look at those.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -89,6 +90,22 @@ export type TablePaginatorRendererProps = {
...
+type ScrollToItemProps<T extends Record<string, unknown>> =
+  | {
+      /**
+       * Initialize the table scrolled to this item.
+       *
+       * Must be used in conjunction with `getRowId` prop.
+       *
+       * Does not support tables with pagination.
+       */
+      scrollToItem: T;
```
bentleyvk, line 102: Shouldn't it accept a string instead which is row id?

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,67 @@
...
+  }, [enableVirtualization, index]);
+
+  const tableRowRef = React.useCallback((row: Row<T>) => {
+    function addReference(element: HTMLDivElement): void {
+      rowRefs.current[row.index] = element;
```
bentleyvk, line 60: Shouldn't we use id here instead of index? In case table is filtered or sorted.

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,67 @@
...
+  paginatorRenderer,
+}: ScrollToRowProps<T>): ScrollToRow<T> {
+  const rowRefs = React.useRef<Record<number, HTMLDivElement>>({});
+
+  const index = React.useMemo((): number => {
+    if (!scrollToItem || !getRowId || paginatorRenderer) {
+      return -1;
+    }
+
+    return data.findIndex(
```
bentleyvk, line 33: Maybe instead of `data` we should use `page` here therefore we wouldn't need to call `getRowId` just `row.id` as I see it has 3 parameters that we are not passing. But we also need to make `scrollIntoItem` to be row id instead of row data.

HaveSpacesuit: Absolutely! I swear I tested against sorted and filtered data before, but must have changed something and now it's off. Using `page` is the right way; `data` causes havoc on sorted or filtered tables.

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
HaveSpacesuit: This was a good catch; I'll update it.

packages/iTwinUI-react/src/core/Table/Table.tsx
HaveSpacesuit: It could work either way. I figured it would be easier for the consumer to pass the item. Possibly more performant if we pass the row ID, a we don't need to run `getRowId` on every item until we find it, but that may be premature optimization.
https://github.com/iTwin/iTwinUI-react/pull/689#discussion_r891675269

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,79 @@
...
+      !enableVirtualization ||
+      !getRowId ||
+      !scrollToItem ||
+      paginatorRenderer
+    ) {
+      return undefined;
+    }
+
+    const index = page.findIndex(
+      (row) => getRowId(row.original, 0) == getRowId(scrollToItem, 0),
```
bentleyvk, line 41: Ugh, I started to think that this might cause us problems because we are not calling `getRowId` as it is intended because first parameter is row data, second is row index and third is row parent if table has subrows. So, if someone uses all these 3 parameters in their `getRowId`, we are screwed. This is how default `getRowId` looks like:
image
What if we make `scrollToItem?: (rows: Rows<T>) => Row<T>`. Then users could use however they want e.g.
 `scrollToItem={(rows) => rows.find(row => row.id === item.id)}` 
or if `data` is not being changed
`scrollToItem={(rows) => rows.find(row => row.original === item)}`

bentleyvk: Hey, I started to work on this, in case you won't implement the same

apps/storybook/src/Table.stories.tsx
```diff
@@ -2422,6 +2422,75 @@ Virtualized.argTypes = {
...
+  return (
+    <Table
+      enableVirtualization
+      columns={columns}
+      emptyTableContent='No data.'
+      {...args}
+      style={{ maxHeight: '90vh' }}
+      data={data}
+      getRowId={(row) => row.id}
+      scrollToRow={(rows) => rows.findIndex((row) => row.id === data[12345].id)}
```
bentleyvk, line 2484: Not sure whether to show this way as an example or better `(rows) => rows.findIndex(row => row.original === data[12345])`

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
HaveSpacesuit: I tested this out in my app, both virtualized and non-virtualized, sorted and filtered. Seems to be working just a-ok! Thanks! 👍

bentleyvk:
Would be great to get others' reviews since I changed an implementation a bit @veekeys @mayank99 @adamhock

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -243,6 +244,24 @@ export type TableProps<
...
+   * @example
+   * <Table
+   *   data={data}
+   *   scrollToRow={(rows) => rows.findIndex((row) => row.original === data[250])}
+   *   {...restProps}
+   * />
+   * @example
+   * <Table
+   *   scrollToRow={(rows) => rows.findIndex((row) => row.id === item.id)}
+   *   getRowId={(rowData) => row.original.id}
```
veekeys, line 260: so `getRowId` is not needed anymore, right?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -243,6 +244,24 @@ export type TableProps<
...
+   * It doesn't work with paginated tables.
+   * @example
+   * <Table
+   *   data={data}
+   *   scrollToRow={(rows) => rows.findIndex((row) => row.original === data[250])}
+   *   {...restProps}
+   * />
+   * @example
+   * <Table
+   *   scrollToRow={(rows) => rows.findIndex((row) => row.id === item.id)}
```
veekeys, line 259: Maybe enough the first example? here is not even clear, what is that `item`

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
```diff
@@ -0,0 +1,59 @@
...
+      enableVirtualization ||
+      scrollToIndex === undefined ||
+      scrollToIndex === null ||
+      scrollToIndex < 0
+    ) {
+      return;
+    }
+
+    rowRefs.current[page[scrollToIndex].id]?.scrollIntoView();
+  }, [enableVirtualization, page, scrollToIndex]);
```
veekeys, line 50: ok, I guess this might not work as expected, because `useEffect` will be run after `page` update, which is not what we intend. I have added `scrollToRow` prop to `LazyLoading` story and disabled virtualization. Whenever new data is loaded, you are being scrolled to the item back.

packages/iTwinUI-react/src/core/Table/Table.tsx
bentleyvk: Yes, it is up to the users how they want to implement this.

packages/iTwinUI-react/src/core/Table/Table.tsx
bentleyvk: Just now realized that I could write this without using `getRowId`:
`scrollToRow={(rows) => rows.findIndex((row) => row.original.id === data[12345].id)}`

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
veekeys: I also thought so, but would it work in the case like...
Table is already rendered, and then user selects and element in the viewer and wants the table to scroll to that element? (technically, it is not run on init).

HaveSpacesuit: It should work, if we leave the `scrollToIndex` dependency. If the user selects a row, then their scrollToIndex function should change and retrigger this, even though it would be using the original page data, right?

veekeys: Even if we remove `page` dep, it will be triggered from user, because his `scrollToIndex` will be updated with new data loaded (data is a dep of memoized `scrollToIndex` function).
So... seems its more complex now :D I know @bentleyvk was having a look today at this, but not sure he found a solution yet

HaveSpacesuit: Yeah, that's a good point, it is some other variable. I guess it was a lazy story. =~)
For me, it would something like `rows.findIndex((row) => row.original.id == selectedIds[0])`, or in an earlier implementation, `selectedItem={selectedIds[0]}`

bentleyvk: I also have another idea, what if we pass `data` to `scrollToRow`?
```
scrollToRow={React.useCallback(
  (rows: Row<TableStoryDataType>[], data:<TableStoryDataType>) =>
    rows.findIndex((row) => row.original.id === data[12345].id),
  [],
)}
```
Then users could use `data` in their function but it won't cause rerenders -> no unnecessary scrolls.

bentleyvk: Now it is passing `data` to scroll function prop.

veekeys: @HaveSpacesuit @bentleyvk so what do we do with this one?
With current implementation, it looks fine to me. As there are some limitations (lazy load, paginated tables do not work), we could add maybe `@beta` tag as it is not fully implemented feature and just go with it?
@HaveSpacesuit, do you think we should just get rid of data in the callback and rely on user if he is going to include data in his calculactions, that everything is memoized and it wont cause the rerender and as a result scrolling everytime to the same row..?

bentleyvk: I would agree to have `@beta` tag and merge it as it is.

HaveSpacesuit: I'm fine either way. If it doesn't cause issues to pass the data back, I can always just ignore it =~)

veekeys (approved):

packages/iTwinUI-react/src/core/Table/hooks/useScrollToRow.ts
bentleyvk: Added `@beta` tag

adamhock (approved):

PR #693, opened 2022-05-27
chore: Update itwinui-css to 0.59.1

veekeys:
The only thing I am not sure is Table images and the changes.
If anyone could verify that, would be good.


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -694,10 +694,7 @@ export const Table = <
                       { 'iui-sorted': column.isSorted },
                       column.columnClassName,
                     ),
-                    style: {
-                      ...getCellStyle(column, !!state.isTableResizing),
-                      flexWrap: 'unset',
-                    },
+                    style: getCellStyle(column, !!state.isTableResizing),
```
mayank99, line 697: In CSS we are adding a new element `<div className='iui-table-header-wrapper'>`. Don't we need that here?

mayank99, line 697: In CSS we are adding a new element `<div className='iui-table-header-wrapper'>`. Don't we need that here?

veekeys: yeah... missed 2 classes actually.. Added

bentleyvk:
If I remember correctly, content moved a bit horizontally to the left because now side borders are in the first and last cells while previously it was in the row.

mayank99 (approved):

PR #698, opened 2022-05-31
fix(ComboBox): Fix keyboard navigation in virtualized list

veekeys:
-- Disabled going around the list if virtualization is enabled as it was causing some issues.
-- Added new attribute on menu item to save actual index in filtered list, so virtual scroll works correctly (previously it was taking index of original list, without checking if it is filtered)
-- Removed separate `useEffect` to reset focus on every `filteredOptions` change and instead dispatching event on every update , cause it needs to be different when you type in input (we fully reset it) compared with others (keep on selected value or reset)


packages/iTwinUI-react/src/core/ComboBox/ComboBox.test.tsx
```diff
@@ -203,106 +203,161 @@ it('should select value on click', async () => {
...
+    // 1 -> 2
+    await userEvent.keyboard('{ArrowDown}');
+    items = document.querySelectorAll('.iui-menu-item');
+    expect(input).toHaveAttribute(
+      'aria-activedescendant',
+      `${id}-option-Item-2`,
+    );
+    expect(items[2]).toHaveClass('iui-focused');
+
+    // 2 -> 0 (or stay if virtualization enabled)
```
bentleyvk, line 252: Looks a bit odd maybe just create another test case

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -226,9 +226,12 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             option.label.toLowerCase().includes(value.toLowerCase()),
           ),
       );
+      if (focusedIndex != -1) {
```
bentleyvk, line 229: Why needed?

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -52,13 +54,25 @@ export const ComboBoxInput = React.forwardRef(
...
                 'focus',
-                Object.values(optionsExtraInfoRef.current)?.[0]
-                  .__originalIndex ?? -1,
+                Number(currentElement?.getAttribute('data-iui-index') ?? 0),
               ]);
             }
 
+            // If virtualization is enabled, dont let round scrolling
+            if (
+              enableVirtualization &&
```
bentleyvk, line 68: Can't you dispatch focus action with 0? Then virtualized menu would set correct `scrollToIndex`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Decided to reset focus whenever user types, so we dont need to handle things like if focused element is visible after typing or not and then do something about `focusedIndex`

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
veekeys: amm.. what? I jsut dont want to move at all, so keeping current element focused.
Or I misunderstood your comment

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -295,6 +296,7 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
           isSelected={selectedIndex === __originalIndex}
           onClick={() => dispatch(['select', __originalIndex])}
           index={__originalIndex}
+          data-iui-filtered-index={filteredIndex}
```
bentleyvk, line 299: Just thinking, can this attribute be used to simplify regular keyboard navigation?

bentleyvk: Ok, maybe it doesn't help

bentleyvk (approved):

PR #700, opened 2022-06-06
feat(Select): Multiple selection support

### LINKED ISSUE #344: Select: multi select
Need to support something like this
image

veekeys:
~~Currently, need to link with my css branch from [here](https://github.com/iTwin/iTwinUI/pull/619)~~

Initial implementation on multi support.
- Have 2 stories: one with multi values using default implementation and one with custom selected items renderer
- ~~Exposed `setFocus` in `DropdownMenu` cause with multi selection I dont need to focus first focused element, otherwise after selecting any other, scroll will always go up to the first one~~ Removed hook to reset focused index on children change in `Menu`
- I have added `T[]` type to `value`, so now you can either have single value or multiple values.. Seemed not to break anything in stories, so hopefully its fine? 
- Added `multiple` prop
- ~~Added `selectedItemsRenderer` prop for user to customize the way values are displayed. I tried to reuse the existing `selectedItemRenderer`, but changing argument type ends up in errors (different type, so breaking change) and adding just 2nd argument did not look good. Maybe could have changed `selectedItemRenderer` prop to something like `(option: SelectOption<T>) => JSX.Element | (options: SelectOption<T>[]) => JSX.Element`, but havent tried that one.
If there are some more options, I'd hear them out.~~
- Added `SelectTag` component to display multi values.
- I have custom content for overflow, but ideally we would have something default and give user prop to override it, I guess.

Issues:
- ~~Somehow if I added (currently commented out) custom tag to display overflow info, calculation of visible items gets messed up and always shows 0.. Will need to have a look there, whats happeing.. Without custom button, it is being handled just fine~~


packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -110,6 +117,15 @@ export type SelectProps<T> = {
    * @see [tippy.js props](https://atomiks.github.io/tippyjs/v6/all-props/)
    */
   popoverProps?: Omit<PopoverProps, 'onShow' | 'onHide' | 'disabled'>;
+  /**
+   * Enable multiple selection.
+   * @default false
+   */
+  multi?: boolean;
```
bentleyvk, line 124: Maybe name it as a native select has - `multiple`

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -110,6 +117,15 @@ export type SelectProps<T> = {
...
+  /**
+   * Custom renderer for the selected items in select.
+   */
+  selectedItemsRenderer?: (options: SelectOption<T>[]) => JSX.Element;
```
bentleyvk, line 128: You could try fancy TS where if `multi` is true, then `selectedItemRenderer` takes options list if false only single option

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -255,25 +275,53 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+    [itemRenderer, multi, onChange, options, value],
   );
 
-  const selectedItem = React.useMemo(() => {
+  const selectedItems = React.useMemo(() => {
     if (value == null) {
       return undefined;
     }
-    return options.find((option) => option.value === value);
+    return Array.isArray(value)
```
bentleyvk, line 298: Maybe better check whether `multi` is true if for some reason someone decides to have single value that is an array.

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Text inside the tag.
+   */
+  children: React.ReactNode;
+} & CommonProps;
+
+/**
+ * Tag for showing selected value in `Select`.
+ */
+export const SelectTag = (props: SelectTagProps) => {
```
mayank99, line 27: Not sure about this name `SelectTag`. Have you looked at other libs?

mayank99: Also it doesn't look very customizable to me, so does the user even need it? Or if it's private, that's fine by me.

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: no idea how to achieve now (at least now), but I will try something 😄

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
veekeys: Name up for discussions (would be nice to change it in css too then).
I was not planning to expose this to users.. So havent thought about customizations (unless it might be needed?)
Now also got idea, should we start marking these components with additoinal tags like `@private` or `@internal`. 
We dont do it everywhere, I think

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: renamed

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -216,10 +234,10 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
   );
 
   React.useEffect(() => {
-    if (selectRef.current && !disabled && setFocus) {
+    if (selectRef.current && !disabled && setFocus && !multi) {
       selectRef.current.focus();
     }
-  }, [setFocus, disabled]);
+  }, [setFocus, disabled, multi]);
```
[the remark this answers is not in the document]
veekeys: it had something.. removed and lets see 😄

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -232,7 +250,7 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
       case 'Enter':
       case ' ':
       case 'Spacebar':
-        toggle();
+        !multi && toggle();
```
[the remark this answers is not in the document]
veekeys: hmm...removed

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Ok, I pushed something..
Guards are up for normal naming, but... it works, I think.. just not in storybook, cause for me it seems to be fully broken and cannot see any types at all (monorepo, I can bet)

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: I have type guard, so user should not be able to pass array without `multiple`

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
veekeys: I see antd calling them just `tag` and mui says its `chip`.
react-select I found only `multi-value`
atlsassian design is using react-select for demo, so `multi-value` too

so idk.. I like tag or chip? just we cannot use simply `tag`, so I udnerstand `SelectTag` is confusing :/

mayank99: If we mark it as private, I'm fine with whatever name, including `SelectTag`. Just maybe improve the description a little to mention multi select

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -55,15 +89,35 @@ export type SelectOption<T> = {
...
+      selectedItemRenderer?: (option: SelectOption<T>) => JSX.Element;
+      /**
+       * Selected option value.
+       * If `multiple` is enabled, it is an array of values.
+       */
+      value?: T;
+    }
+  | {
+      multiple: true;
+      selectedItemRenderer?: (options: SelectOption<T>[]) => JSX.Element;
```
bentleyvk, line 112: No need JSDocs?

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -255,25 +309,54 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
           isSelected,
           onClick: () => {
             !option.disabled && onChange?.(option.value);
-            close();
+            !multiple && close();
+          },
+          ref: (el: HTMLElement) => {
+            if (isSelected && !multiple) {
+              el?.scrollIntoView();
```
bentleyvk, line 316: No way to make it work? Maybe save first selected index or something

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -255,25 +309,54 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
       return undefined;
     }
-    return options.find((option) => option.value === value);
+    return Array.isArray(value)
+      ? options.filter((option) => value.some((val) => val === option.value))
+      : options.find((option) => option.value === value);
   }, [options, value]);
 
+  const selectedItemsArray = React.useMemo(() => {
+    if (!selectedItems || !Array.isArray(selectedItems)) {
```
bentleyvk, line 338: Why not checking whther `multiple` is enabled`

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -303,29 +386,65 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
-          {selectedItem &&
-            selectedItemRenderer &&
-            selectedItemRenderer(selectedItem)}
-          {selectedItem && !selectedItemRenderer && (
+          {(!selectedItems || selectedItems.length === 0) && (
+            <span className='iui-content'>{placeholder}</span>
+          )}
+          {isMultipleEnabled(selectedItems, multiple) ? (
+            <>
+              {multipleSelectedItemRendererType(
```
bentleyvk, line 408: Maybe add some comments

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Amm... I tried, but it is getting triggered on every new item select...
So I would need to add some more logic and make it scroll only on open or something :?

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Cause I am trusting the props guard, which should not allow to pass array of values, when disabled `multiple`

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: Yeah... then you would need to save whether it is first render, maybe enough for this time 😀

gretanausedaite (commented):
Code could have some comments explaining what it does (eg. Select.tsx).
Was this white gap on "delete tag" button hover discussed in css PR? 
image

veekeys:
> Code could have some comments explaining what it does (eg. Select.tsx). Was this white gap on "delete tag" button hover discussed in css PR? image

No it wasnt, but it is same on our `Tags` outside `Select`

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: ammm... i dont know then :/ what you suggest

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import SvgCloseSmall from '@itwin/itwinui-icons-react/cjs/icons/CloseSmall';
+import { useTheme, CommonProps } from '../utils';
+import '@itwin/itwinui-css/css/inputs.css';
+
+export type SelectTagProps = {
+  /**
+   * Callback function that handles click on close icon.
+   * Close icon is shown only when this function is passed.
+   * Use only with 'default' Tag.
```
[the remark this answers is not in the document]
veekeys: ups.. removed

gretanausedaite (commented):
Code makes sense 🌵 
Do we want to add unit tests for new functionality?
Do we want to add stories with select tags overflow (can users add custom overflow tag)?

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: Just check `multiple` 😀

veekeys: but then I need again some kind of a guard, cause types wont be recognized...

bentleyvk: Do you? Isn't `selectedItems` type `SelectOption[] | SelectOption`?

bentleyvk: and I see you already have type guard 
image

veekeys: But here we are talking about value :/

bentleyvk: Oh, now I understand.

veekeys:
> Code makes sense 🌵 Do we want to add unit tests for new functionality? Do we want to add stories with select tags overflow (can users add custom overflow tag)?

I left comment above, that tests are missing, I will add those.
In added stories you can achieve overflow by selecting more items and resizing window, so not sure will add separate story.
Yeah, for custom overflow tag is an open question.. Need to provide some kind of an API for this too, I guess

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -6,9 +6,43 @@ import React from 'react';
...
+import SelectTag from './SelectTag';
+
+const isMultipleEnabled = <T,>(
+  selectedItems: SelectOption<T> | SelectOption<T>[] | undefined,
+  multiple: boolean,
+): selectedItems is SelectOption<T>[] | undefined => {
+  return multiple;
+};
+
+const multipleSelectedItemRendererType = <T,>(
```
mayank99, line 27: These should be prefixed by `is`, right? Because it is not a `type`, and we are returning a boolean.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -55,19 +89,42 @@ export type SelectOption<T> = {
   [key: string]: unknown;
 } & CommonProps;
 
+export type SelectValueChangeEvent = 'Added' | 'Removed';
```
mayank99, line 92: any reason to go with `Added` instead of `added`?

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -55,19 +89,42 @@ export type SelectOption<T> = {
...
+
+export type MultipleType<T> =
```
mayank99, line 94: if we are exporting this, we should use a more descriptive name. `MultipleType` sounds too vague

or don't export it

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -254,26 +311,61 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
     if (value == null) {
       return undefined;
     }
-    return options.find((option) => option.value === value);
+    return Array.isArray(value)
+      ? options.filter((option) => value.some((val) => val === option.value))
+      : options.find((option) => option.value === value);
   }, [options, value]);
 
+  const selectedItemsArray = React.useMemo(() => {
```
mayank99, line 346: Maybe rename this to be `selectedItemsElements` or `selectedItemsJsx` or `selectedItemsTags`.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -303,29 +395,69 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
             setIsOpen(false);
           }
         }}
+        setFocus={!multiple}
```
mayank99, line 398: shouldn't we manually focus first item if we are disabling this?

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * Tag for showing selected value in `Select`.
+ * @private
+ */
+export const SelectTag = (props: SelectTagProps) => {
+  const { className, children, onRemove, ...rest } = props;
+  useTheme();
+
+  return (
+    <span
+      className={cx('iui-select-tag', className)}
```
mayank99, line 33: Why is this truncating even when there is space available?

veekeys: ohh, can you tell me how you reproduced this one??? Test captures this too

apps/storybook/src/Select.test.ts
```diff
@@ -37,7 +39,13 @@ describe('Select', () => {
...
       cy.compareSnapshot(`${testName} (Open)`);
 
-      if (testName.includes('Custom')) {
+      if (testName.includes('Multi')) {
+        cy.get('.iui-menu-item').first().click();
+        cy.get('.iui-menu-item').eq(1).click();
+        cy.compareSnapshot(`${testName} (Multi selected)`);
+      }
+
+      if (testName.includes('Custom') && !testName.includes('Multi')) {
```
bentleyvk, line 48: Don't you want just write `testName === 'Custom'`?

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: I was not sure.. tried to look for guidance over type guard naming, but nothing specific found.
Prefixed with `is`

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: I exported only for tests :/
Changed the naming a bit, not sure if its better now

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: renamed to `selectedItemsElements`

apps/storybook/src/Select.test.ts
veekeys: ohh, haha 
Good we have reviewers

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -254,26 +311,61 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
           isSelected,
           onClick: () => {
-            !option.disabled && onChange?.(option.value);
-            close();
+            !option.disabled &&
+              onChange?.(
+                option.value,
+                Array.isArray(value) && value.includes(option.value)
+                  ? 'Removed'
+                  : 'Added',
```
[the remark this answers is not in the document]
veekeys: amm.. you want me to pass `CustomEvent` to the `onChange` handler or create new ones separate for addition and removal?
Cause if its just a type of the argument in function, whats the advantage we get versus we just have a custom type in there..?
But also not sure if we want to dispatch separate events and kinda change the existing API

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -254,26 +311,61 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+              );
+            !multiple && close();
+          },
+          ref: (el: HTMLElement) => {
+            if (isSelected && !multiple) {
+              el?.scrollIntoView();
```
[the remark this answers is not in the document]
veekeys: yes, but the trick here is we need to do that only on opening..

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: changed

packages/iTwinUI-react/src/core/Select/Select.test.tsx
```diff
@@ -446,3 +448,88 @@ it('should pass custom props to menu item', () => {
...
+
+  const selectedValues = container.querySelectorAll(
+    '.iui-select-button > span',
+  );
+  expect(selectedValues.length).toBe(2);
+  expect((selectedValues[0] as HTMLElement).style.color).toEqual('green');
+  expect((selectedValues[1] as HTMLElement).style.color).toEqual('red');
+});
+
+it('should call onChange if tag is removed', () => {
```
bentleyvk, line 524: nit: You tested that in the previous test.

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: Or can you have a typeguard with generics where you pass `multiple` and other field that you care about.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -254,26 +311,61 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
           key: `${option.label}-${index}`,
           isSelected,
           onClick: () => {
-            !option.disabled && onChange?.(option.value);
-            close();
+            !option.disabled &&
+              onChange?.(
```
bentleyvk, line 315: So now for single selectable Select it will call `onChange` with value and `added` as second parameter? If yes, that's can be a breaking change.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -303,29 +395,66 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
-          {selectedItem && !selectedItemRenderer && (
+          {(!selectedItems || selectedItems.length === 0) && (
+            <span className='iui-content'>{placeholder}</span>
+          )}
+          {isMultipleEnabled(selectedItems, multiple) ? (
+            <>
+              {/* Either render custom multiple selected items provided by user */}
+              {isMultipleSelectedItemRenderer(selectedItemRenderer, multiple) &&
+                selectedItems &&
+                selectedItemRenderer(selectedItems)}
```
bentleyvk, line 420: Maybe we should pass `visibleCount` so users could handle that on their end if they wish.

gretanausedaite (commented):
Code looks OK, I don't have any additional comments.
Storybook `remove` functionality doesn't work.
My main concern is what if users want to have custom `+2 items selected` tag/button/etc. But that could be implemented later.

apps/storybook/src/Select.stories.tsx
```diff
@@ -341,3 +341,75 @@ export const TruncateMiddleText: Story<SelectProps<string>> = (args) => {
...
+      <Select<number>
+        {...rest}
+        options={[...Array(20).fill(null)].map((_, index) => ({
+          label: `Item #${index}`,
+          value: index,
+        }))}
+        value={value}
+        onChange={(val, event) =>
+          setValue((prev) =>
+            event === 'Removed'
```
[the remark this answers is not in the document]
bentleyvk: Ok, it shows for me 
image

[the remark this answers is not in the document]
veekeys: good, fixed

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Yes, it will, but I did not update interface that value is coming and also did not update any examples with it (except unit test).
Why that would be a breaking change? It comes as invisible parameter.

bentleyvk: If someone is passing a function that accepts two parameters, second one were always getting undefined and nothing happened but now it will get "added". I don't remember exactly but I have read somewhere to avoid doing that.

veekeys: what? Why they would accept 2nd parameter from `onChange`, if they know its giving only one?
Or I misunderstood the use case

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Just then need to figure where to put container ref, cause its needed for overflow to work.
Currently it goes on tag container, which is not present in custom renderer case

packages/iTwinUI-react/src/core/Select/Select.test.tsx
veekeys: no, there I tested clicking on same item, whereas here testing the close button on tag.

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: Hm, if users are passing the container, maybe then we should leave overflow handling to them.

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
mayank99: i can't reproduce it either but i noticed in tests

packages/iTwinUI-react/src/core/Select/Select.tsx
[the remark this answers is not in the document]
mayank99: maybe we could keep the old API for single select and pass CustomEvent for multi select? but i'm not sure how it would work with react events so maybe not 🤔

one advantage we get is being able to `dispatchEvent` in tests for example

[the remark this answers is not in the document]
veekeys: But I want to `dispatchEvent` in `Select`, not in tests :?
hmm.. not sure. I only see that maybe instead of string, we could pass an object so in future we could add more data, if needed?

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: We should, but as for now, dont know how

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: they could pass, or they could not (like in story example).
So I dont know...

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Ok, I have moved weird hook from `Menu` which was reseting focused index. This fixes issue in multiple select.
I have tried to go through all other stories and havent noticed keyboard nav to be broken anywhere.

packages/iTwinUI-react/src/core/Select/Select.tsx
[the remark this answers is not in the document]
veekeys: Fixed with the reseting hook removal from `Menu`

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: If they are reusing some function e.g. `handleItem(value: string, forceUpdate?: boolean)`, second parameter previously would always be undefined/falsie but not it will be truthy.

veekeys: sounds pretty weird...
But I guess I can try to add 2nd parameter only when multiple is enabled..

veekeys: fixed

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: added

veekeys:
Added keyboard nav on tags.
Last thing is that test image with cut out things.. I cannot reproduce it and not sure yet, why that happened..

mayank99:
> Added keyboard nav on tags. Last thing is that test image with cut out things.. I cannot reproduce it and not sure yet, why that happened..

I think we can leave the cut out image if it's at least consistent. Works fine in real usage so should be ok.

Keyboard navigation looks pretty good, apart from one issue: removing an item loses the focus. Would be nice to keep the focus on a nearby tag or at least on the select button. Maybe the `onBlur`/`onRemove` callbacks on `SelectTag` will help?

veekeys:
> > Added keyboard nav on tags. Last thing is that test image with cut out things.. I cannot reproduce it and not sure yet, why that happened..
> 
> I think we can leave the cut out image if it's at least consistent. Works fine in real usage so should be ok.
> 
> Keyboard navigation looks pretty good, apart from one issue: removing an item loses the focus. Would be nice to keep the focus on a nearby tag or at least on the select button. Maybe the `onBlur`/`onRemove` callbacks on `SelectTag` will help?

Added logic whenever tag is removed using keyboard, focus goes to the previous one... Just something feels wrong there, not sure :/

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -345,4 +452,91 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+    if (!selectedItems) {
+      return [];
+    }
+
+    return selectedItems.map((item) => (
+      <SelectTag
+        key={item.label}
+        onRemove={(e) => {
+          // assume if coords are zero, keyboard was used to remove tag
+          onChange?.(item.value, 'removed', e.screenX === 0 && e.screenY === 0);
```
bentleyvk, line 507: Yeah, looks a bit hacky. Maybe better handle `onKeyDown` explicitly.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -307,26 +400,40 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+            <MultipleSelectButton
+              selectedItems={selectedItems}
+              selectedItemsRenderer={
+                selectedItemRenderer as (
+                  options: SelectOption<T>[],
+                ) => JSX.Element
+              }
+              onChange={(value, event, isKeyboard) => {
+                if (isKeyboard) {
+                  focusPreviousElement(selectRef.current);
```
bentleyvk, line 425: It is not focusing previous tag on removal.

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: changed a little..

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: should be focusing firstly next and then previous, if not available

packages/iTwinUI-react/src/core/Select/Select.tsx
bentleyvk: For some reason this seems to work on multiple although it is disabled.

packages/iTwinUI-react/src/core/Select/Select.tsx
```diff
@@ -254,25 +344,71 @@ export const Select = <T,>(props: SelectProps<T>): JSX.Element => {
...
+  }, [multiple, options, value]);
+
+  const tagRenderer = React.useCallback(
+    (item: SelectOption<T>) => {
+      const onCloseKeyDown = (e: React.KeyboardEvent) => {
+        switch (e.key) {
+          case 'Enter':
+          case ' ':
+          case 'Spacebar':
+            focusNextElement(selectRef.current, () =>
```
bentleyvk, line 387: I know that you already know about this but in or der not to forget bringing focus to next tag doesn't work when there overflowed tags.

veekeys: I have fixed it.. 
Need to apply the same fix for `Breadcrumbs` and `ButtonGroup`. 
WIll create an issue.

packages/iTwinUI-react/src/core/Select/Select.tsx
veekeys: Yes. In single select, it also works, if I remove that line 😄

bentleyvk (approved):

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
```diff
@@ -0,0 +1,67 @@
...
+      onClick={(e) => e.stopPropagation()}
+      {...rest}
+    >
+      <span className='iui-select-tag-label'>{children}</span>
+      {isRemovable && (
+        <button
```
mayank99, line 53: I was testing on a screen reader and noticed two issues:
- the left/right arrow key navigation conflicts with the VoiceOver commands
- the "remove" buttons are not _connected_ with the select tags, so it's hard to tell what is getting deleted.

For the first issue, I think we might be missing some `role`, because in other components like Tabs, the left/right keys don't conflict.

For the second issue, I feel like the whole tag should be a focusable button so that it action stays "connected" with the item, it might help with the first issue to some extent as well.

veekeys: I have added roles, which made the keyboard nav work, but they are not the best (couldnt find any, which could make sense).
When added those, screen reader stopped reading the `aria-label` from the buttom, which is weird....
If I moved focus to tag, there need more work, starting from the styling (no focus style on tag) and finishing with the action, cause then it is not abvious that clicking on tag will remove it :/

veekeys: @mayank99 can you check it up? Either we proceed with this one and later could come back for improvements or please suggest, how to improve it now or we get rid of closing icons from tags and remove a bunch of code.. Just with the last one kinda a pitty, cause most of the stuff works fine :/

mayank99: Maybe let's remove it and create a separate issue for this. Because it would be much harder to change things later after we ship. We could fork the branch and keep it around (with a link in the issue) if the concern is that all this work would be "lost".

I was trying to research and came across this detailed article which includes real user testing, and they have different suggestions to what we are currently doing.
https://www.24a11y.com/2019/select-your-poison-part-2/

veekeys: I have gotten rid of close icon on tags and all the removal logic.

gretanausedaite (approved):
Code looks good. 
Let's not forget about "remove" button on select tags.

packages/iTwinUI-react/src/core/Select/SelectTag.tsx
mayank99: Can you create an issue to improve this part later?

veekeys: #790

mayank99: Testing on VO again, the select-button part is definitely better but i noticed that there is no feedback when you select/deselect there is no feedback

https://user-images.githubusercontent.com/9084735/185974811-fa3fdf54-9ed3-4671-9292-165ca863c221.mov

mayank99 (approved):
Works for an initial implementation but since there are issues with semantics and screen readers, let's come back to it later (https://github.com/iTwin/iTwinUI-react/issues/794, https://github.com/iTwin/iTwinUI-react/issues/790)

PR #706, opened 2022-06-09
fix(Table): Overriding min-width for the whole table

### LINKED ISSUE #643: Table: min-width on Table is bigger than its container width
Check this codesandbox without `style={{ minWidth: 0 }}`: https://codesandbox.io/s/itwinui-react-minimal-example-forked-8f3ecq?file=/src/App.tsx
`react-table` sets `min-width` to 1200px when container can be much smaller and that prevents horizontal scrollbar to appear inside Table. It appears on the whole page.

bentleyvk:
Closes #643 

When there is no column with `width` or `minWIdth` set, table `minWidth` is `0px` but when there are some columns that have `width` or `minWidth` then react-table sums them up and sets as table `minWidth` but that sometimes prevents horizontal scroll from working properly.

## Checklist

- [x] ~Add meaningful unit tests for your component (verify that all lines are covered)~

### DIFF
--- a/packages/iTwinUI-react/src/core/Table/Table.tsx
+++ b/packages/iTwinUI-react/src/core/Table/Table.tsx
@@ -711,7 +711,10 @@ export const Table = <
             { [`iui-${density}`]: density !== 'default' },
             className,
           ),
-          style,
+          style: {
+            minWidth: 0, // Overrides the min-width set by the react-table but when we support horizontal scroll it is not needed
+            ...style,
+          },
         })}
         {...ariaDataAttributes}
       >


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -711,7 +711,10 @@ export const Table = <
             { [`iui-${density}`]: density !== 'default' },
             className,
           ),
-          style,
+          style: {
+            minWidth: 0, // Overrides the min-width set by the react-table but when we support horizontal scroll it is not needed
```
veekeys, line 715: Is there a point to have a test for some of these scenarios? If its not too complex

bentleyvk: Idk, it is pretty dumb test also since our tests are completely mocked there won't be any benefit. Maybe visual tests are better for this.

veekeys (approved):

elephantcatdog (approved):

PR #708, opened 2022-06-10
fix(Table): Setting correct intersection root for margin to work

### LINKED ISSUE #699: intersectionMargin doesn't work
intersectionMargin values are ignored by both onBottomReached and onRowInViewport. Both events trigger only when the specific row is in viewport no matter what intersectionMargin value is written.

#### Expected behavior
Setting intersectionMargin to a higher value should trigger above-mentioned events sooner when scrolling.

#### Reproduction
https://codesandbox.io/s/itwinui-react-minimal-example-forked-7c3syr?file=/src/App.tsx

bentleyvk:
Closes #699 

 - Apparently correct intersection root element must be set for margin to work.
 - I had to use `useState` instead of `useRef` to make `useIntersection` hook rerender when row ref is get.
 - I can't just use `getScrollableParent` on a row because Table body has `overflow` set but for it to work, Table has to have a height set or using other methods to limit its height. So I am just comparing `scrollHeight` and `offsetHeight`.

## Checklist


packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -49,8 +54,30 @@ export const TableRow = <T extends Record<string, unknown>>(props: {
...
+
+    const scrollableParent = getScrollableParent(
+      rowRef?.closest('.iui-table'),
+      rowRef?.ownerDocument,
+    );
+    // If there is no scrollable parent besides body, then return undefined to make viewport as a root
+    if (scrollableParent === rowRef?.ownerDocument.body) {
+      return undefined;
+    }
+    return scrollableParent;
```
mayank99, line 75: Do we ever make the table scrollable but not the table body? Feels like we can just leave it as before if it's not table body.

```suggestion
    return undefined;
```

bentleyvk: Maybe it doesn't make sense if users have Table inside some div, idk.

mayank99: Even if it's inside a div, we still make only the table body scrollable, right?

Plus I feel like it makes sense that if you want to use our built in intersectionMargin that you need to use our scroll method. And if you control the scrolling outside the table, you should bring your own intersection observer.

bentleyvk: No, only if the table has something like `height: 100%` because else scroll will go on the wrapper div. It is kind of the same as in Storybook some Table stories.

packages/iTwinUI-react/src/core/utils/functions/dom.ts
```diff
@@ -38,3 +38,27 @@ export const getDocument = () => {
...
+
+  return isElementScrollable(element)
+    ? element
+    : getScrollableParent(element.parentElement, ownerDocument);
+};
+
+export const isElementScrollable = (element: HTMLElement) => {
+  return /(auto|scroll|overlay)/.test(
+    getElementStyle(element, 'overflow') +
+      getElementStyle(element, 'overflow-y'),
```
mayank99, line 58: Should we change this function to just compare `scrollHeight`/`offsetHeight`?

@veekeys Was there any reason you went with this logic instead of heights?

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
bentleyvk: Simplified things, if there will be a need, we can add that handling later.

packages/iTwinUI-react/src/core/utils/functions/dom.ts
veekeys: Hm... SOmething was not working with heights, I think...
What is the issue with this one?

mayank99: @bentleyvk had to create this same function with heights comparison, because overflow does not automatically imply scrolling. But not relevant anymore.

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -49,8 +49,23 @@ export const TableRow = <T extends Record<string, unknown>>(props: {
     isLast && onBottomReached.current?.();
   }, [isLast, onBottomReached, onRowInViewport, row.original]);
 
-  const rowRef = useIntersection(onIntersect, {
+  const [rowRef, setRowRef] = React.useState<HTMLElement | null>();
+  const getIntersectionRoot = () => {
```
veekeys, line 53: Would it make sense to memoize this?

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -49,8 +49,23 @@ export const TableRow = <T extends Record<string, unknown>>(props: {
     isLast && onBottomReached.current?.();
   }, [isLast, onBottomReached, onRowInViewport, row.original]);
 
-  const rowRef = useIntersection(onIntersect, {
+  const [rowRef, setRowRef] = React.useState<HTMLElement | null>();
```
veekeys, line 52: Weird to save ref in state... Just useing ref would not work here?

veekeys (commented):
Seems it works fine now.
Test would be helpful here, so we would catch in case it breaks again.

mayank99 (approved):

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -33,7 +33,8 @@ const intersectionCallbacks = new Map<Element, () => void>();
 jest
   .spyOn(IntersectionHooks, 'useIntersection')
   .mockImplementation((onIntersect) => {
-    return (el: HTMLElement) => intersectionCallbacks.set(el, onIntersect);
+    return (el: HTMLElement) =>
```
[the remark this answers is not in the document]
veekeys: ok..I tried to do it, but without the real dom is not really useful..

apps/storybook/src/Table.stories.tsx
```diff
@@ -841,7 +841,6 @@ export const LazyLoading: Story<Partial<TableProps>> = (args) => {
 
   return (
     <Table
-      enableVirtualization
```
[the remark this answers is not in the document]
veekeys: am... I think it does not work with virtualization enabled.
Parent element of `rowRef` there is not the body one, so the check is failing...
Should we add same logic of scrollable element as in VS or handle virtualized table explicitly (by going further than parent element)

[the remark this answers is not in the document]
mayank99: does intersection observer even make sense with virtualization? it's supposed to be for elements that are outside the viewport and in the DOM, but virtualization does not keep those elements in the DOM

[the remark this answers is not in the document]
veekeys: well, we do have at least 10 elements by default (can change that) which are outside of dom.. SO that gives a little bit of room, IMO..
Its not that complex to adjust parent if virtualization is enabled.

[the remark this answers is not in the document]
veekeys: added handling if virtualization is enabled

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
veekeys: using `useMemo`.
Checked, seems to still work.

veekeys (approved):

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
```diff
@@ -50,22 +52,24 @@ export const TableRow = <T extends Record<string, unknown>>(props: {
   }, [isLast, onBottomReached, onRowInViewport, row.original]);
 
   const [rowRef, setRowRef] = React.useState<HTMLElement | null>();
-  const getIntersectionRoot = () => {
+  const intersectionRoot = React.useMemo(() => {
+    const rowParent = enableVirtualization
+      ? rowRef?.parentElement?.parentElement?.parentElement
+      : rowRef?.parentElement;
```
mayank99, line 58: Should we flip this to have a more dynamic check that recursively looks upwards for a scrollable container? `parentElement.parentElement.parentElement` looks hacky/fragile. Also `isTableBodyScrollable` sounds wrong now.

veekeys: so what you suggest instead? I think as it is specific to table and we know the actual dom strucre, kinda makes sense.
I guess my implementation in VS is not really nice, so you just want to go up the parent until you find scrollable element? additional cycle not something I look up for every row, but.. maybe?

elephantcatdog (approved):

packages/iTwinUI-react/src/core/Table/TableRowMemoized.tsx
mayank99: Previously @bentleyvk had the same check in case the user had the table in a scrollable container.

Should we somehow check this only once per table and set a flag that every row can access?

veekeys: I moved ref state to table

bentleyvk:
New changes look good to me!

mayank99 (approved):

PR #711, opened 2022-06-15
fix(Table): Horizontal scroll is working when virtual scroll is enabled

### LINKED ISSUE #676: Virtualization disables horizontal scroll
When virtualization is enabled and table is wider than its parent, no horizontal scroll is enabled.

#### Expected behavior
Horizontal scroll should be visible and the table should be scrollable.

#### Reproduction
https://codesandbox.io/s/itwinui-react-minimal-example-forked-p1q6ih?file=/src/App.tsx

#### Additional information
VirtualScroll sets div style to `overflow: 'hidden'`, which causes the issue.

bentleyvk:
Closes #676 

`overflow: hidden` on outer div seems to do nothing, kind of obvious because it also has `minHeight` of the full list.
So removing it, seems to fix the issue.

## Checklist

- [x] ~Add meaningful unit tests for your component (verify that all lines are covered)~


packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -389,14 +389,13 @@ export const useVirtualization = (props: VirtualScrollProps) => {
   return {
     outerProps: {
       style: {
-        overflow: 'hidden',
```
mayank99, line 392: I'm not sure about this change.

@veekeys Do you remember why this was added?

veekeys: I was trying to find the case, but could not.. So I am not sure anymore :/

mayank99 (approved):

gretanausedaite (approved):
🌵

PR #713, opened 2022-06-15
feat(Tree): Virtualization

### LINKED ISSUE #712: Tree: virtualization support
There is a need to add virtualization in `Tree` component too.

veekeys:
Closes #712 

In order to preserve semantics (`li` inside `ul`), when virtualization is enabled I wrapped `ul` with `div` that gets `outterProps`, `className` and `style`.
To make keyboard navigation to work when parent node is not in the DOM, I used `useState` for scroll index and had to use `setTimeout` inside `useEffect` to focus the node when it is visible.

## Checklist

- [X] Add meaningful unit tests for your component (verify that all lines are covered)


packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,141 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
                 ? firstLevelNodesList.length
-                : flatNode.parentNode?.subNodeIds?.length ?? 0,
-            indexInGroup: flatNode.indexInGroup,
-            parentNodeId: flatNode.parentNode?.nodeProps.nodeId,
+                : node.parentNode?.subNodeIds?.length ?? 0,
+            indexInGroup: node.indexInGroup,
+            parentNodeId: node.parentNode?.nodeProps.nodeId,
+            scrollToParent: node.parentNode
+              ? () => {
+                  const parentNodeId = node.parentNode?.nodeProps.nodeId;
```
veekeys, line 264: you check if `node.parentNode` is defined above, no need here `?`

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,141 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
-      ))}
-    </ul>
+      );
+    },
+    [firstLevelNodesList.length, flatNodesList, nodeRenderer],
+  );
+
+  const [scrollToIndex, setScrollToIndex] = React.useState<number>();
+  const flatNodesListRef = React.useRef<FlatNode<T>[]>(flatNodesList);
+  flatNodesListRef.current = flatNodesList;
```
veekeys, line 282: Why this instead of `useEffect` ?

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,141 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+        nodeElement?.focus();
+        // Need to reset that if navigating with mouse and keyboard,
+        // e.g. going to the previous node with keyboard
+        // and then clicking the next node with mouse.
+        setScrollToIndex(undefined);
+      }
+    });
+  }, [scrollToIndex]);
+
+  const { outerProps, innerProps, visibleChildren } = useVirtualization({
```
veekeys, line 300: No need to initiate all the virtualization logic if it is not even enabled. Could you create a separate component with virtualization hook in there? Same as its done in `Combobox` for example

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,141 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+      {!enableVirtualization && (
+        <TreeElement
+          onKeyDown={handleKeyDown}
+          onFocus={handleFocus}
+          className={className}
+          style={style}
+          ref={mergeRefs(treeRef, innerProps.ref)}
+          {...rest}
+        >
+          {!enableVirtualization &&
```
veekeys, line 346: checked already above

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: Tell that to lint 
image

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: Moved to separate component

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,178 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
 };
 
+const TreeElement = React.forwardRef(
+  (
+    {
+      children,
+      className,
+      ...rest
+    }: {
+      children: React.ReactNode;
```
veekeys, line 348: maybe have `TreeElementProps` type

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,178 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+      onKeyDown: React.KeyboardEventHandler<HTMLUListElement>;
+      onFocus: React.FocusEventHandler<HTMLUListElement>;
+      className?: string;
+      style?: React.CSSProperties;
+      id?: string;
+    },
+    ref: React.ForwardedRef<HTMLUListElement>,
```
veekeys, line 355: oh, didnt know you can define types like that too :D
Usually we put those as generics using `<>`

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +244,178 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+    }
+  };
+
+  return (
+    <>
+      {enableVirtualization ? (
+        <VirtualizedTree
+          flatNodesList={flatNodesList}
+          itemRenderer={itemRenderer}
+          scrollToIndex={scrollToIndex}
```
veekeys, line 317: damn...keyboard nav is not nice.. it should always go by 1 element down

bentleyvk: What you mean?

bentleyvk: If you press arrow right, it goes to the parent that isn't always in the DOM. You can try by reducing buffer items to 1-3.

veekeys: hm... seems its native `focus` which is scroling half a list :/
Maybe we can leave it for now like that

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: Added

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -368,6 +363,16 @@ const TreeElement = React.forwardRef(
   },
 );
 
+type VirtualizedTreeProps<T> = {
+  flatNodesList: FlatNode<T>[];
+  itemRenderer: (index: number) => JSX.Element;
+  scrollToIndex?: number;
+  onKeyDown: React.KeyboardEventHandler<HTMLUListElement>;
+  onFocus: React.FocusEventHandler<HTMLUListElement>;
+  className?: string;
```
veekeys, line 372: `StylingProps`

bentleyvk (approved):
Well done!

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +245,173 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
-              flatNode.depth === 0
+              node.depth === 0
                 ? firstLevelNodesList.length
-                : flatNode.parentNode?.subNodeIds?.length ?? 0,
-            indexInGroup: flatNode.indexInGroup,
-            parentNodeId: flatNode.parentNode?.nodeProps.nodeId,
+                : node.parentNode?.subNodeIds?.length ?? 0,
+            indexInGroup: node.indexInGroup,
+            parentNodeId: node.parentNode?.nodeProps.nodeId,
+            scrollToParent: node.parentNode
```
mayank99, line 263: is this needed in the context for non-virtualized tree?

packages/iTwinUI-react/src/core/Tree/Tree.tsx
mayank99: See https://github.com/iTwin/iTwinUI-react/pull/738#discussion_r927826147

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +245,173 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+      if (scrollToIndex !== undefined) {
+        const nodeId = flatNodesListRef.current[scrollToIndex].nodeProps.nodeId;
+        const nodeElement = treeRef.current?.ownerDocument.querySelector(
+          `#${nodeId}`,
+        ) as HTMLElement;
+        nodeElement?.focus();
+        // Need to reset that if navigating with mouse and keyboard,
+        // e.g. going to the previous node with keyboard
+        // and then clicking the next node with mouse.
+        setScrollToIndex(undefined);
```
mayank99, line 296: Not sure I wonderstand what this is really doing. I didn't notice any difference when I commented it out and tried going to previous node then clicking next node with mouse.

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +245,173 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+      itemRenderer: itemRenderer,
+      scrollToIndex,
+    });
+
+    return (
+      <div
+        {...{
+          ...outerProps,
+          className: cx(className, outerProps.className),
+          style: { ...style, ...outerProps.style },
```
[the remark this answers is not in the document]
bentleyvk: I thought it should go on the outter div for easier styling because users would need to have some wrapper or something to target the top element e.g. `.tree-wrapper > div`.

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: I am just using the same keyboard navigation logic for virtualized and non-virtualized tree

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: Bad comment, changed it a bit. It was needed when using arrow left to go back to the parent node. So if you press arrow left and go to the parent node, then click with the mouse some child node and try pressing arrow left again, without this reset it would do nothing because `scrollToIndex` did not change.

packages/iTwinUI-react/src/core/Tree/Tree.tsx
bentleyvk: That was new to me. Changed.

packages/iTwinUI-react/src/core/Tree/Tree.tsx
```diff
@@ -224,40 +245,173 @@ export const Tree = <T,>(props: TreeProps<T>) => {
...
+        }}
+      >
+        <TreeElement
+          {...{ ...innerProps, ...rest }}
+          onKeyDown={onKeyDown}
+          onFocus={onFocus}
```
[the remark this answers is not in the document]
bentleyvk: Changed.

packages/iTwinUI-react/src/core/Tree/Tree.tsx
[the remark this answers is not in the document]
mayank99: That's a good point, but the rest props go on the tree element so it's really confusing.

I think we should probably support both. What is the currently expected way for the user to target the tree element when they use virtualization? `.my-outer-class > *`? Then we should document that.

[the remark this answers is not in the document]
bentleyvk: Added some comments.

mayank99 (approved):

PR #719, opened 2022-06-24
chore: Fix `createComponent` script

### LINKED ISSUE #692: `createComponent` script needs to be updated
The `createComponent` script has become obsolete after the addition of monorepo/storybook/cypress. It should be updated to create the correct files in the correct folders.

We might also want to move this script out of the `itwinui-react` folder into root or a shared package. Unclear.

marashad001:
Updated `createComponent` script to create files in new directories
Still need to move script to root folder
Closes #692


packages/iTwinUI-react/scripts/createComponent.js
```diff
@@ -81,7 +85,7 @@ const storiesFactory = (directory, name) => {
     template: `${copyrightHeader}
 import { Story, Meta } from '@storybook/react';
 import React from 'react';
-import { ${name}, ${name}Props } from '../../src/core';
+import { ${name}, ${name}Props } from '../../../packages/iTwinUI-react/src/core';
```
veekeys, line 88: Here should be
`from '@itwin/itwinui-react'`.
Check already existing stories and the import path.

packages/iTwinUI-react/package.json
```diff
@@ -35,7 +35,6 @@
     "clean:coverage": "rimraf coverage",
     "test": "jest",
     "test:watch": "jest --watch",
-    "createComponent": "node scripts/createComponent.js",
```
veekeys, line 38: not sure about this.. Maybe it should be possible to create component from both, root and itwinui-react package?
Even though now as storybook is moved onto separate folder, it might make less sense :/

marashad001: It's currently set up to make components from the root (even though the file itself is inside packages, i'll move to root). Do you think if we had two scripts (one for root and one for package) it would cause any issues? The scripts shouldn't interfere with any existing code, but you would run the command the same in either directory and have the same output.

veekeys: oh no no.. only single script.
I meant to add `createComponent` command in both `package.json` (root and package itself).
But it might make sense to have it only in root, so its fine.

packages/iTwinUI-react/scripts/createComponent.js
```diff
@@ -175,5 +199,6 @@ inquirer
     writeFile(componentTest);
     writeFile(componentIndex);
 
-    writeFile(storiesFactory('stories/core', component));
+    writeFile(storiesFactory('apps/storybook/src', component));
+    writeFile(storyTestsFactory('apps/storybook/src', component));
   });
```
elephantcatdog, line 204: You can use the following code here so that this script can be run from the root and from packages. It uses `process.cwd()` to get the location where the script is run from, and then adjusts the directory paths accordingly.

(couldn't get the formatting right)

elephantcatdog: .then(({ component }) => {
    const currentDirectory = process.cwd();
    const isInPackage = currentDirectory.includes('packages');

    let directory = '';
    let storyDirectory = '';
    let levels = [];

    if (isInPackage) {
      directory = `src/core/${component}`;
      storyDirectory = '../../apps/storybook/src';
    } else {
      directory = `packages/iTwinUI-react/src/core/${component}`;
      storyDirectory = 'apps/storybook/src';
    }
    levels.push(directory);

    let componentOut = componentFactory(directory, component);
    let componentTest = componentTestFactory(directory, component);
    let componentIndex = componentIndexFactory(directory, component);
    let indexAppend = indexAppendFactory(directory, component);

    for (const level of levels) {
      makeDir(level);
    }

    appendFile(indexAppend);

    writeFile(componentOut);
    writeFile(componentTest);
    writeFile(componentIndex);

    writeFile(storiesFactory(storyDirectory, component));
    writeFile(storyTestsFactory(storyDirectory, component));
  });

packages/iTwinUI-react/package.json
elephantcatdog: I left a comment below about how to achieve this. Then you can just leave this line as it was before.

marashad001: Ok added this in, thanks!

packages/iTwinUI-react/scripts/createComponent.js
```diff
@@ -154,7 +178,7 @@ inquirer
   ])
   .then(({ component }) => {
     console.log();
```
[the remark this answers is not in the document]
marashad001: Removed

packages/iTwinUI-react/package.json
elephantcatdog: You'll need to add `"createComponent": "node scripts/createComponent.js",` back in the file `packages/iTwinUI-react/package.json`.

packages/iTwinUI-react/scripts/createComponent.js
elephantcatdog: Great! Don't forget to remove the commented out code

createComponent.js
```diff
@@ -2,14 +2,14 @@
...
-const fs = require('fs');
+const inquirer = require("inquirer");
+const fs = require("fs");
 
 const copyrightHeader =
-  '/*---------------------------------------------------------------------------------------------\n ' +
-  '* Copyright (c) Bentley Systems, Incorporated. All rights reserved.\n ' +
-  '* See LICENSE.md in the project root for license terms and full copyright notice.\n ' +
-  '*--------------------------------------------------------------------------------------------*/';
+  "/*---------------------------------------------------------------------------------------------\n " +
```
veekeys, line 9: What happened with quotes here?

packages/iTwinUI-react/scripts/createComponent.js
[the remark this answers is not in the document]
marashad001: Removed

packages/iTwinUI-react/scripts/createComponent.js
marashad001: Thanks

createComponent.js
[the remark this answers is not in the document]
marashad001: OK fixed this

marashad001: it looks regular in my vscode

packages/iTwinUI-react/package.json
marashad001: Got it

createComponent.js
veekeys: What?
I mean you shouldnt have made any additional changes in here (i.e. replace single quotes with double quotes), so github would show only changes we care.
Can you try to revert that?
Could it be the case, that you dont have `prettier` installed in your vscode?

scripts/createComponent.js
```diff
@@ -136,7 +162,7 @@ it('should be improved', () => {
 
 const indexAppendFactory = (directory, name) => {
   return {
-    path: `${directory.split('/').slice(0, -1).join('/')}/index.ts`,
+    path: `${directory.split("/").slice(0, -1).join("/")}/index.ts`,
```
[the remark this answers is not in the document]
marashad001: Ok I got prettier installed in the config it's working now

veekeys (approved):

PR #724, opened 2022-06-29
feat: Update itwinui-css to 0.61.0

veekeys:
Updated itwinui-css to 0.61.0

- Updated `Modal` code
- Border radius change was caught only by some tests, so I have regenerated all of the images


veekeys:
@mayank99 @adamhock can I get a review on this?

packages/iTwinUI-react/src/core/Modal/Modal.tsx
```diff
@@ -177,52 +178,55 @@ export const Modal = (props: ModalProps) => {
...
+                  { 'iui-dialog-default': styleType === 'default' },
+                  { 'iui-dialog-full-page': styleType === 'fullPage' },
+                  { 'iui-dialog-visible': isOpen },
+                  className,
                 )}
+                id={id}
+                style={style}
+                role='dialog'
+                aria-modal='true'
+                {...rest}
```
mayank99, line 209: style and className were previously applied on the backdrop/wrapper whereas now they will be on the dialog itself. This makes sense but we should mention this in the changelog as it's is a breaking change for user CSS.

mayank99 (approved):
Looks good. A lot of the images look unchanged so it would have been nice to not have them show as changed, but the content looks fine.

Issue #709, opened 2022-06-10
ComboBox does not handle change of options correctly


saskliutas:
#### Describe the bug (actual behavior)
ComboBox end up in infinite loop when one of the options is selected and new list of options is passed to ComboBox

#### Expected behavior
ComboBox should not end up in infinite loop.

#### Reproduction
Link to a minimal repro: [SandBox](https://codesandbox.io/s/itwinui-react-example-forked-8frs6w?file=/src/App.tsx)

##### Steps to reproduce
1. Select 'Label 4' in ComboBox
1. Press 'Remove First' button
1. Warnings start appearing in the console: `Maximum update depth exceeded`

mayank99:
Hi @saskliutas, thanks for the issue. This looks like a duplicate of #697 which we are currently working on, so you can subscribe to it for updates.

mayank99:
Actually this might not be a duplicate, as #697 is about changing the `value` whereas here you are changing the `options` array. We're still looking into it!

JonasDov:
Facing the same issue, need a fix of this ASAP.

PR #728, opened 2022-07-01
fix(Combobox): Allow users to reset value

### LINKED ISSUE #697: ComboBox: Infinite loop
Externally changing the controlled value of the ComboBox is not respected or an infinite loop occurs. 

#### Reproduction
https://codesandbox.io/s/itwinui-react-example-forked-v8pllg?file=/src/App.tsx

gretanausedaite:
Fixes bug #697 

Removed line 273: Allow users to set value as undefined.
Removed line 275 (part of 280): Fix infinite loop bouncing between two useEffects.


packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -151,6 +151,7 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
   const inputRef = React.useRef<HTMLInputElement>(null);
   const menuRef = React.useRef<HTMLUListElement>(null);
   const toggleButtonRef = React.useRef<HTMLSpanElement>(null);
+  const firstRenderRef = React.useRef(true);
```
mayank99, line 154: Maybe invert and rename this to something like `justMounted` or `mounted`.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -270,14 +273,13 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   React.useEffect(() => {
-    if (selectedIndex != undefined && selectedIndex >= 0) {
-      const value = options[selectedIndex]?.value;
-      if (value === valueProp) {
-        return;
-      }
-      onChangeProp.current?.(value);
+    if (firstRenderRef.current) {
+      firstRenderRef.current = false;
+      return;
```
mayank99, line 278: Do you know why this fixes the issue? Can you add a comment then?

Also add a test

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -270,14 +273,13 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
 
   // Call user-defined onChange when the value actually changes
   React.useEffect(() => {
-    if (selectedIndex != undefined && selectedIndex >= 0) {
-      const value = options[selectedIndex]?.value;
-      if (value === valueProp) {
```
mayank99, line 275: Why removed the `value === valueProp` check?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Why? What is the difference?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: This solves only issue where useEffect is called on first render. It's called because combobox value is set (undefined or actual value). And we want to allow undefined value for users to reset combobox.
Will add comments.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: @mayank99 this is same as I had in multi select.
2nd use effect has the `old` value of selectedIndex!
Even though I thought if we use reducer, it should get the newest one :O

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Yeah, I think we used flag like that in somewhere else already.
@gretanausedaite `mounted` came from the original react in classes.
https://reactjs.org/docs/react-component.html#componentdidmount
so `useEffect` first call is 'almost' same as that lifecycle method in class components.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: If i remember correctly, this check was added because people were not always memoizing the onChange callback and it was leading to infinite loops 🤔 Should we use the same `queueMicrotask` logic here?

elephantcatdog (commented):
I agree with Mayank on adding a test for resetting the combobox, but otherwise have no comments.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
veekeys: Would be nicer if we could avoid `queueMicrotask`... Looks hacky.
Dont we add ref of `onChange` ourselves, so we dont care about users memoizing it...? Or its different

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Added test to programmatically reset value.

veekeys (approved):

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: Hmm yeah, I think `onChange` might be fine because of the ref but people also don't always memoize `options`. If you try to recreate #697 and #709 in playground, you will find that it still loops infinitely.

gretanausedaite: I would leave `options` issue to other PR. This one fixes #697 so maybe merge?

veekeys: But does it?
@mayank99 is saying #697 is still reproducible with your fix?

mayank99: yeah #697 is reproducible with current changes if `options` is not memoized

PR #732, opened 2022-07-08
fix(Table): Prevents crashing when clearing editable cell

### LINKED ISSUE #703: Table: Crash when leaving empty editable cell
Crash when using "tab" key or mouse click to leave a empty editable cell

**Note:  Working in Firefox, Crashing in Chrome**

#### Reproduction
Use story book "editable" example

##### Steps to reproduce
1. Enter "Description1" cell
2. Remove its content
3. Press "Tab" key or click outside to leave the cell

bentleyvk:
Closes #703  

Error was throw because when clearing editable cell, text node is being removed and React doesn't like DOM manipulation outside React. So giving editable cell a new key will make React reattach with the cell element in the DOM and it won't complain.

## Checklist

### DIFF
--- a/packages/iTwinUI-react/src/core/Table/cells/EditableCell.tsx
+++ b/packages/iTwinUI-react/src/core/Table/cells/EditableCell.tsx
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import React from 'react';
 import { CellRendererProps } from 'react-table';
+import { getRandomValue } from '../../utils';
 
 export type EditableCellProps<
   T extends Record<string, unknown>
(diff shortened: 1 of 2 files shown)


packages/iTwinUI-react/src/core/Table/cells/EditableCell.tsx
```diff
@@ -80,6 +84,9 @@ export const EditableCell = <T extends Record<string, unknown>>(
           onCellEdit(cellProps.column.id, value, cellProps.row.original);
         }
         props.onBlur?.(e);
+        // Prevents error when text is cleared.
+        // New key makes React to reattach with the DOM so it won't complain about deleted text node.
+        setKey(getRandomValue(10));
```
mayank99, line 89: So I know React treats new key as a new node but why does this fix the error? Any reading material on this?

mayank99 (approved):
Interesting fix but works so LGTM 👍

packages/iTwinUI-react/src/core/Table/cells/EditableCell.tsx
bentleyvk: I found about this here: https://stackoverflow.com/a/67686478/18045303

gretanausedaite (approved):

PR #735, opened 2022-07-13
fix(Small): Use `small` element instead of `p`

### LINKED ISSUE #734: Small: Wrong HTML tag

When `Small` is used inside another *Typography* (for example `Body`) a warning is raised:

image

#### Expected behavior

According to [HTML standards for p](https://www.w3.org/TR/html401/struct/text.html#h-9.3.1):

> The [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) element represents a paragraph. It cannot contain [block-level elements](https://www.w3.org/TR/html401/struct/global.html#block-inline) (including [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) itself).

The `<Small>` component is expected to generate the `<small>` tag.

#### Reproduction
```jsx
<Body>
  Some <Small>text</Small>
</Body>
```

#### Additional information
References:
[W3C - Paragraphs: the P element](https://www.w3.org/TR/html401/struct/text.html#h-9.3.1)
[W3C - Block-level and inline elements](https://www.w3.org/TR/html4/struct/global.html#h-7.5.3)
[MDN Content Categories](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories)
[MDN page for `small`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/small)

mayank99:
Fixes #734


packages/iTwinUI-react/src/core/Typography/Small/Small.tsx
```diff
@@ -28,7 +28,7 @@ export const Small = React.forwardRef<HTMLParagraphElement, SmallProps>(
     useTheme();
 
     return (
-      <p
+      <small
```
veekeys, line 31: Why images have changed?

packages/iTwinUI-react/src/core/Typography/Small/Small.tsx
```diff
@@ -28,7 +28,7 @@ export const Small = React.forwardRef<HTMLParagraphElement, SmallProps>(
...
         ref={ref}
```
veekeys, line 32: Need to make same update in css demos
@FlyersPh9

veekeys (approved):

packages/iTwinUI-react/src/core/Typography/Small/Small.tsx
mayank99: i think it's because small is an inline element, p is a block element

PR #738, opened 2022-07-15
fix(ComboBox): Prevent infinite loop when `options` change

### LINKED ISSUE #709: ComboBox does not handle change of options correctly

gretanausedaite:
Closes #709 

- Added ref to `options` prop;
- Added ref to `valueProp` ;
- Added check if valueProp has same value as selectedIndex;
- Added tests;


gretanausedaite (commented):
I need some help with implementation:
What do we do with selected value when options change?

Edit: fixed.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,46 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
   const toggleButtonRef = React.useRef<HTMLSpanElement>(null);
   const mounted = React.useRef(false);
 
+  // Record to store all extra information (e.g. original indexes), where the key is the id of the option
```
marashad001, line 156: Does the order where this bit of code appears matter in the file? (curious)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,46 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-    Record<string, { __originalIndex: number }>
-  >({});
+  // Latest value of the options prop
+  const optionsProp = React.useRef(options);
 
   // Clear the extra info when the options change so that it can be reinitialized below
   React.useEffect(() => {
     optionsExtraInfoRef.current = {};
-  }, [options]);
+  }, [optionsProp]);
```
marashad001, line 182: What is the difference between `options` and `optionsProp`?

marashad001 (commented):
Your code works very well in the test app, I don't have any new suggestions right now. Good job!

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Computer reads code from top to bottom so order where this bit of code appears in the file depends where I'm using it. 
To use variable/function/etc., it firstly needs to be declared.
I created a function (to reuse code) `initializeExtraInfo` which uses `optionsExtraInfoRef`. Therefore `optionsExtraInfoRef` needs to be above said function.

bentleyvk (commented):
Could you explain why the fix works?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
gretanausedaite: Changed back to options.

gretanausedaite: `options` is array of options user provides.
`optionsProp` is mutable ref object we create.
Having `optionsProp` helps us to control when we change options and selected value in combobox, helps us prevent infinite bouncing loops between useEffects.
You can read more about React hooks [here](https://reactjs.org/docs/hooks-reference.html#useref).

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -278,13 +298,19 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    if (currentValue === localValueProp.current || selectedIndex === -1) {
+      return;
+    }
+    onChangeProp.current?.(currentValue);
+  }, [selectedIndex]);
 
   const getMenuItem = React.useCallback(
     (option: SelectOption<T>, filteredIndex?: number) => {
       const optionId = getOptionId(option, id);
+
```
bentleyvk, line 313: Remove new line

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -223,17 +241,19 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
   }, [isOpen]);
 
   // Update filtered options to the latest value options according to input value
-  const [filteredOptions, setFilteredOptions] = React.useState(options);
+  const [filteredOptions, setFilteredOptions] = React.useState(
```
bentleyvk, line 244: This change is not necessary as `useState` uses value passed only for initialization later it is not used.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -223,17 +241,19 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    optionsProp.current,
+  );
   React.useEffect(() => {
     if (inputValue) {
       setFilteredOptions(
-        filterFunction?.(options, inputValue) ??
+        filterFunction?.(optionsProp.current, inputValue) ??
```
bentleyvk, line 250: I see `options` are still in the deps array, so is this change needed for this `useEffecct`?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -169,16 +163,41 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     optionsExtraInfoRef.current = {};
```
bentleyvk, line 163: Maybe you can move this to your newly created useEffect?

bentleyvk:
@mayank99 can you still check this out as you have better understanding of Combobox than me?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: undo

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: moved

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -169,16 +163,41 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    optionsProp.current.forEach((option, index) => {
       optionsExtraInfoRef.current[getOptionId(option, id)] = {
         __originalIndex: index,
       };
     });
+  }, [id]);
+
+  // Latest value of the valueProp prop
+  const localValueProp = React.useRef(valueProp);
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
gretanausedaite: like simply `localValueProp.current = valueProp`?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,48 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  const initializeExtraInfo = React.useCallback(() => {
+    optionsProp.current.forEach((option, index) => {
+      optionsExtraInfoRef.current[getOptionId(option, id)] = {
+        __originalIndex: index,
+      };
+    });
+  }, [id]);
+
+  // Latest value of the valueProp prop
+  const localValueProp = React.useRef(valueProp);
```
bentleyvk, line 171: Hm, maybe rename to `valuePropRef`.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,48 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   }, [onChange]);
 
-  // Record to store all extra information (e.g. original indexes), where the key is the id of the option
-  const optionsExtraInfoRef = React.useRef<
-    Record<string, { __originalIndex: number }>
-  >({});
-
-  // Clear the extra info when the options change so that it can be reinitialized below
+  // Latest value of the options prop
+  const optionsProp = React.useRef(options);
```
bentleyvk, line 183: Rename to `optionsRef` since there is `valueProp` and means completely different thing.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: renamed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: renamed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,47 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  const optionsRef = React.useRef(options);
+  optionsRef.current = options;
 
-  // Clear the extra info when the options change so that it can be reinitialized below
+  // Change options ref on options change
   React.useEffect(() => {
+    // Clear the extra info when the options change so that it can be reinitialized below
     optionsExtraInfoRef.current = {};
-  }, [options]);
+    initializeExtraInfo();
```
mayank99, line 188: I don't think we need this change. When useEffect clears extra info, it will automatically run the logic below which checks `Object.keys(optionsExtraInfoRef.current).length === 0`, so no extra info needed.

Lets remove this function and bring back the original order

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -278,9 +292,14 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
       mounted.current = true;
       return;
     }
-    const value = options[selectedIndex]?.value;
-    onChangeProp.current?.(value);
-  }, [options, selectedIndex]);
+    const currentValue = optionsRef.current[selectedIndex]?.value;
+
+    // If current value is the same as selected value, do not call user defined onChange
```
mayank99, line 297: nit: This comment is redundant as there is a similar comment right above the `useEffect`.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -278,9 +292,14 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    if (currentValue === valuePropRef.current || selectedIndex === -1) {
+      return;
+    }
```
mayank99, line 300: I think this check is also not needed now that we don't have `options` or `value` in the deps array

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,47 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+
+  // Latest value of the valueProp prop
+  const valuePropRef = React.useRef(valueProp);
+  valuePropRef.current = valueProp;
+
   // Latest value of the onChange prop
   const onChangeProp = React.useRef(onChange);
   React.useEffect(() => {
     onChangeProp.current = onChange;
   }, [onChange]);
```
mayank99, line 178: I like the changes you have done with the options and vaue refs, and it made me realize we don't need this useEffect either. We _always_ want the latest value.

```suggestion
  onChangeProp.current = onChange;
```

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -153,32 +153,47 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  // Function to initialize extra info
+  const initializeExtraInfo = React.useCallback(() => {
+    optionsRef.current.forEach((option, index) => {
+      optionsExtraInfoRef.current[getOptionId(option, id)] = {
+        __originalIndex: index,
+      };
+    });
+  }, [id]);
+
+  // Latest value of the valueProp prop
```
mayank99, line 170: nit: there is no `valueProp` prop

```suggestion
  // Latest value of the value prop
```

mayank99 (commented):
The bug is fixed but there is a lot of extra code. I think all we really need is `valuePropRef` and `optionsRef` because previously those were running the useEffects unnecessarily. Everything else looks unnecessary (from my testing)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: This check is very much needed because user defined onChange is being called two times:
1. User programmatically changes valueProp and useEffect with [options, valueProp] changes selectedIndex. 
2. This useEffect is called when selectedIndex change and calls user defined onChange.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed and restored original order

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: But you removed this check in the previous PR, does it not reintroduce that bug?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Added new private hook to save latest value.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Previous check used not "ref'ed" value of valueProp. That's why it caused infinite loop. Now it should work as expected.
I probably missed this double onChange call in previous PR, so fixing my own bug.

mayank99 (approved):
LGTM 🚀

bentleyvk (approved):

PR #746, opened 2022-07-22
chore: Use published docker image

veekeys:
Same as in https://github.com/iTwin/iTwinUI/pull/716

Published 2 separate images for AMD and ARM architecture. That means, whenever you need to update those, need to have 2 different machines, build locally and publish using `push-docker` command.
Couldnt make the multi-arch build successful.

image


apps/storybook/scripts/run-tests.js
```diff
@@ -13,7 +13,7 @@ const dockerProcess = spawn(
   // -w /e2e - makes `e2e` the working directory
   // scripts/entrypoint.sh - entrypoint script to run
   // args - cli args to forward to entrypoint scripts, e.g. for testing only one component using --spec Component.test.ts
-  `docker run --rm --entrypoint /bin/bash -v "${__dirname}/../../..":/e2e -w /e2e itwinui/cypress:latest apps/storybook/scripts/entrypoint.sh ${args}`,
+  `docker run --rm --entrypoint /bin/bash -v "${__dirname}/../../..":/e2e -w /e2e bentleysystemsinc/itwinui:cypress apps/storybook/scripts/entrypoint.sh ${args}`,
```
mayank99, line 16: Why are we using the same repo for both images? What happens if we need to update one of these images to a newer version? We overwrite the same tag?

It would make more sense to have `bentleysystemsinc/itwinui-cypress:latest`.

mayank99 (approved):

PR #747, opened 2022-07-25
feat(Slider): Adding vertical orientation support

### LINKED ISSUE #744: Slider: Support for vertical slider
#### Feature
[Issue #699](https://github.com/iTwin/iTwinUI/issues/699) in the [iTwinUI repo](https://github.com/iTwin/iTwinUI) requested a vertical slider. [PR #707](https://github.com/iTwin/iTwinUI/pull/707) tries to fix the previously mentioned issue.

The request in this issue is to implement that vertical slider that is in the iTwinUI repo in this iTwinUI-react repo

#### Examples

https://user-images.githubusercontent.com/45748283/180448560-d753963b-9820-4a5c-ad65-9350dc4c389e.mp4

r100-stack:
Closes #744. I am making this PR Ready for Review since the coding part is complete. I hope to get your advice and help on why five vertical slider tests are failing. I have added comments for my questions below.

## Checklist

  - [x] Pass these new tests (Last 5 vertical unit tests are not passing)

## TODOs

- [x] Fix the unexpected bug of vertical slider still displaying as horizontal slider. (Some change I made in the CSS repo might have broken something here)
- [x] Complete and remove all TODOs
- [x] Remove unnecessary vertical slider tests (tests that aren't vertical specific and have already been tested for horizontal)

## Screenshots/Videos

https://user-images.githubusercontent.com/45748283/181840510-3bba7266-639e-4535-a70e-ed3bdce2bd69.mp4


apps/storybook/src/Slider.stories.tsx
```diff
@@ -20,19 +20,36 @@ export default {
...
+      style={{
+        ...(args.orientation == 'vertical'
+          ? { height: '400px', width: 'fit-content' }
+          : {}),
+      }}
+    >
+      {element}
+    </div>
+  );
+};
```
rohan-kadkol, line 41: I have a `sliderWrapper()` method that wraps the slider in container of `height: 400px` when the orientation is vertical. I have it hardcoded in the example to 400px. Is this okay? I tried using a height of window.innerHeight, but it doesn't refresh/rerender on dimension changes (of the storybook iframe or even the browser). `height: 100%` also doesn't work. Can you give any ideas to avoid the hardcoded height? Or is the hardcoding fine?

apps/storybook/src/Slider.stories.tsx
```diff
@@ -65,10 +82,19 @@ MultiThumbsAllowCrossing.args = {
...
+  const thumbProps = args.thumbProps();
+  thumbProps.style = {
+    ...thumbProps.style,
+    ...(args.orientation === 'horizontal' // TODO: Is there a way to get the args inside WithCustomThumb.args? Because just to gets args, I have to move those lines of code from WithCustomThumb.args to WithCustomThumb
+      ? { transform: 'translateX(-19.2px)', top: 0 }
+      : { transform: 'translate(-25%, 50%)' }),
+  };
+  args.thumbProps = () => thumbProps;
+  return sliderWrapper(<Slider {...args} />, args);
 };
```
rohan-kadkol, line 94: I am calculating the `transform` and `top` properties that depend on the orientation. I am doing this in the `WithCustomThumb: Story<SliderProps> = ...` part. Is there a way to do this in the `WithCustomThumb.args` part? I don't think that's possible because I think `WithCustomThumb.args` can't change, but I wanted to confirm.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -65,10 +82,19 @@ MultiThumbsAllowCrossing.args = {
...
 
 WithCustomThumb.args = {
+  // TODO: Should we remove it from the function since it cannot be edited in storybook?
   thumbProps: () => {
     return {
       style: {
```
rohan-kadkol, line 100: Just to confirm, with the way the code was, `thumbProps` cannot be edited from Storybook. It just This is because `thumbProps` is a function that returns an object and is not the object itself. Is that okay? Or should I unwrap `thumbProps` to an object and then rewrap it into a function before passing the args into the slider?

image

Example code:
```javascript
let thumbProps = args.thumbProps;
  thumbProps = {
    ...thumbProps,
    style: {
      ...thumbProps.style,
      ...(args.orientation === 'horizontal'
        ? { transform: 'translateX(-19.2px)', top: 0 }
        : { transform: 'translate(-25%, 50%)' }),
    },
  };
  args.thumbProps = () => thumbProps; // Rewrap the object into a method
```

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -437,6 +467,8 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
         ref={ref}
         className={cx(
           'iui-slider-component-container',
+          { 'iui-slider-horizontal': orientation === 'horizontal' }, // TODO: Is a ternary operator possible to select only horizontal or vertical?
+          { 'iui-slider-vertical': orientation === 'vertical' },
           { 'iui-disabled': disabled },
           className,
         )}
```
rohan-kadkol, line 474: Is there a way to use a ternary operator here? I was thinking, in case there is a typo and on both lines 470 and 471, we write `orientation === 'horizontal'`, then both `iui-slider-horizontal` and `iui-slider-vertical` classes will be included. In a ternary operator, this cannot happen, because it always only does either the true case or the false case.

packages/iTwinUI-react/src/core/Slider/Track.tsx
```diff
@@ -46,14 +46,15 @@ export type TrackProps = {
   sliderMin: number;
   sliderMax: number;
   values: number[];
+  orientation: Orientation; // TODO: Confirm if need documentation for this line
```
rohan-kadkol, line 49: I noticed that these variables did not have any documentation. Variables in `Slider.tsx` had documentation. So I just wanted to confirm whether not having any documentation here was alright or not.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -20,19 +20,36 @@ export default {
...
+    orientation: 'horizontal',
   },
 } as Meta<SliderProps>;
 
+const sliderWrapper = (element: ReactNode, args: object) => {
+  return (
+    // TODO: Is there a way to avoid the hardcoded height of 400px? height of 100% does not work
+    <div
+      className='slider-wrapper'
+      style={{
```
rohan-kadkol, line 32: On line 27, I had to change the type of args from `any` to `object`. This was because of the below error when `yarn lint` was run by GitHub actions. Why am I receiving this error? Why is any wrong here?

image

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -23,6 +23,15 @@ const createBoundingClientRect = (
...
 });
 
+// TODO: Is it needed to wrap all vertical aliders in this height: 500px <div>? Because without it, the slider has no height
+const verticalSliderWrapper = (verticalSlider: ReactNode) => {
+  return (
+    <div style={{ height: '500px', width: 'fit-content' }}>
+      {verticalSlider}
+    </div>
+  );
+};
```
rohan-kadkol, line 33: Since the slider expands to the height of the container, the slider would have no height if the container doesn't have an explicit height. Thus, I am wrapping all vertical sliders in a `<div>` with an explicit height. Is this alright? Or there is a better way?

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -532,6 +1131,53 @@ it('should move thumb when pointer down on rail', () => {
...
+      clientY: 30,
+    });
+  });
+
+  const thumb = container.querySelector('.iui-slider-thumb') as HTMLDivElement;
+  expect(thumb.getAttribute('aria-valuenow')).toEqual('30');
+
+  expect(handleOnChange).toHaveBeenCalledWith([30]);
+  expect(handleOnUpdate).toHaveBeenCalledWith([30]);
+});
```
rohan-kadkol, line 1179: I was having trouble understanding what this test actually does and why is it failing for the vertical slider case. The below four other vertical slider test cases that have to do something with `clientX` or `clientY` too are also failing because of the same reason. I am trying to debug it in more detail, but is there any mistake that stands out? Do you have any advice?

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -458,6 +983,30 @@ it('should show tooltip on thumb hover', () => {
...
+  });
+  expect(document.querySelector('.iui-tooltip')?.textContent).toBe('50');
+
+  // TODO: What is data-tippy-root? What is the square bracket notation?
+  const tippy = document.querySelector('[data-tippy-root]') as HTMLElement;
+  act(() => {
+    fireEvent.mouseLeave(thumb);
+  });
+  expect(tippy).not.toBeVisible();
+});
```
rohan-kadkol, line 1008: What is tippy here? What is the meaning of the square brackets notation?

apps/storybook/src/Slider.stories.tsx
bentleyvk: It is not bad to hardcode height in stories. Check `SideNavigation.stories.tsx` how to wrap stories properly by using `decorators` 
image

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -24,9 +24,23 @@ export type TrackDisplayMode =
   | 'odd-segments'
   | 'even-segments';
 
-const getPercentageOfRectangle = (rect: DOMRect, pointer: number) => {
-  const position = getBoundedValue(pointer, rect.left, rect.right);
-  return (position - rect.left) / rect.width;
+/**
+ * Determines the orientation of the vertical slider.
+ */
+export type Orientation = 'horizontal' | 'vertical';
```
bentleyvk, line 30: We usually don't move such types to a separate exported type, instead just inline them in the props and use `Pick<SliderProps, 'orientation'>`

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -405,9 +433,11 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
         return (
           <div className='iui-slider-ticks'>
             {tickLabels.map((label, index) => (
-              <span key={index} className='iui-slider-tick'>
-                {label}
-              </span>
+              // TODO: Is passing index alright to pass as a key? I guess it is since there's nothing better, but just confirming.
```
bentleyvk: Yeah, it is not recommended to use index as a key but sometimes there is no better property to use. Here you could combine index with label ``{`${index-label}`}`` if label was a string but it is `React.ReactNode`.

packages/iTwinUI-react/src/core/Slider/Slider.tsx
bentleyvk: It is ok do as you have here since only can be horizontal or vertical. And Typescript wouldn't allow to type any other value here.

packages/iTwinUI-react/src/core/Slider/Thumb.tsx
```diff
@@ -144,7 +153,12 @@ export const Thumb = (props: ThumbProps) => {
         {...rest}
         data-index={index}
         ref={thumbRef}
-        style={{ ...style, left: `${leftPercent}%` }}
+        style={{
+          ...style,
+          ...(orientation === 'horizontal'
+            ? { left: `${leftPercent}%` }
```
bentleyvk, line 159: Rename `leftPercent` to something more generic like `positionPercent` or something better.

packages/iTwinUI-react/src/core/Slider/Track.tsx
```diff
@@ -80,7 +81,17 @@ export const Track = (props: TrackProps) => {
...
                   className='iui-slider-track'
-                  style={{ left: `${leftPercent}%`, right: `${rightPercent}%` }}
+                  style={{
+                    ...(orientation === 'horizontal'
+                      ? {
+                          left: `${leftPercent}%`,
+                          right: `${rightPercent}%`,
+                        }
+                      : {
+                          top: `${rightPercent}%`,
```
bentleyvk, line 91: Same as in Thumb, would be nice to rename to more generic name.

packages/iTwinUI-react/src/core/Slider/Track.tsx
bentleyvk: Since this component is only for internal use and props are kind of obvious, you don't need to always add comments.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -213,6 +435,24 @@ it('should take class and style', () => {
   expect(slider.style.width).toBe('350px');
 });
 
+it('should take class and style (vertical)', () => {
```
bentleyvk, line 438: Not needed as it is not vertical orientation specific and is already tested.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -229,6 +469,25 @@ it('should take railContainerProps', () => {
   expect(railContainer.style.marginRight).toBe('8px');
 });
 
+it('should take railContainerProps (vertical)', () => {
```
bentleyvk, line 472: Not needed as it is not vertical orientation specific and is already tested.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -237,10 +496,25 @@ it('should render tick marks', () => {
     />,
   );
   assertBaseElement(container);
+  // TODO: Do we also need to check the content of the slider ticks?
+  expect(container.querySelectorAll('.iui-slider-tick').length).toBe(5);
+});
+
+it('should render tick marks (vertical)', () => {
```
bentleyvk, line 503: Not needed as it is not vertical orientation specific and is already tested.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -251,19 +525,54 @@ it('should render custom tick marks as defined by ReactNode.', () => {
...
+  expect(container.querySelector('.custom-tick-mark')).toBeTruthy();
+});
+
 it('should render single track', () => {
   const { container } = render(<Slider values={defaultSingleValue} />);
   assertBaseElement(container);
   expect(container.querySelectorAll('.iui-slider-track').length).toBe(1);
 });
 
+it('should render single track (vertical)', () => {
```
bentleyvk, line 548: Not needed as it is not vertical orientation specific and is already tested. You should check whether other tests are really needed.

bentleyvk (commented):
I will finish to review tommorow.

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Thanks for the idea!

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Thanks

packages/iTwinUI-react/src/core/Slider/Track.tsx
rohan-kadkol: **[DONE]**

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: I tried implementing what you mentioned but I'm not sure if I've done it correctly. Can you please confirm it by looking the commit  with hash `628f8b311c1d8eaeb5ef695bd1de0883a02d8d44`?

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Thanks

packages/iTwinUI-react/src/core/Slider/Thumb.tsx
rohan-kadkol: I renamed `leftPercent` to `lowPercent` and `rightPercent` to `highPercent`

packages/iTwinUI-react/src/core/Slider/Track.tsx
rohan-kadkol: Done in [https://github.com/iTwin/iTwinUI-react/pull/747#discussion_r931528071](here)

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: I can remove it, but just to confirm, would it be safer to test everything we tested for the horizontal cases in the vertical also? Would removing some of these tests from the vertical side make it less robust?

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
bentleyvk: tippy is a library that we are using for positioning dynamic components like Dropdown menu or Tooltip. In this case it is testing whether tooltip is not visible after focus was moved out from the thumb. And square brackets notation is used to query elements by their attributes so it is querying `<div data-tippy-root>...</div>`. Here is a bit info what you can do with attribute selectors https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
bentleyvk: We talked yesterday that unit tests are rendered in mocked browser so no CSS or any styling is being taken into account. So you need to mock functions that return sizes or styling.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
bentleyvk: Have you found the answer because I see tests are passing?

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -28,6 +28,7 @@ const createBoundingClientRect = (
  */
 const getBoundingClientRect = Element.prototype.getBoundingClientRect;
 const sliderContainerSize = createBoundingClientRect(10, 0, 1010, 60);
+const sliderContainerVerticalSize = createBoundingClientRect(10, 0, 70, 1000);
 Element.prototype.getBoundingClientRect = () => sliderContainerSize;
```
bentleyvk, line 32: I would recommend to change this block (29-36) to something like this:
```ts
const getBoundingClientRectMock = jest.spyOn(HTMLElement.prototype, 'getBoundingClientRect');
const sliderContainerHorizontalSize = createBoundingClientRect(10, 0, 1010, 60);
const sliderContainerVerticalSize = createBoundingClientRect(10, 0, 70, 1000);

beforeEach(() => {
  getBoundingClientRectMock.mockReturnValue(sliderContainerHorizontalSize);
});

afterAll(() => {
  jest.clearAllMocks();
});
```
And then inside vertical tests you need change mock to:
```ts
getBoundingClientRectMock.mockReturnValue(sliderContainerVerticalSize);
```

apps/storybook/src/Slider.stories.tsx
bentleyvk: Have you forgot to push? 😁

apps/storybook/src/Slider.stories.tsx
```diff
@@ -20,19 +20,36 @@ export default {
   args: {
     thumbMode: 'inhibit-crossing',
     trackDisplayMode: 'auto',
+    orientation: 'horizontal',
```
bentleyvk, line 23: Have one story for vertical case.

packages/iTwinUI-react/src/core/Slider/Slider.tsx
adamhock: `` `iui-slider-${orientation}` ``

Using this line of code will work for slider orientation since orientation will never be undefined

apps/storybook/src/Slider.stories.tsx
adamhock: Change the type here from `object` to `SliderProps` since you're only passing in `SliderProps` with this function. It should also correct the error when trying to access `args.orientation`

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -174,6 +183,13 @@ export type SliderProps = {
    * high-volume of updates will occur when dragging.
    */
   onUpdate?: (values: ReadonlyArray<number>) => void;
+  /**
+   * The orientation of slider
+   * 'horizontal' - Slider is horizontal and fills from left to right. Default.
+   * 'vertical'- Slider is vertical and fills from bottom to top.
+   * @default 'horizontal'
+   */
```
adamhock, line 191: I'd remove the two comments describing what horizontal and vertical are

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -405,9 +428,10 @@ export const Slider = React.forwardRef<HTMLDivElement, SliderProps>(
         return (
           <div className='iui-slider-ticks'>
             {tickLabels.map((label, index) => (
-              <span key={index} className='iui-slider-tick'>
-                {label}
-              </span>
+              <div key={index} className='iui-slider-tick'>
```
adamhock, line 431: This change needs to be reverted

apps/storybook/src/Slider.stories.tsx
```diff
@@ -20,19 +20,36 @@ export default {
...
   },
 } as Meta<SliderProps>;
 
+const sliderWrapper = (element: ReactNode, args: object) => {
+  return (
+    // TODO: Is there a way to avoid the hardcoded height of 400px? height of 100% does not work
+    <div
+      className='slider-wrapper'
```
adamhock, line 31: I don't believe this div really needs a class name

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
```diff
@@ -512,6 +1040,7 @@ it('should move thumb when pointer down on rail', () => {
   const sliderContainer = container.querySelector(
     '.iui-slider-container',
   ) as HTMLDivElement;
+
```
adamhock, line 1043: It seems nitpicky, but make sure you don't make any unnecessary file changes, even if it's as small as an extra space 😄.

adamhock (commented):
It looks good for the most part. I'd definitely remove a lot of those vertical slider tests as well

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: Haha, I've done that mistake before! But this time, I think I did push the code(commit hash: `5bc5bcf7e60b7eb9de27d403d44fc9e6e9d56695`). Are you sure you referred to the latest version or looked at an older version of this code?

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: @adamhock Did you mean something like this? If so, I don't think it's working, unfortunately.
image

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Ohh, right, makes sense, thanks

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Right, thanks for teaching me about that. I removed `verticalSliderWrapper` as it was not needed since there is no CSS taken into account

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Yes, I got them to pass! After you told where the size is being mocked, I understood why the five failing test were failing. It was failing because of the dimension checks mismatch and the mock clicks not being in the correct screen location and hence causing wrong values in the slider. I correct the expected dimensions, put the clicks in the correct `clientY` value that caused the actual slider values match the expected slider value after the click.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Ohh, that's a useful selector, thanks

apps/storybook/src/Slider.stories.tsx
bentleyvk: But you are not using decorators that's why I mistaken it.

bentleyvk (commented):
Ugh, yesterday GitHub showed that I have no pending comments...

packages/iTwinUI-react/src/core/Slider/Slider.tsx
adamhock: You shouldn't need the `: true,` portion of the code.

apps/storybook/src/Slider.stories.tsx
adamhock: I don't see the change, but it could be an issue on my end

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: Ohh, yes, I made this change near line 34 of `Slider.stories.tsx`. Is this my change alright, or I have to make some change to use decorators?

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Yes, I felt so but just wanted to confirm, thanks

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Thanks

apps/storybook/src/Slider.stories.tsx
bentleyvk: Better use decorators as we are using them mostly for this purpose. I also think you don't really need to check whether it is vertical or not, just set height on decorator/wrapper as it won't really cause any problems.

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Ohh, I totally forgot that we can pass classes as plain text also! Yes, after removing it from the curly braces and removing the `: true`, it worked, thanks

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: Yes, sorry about that, I forgot to push it 😁. Please let me know if it is still not visible

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Thanks, was just confirming. Removed this test.

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Thanks, removed this test

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Thanks, removed this test

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Thanks, removed this test

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Yes, that makes sense, removed lines 188 and 189

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: **[DONE]** Ohh, right, I guess I accedentally added a line. Will remember to avoid unnecessary changes

rohan-kadkol:
> Ugh, yesterday GitHub showed that I have no pending comments...

Ohh, I'm not sure why that was. I know I forgot to push a few small changes towards the end, but I guess I may not have pushed many of my commits. Really sorry about that, I'll remember to push commits frequently from now on.

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Ohh, right, thanks for reminding

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Thanks, removed it

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: Added a story at the end called `SomeVerticalExamples`. It contains three vertical sliders as examples. Please let me know if you want me to change the contents of this story as I wasn't sure what would be the best way to demo vertical sliders

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** I looked at the documentation, wow, decorators are very useful. I replaced `sliderWrapper()` with using decorators

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: I replaced the lines as you suggested. To confirm, would it be alright to add `getBoundingClientRectMock.mockReturnValue(sliderContainerVerticalSize);` to only the tests that depend on the dimension of the slider container? As of now, I only added that line to the last five vertical tests since they are the only vertical tests that fail with this line (depend on the dimension of the slider container)

apps/storybook/src/Slider.stories.tsx
bentleyvk: Maybe just have one basic vertical example and name story `Vertical`

apps/storybook/src/Slider.stories.tsx
```diff
@@ -65,6 +81,14 @@ MultiThumbsAllowCrossing.args = {
 };
 
 export const WithCustomThumb: Story<SliderProps> = (args) => {
+  const thumbProps = args.thumbProps();
```
bentleyvk, line 84: Here `thumbProps` expects index. So maybe just undo your changes and disable orientation control for this story.
image

packages/iTwinUI-react/src/core/Slider/Slider.tsx
```diff
@@ -24,9 +24,18 @@ export type TrackDisplayMode =
...
   | 'even-segments';
 
-const getPercentageOfRectangle = (rect: DOMRect, pointer: number) => {
-  const position = getBoundedValue(pointer, rect.left, rect.right);
-  return (position - rect.left) / rect.width;
+const getPercentageOfRectangle = (
+  rect: DOMRect,
+  pointerX: number,
+  pointerY: number,
+  orientation: SliderProps['orientation'], // TODO: Confirm if this usage is correct
```
bentleyvk, line 31: Remove comment.

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: I added one example story named `Vertical` that is just the `Basic` example with `orientation: 'vertical'`

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: I'm not sure why, but I'm not getting that error.

image

Also, `args.thumbProps()` for `WithCustomThumb` does not take any parameter

image

But actually, it might be easier to just allow horizontal orientation for the with custom thumb example. This is because in the horizontal and vertical examples, we need slightly different styling of the thumb. Below is the code example from lines 83-93 from `Slider.stories.tsx`.

```tsx
export const WithCustomThumb: Story<SliderProps> = (args) => {
  const thumbProps = args.thumbProps();
  thumbProps.style = {
    ...thumbProps.style,
    ...(args.orientation === 'horizontal'
      ? { transform: 'translateX(-19.2px)', top: 0 }
      : { transform: 'translate(-25%, 50%)' }),
  };
  args.thumbProps = () => thumbProps;
  return <Slider {...args} />;
};
```

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: Sorry, I added this comment/`TODO` to remind me to ask you if the parameter type I mentioned is correct. Can you please confirm if it is correct? Thanks

bentleyvk: Yup, it's ok.

apps/storybook/src/Slider.stories.tsx
bentleyvk: I would go with horizontal only in order not to make our stories' code too complicated.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -189,3 +219,18 @@ DecimalIncrement.args = {
   step: 2.5,
   values: [25],
 };
+
+export const Vertical: Story<SliderProps> = (args) => {
+  return <Slider {...args} />;
+};
+
+Vertical.args = {
+  ...Basic.args,
```
bentleyvk, line 228: `Basic.args` only has one value, so just hardcode it here as it will be easier for the user to see.

apps/storybook/src/Slider.stories.tsx
```diff
@@ -155,10 +177,18 @@ export const CustomTickNoTooltip: Story<SliderProps> = (args) => {
             display: 'flex',
             alignItems: 'center',
             justifyContent: 'center',
-            marginTop: '20px',
+            ...(args.orientation === 'horizontal'
```
bentleyvk, line 180: I would say to make this only horizontal too in order not to make it too complicated.

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Right, that makes sense

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: I removed the option to change the orientation for `CustomTickNoTooltip`. I was also wondering, should I do the same for `` since that also involves conditionally applying styling? Below is the code:

```tsx
export const WithCustomThumb: Story<SliderProps> = (args) => {
  const thumbProps = args.thumbProps();
  thumbProps.style = {
    ...thumbProps.style,
    ...(args.orientation === 'horizontal'
      ? { transform: 'translateX(-19.2px)', top: 0 }
      : { transform: 'translate(-25%, 50%)' }),
  };
  args.thumbProps = () => thumbProps;
  return <Slider {...args} />;
};
```

packages/iTwinUI-react/src/core/Slider/Slider.tsx
rohan-kadkol: **[DONE]** Thanks

apps/storybook/src/Slider.stories.tsx
bentleyvk: I see you did that change, so all good.

bentleyvk:
`Slider.test.ts-With Custom Thumb.png` might need to be undo.

apps/storybook/src/Slider.stories.tsx
rohan-kadkol: **[DONE]** Thanks, I removed the vertical option from `WithCustomThumb`

image

packages/iTwinUI-react/src/core/Slider/Slider.test.tsx
rohan-kadkol: @bentleyvk I had one quick question. I ended up just keeping five vertical tests near the end. I chose those because they involved clock coordiates which were vertical specific. All the other vertical tests didn't seem as vertical specific and seemed testable by the horizontal slider case. Can you please confirm if just having the five vertical test cases at the end of the file are okay? Or do I need more test cases?

adamhock (approved):
Everything looks good to me. Great job 🎉

bentleyvk (approved):
Great job!

PR #754, opened 2022-07-28
feat(Footer): Allows to use composition to customize Footer as you wish

### LINKED ISSUE #705: Not able to add icon in FooterElement
I'm using Footer component in my react application. I'm creating a custom footer passing FooterElement[] to customElements prop of Footer component. I want to add an icon at the start of the footer. But the definition for FooterElement doesn't allow to do so. It accepts only following:
type FooterElement = {
    title: string;
    url?: string | undefined;
    key?: keyof TitleTranslations | "copyright" | (string & Record<never, never>) | undefined;
}

#### Steps to reproduce
1. Write footer with Footer component like - `<Footer className="footer-class" customElements={() => footerElement} />`
1. Create an array of footerElement as -
`const footerElement: FooterElement[] = [
    {
      title: "xyz",
      url: "url to xyz"
    },
    {
      title: "abc",
      url: "url to abc"
    }    
  ];`

#### Expected behavior
My requirement if that this array should accept app icon.
So, I want above array to be like-
`const footerElement: FooterElement[] = [
    {
      icon:<svgIcon />
    },
    {
      title: "xyz",
      url: "url to xyz"
    },
    {
      title: "abc",
      url: "url to abc"
    }    
  ];`

@FlyersPh9

bentleyvk:
Closes #705  

Until CSS is released you need to link iTwinui-css.
Footer now accepts `children` prop and if it is passed then only its result will be used for rendering.
All visual test images changed because now separators are properly centered.

## Checklist


packages/iTwinUI-react/src/core/Footer/FooterSeparator.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * Footer separator. Recommended to use with `Footer`.
+ */
+export const FooterSeparator = (props: FooterSeparatorProps) => {
+  const { className, ...rest } = props;
+  useTheme();
+
+  return (
+    <li
+      className={cx('iui-legal-footer-separator', className)}
+      role='separator'
```
mayank99, line 25: i think `aria-hidden` would be better as these are not real separators and there are two many of them

packages/iTwinUI-react/src/core/Footer/FooterItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const FooterItem = (props: FooterItemProps) => {
+  const { url, title, children, className, ...rest } = props;
+  useTheme();
+
+  return (
+    <li className={cx('iui-legal-footer-item', className)} {...rest}>
+      {children ? (
+        children
+      ) : url ? (
+        <a href={url} target='_blank' rel='noreferrer'>
```
mayank99, line 37: Do we really need all this extra url handling in this component? children should be enough

packages/iTwinUI-react/src/core/Footer/Footer.tsx
```diff
@@ -28,6 +30,11 @@ export type FooterProps = {
    * Provide localized strings.
    */
   translatedTitles?: TitleTranslations;
+  /**
+   * Custom footer content. If provided, it will render only what you passed.
+   * If function provided, it should accept an array of default footer elements and return `React.ReactNode`.
+   */
+  children?: React.ReactNode | ((elements: FooterElement[]) => React.ReactNode);
```
mayank99, line 37: What does this render function provide that is not possible with just regular children?

bentleyvk: It provides default footer elements else if you want to customize anything a bit, you would need to take them from somewhere.

packages/iTwinUI-react/src/core/Footer/FooterItem.tsx
bentleyvk: At first it was a very basic component with just a class name so I added url and title handling because it seemed kind of duplicated work to do that ourselves.

packages/iTwinUI-react/src/core/Footer/FooterSeparator.tsx
bentleyvk: Changed

packages/iTwinUI-react/src/core/Footer/Footer.tsx
```diff
@@ -67,74 +74,87 @@ const footerTranslations: TitleTranslations = {
...
-      </ul>
-    </footer>
-  );
-};
+    const childrenContent =
+      typeof children === 'function' ? children(elements) : children;
+
+    return (
+      <footer className={cx('iui-legal-footer', className)} {...rest}>
+        <ul className='iui-legal-footer-list'>
```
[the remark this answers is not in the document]
bentleyvk: Changed

packages/iTwinUI-react/src/core/Footer/FooterItem.tsx
mayank99: but now there is no way to customize this link. What if i don't want `target=_blank` or `rel='noreferrer'` in my footer?

bentleyvk: Then use `children` and pass everything yourself.

mayank99: then why even handle this? plus i don't think those attributes make sense as defaults.

if it's just to reduce code in our component, we can make an internal helper.

packages/iTwinUI-react/src/core/Footer/Footer.tsx
bentleyvk: What is the benefit of doing that?

mayank99: if we change something about the default elements in the future, we won't need to think about `children` handling

bentleyvk: We won't be able to make this change without major change anyways.

bentleyvk: Because I don't like this not working code either 
image

mayank99: Not sure I understand that code snippet. Where is that from?

bentleyvk: Changes required to reuse default elements if we want to export them.

bentleyvk: Did something

bentleyvk:
Need to wait before https://github.com/iTwin/iTwinUI-react/pull/747 is merged.

packages/iTwinUI-react/src/core/Footer/Footer.tsx
```diff
@@ -28,6 +31,11 @@ export type FooterProps = {
    * Provide localized strings.
    */
   translatedTitles?: TitleTranslations;
+  /**
+   * Custom footer content. If provided, it will render only what you passed.
+   * If function provided, it should accept an array of default footer elements and return `React.ReactNode`.
```
mayank99, line 36: Comment is out of date now

packages/iTwinUI-react/src/core/Footer/Footer.tsx
```diff
@@ -67,74 +108,75 @@ const footerTranslations: TitleTranslations = {
...
+    const titles = { ...footerTranslations, ...translatedTitles };
+    const translatedElements = defaultFooterElements.map((element) => {
+      if (element.key && titles.hasOwnProperty(element.key)) {
+        const key = element.key as keyof TitleTranslations;
+        return {
+          ...element,
+          title:
+            element.key && titles.hasOwnProperty(element.key)
+              ? titles[key]
+              : element.title,
```
mayank99, line 132: we already checked for this condition so `element.title` will never be used

if typescript is complaining maybe we can use null assertion `!` rather than adding more code

yarn.lock
```diff
@@ -3688,11 +3688,31 @@ ansi-html-community@0.0.8, ansi-html-community@^0.0.8:
...
+  version "2.1.1"
+  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-2.1.1.tgz#c3b33ab5ee360d86e0e628f0468ae7ef27d654df"
+  integrity sha512-TIGnTpdo+E3+pCyAluZvtED5p5wCqLdezCyhPZzKPcxvFplEt4i+W7OONCKgeZFT3+y5NZZfOOS/Bdcanm1MYA==
+
+ansi-regex@^3.0.0:
+  version "3.0.1"
+  resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-3.0.1.tgz#123d6479e92ad45ad897d4054e3c7ca7db4944e1"
+  integrity sha512-+O9Jct8wf++lXxxFc4hc8LsjaSq0HFzzL7cVsw8pRDIPdjKD2mT4ytDZlLuSBZ4cLKZFXIrMGO7DbQCtMJJMKw==
+
+ansi-regex@^4.1.0:
```
mayank99, line 3701: looks like it's not taking yarn resolutions. did you run `yarn` from root?

apps/storybook/src/Footer.stories.tsx
```diff
@@ -107,3 +108,50 @@ export const BottomFixed: Story<FooterProps> = ({ ...rest }: FooterProps) => {
...
+  return (
+    <Footer {...rest}>
+      {(elements) => {
+        return (
+          <Footer.List>
+            {[
+              customItem,
+              ...elements
+                .filter((el) => el.key !== 'copyright')
+                .map((element, index) => {
```
mayank99, line 128: nit: would it be more readable to use `flatMap`? especially since we need to return multiple nodes for each element

packages/iTwinUI-react/src/core/Footer/Footer.tsx
bentleyvk: Changed. Now mentioned `defaultFooterElements`

packages/iTwinUI-react/src/core/Footer/Footer.tsx
bentleyvk: Ups, leftovers, not needed.

yarn.lock
bentleyvk: Forgot that `yarn add` doesn't take care of resolutions. Fixed.

mayank99 (approved):

PR #755, opened 2022-07-28
feat(Header): Update to iTwinUI-css 1.0.0

### LINKED ISSUE #447: Header thumbnail is displaying at a smaller size than expected
When the header is in the large size, if a thumbnail / photo is available it only displays at 16x16.
Screen Shot 2021-11-29 at 9 40 11 AM

#### Expected behavior
When the header is in the large size, the thumbnail / photo should take the full 24x24 space.  It is displaying properly in [the HTML/CSS demo page](https://itwin.github.io/iTwinUI/backstop/minified/header.html).
Screen Shot 2021-11-29 at 9 41 32 AM

adamhock:
- Updated changes made in this [comment](https://github.com/iTwin/iTwinUI/pull/639#issuecomment-1196780223)
- Due to the `iui-button iui-borderless` -> `iui-header-breadcrumb-button` replacement, there are three new classes, `HeaderBasicButton`, `HeaderDropdownButton`, and `HeaderSplitButton` which replace the use of `Button`, `DropdownButton`, and `SplitButton` in `HeaderButton.tsx` respectively.
- Visual tests are not passing because of the icon buttons on the right side of header which will be updated once the Button Component is updated for iTwinUI-css 1.0.0. Any other visual differences are a result of the `box-sizing` changes which are detailed [here](https://github.com/iTwin/iTwinUI/pull/639#issuecomment-1156527016)


bentleyvk:
What the hell? 😁
image

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+import React from 'react';
+import { DropdownMenu } from '../DropdownMenu';
+import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
+import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
+
+import { PolymorphicForwardRefComponent, useTheme } from '../utils';
+import { DropdownButtonProps } from '../Buttons';
+import { useMergedRefs } from '../utils';
+export type HeaderDropdownButtonProps = DropdownButtonProps;
+type HeaderDropdownButtonComponent = PolymorphicForwardRefComponent<
```
bentleyvk, line 15: Add new line space from imports.

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+  'button',
+  HeaderDropdownButtonProps
+>;
+
+/**
+ * Dropdown button component for Header
+ *
```
bentleyvk, line 22: Remove this line. In other components too.

adamhock:
> What the hell? 😁 image

Icon/Avatar refactor is in CSS but not in React. I didn't want to include those changes in this PR

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
adamhock: Removed

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
adamhock: Fixed

packages/iTwinUI-react/src/core/Header/HeaderButton.tsx
```diff
@@ -79,41 +78,56 @@ export const HeaderButton: HeaderButtonComponent = React.forwardRef(
...
       ),
       ref: ref,
       ...(!!menuItems && { menuItems }),
       ...rest,
     } as const;
 
     if (isSplitButton(buttonProps)) {
-      return <SplitButton {...buttonProps} />;
+      return (
+        <li
```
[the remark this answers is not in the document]
adamhock: Fixed. I created the correct button type inside of a `const` and then put that inside of the `<li>` statement. It achieves the same thing

bentleyvk (approved):

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const [isMenuOpen, setIsMenuOpen] = React.useState(false);
+
+    const [menuWidth, setMenuWidth] = React.useState(0);
+    const ref = React.useRef<HTMLLIElement>(null);
+
+    React.useEffect(() => {
+      if (ref.current) {
+        setMenuWidth(ref.current.offsetWidth);
+      }
+    }, [children, size]);
```
veekeys, line 48: I dont think we should react to `size` prop change... actually, changing button size should not do anything.

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+      menuItems,
+      className,
+      menuPlacement = 'bottom-end',
+      size,
+      children,
+      style,
+      title,
+      type = 'button',
+      startIcon,
+      ...rest
```
veekeys, line 34: In here we also have `endIcon` and `styleType`, which are now not working, but if you pass those through on `button` they will be set in HTML and will be as invalid attributes :/
So I think we need to destruct them out of props and not use?
Oh, there is also polymorphism, which is now also not working anymore :/

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export type HeaderDropdownButtonProps = DropdownButtonProps;
+type HeaderDropdownButtonComponent = PolymorphicForwardRefComponent<
+  'button',
+  HeaderDropdownButtonProps
+>;
+
+export const HeaderDropdownButton: HeaderDropdownButtonComponent = React.forwardRef(
+  (props, ref) => {
+    const { menuItems, className, size, children, startIcon, ...rest } = props;
```
veekeys, line 22: same comment about props as in `HeaderSplitButton`

packages/iTwinUI-react/src/core/Header/HeaderBasicButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+        ref={ref}
+        type={type}
+        style={style}
+        {...rest}
+      >
+        {startIcon &&
+          React.cloneElement(startIcon, {
+            className: startIcon.props.className,
+          })}
+        {children}
```
veekeys, line 45: wont we have breaking change now that we do not support `endIcon` ?

veekeys (commented):
Need to carefully review the props hiding under `rest` in destructing.. Now some custom ones are passed on native `button` and they would not work (previously some of them might have worked in some way, as we reused our `Button` component).
Need to be sure, that we do not remove some features (like `endIcon` support) and might need to review the prop types now, so they are correct ones.

packages/iTwinUI-react/src/core/Header/HeaderBasicButton.tsx
elephantcatdog, line 45: In `Button` the children are included like: `{children && <span>{children}</span>}`. Does it not need to be written the same way? Especially for the check that `children` exists?

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+import cx from 'classnames';
+import React from 'react';
+import { DropdownMenu } from '../DropdownMenu';
+import SvgCaretDownSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretDownSmall';
+import SvgCaretUpSmall from '@itwin/itwinui-icons-react/cjs/icons/CaretUpSmall';
+import { PolymorphicForwardRefComponent, useTheme } from '../utils';
+import { DropdownButtonProps } from '../Buttons';
+import { useMergedRefs } from '../utils';
+
+export type HeaderDropdownButtonProps = DropdownButtonProps;
```
elephantcatdog, line 14: `styleType` is a prop in `DropdownButtonProps` but nothing happens with it in `HeaderDropdownButton`. Is that ok?

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
elephantcatdog: This is probably applicable to line 36 of `HeaderDropdownButton` also.

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+    }, [children, size]);
+
+    return (
+      <DropdownMenu
+        menuItems={menuItems}
+        style={{ minWidth: menuWidth }}
+        onShow={() => setIsMenuOpen(true)}
+        onHide={() => setIsMenuOpen(false)}
+      >
+        <button
```
elephantcatdog, line 45: Why do you use `button` instead of `Button` or `HeaderBasicButton` here (and in `HeaderSplitButton`)? More control/fine-tuning?

packages/iTwinUI-react/src/core/Header/HeaderBasicButton.tsx
elephantcatdog: Also applicable to line 55 in `HeaderDropdownButton` and line 68 in `HeaderSplitButton`.

packages/iTwinUI-react/src/core/Header/HeaderButton.tsx
```diff
@@ -79,41 +78,44 @@ export const HeaderButton: HeaderButtonComponent = React.forwardRef(
...
         : undefined,
+      className: className,
       styleType: 'borderless',
-      className: cx(
-        {
-          'iui-header-button': !isSplitButton(props),
-          'iui-header-split-button': isSplitButton(props),
-          'iui-header-dropdown-button':
-            !isSplitButton(props) && isDropdownButton(props),
-          'iui-active': isActive,
```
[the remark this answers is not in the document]
adamhock: Yes. `iui-header-button` class name will only be used when calling `HeaderBasicButton`. Same with split and dropdown. `iui-active` is not used anymore, and it's styling is applied to the `<li>` element, so the class name does not need to be passed to the children

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
adamhock: I have removed the `size` prop as it is not needed for any `HeaderButton` type. Reducing the size of the button can be done by passing `isSlim` into `Header`. The previous `HeaderButton` did not have a `size` prop, so this should not cause any breaking changes

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
adamhock: `styleType` has been removed as a prop since `HeaderButton` only has the `styleType: borderless`. The previous `HeaderButton` did not have a `styleType` prop, so this should not cause any breaking changes.

`endIcon` support has now been added. It is passed into `HeaderBasicButton` and now `HeaderDropdownButton` and `HeaderSplitButton` both use `HeaderBasicButton` instead of the `<button>` element

packages/iTwinUI-react/src/core/Header/HeaderBasicButton.tsx
adamhock: `endIcon` support has been added to `HeaderBasicButton`. The `size` and `styleType` props are not needed for any of the `HeaderButton` types and were not used in the previous `HeaderButton`, so this should cause no breaking changes

adamhock: The children are defined in the `buttonProps` variable inside of `HeaderButton.tsx`, so they do exist and are already wrapped in a span with custom styling using `iui-header-breadcrumb-button-text`.

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
adamhock: `styleType` is not needed for any of the `HeaderButton` types because they are always `borderless`

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
adamhock: You are 100% correct, and this change solved a lot of the other inital issues raised in this PR. Thank you 😄

adamhock:
I have answered the initial review comments. This is the summary

- The previous `HeaderButton` did not allow for `size` and `styleType` props to be passed in. This is because changing the size of the header button should be done by passing in `isSlim` into the `<Header>` props, and the styleType of `HeaderButton` is always `borderless`. That is why those attributes should not be passed into the new `HeaderButton`. I have removed the `size` attribute from each of the Header button types

- `endIcon` is now passed into `HeaderBasicButton`.

- The biggest change made to solve most of these comments is now `HeaderSplitButton` and `HeaderDropdownButton` each use the `HeaderBasicButton` component instead of their own `<button>` elements similar to how `SplitButton` and `DropdownButton` each use `Button`

packages/iTwinUI-react/src/core/Header/HeaderBasicButton.tsx
veekeys: Its not about being used.. Its about the fact, that users could have passed those props.. 
So now if they pass, you spread them over the element (which is invalid)

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
veekeys: amm... so if you plan to reuse `HeaderBasicButton`, you can get rid of destructing explicitly here `styleType`, `size` and leave those in `rest`, so they are passed to `HeaderBasicButton` anyway..
But whats important, is to remove these props from `rest` in `HeaderBasicButton`. So you need to desctruct them and not use, otherwise, you get button rendered in DOM like that
image

So need to review all props which do not work and take care of those.

Also, I do not see here `endIcon` being passed to `HeaderBasicButton` as you mentioned in your comment.

adamhock: `styleType` and `size` are destructed in `HeaderBasicButton` and not used. I decided to just call the variables inside of `HeaderBasicButton` instead of adding eslint ignore to the file.

[Here](https://github.com/iTwin/iTwinUI-react/pull/755/files#diff-d07cba2fdef199f25e834f5c92fe6046a74a905ed2e0e07148b7bbd20b52f6f5R26) is where I destructed `endIcon` inside of `HeaderBasicButton`.

veekeys: What about `disabled` ?

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+        onHide={() => setIsMenuOpen(false)}
+      >
+        <HeaderBasicButton
+          className={cx('iui-header-breadcrumb-button', className)}
+          ref={refs}
+          aria-label='Dropdown'
+          {...rest}
+        >
+          {children}
+          {isMenuOpen ? (
```
veekeys, line 53: so why didnt you `endIcon` prop for this one?

adamhock: `endIcon` is now used in `HeaderDropdownButton`

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
```diff
@@ -54,19 +52,9 @@ export const HeaderSplitButton: HeaderSplitButtonComponent = React.forwardRef(
...
-          onClick={onClick}
-          ref={forwardedRef}
-          type={type}
-          {...rest}
-        >
-          {startIcon &&
-            React.cloneElement(startIcon, {
-              className: startIcon.props.className,
-            })}
+        <HeaderBasicButton onClick={onClick} ref={forwardedRef} {...rest}>
```
elephantcatdog, line 55: You aren't including `startIcon` anymore (or `endIcon`). Why is that?

packages/iTwinUI-react/src/core/Header/HeaderDropdownButton.tsx
```diff
@@ -42,29 +43,27 @@ export const HeaderDropdownButton: HeaderDropdownButtonComponent = React.forward
...
+                aria-hidden
+              />
+            )
+          }
           {...rest}
         >
-          {startIcon &&
-            React.cloneElement(startIcon, {
-              className: startIcon.props.className,
-            })}
```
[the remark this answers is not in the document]
adamhock: The `startIcon` prop isn't destructed in `HeaderDropdownButton` since `HeaderDropdownButton` does not change the `startIcon`, so it is passed to `HeaderBasicButton` through `...rest`

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
adamhock: The `startIcon` isn't being destructed in `HeaderSplitButton`, so it is passed to be destructed in `HeaderBasicButton` through `...rest`

The `endIcon` for the button is being passed through `...rest` as well. The caret icon in `HeaderSplitButton` is handled in the `<button>` element that takes care of the split part of the `HeaderSplitButton`

I also just realized that `onClick` isn't being handled in `HeaderSplitButton` either, so I've updated the code to have it pass through `...rest` as well

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
veekeys: @adamhock how is it going with `disabled` support? are you facing any issues?

elephantcatdog (approved):
Looks good to me! I'd just check with @veekeys that the props are correct.

gretanausedaite (commented):
As @veekeys mentioned previously, currently we can't disable header buttons. 
Other than that code looks ok.

packages/iTwinUI-react/src/core/Header/HeaderSplitButton.tsx
adamhock: `disabled` support does exist for `HeaderBasicButton`. I added `disabled` support for the split menu of `HeaderSplitButton` by destructing it in `HeaderSplitButton` and passing the prop to the `HeaderBasicButton` and `<button>` elements.

The `disabled` state for `HeaderButton` is not the same as it was because the CSS changes for `Header` only apply `cursor: not-allowed` to `iui-header-breadcrumb-button`. The text and background colors are not changed, and we should fix this in CSS

adamhock:
Visual tests have been updated

veekeys:
> Visual tests have been updated

Do you know what happened with `Version` active button thumbnail? Now it is white.
Also weird, that now `Full` header is fitting well and logo with text is visible. Do you know if we changed breakpoints in there?

adamhock:
> Do you know what happened with `Version` active button thumbnail? Now it is white. 

I asked Jon about this last week and he said this design change is intentional

> Also weird, that now `Full` header is fitting well and logo with text is visible. Do you know if we changed breakpoints in there?

I know the Header buttons became less wide as a result of the CSS changes, so maybe that's why the text is now visible?

veekeys (approved):

PR #756, opened 2022-07-29
feat(Table): Style `DropdownMenu` inside of `ActionColumn` in `Table`

### LINKED ISSUE #730: Need ability to pass props to DropdownMenu through ActionColumn
Can't set className for DropdownMenu in ActionColumn in Table. Can't change its style because of this.

#### Expected behavior
Have ability to pass props to DropdownMenu through ActionColumn.

#### Reproduction
https://codesandbox.io/s/itwinui-react-minimal-example-forked-zxyp3w?file=/src/App.tsx

r100-stack:
Co-authored-by: adamhock <adamhock@users.noreply.github.com>

Closes #730.

* Can now pass `className` or `style` as props to `ActionColumn` to style the inner `DropdownMenu`
```tsx
ActionColumn({
  columnManager: {
    dropdownMenuProps: {
      className: '', // To customize styling of DropdownMenu in ActionColumn
      style: {}, // To customize styling of DropdownMenu in ActionColumn
    },
  },
}),
```
* The DropdownMenu has a default `max-height: 315px`. This can be overridden if a custom `style` or `className` is provided
  * **NOTE:** This hardcoded max-height will hopefully be included in the CSS repo's `.iui-scroll` class. The DropdownMenu already uses the `.iui-scroll` class, so migrating to this way is easy.

## Checklist

### `style: {}`: If no custom style/className is passed, the DropdownMenu has a `max-height: 315px`.

image

### `style: { maxHeight: 'unset', backgroundColor: 'purple' }`

image


packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -2727,6 +2727,156 @@ it('should render action column with column manager', () => {
   expect(actionColumn[3].textContent).toBe('View');
 });
 
+it('should render dropdown menu with specified class and style in action column with column manager', async () => {
```
rohan-kadkol, line 2730: Have I placed the new tests in a correct place in the file? Is there a better place within the file to place these tests?

rohan-kadkol, line 2730: Should the new tests be made shorter? Or is the current length alright? I was thinking of making a common variable for the columns, or something similar, but I was not sure if that would be necessary.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -36,6 +36,8 @@ const ACTION_CELL_ID = 'iui-table-action';
  */
 export const ActionColumn = <T extends Record<string, unknown>>({
   columnManager = false,
+  className = '',
+  style = {},
 } = {}) => {
   return {
     id: ACTION_CELL_ID,
```
rohan-kadkol, line 43: To confirm, is this a correct way to add new parameters? Because I was expecting to do something like

```tsx
export const ActionColumn = <T extends Record<string, unknown>>({
  columnManager: boolean,
  className: string,
  style: object,
} = {}) => {
```

i.e. I thought we have to pass the variable types in tsx, but they were giving this error

image

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -99,6 +101,8 @@ export const ActionColumn = <T extends Record<string, unknown>>({
           menuItems={headerCheckBoxes}
           onHide={() => setIsOpen(false)}
           onShow={() => setIsOpen(true)}
+          className={className}
+          style={{ maxHeight: '300px', overflow: 'scroll', ...style }} // Default style (max height) that can be overridden/unset
```
rohan-kadkol, line 105: Are the default values okay for when the `DropdownMenu` when no custom style is provided?

mayank99, line 105: While it's nice that the user can override `style`, this isn't the best way to do it. We should try to add scrolling through CSS. We already have a class for this purpose.

```jsx
className='iui-scroll'
```

And we can also merge the user's name using `import cx from 'classnames'`:
```suggestion
          className={cx('iui-scroll', className)}
```

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -36,6 +36,8 @@ const ACTION_CELL_ID = 'iui-table-action';
  */
 export const ActionColumn = <T extends Record<string, unknown>>({
   columnManager = false,
+  className = '',
+  style = {},
 } = {}) => {
```
mayank99, line 41: I don't think these props make sense as top level arguments of `ActionColumn` because the dropdown is only a small part of the column.

Also, instead of plucking individual props we should probably expose either the whole dropdown component or all of its props.

I'm thinking we could convert `columnManager` into an object that has a property called `dropdownMenuProps`. The end usage could look something like:
```ts
ActionColumn({
  columnManager: {
    dropdownMenuProps: { className: 'my-custom-class' }
  }
})
```

Here's how you would change the type to be polymorphic (accept either a boolean or an object):
```suggestion
export const ActionColumn = <T extends Record<string, unknown>>({
  columnManager = false,
}: {
  columnManager?: boolean | { dropdownMenuProps: PopoverProps }
} = {}) => {
```

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: Thanks for the idea! I will not add the `style` parameter to the `ActionColumn`, but should I at least have a style with `max-height: 300px`? Because without it, the dropdown menu always expands to fit and so there will not be any scrolling

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: Thanks, yes this is a great idea. Just to confirm, are you meaning that I should also remove `className`? Or should I keep className as one of the props but also add `columnManager` as an object of type `PopoverProps`? I wanted to confirm because `PopoverProps` did not have `className` as a property.

mayank99: > `PopoverProps` did not have `className` as a property.

oh then you can use `DropdownMenuProps` but might need to `Omit` some props from it. We definitely want to keep `className` as that's the main reason for this change.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
mayank99: That's interesting. I thought our `.iui-scroll` class includes the max-height, but apparently not. I feel like it _should_ though. Maybe lets add it in `itwinui-css`?

@bentleyvk What do you think?

mayank99: In the interest of not blocking this PR:
- Lets add an inline style that sets max-height to `315` (same as our other dropdown components). 
- Lets open an issue in the `iTwinUI` repo to handle max-height there.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: I actually just realized that I might not have understood the way these `type`s work. Looks like the hierarchy for `PopoverProps` is `PopoverProps -> TippyProps -> className`. `DropdownMenu` uses `PopoverProps` and also uses `CommonProps -> StylingProps -> ClassNameProps -> className`. But I'm not sure if this is correct.

```tsx
export type DropdownMenuProps = {...} & PopoverProps & Omit<CommonProps, 'title'>;
export type PopoverProps = {...} & Omit<TippyProps, 'placement' | 'trigger' | 'visible'>;
export type CommonProps = {...} & StylingProps;
export type StylingProps = {...} & ClassNameProps;

export type ClassNameProps = { className?: string; };
export interface TippyProps extends Partial<Omit<Props, 'content' | 'render'>> {
  ...
  className?: string;
  render?: (
    ...
  ) => React.ReactNode;
}


```

What I ended up doing was to create a new type called `ActionColumnDropdownMenuProps` that extended `DropdownMenuProps`. I omitted the three properties that were being passed by the ActionColumn to the `DowndownMenu`.

```tsx
export type ActionColumnDropdownMenuProps = Omit<
  DropdownMenuProps,
  'menuItems' | 'onHide' | 'onShow'
>;
```

I just pushed my changes. Cann you please confirm if I did everything correctly like you suggested?

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: Thanks, I created an [issue in the iTwinUI repo](https://github.com/iTwin/iTwinUI/issues/733) and added an inline style `maxHeight: 315px` to the `DropdownMenu` in `ActionColumn`. I will edit the new tests to work with this new props structure

mayank99: Make sure to merge user's `style` prop (from `columnManager.dropdownProps`) into it

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
mayank99: that looks fine. two minor suggestions:
- I would also probably omit `children`
- no need to `export` this type, it can stay internal

rohan-kadkol: **[DONE]** Thanks, I made these changes. But just curious for my understanding, what does the `export` do here? What are the benefits of not exporting it here? Would it affect the autocomplete of the available parameters for the `ActionColumnDropdownMenuProps` in any way?

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: **[DONE]** Thanks for reminding

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -99,6 +109,18 @@ export const ActionColumn = <T extends Record<string, unknown>>({
           menuItems={headerCheckBoxes}
           onHide={() => setIsOpen(false)}
           onShow={() => setIsOpen(true)}
+          style={{
```
mayank99, line 112: You are currently passing `style` and `className` but missing the rest of the `dropdownMenuProps`.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
mayank99: `export` makes it available outside the file (so it can be `import`ed in another file). not exporting allows us to make changes to it in the future without potentially breaking anything on the outside.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: Oops, right, looks like I might have deleted that. I included all the other props also and included a test to check if other props are also being included from `ActionColumn` to `DropdownMenu`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -2883,6 +2883,66 @@ it('should render dropdown menu with custom styling and override default styling
...
+        },
+        ActionColumn({
+          columnManager: {
+            dropdownMenuProps: {
+              className: 'testing-classname',
+              style: {
+                maxHeight: '500px',
+                backgroundColor: 'purple',
+              },
+              role: 'listbox', // Sample property other than className and style
```
mayank99, line 2914: You could just include this in the other test, but upto you

mayank99 (approved):
Looks good 👍

packages/iTwinUI-react/src/core/Table/Table.test.tsx
mayank99: I think you don't need 3 new tests tbh. You can modify the existing column manager test to check for the inline style and iui-scroll, and you only need one new test for the dropdownMenuProps

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -99,6 +109,21 @@ export const ActionColumn = <T extends Record<string, unknown>>({
           menuItems={headerCheckBoxes}
           onHide={() => setIsOpen(false)}
           onShow={() => setIsOpen(true)}
+          {...(typeof columnManager !== 'boolean'
```
veekeys, line 112: What if instead of duplicating this `typeof` check in 3 places, you would create an object above and reuse it everywhere?
`const dropdownMenuProps = typeof columnManager !== 'boolean' ? columnManager.dropdownMenuProps : {};`

and then just
`...dropdownMenuProps.style`

veekeys (commented):
As I mentioned in other comment, I am concerned about the actual value we are setting here, cause it feels it should be relative to table height (amount of rows, density of table)..
But I am not sure if we could easily calculate this.

Another question, if we add css variable for height, how we expect this to be delivered? Leave a jsdoc about existence of it?

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -6,15 +6,21 @@ import React from 'react';
...
 import { tableResizeStartAction } from '../Table';
 import { SELECTION_CELL_ID } from './selectionColumn';
 import { EXPANDER_CELL_ID } from './expanderColumn';
+import cx from 'classnames';
 
 const ACTION_CELL_ID = 'iui-table-action';
 
+type ActionColumnDropdownMenuProps = Omit<
+  DropdownMenuProps,
+  'menuItems' | 'onHide' | 'onShow' | 'children'
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, I omitted them here because the `ActionColumn` already specifies the values for `menuItems`, `onHide`, and `onShow`.

image

So I felt allowing users to override them might cause some unexpected problems since the expected ActionColumn specified functionality may not happen and lead to state inconsistencies because of `ActionColumn` callbacks not being called when expected.

But I can not omit them if they are useful. Or maybe call both the `ActionColumn`'s onHide and onShow along with the user specified onHide and onShow?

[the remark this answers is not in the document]
veekeys: Yeah, so not override, but extend. You can still keep `isOpen` functionality, but also call users `onHide`/`onShow` if they have provided.
With `menuItems` might be more interesting... Maybe for now we can exclude `menuItems` and `children`. We could add those later, if users need.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: [DONE] Ohh right! Thanks for noticing that

packages/iTwinUI-react/src/core/Table/Table.test.tsx
mayank99: Oh that's a good point. When passing props to a react component in JSX, the data attributes are automatically allowed, but since we are accepting these props as a plain object, TS can't make the connection.

I think the problem might be that `TippyProps` (which `PopoverProps` is based on) doesn't extend `React.ComponentProps`. I wonder if we should try to find a "proper" fix or just allow any arbitrary props using `[k: string]: unknown`. @veekeys What do you think?

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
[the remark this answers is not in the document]
rohan-kadkol: I've been trying this to do for quite a while, but I don't think I'm understand what's going on behind the scenes to get this to work. I keep getting this error when running `yarn build --filter=storybook`.

```bash
src/core/Table/columns/actionColumn.tsx(121,31): error TS2554: Expected 1 arguments, but got 0.
```

My code looks like this


image

I do not know what parameter to provide. When I provide a random number as a param, I get this error.

```bash
src/core/Table/columns/actionColumn.tsx(121,40): error TS2345: Argument of type 'number' is not assignable to parameter of type 'Instance<Props>'.
```

`ActionColumnProps` now looks like this

```tsx
type ActionColumnProps = {
  columnManager: boolean | Omit<DropdownMenuProps, 'menuItems' | 'children'>;
};
```

@mayank99 Can you please help?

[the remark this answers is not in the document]
mayank99: i think you need to forward the parameter:
```ts
onHide={(i) => {
  setIsOpen(false);
  dropdownMenuProps.onHide?.(i);
}}
```

packages/iTwinUI-react/src/core/Table/Table.test.tsx
veekeys: Hmm... We have few components, which do not extend all native elements props and we never really agreed, which path we should follow
- Just extend native elements props on root element
- Allow to pass any prop with `[k: string]: unknown`
- Provide a separate prop for native props, not to flood with all the options so its hard to find our custom props in a huge list

I think we have all 3 options used on different parts in the library.
`[k: string]: unknown` option would seem the easiest if we want to cover only `data` props right now.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Thanks

apps/storybook/src/Table.stories.tsx
```diff
@@ -2953,7 +2953,37 @@ export const ColumnManager: Story<Partial<TableProps>> = (args) => {
...
+            accessor: 'rating',
+          },
+          {
+            id: 'Location',
+            Header: 'Location',
+            accessor: 'location',
+          },
+          ActionColumn({
+            columnManager: {
+              className: '', // To customize styling of DropdownMenu in ActionColumn
```
adamhock, line 2983: If `style` and `className` are optional, why are we passing blank parameters for them when we can just pass `columnManager: true`?

adamhock (commented):
I agree with Mayank about only having one extra test for `dropdownMenuProps`. Other than that and my additional comment, everything looks good

apps/storybook/src/Table.stories.tsx
rohan-kadkol: I left that so that people going through the storybook code can know that they can pass an object too for `columnManager`  to customize the styles instead of just passing a boolean. But I could remove it too if it feels unnecessary

adamhock: I don't think it's necessary. Examples such as `DatePicker` and `DropdownMenu` can be given custom styling and there are unit tests to check for that, but it's not stated anywhere in their stories.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
veekeys: yeah, cleanup tests a little

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -34,9 +42,9 @@ const ACTION_CELL_ID = 'iui-table-action';
  *   ),
  * },
  */
-export const ActionColumn = <T extends Record<string, unknown>>({
-  columnManager = false,
-} = {}) => {
+export const ActionColumn = <T extends Record<string, unknown>>(
+  { columnManager }: ActionColumnProps = { columnManager: false },
```
veekeys, line 46: `{columnManager = false}: ActionColumnProps`

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -6,15 +6,23 @@ import React from 'react';
...
 import { MenuItem } from '../../Menu';
 import { tableResizeStartAction } from '../Table';
 import { SELECTION_CELL_ID } from './selectionColumn';
 import { EXPANDER_CELL_ID } from './expanderColumn';
+import cx from 'classnames';
 
 const ACTION_CELL_ID = 'iui-table-action';
 
+type ActionColumnProps = {
+  [k: string]: unknown;
```
veekeys, line 20: Sure needed? you are not using it anywhere

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: I actually was trying this way because I was getting the error on places where just `ActionColumn()` was used that it expected 1 argument but got 0. That is why I tried changing it to the code I have currently to avoid that error.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: I was just experimenting with putting `[k: string]: unknown;` in every hierarchy level of the `ActionColumnProp` to try to fix [this comment thread](https://github.com/iTwin/iTwinUI-react/pull/756#discussion_r941024460). But unfortunately, wherever I tried to put a data parameter (`data-abc: 'value'`), it did not appear in the tag of the `DropdownMenu`/`ActionColumn` (as suggested [here](https://github.com/iTwin/iTwinUI-react/pull/756#discussion_r940414448)). When I figure out the right syntax, I will remove it from here and put it in the right place.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: > Oh that's a good point. When passing props to a react component in JSX, the data attributes are automatically allowed, but since we are accepting these props as a plain object, TS can't make the connection.
> 
> I think the problem might be that `TippyProps` (which `PopoverProps` is based on) doesn't extend `React.ComponentProps`. I wonder if we should try to find a "proper" fix or just allow any arbitrary props using `[k: string]: unknown`. @veekeys What do you think?

I think I misunderstood this React concept, because I tried passing a testing data value to `columnManager`, `dropdownMenu`, or even directly `ActionColumn`, but I couldn't see the data value being added to the HTML tag. To just test this concept of automatically adding data props, I tried it in a simple [React app](https://codesandbox.io/s/react-playground-forked-q09bbh?file=/index.js) on CodeSandbox. I unfortunately couldn't do it there too. @mayank99 @veekeys Can you please help?

image

apps/storybook/src/Table.stories.tsx
rohan-kadkol: **[DONE]** Thanks

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: **[OUTDATED]**

packages/iTwinUI-react/src/core/Table/Table.test.tsx
mayank99: In that example you are not spreading props from `MyComp` in the rendered `<div>`. It should show up correctly if you do it like this:
```js
const MyComp = (props) => {
 return <div {...props}>MyComp123</div>;
};
```

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[DONE]** Thanks, now we have only one new test and one edited existing test. I changed the existing test to test for the default style (`max-height: 315px`) and className of `iui-scroll`. The new test sees if the passed custom style is added and overrides the default style (`max-height`)

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[DONE]** Thanks. After taking help for Vykintas, we found out that passing the data prop and adding it to the HTML element is not possible due to the following:

* After going up the hierarchy, we noticed that there is a Tippy element (external library) that may not include all remaining props `{...rest}` in the attributes list.
* We also might need to change our `Menu` component, which is not intended in this PR

So for now, the new test just tests for `role: 'listbox'` and not for any data properties.

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: **[DONE]** Thanks. I reverted back line 46 to the code you recommended.

veekeys (approved):

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
rohan-kadkol: **[DONE]** Since we don't need to test with data attributes that were suggested [here](https://github.com/iTwin/iTwinUI-react/pull/756#discussion_r940125334), we don't need to allow other props as of now, and so don't need `[k: string]: unknown;`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[SOLVED BY SELF]** Since now there is just one new test instead of nearly four, I think putting the new test here is alright.

PR #760, opened 2022-08-03
fix(Table): Syncing header and body horizontal scroll state

### LINKED ISSUE #757: Table: Resize after column toggle leads to distorted columns
Using a resizable table with the column manager:
* Resize a column
* Hide and then show a column with the column manager
* Resize any column - the headers no longer line up with the table data

Alternatively:
* Hide a column
* Resize any column
* Show the hidden column
* Resize any column - the headers no longer line up with the table data

This does not seem to occur if the user hasn't resized a column before unhiding a column.

#### Expected behavior
The columns should line up with the headers when resizing after using the column manager.

#### Reproduction
https://codesandbox.io/s/itwinui-react-minimal-example-forked-xbs4jm?file=/src/App.tsx
Column alignment

bentleyvk:
Closes #757 

This one was a really strange bug. Turns out after you enable some column, it won't fully fit and horizontal scrollbar will appear and everything will be scrolled to the left although column manager dropdown menu will still be visible (check GIF) so clicking anywhere else will close the menu and focus to icon button inside header column. This focus will scroll header and that makes header not aligned with the body.
table-header-shift2

So what I did, I fully synced header scroll position with the body. I also made it scroll into column manager button to prevent content shift for the user.

## Checklist


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -752,7 +752,16 @@ export const Table = <
         })}
         {...ariaDataAttributes}
       >
-        <div className='iui-table-header-wrapper' ref={headerRef}>
+        <div
+          className='iui-table-header-wrapper'
+          ref={headerRef}
+          onScroll={() => {
+            if (headerRef.current && bodyRef.current) {
```
bentleyvk, line 759: This change was also needed for keyboard navigation.
Before:
table-header-shift3-before

After:
table-header-shift3

packages/iTwinUI-react/src/core/Table/columns/actionColumn.tsx
```diff
@@ -49,6 +49,16 @@ export const ActionColumn = <T extends Record<string, unknown>>({
     disableReordering: true,
     Header: ({ allColumns, dispatch, state }: HeaderProps<T>) => {
       const [isOpen, setIsOpen] = React.useState(false);
+
+      const buttonRef = React.useRef<HTMLButtonElement>(null);
+      const hasToggledVisibility = React.useRef(false);
+      React.useEffect(() => {
+        if (hasToggledVisibility.current) {
```
bentleyvk, line 56: Kind of a hacky way but I tried to add `scrollIntoView` inside `onClick` but it didn't work because I guess DOM wasn't still changed, also tried to set body scroll inside `Table.tsx` in several places but also didn't work because of the same - DOM wasn't changed yet.

mayank99: Does `queueMicrotask` help?

bentleyvk: `setTimeout` did not help, so I guess this one neither.

bentleyvk: Hm, ok, I needed to refresh the page for my changes to work. So `setTimeout` works but there is visible scroll and `queueMicrotask` also worked, so gonna change to that.

mayank99 (approved):

PR #765, opened 2022-08-04
feat(Stepper): Update to iTwinUI-css 1.0.0

adamhock:
- Updated changes made in this [comment]( https://github.com/iTwin/iTwinUI/pull/639#issuecomment-1185950337)

- Renamed `Wizard` component to `Stepper`
- Moved `WorkflowDiagram` from `Wizard` into its own component
- Moved `Step.tsx` to `utils` so it can be used by both `Stepper` and `WorkflowDiagram`


packages/iTwinUI-react/src/core/Stepper/Stepper.tsx
```diff
@@ -2,78 +2,63 @@
...
  *
  * @example
- * <Wizard
+ * <Stepper
  *  steps=[{name: "Step One"}, {name: "Step Two"}, {name: "Step Three"}]
  *  currentStep={0}
  *  type='long'
  *  />
  */
-export const Wizard = React.forwardRef<HTMLDivElement, WizardProps>(
```
mayank99, line 70: We cannot just remove the existing component or its props as that would constitute a breaking change. So we definitely need to keep the "old" component until we hit 2.0. 

For this PR, I would suggest only changing the classnames inside `Wizard` and making no other changes. Or at the very least don't export `Stepper` and `WorkflowDiagram` publicly until we discuss the public API. Would be much easier to review that way.

adamhock:
A new PR which only includes the CSS changes is here #769

Closing

PR #771, opened 2022-08-10
feat(Dialog): Draggable dialog

bentleyvk:
Partly #614 

dialog-drag-demo

## Checklist


packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -0,0 +1,68 @@
+/*---------------------------------------------------------------------------------------------
```
bentleyvk, line 1: Decided to move drag logic to a separate hook if we need it in the future and it was nice to use it inside dialog.

packages/iTwinUI-react/src/core/Dialog/DialogWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Dialog wrapper used for draggable dialogs in order to make them draggable relatively to this container.
+ * @example
+ * <DialogWrapper>
+ *   <Dialog>
+ *    ...
+ *  </Dialog>
+ * </DialogWrapper>
+ */
+export const DialogWrapper = React.forwardRef<
```
bentleyvk, line 21: It is not `Dialog.Wrapper` because I wanted to have only dialog inner sub-components under `Dialog.*`.

packages/iTwinUI-react/src/core/Dialog/DialogDragContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+export type DialogDragContextProps = {
+  onPointerDown: (e: React.PointerEvent<HTMLElement>) => void;
+};
+
+export const DialogDragContext = React.createContext<
+  DialogDragContextProps | undefined
+>(undefined);
+
+export const useDialogDragContext = () => {
+  const context = React.useContext(DialogDragContext);
```
veekeys, line 16: We have this `useSafeContext` hook, if you wanna reuse...

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
```diff
@@ -39,6 +39,11 @@ export type DialogContextProps = {
    * @default false
    */
   preventDocumentScroll?: boolean;
+  /**
+   * Flag whether dialog is draggable.
```
veekeys, line 43: Maybe add additional info, that you need `DialogWrapper` or put even an example in here?

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -0,0 +1,71 @@
...
+export const useDragAndDrop = (elementRef: React.RefObject<HTMLElement>) => {
+  const grabOffsetX = React.useRef(0);
+  const grabOffsetY = React.useRef(0);
+  const translateX = React.useRef(0);
+  const translateY = React.useRef(0);
+
+  const [transform, setTransform] = React.useState('');
+
+  const onPointerMove = React.useCallback(
+    (event: PointerEvent) => {
```
veekeys, line 23: why here `event: PointerEvent` and below `e: React.PointerEvent<HTMLElement>` ?

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -0,0 +1,71 @@
...
+      if (elementRef.current) {
+        translateX.current = event.clientX - grabOffsetX.current;
+        translateY.current = event.clientY - grabOffsetY.current;
+        elementRef.current.style.transform = `translate(${translateX.current}px, ${translateY.current}px)`;
+      }
+    },
+    [elementRef],
```
veekeys, line 30: I know its linter suggesting it here... but maybe it would make sense to accept `HTMLElement | null` instead of the ref..?

packages/iTwinUI-react/src/core/Dialog/DialogWrapper.tsx
```diff
@@ -0,0 +1,34 @@
...
+ *--------------------------------------------------------------------------------------------*/
+import React from 'react';
+import cx from 'classnames';
+import { useTheme } from '../utils';
+import '@itwin/itwinui-css/css/dialog.css';
+
+export type DialogWrapperProps = React.ComponentPropsWithRef<'div'>;
+
+/**
+ * Dialog wrapper used for draggable dialogs in order to make them draggable relatively to this container.
```
veekeys, line 13: If its realtive to container, shouldnt it be limited not to go outside of the container boundaries? Otherwise, I am not sure, what it means by `relative to container`

veekeys: Also, there is this weird behavior then too:
drag

packages/iTwinUI-react/src/core/Dialog/DialogDragContext.tsx
bentleyvk: It does a different thing. My hook doesn't throw it if component is not under context e.g. custom dialog body.

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
bentleyvk: But you don't `DialogWrapper` to make dialog draggable. Or you mean to mention `DialogWrapper` here?

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
bentleyvk: Because `onPointerMove` is native event handler (added to the document) and `onPointerDown` is React event handler.

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
bentleyvk: What it would give? I am afraid it will make too many unnecessary rerenders because of element changes.

packages/iTwinUI-react/src/core/Dialog/DialogWrapper.tsx
bentleyvk: > If its realtive to container, shouldnt it be limited not to go outside of the container boundaries? Otherwise, I am not sure, what it means by `relative to container`

I thought relative means that it can't go outside the container, e.g. can't go outside the viewer and cover header, sidenav.

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
veekeys: Yeah, sorry, meant leaving info about availability of `DialogWrapper` if you need to have dialog draggable relative to container

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
veekeys: just whenever see ref in deps list, looks like something is wrong..
but I guess leave it, as it works fine.

packages/iTwinUI-react/src/core/Dialog/DialogWrapper.tsx
veekeys: But it does go, just its not visible (why would you need that?)

mayank99: I also think it's weird. I would except the dialog's outer edges to not go outside the container.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -48,14 +49,30 @@ export const DialogTitleBar = Object.assign(
...
+      isDraggable = dialogContext.isDraggable,
       className,
+      onPointerDown: onPointerDownProp,
       ...rest
     } = props;
+
+    const { onPointerDown } = useDialogDragContext();
+    const handlePointerDown = React.useCallback(
+      (event: React.PointerEvent<HTMLDivElement>) => {
+        onPointerDown?.(event);
```
mayank99, line 61: Maybe guard this inside a `defaultPrevented` so that the user can cancel this if they want.

packages/iTwinUI-react/src/core/Dialog/DialogDragContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return {
+    ...context,
```
mayank99, line 18: is this hook even necessary?

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
mayank99, line 61: Maybe guard this inside a `defaultPrevented` so that the user can cancel this if they want.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -109,6 +113,16 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
       onKeyDown?.(event);
     };
 
+    const { onPointerDown, transform } = useDragAndDrop(dialogRef);
+    const handlePointerDown = React.useCallback(
```
mayank99, line 117: Not sure I understand what's going on here. You grab the `onPointerDown` callback from context, wrap it in `handlePointerDown` and pass it back into the provider?

apps/storybook/src/Dialog.stories.tsx
```diff
@@ -0,0 +1,270 @@
...
+      <Button styleType='high-visibility' onClick={() => setIsDialogOpen(true)}>
+        Open Dialog
+      </Button>
+      <DialogWrapper
+        style={{
+          width: '70vw',
+          height: '70vh',
+          border: '1px solid red',
+          marginTop: 12,
+        }}
```
mayank99, line 238: Looking at the dialog wrapper css, it looks like it includes `top`, `left`, `width` and `height` already. Do you remember the reasoning for that? Maybe we intended for the wrapper to be `position: absolute` that the user would place inside their own container?

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -0,0 +1,71 @@
...
+      translateX.current = matrix.m41;
+      translateY.current = matrix.m42;
+
+      grabOffsetX.current = e.clientX - translateX.current;
+      grabOffsetY.current = e.clientY - translateY.current;
+
+      elementRef.current.ownerDocument.addEventListener(
+        'pointermove',
+        onPointerMove,
+        false,
```
mayank99, line 52: did you mean to put `{ once: true }` here as well? `false` does nothing afaik

bentleyvk: Definitely not `{ once: true }`, removed.

apps/storybook/src/Dialog.stories.tsx
bentleyvk: Nope, I don't remember. Aren't you the author of these dialog css changes? 😁

packages/iTwinUI-react/src/core/Dialog/DialogWrapper.tsx
bentleyvk: Dialog now can't go out of bounds, that means that bug is also gone.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: No, I take pass main dialog ref and wrapper bounding rect to the `useDragAndDrop` and take `onPointerDown` from it, then pass it down the context.

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
bentleyvk: Added

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
bentleyvk: Mentioned `DialogWrapper` in comments.

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -11,24 +13,64 @@ import React from 'react';
  * `onPointerDown` - handler that is called when pointer is down and handles all the dragging logic.
  * `transform` - current transform of the element, it is used to preserve drag position when element visibility is being toggled.
  */
-export const useDragAndDrop = (elementRef: React.RefObject<HTMLElement>) => {
+export const useDragAndDrop = (
+  elementRef: React.RefObject<HTMLElement>,
+  containerRect?: DOMRect,
```
veekeys, line 18: document new param

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -11,24 +13,64 @@ import React from 'react';
...
-    (event: PointerEvent) => {
-      if (elementRef.current) {
-        translateX.current = event.clientX - grabOffsetX.current;
-        translateY.current = event.clientY - grabOffsetY.current;
-        elementRef.current.style.transform = `translate(${translateX.current}px, ${translateY.current}px)`;
-      }
-    },
-    [elementRef],
-  );
+  const onPointerMove = React.useRef((event: PointerEvent) => {
```
veekeys, line 41: why using `ref`? Why didnt make same with the callback below ?

veekeys (commented):
Not allowing to drag dialog outside the container looks great!
But now this title bar is bothering me..
Feels like we should have title below the draggable title bar or we need not to allow selection or any other clicks on the title bar if its draggable.
Pay attention to the cursor to see, when I actually click it.
dragg

packages/iTwinUI-react/src/core/Dialog/DialogTitleBar.tsx
```diff
@@ -48,14 +49,32 @@ export const DialogTitleBar = Object.assign(
...
       className,
+      onPointerDown: onPointerDownProp,
       ...rest
     } = props;
+
+    const { onPointerDown } = useDialogDragContext();
+    const handlePointerDown = React.useCallback(
+      (event: React.PointerEvent<HTMLDivElement>) => {
+        if (!event.defaultPrevented) {
+          onPointerDown?.(event);
```
veekeys, line 62: this needs to be switched with the below one.. otherwise that `defaultPrevented` will do nothing and user has no chance to set it (unless he applies event on child, but still.. wont fully work)

bentleyvk: Switched

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
bentleyvk: Just to avoid bugs and memory leaks when deps changes and this function is not being removed from event listeners.

bentleyvk:
> Not allowing to drag dialog outside the container looks great! But now this title bar is bothering me.. Feels like we should have title below the draggable title bar or we need not to allow selection or any other clicks on the title bar if its draggable. Pay attention to the cursor to see, when I actually click it. 

Fixed

veekeys (approved):
Seems to work fine

packages/iTwinUI-react/src/core/Dialog/DialogDragContext.tsx
mayank99: how is it better than just doing `React.useContext(DialogDragContext)` in those two places?

apps/storybook/src/Dialog.stories.tsx
mayank99: I am not the author, these changes came from before (all i did is rename -backdrop to -wrapper). But looking at the html, it seems like we expect the user to provide their own container (with custom sizing) and then put this wrapper inside it (with no sizing).
https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/backstop/tests/dialog.html#L261

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
mayank99: should we just pass the ref down in context then? because this `handlePointerDown` is not used in this component

mayank99: wait i see we are using the `transform` in here. hmm, we could do `dialogMainRef.current.style.transform = transform` in title bar, but maybe that's confusing.

elephantcatdog (commented):
The modal test images look like this too, but do you know why the test images have a yellow border on the dialogs?


The draggable movement/stopping against the sides is really nice. There's a bug when moving the dialogs after resizing the window. The 'box' for the dialog to move around in doesn't seem to update immediately, only after the second click-and-move. Hopefully the video shows what I mean. This happens for draggable in or out of a container.

https://user-images.githubusercontent.com/37936169/185249319-8c5c4b5e-ea07-47b6-8ad5-daf0c11c74f8.mp4

bentleyvk:
> The modal test images look like this too, but do you know why the test images have a yellow border on the dialogs?
> 
> The draggable movement/stopping against the sides is really nice. There's a bug when moving the dialogs after resizing the window. The 'box' for the dialog to move around in doesn't seem to update immediately, only after the second click-and-move. Hopefully the video shows what I mean. This happens for draggable in or out of a container.

Good catch! I hope you haven't found it because I had to redo everything to make it work 😁

packages/iTwinUI-react/src/core/Dialog/DialogDragContext.tsx
bentleyvk: I know that context is always truthy and it is just inside code so more comfortable to use.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: We can't just apply `transform` to the html element because if dialog is hidden by setting `isOpen={false}` then element disappears and `transform` is lost.

elephantcatdog (approved):
😁 It works beautifully now!

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -117,13 +132,18 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
             'iui-dialog-default': styleType === 'default',
             'iui-dialog-full-page': styleType === 'fullPage',
             'iui-dialog-visible': isOpen,
+            'iui-dialog-draggable': isDraggable,
```
mayank99, line 135: i think this needs a CSS fix. it unnecessarily applies `left` and `top` which impacts positioning when resizing

https://user-images.githubusercontent.com/9084735/185697821-c3d9de81-e57b-4072-94f1-c50dceac1b36.mov

mayank99:
Thinking about the API a bit more, I think it makes sense to reuse `<Dialog>` for the wrapper. In #776, there is already a `<div>` added in `<Dialog>` so here we can just add the classname.

Additionally, re: my concern mentioned above in https://github.com/iTwin/iTwinUI-react/pull/771#discussion_r946069647, I think the wrapper should have `position: absolute`, not `position: relative` so that it doesn't occupy any space. Because otherwise the user needs to put their app content (unrelated to dialog) inside the dialog wrapper, which doesn't make sense.

Lastly, I think we'll need a flag to toggle whether the dialog should be relative to the container or to the page. This flag could toggle the class on the wrapper and also on the backdrop (changing `position: absolute` vs `position: fixed`) to fix https://github.com/iTwin/iTwinUI/issues/738.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: But if you don't move the dialog it will still resize according to the screen. So decided that it would do the same after dragging. Users can set concrete size if they wish.

bentleyvk:
> Thinking about the API a bit more, I think it makes sense to reuse `<Dialog>` for the wrapper. In #776, there is already a `<div>` added in `<Dialog>` so here we can just add the classname.
> 
> Additionally, re: my concern mentioned above in [#771 (comment)](https://github.com/iTwin/iTwinUI-react/pull/771#discussion_r946069647), I think the wrapper should have `position: absolute`, not `position: relative` so that it doesn't occupy any space. Because otherwise the user needs to put their app content (unrelated to dialog) inside the dialog wrapper, which doesn't make sense.

With `position: absolute` our backdrop won't work after we change it to have `position: absolute`, so wrapper must have `position: relative`.

> With that in mind, I think we'll need a flag to toggle whether the dialog should be relative to the container or to the page. This flag could toggle the class on the wrapper and also on the backdrop (changing `position: absolute` vs `position: fixed`) to fix [iTwin/iTwinUI#738](https://github.com/iTwin/iTwinUI/issues/738).
> 
> Lastly, if we want to keep `DialogWrapper` as a utility (for multiple dialogs or whatever), we can do that but I don't think it should be _required_, since it's simpler to toggle a boolean prop and will result in one less element in the DOM.

idk, for me wrapper still feels like a separate component that should remain always as a separate one, not a prop because we are also doing composition here so prop doesn't make sense to me.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
mayank99: > But if you don't move the dialog it will still resize according to the screen. So decided that it would do the same after dragging.

What if we remove the `left`/`right` in CSS and calculate the initial position using JS (only for draggable dialog)? I think it's nicer to preserve the position of draggable dialogs when the containers get resized.

> Users can set concrete size if they wish.

How would this help?

bentleyvk: Draggable now has static size

bentleyvk:
> > With `position: absolute` our backdrop won't work after we change it to have `position: absolute`, so wrapper must have `position: relative`.
> 
> I think `position: absolute` works as long as the parent has `position` set to anything other than `static`, so this should not be an issue.
>
> I recorded a video to demonstrate why it doesn't make sense for the wrapper to occupy space.
> 
>  Dialog._.iTwinUI.-.29.August.2022.mp4 
> > idk, for me wrapper still feels like a separate component that should remain always as a separate one, not a prop because we are also doing composition here so prop doesn't make sense to me.
> 
> How does it not make sense to use the same `<div>` that `<Dialog>` already has?
> 
> Composition means constructing the DOM yourself with one element per component, which is exactly what we'd be doing here.
> 
> This
> 
> ```js
> <Dialog isDraggable>
>   <Dialog.Backdrop />
>   <Dialog.Main>...</Dialog.Main>
> </Dialog>
> ```
> 
> would map to this:
> 
> ```
> <div class='iui-dialog-wrapper'>
>   <div class='iui-backdrop'></div>
>   <div class='iui-dialog iui-dialog-default iui-dialog-draggable'>...</div>
> </div>
> ```

I think it works because you are still wrapping everything with a div that has `position: relative` and backdrop is positioned relatively to it. So in the end users still need to have `position: relative` on their end.
I still don't like using wrapper inside `Dialog` and having some prop, sorry.

mayank99:
> I think it works because you are still wrapping everything with a div that has `position: relative` and backdrop is positioned relatively to it.

This is not true according to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#:~:text=If%20the%20position%20property%20is%20absolute%2C%20the%20containing%20block%20is%20formed%20by%20the%20edge%20of%20the%20padding%20box%20of%20the%20nearest%20ancestor%20element%20that%20has%20a%20position%20value%20other%20than%20static%20(fixed%2C%20absolute%2C%20relative%2C%20or%20sticky).)

> So in the end users still need to have `position: relative` on their end.

This is true but it's much easier to do than using `position: absolute` to place all custom content (like how the `<Button>` in the CSS demo is currently placed) or adding `pointer-events: auto` on everything.

> I still don't like using wrapper inside `Dialog` and having some prop, sorry.

I explained to you what the problem is, recorded a video, demonstrated how it actually makes sense with composition, and even suggested having `DialogWrapper` as an additional separate component if you want to keep it. If, after all that, your response is "I don't like it", then I'm not sure what else to say.

> Then you could say that we need to change `InformationPanel` and its wrapper...

InformationPanelWrapper is much different because it only sits on top of one element (usually table) and has no problematic side effects (unlike `pointer-events: none` in the case of DialogWrapper).

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: Because of `left` and `top` it has size what it is. So without `left`, `top`, `max-width`, `max-height` it will span through entire screen.

bentleyvk:
> > I think it works because you are still wrapping everything with a div that has `position: relative` and backdrop is positioned relatively to it.
> 
> This is not true according to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#:~:text=If%20the%20position%20property%20is%20absolute%2C%20the%20containing%20block%20is%20formed%20by%20the%20edge%20of%20the%20padding%20box%20of%20the%20nearest%20ancestor%20element%20that%20has%20a%20position%20value%20other%20than%20static%20(fixed%2C%20absolute%2C%20relative%2C%20or%20sticky).)
> 
> > So in the end users still need to have `position: relative` on their end.
> 
> This is true but it's much easier to do than using `position: absolute` to place all custom content (like how the `<Button>` in the CSS demo is currently placed) or adding `pointer-events: auto` on everything.
> 
> > I still don't like using wrapper inside `Dialog` and having some prop, sorry.
> 
> I explained to you what the problem is, recorded a video, demonstrated how it actually makes sense with composition, and even suggested having `DialogWrapper` as an additional separate component if you want to keep it. If, after all that, your response is "I don't like it", then I'm not sure what else to say.
> 
> > Then you could say that we need to change `InformationPanel` and its wrapper...
> 
> InformationPanelWrapper is much different because it only sits on top of one element (usually table) and has no problematic side effects (unlike `pointer-events: none` in the case of DialogWrapper).

I was planning to after all implement your changes but didn't like the part that wrapper is always there when `isOpen` is `false`:
image

bentleyvk:
> > I think it works because you are still wrapping everything with a div that has `position: relative` and backdrop is positioned relatively to it.
> 
> This is not true according to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#:~:text=If%20the%20position%20property%20is%20absolute%2C%20the%20containing%20block%20is%20formed%20by%20the%20edge%20of%20the%20padding%20box%20of%20the%20nearest%20ancestor%20element%20that%20has%20a%20position%20value%20other%20than%20static%20(fixed%2C%20absolute%2C%20relative%2C%20or%20sticky).)
> 
> > So in the end users still need to have `position: relative` on their end.
> 
> This is true but it's much easier to do than using `position: absolute` to place all custom content (like how the `<Button>` in the CSS demo is currently placed) or adding `pointer-events: auto` on everything.
> 
> > I still don't like using wrapper inside `Dialog` and having some prop, sorry.
> 
> I explained to you what the problem is, recorded a video, demonstrated how it actually makes sense with composition, and even suggested having `DialogWrapper` as an additional separate component if you want to keep it. If, after all that, your response is "I don't like it", then I'm not sure what else to say.
> 
> > Then you could say that we need to change `InformationPanel` and its wrapper...
> 
> InformationPanelWrapper is much different because it only sits on top of one element (usually table) and has no problematic side effects (unlike `pointer-events: none` in the case of DialogWrapper).

Did the changes, not sure they are 100% what you wanted, be good if you could check.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
mayank99: Not sure I understand. The default dialog styles would still supply the max-width and max-height. I'm only suggesting removing the `max-height`/`max-width` from the draggable dialog styles.

Check out the styles I have unchecked in this screenshot. It still looks the same without all these styles.


So all that's needed in JS is to calculate the initial `transform` for centering the dialog initially.

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -117,13 +148,19 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
...
           },
           className,
         )}
         role='dialog'
         ref={refs}
         onKeyDown={handleKeyDown}
         tabIndex={-1}
+        style={{
+          transform,
```
mayank99, line 160: It looks like this is being applied to the non-draggable dialog too

apps/storybook/src/Dialog.stories.tsx
```diff
@@ -0,0 +1,272 @@
...
+  const secondaryButtonHandle = (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    action('Secondary button')(event);
+    closeDialog();
+  };
+
+  return (
+    <>
+      <Button styleType='high-visibility' onClick={() => setIsDialogOpen(true)}>
```
mayank99, line 229: Maybe put this button inside the custom div so that it's obvious that the dialog-wrapper will overlay the user content.


Could also add some text explaining that the user needs to specify `position: relative`.

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
```diff
@@ -53,6 +54,7 @@ export const DialogBackdrop = React.forwardRef<
       isVisible={isVisible}
       ref={refs}
       onMouseDown={handleMouseDown}
+      style={{ pointerEvents: 'auto', ...style }}
```
mayank99, line 57: Since https://github.com/iTwin/iTwinUI/pull/739 is merged into the v1 branch and won't be available for a while, maybe add inline `position: fixed` to this.

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
```diff
@@ -39,6 +39,23 @@ export type DialogContextProps = {
...
+   *
+   * If you want to make dialog draggable relatively to the container, you should use set `relativeTo` to `container`.
+   *
+   * @default false
+   */
+  isDraggable?: boolean;
+  /**
+   * Whether dialog should be positioned relatively to a container or the viewport.
```
mayank99, line 51: Could use more explanation.

```suggestion
   * Whether dialog should be positioned relatively to a container or the viewport.
   *
   * Using `'container'` will absolutely position this dialog relative to the closest positioned ancestor.
   * In other words, you must place the dialog as a child of an element that has `position` set to 
   * something other than `static`, e.g. `position: relative`.
```

packages/iTwinUI-react/src/core/Dialog/DialogBackdrop.tsx
bentleyvk: Added

apps/storybook/src/Dialog.stories.tsx
bentleyvk: Changed

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: Added ternary

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: But you also have to undo `iui-dialog-draggable` changes that I am not sure about. I think those max width and height are intentional @FlyersPh9. I guess I will undo my changes related to this comment tree and fix that later because it needs CSS changes then and I don't want to do so much resetting trough `style`.

mayank99: Lets tackle that in a separate PR then.

mayank99 (approved):

PR #774, opened 2022-08-10
feat(Table): Table shift click support

### LINKED ISSUE #632: Table: Shift Click Support
Table supports control-click for selecting multiple rows. We should add support for shift-click, which would allow users to select a range of rows by shift-clicking the first and last elements in the group.

r100-stack:
Closes #632  

## Checklist

## TODOs

- [x] When shift clicking, the browser also selects the content/text between the last clicked position and the current clicked position. How to stop this from happening? When searching for solutions, I only found answers to disable selection altogether, which is not what we want. **We want to allow selection by mouse hold and select but disallow selection by shift clicks.** How to achieve this?
- [x] Solve bug: Selecting a subrow selects all the rows
[vokoscreenNG-2022-08-29_10-07-30.webm](https://user-images.githubusercontent.com/45748283/187220698-4749ad80-314b-4f8f-9587-87dd4be59f1b.webm)
- [x] Solve bug: `r.toggleRowSelected is not a function` [More details](https://github.com/iTwin/iTwinUI-react/pull/774#issuecomment-1225715013)

#### 1. Holding `Shift` selects rows from the `lastSelectedRow` to the clicked row (or subrows)<br>(`Expandable Subrows` Story)

[vokoscreenNG-2022-09-20_08-49-36.webm](https://user-images.githubusercontent.com/45748283/191262195-68522027-18e9-4483-ab83-9f1d90e2a732.webm)

#### 2. Doing a shift click resets the previous selection (like Windows File Explorer)<br>(`SelectableMulti` Story)

[vokoscreenNG-2022-09-20_08-47-48.webm](https://user-images.githubusercontent.com/45748283/191261778-7aded674-d277-4974-9fc9-ee338c175764.webm)

#### 3. If no row selected before shift clicking, then assume first row is the last selectedRow (like Windows File Explorer)<br>(`Selectable Multi` Story)

[vokoscreenNG-2022-09-20_08-46-04.webm](https://user-images.githubusercontent.com/45748283/191261314-dd97414f-c832-4076-b784-77ef5325967c.webm)

#### 4. Text selection is disabled when shift is held (if `isSelectable && selectionMode === 'multi'`) (`Selectable Multi` Story)

[vokoscreenNG-2022-09-20_08-40-30.webm](https://user-images.githubusercontent.com/45748283/191260788-f22d187c-32da-4cf9-b917-2f8f78bfcf06.webm)

#### 5. Can select from lower to upper rows also (instead of upper to lower) (`Full 2` Story)

[vokoscreenNG-2022-09-20_08-51-25.webm](https://user-images.githubusercontent.com/45748283/191262858-7e18fe26-3a2a-47fd-8140-f1f5fc17dd10.webm)

<details>
<summary>Click to see old screenshot</summary>
https://user-images.githubusercontent.com/45748283/184366150-0d3ee646-b457-4928-9a7b-89f815a79048.mp4
</details>


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -52,8 +52,10 @@ import {
 } from './actionHandlers';
 import VirtualScroll from '../utils/components/VirtualScroll';
 import { SELECTION_CELL_ID } from './columns';
+import { onShiftSelectHandler } from './actionHandlers/selectHandler';
```
gretanausedaite, line 55: Move this import to other `actionHandlers` import.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,88 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+      while (nextRow == null && parentId != null) {
+        nextRow = instance.rowsById[getGreaterId(parentId)];
+        parentId = getParentId(parentId);
+      }
+    }
+
+    return nextRow;
+  };
+
+  resetSelection();
```
rohan-kadkol, line 115: Is it a good idea to reset the selections on every shift click? Or is there a better idea? I am doing this reset (deselect any previous selections) since that is what File Explorer does.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,91 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+    }
+    return id.substring(0, lastDotIndex);
+  };
+
+  // Returns true if endId is a parent of currentId
+  // E.g. currentId = 1.4.2.3, endId = 1.4  -> true
+  // E.g. currentId = 1.2.5, endId = 1.3    -> false
+  const isAncestorParent = (currentId: string, endId: string) => {
+    return endId.indexOf(currentId) === 0;
+  };
```
rohan-kadkol, line 89: Is `isAncestorParent()` a good name for this function? Can you suggest a better name?

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,91 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+      currentRow.toggleRowSelected(true);
+      break;
+    }
+
+    if (!isAncestorParent(currentRow.id, endId)) {
+      currentRow.toggleRowSelected(true);
+    }
+
+    currentRow = getNextRow(currentRow, startId, endId);
+  }
```
rohan-kadkol, line 141: Not sure if there is a way, but can you think of a more cleaner code for this? Or is this good enough?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -383,6 +383,88 @@ it('should handle row clicks', async () => {
...
+  await user.keyboard('[ShiftLeft>]'); // Press Shift (without releasing it)
+  await user.click(getByText(data[1].subRows[0].name)); // [shiftKey = true]
+
+  checkboxes = container.querySelectorAll<HTMLInputElement>(
+    '.iui-table-body .iui-checkbox',
+  );
+  selectedIndices = [4, 5, 7];
+  Array.from(checkboxes).forEach((checkbox, index) => {
+    expect(!!checkbox.checked).toBe(selectedIndices.includes(index));
+  });
```
rohan-kadkol, line 464: I am not sure, but is `!!` really needed? Can it just be `checkbox.checked`. I kept it in since I saw it on line 1546.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -404,7 +406,7 @@ export const Table = <
       newState: TableState<T>,
       action: ActionType,
       previousState: TableState<T>,
-      instance?: TableInstance<T>,
+      instance: TableInstance<T>,
```
rohan-kadkol, line 409: Earlier, instance was optional. But since `onShiftSelectHandler` on line 434 needed a non-null instance variable, I made it required. Is this alright?

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,91 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+    } else {
+      // next sibling
+      nextRow = instance.rowsById[getGreaterId(currentRow.id)];
+
+      // If the next sibling doesn't exist, go to the next child at a one upper depth
+      let parentId = getParentId(currentRow.id);
+      while (nextRow == null && parentId != null) {
+        nextRow = instance.rowsById[getGreaterId(parentId)];
+        parentId = getParentId(parentId);
+      }
```
rohan-kadkol, line 109: On line 106, should I use `!= null` or `!== null`? Is the `!=` version better since it covers more nully/falsy cases? What is the difference?

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Ohh, right, thanks

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: this makes sense to me

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: i can't think of a better name but it doesn't really matter since it's an internal helper function. All i would suggest is to convert the comment to a JSDoc comment. This way it will show up when you hover this function name anywhere in the code.

```suggestion
  /**
    * Returns true if endId is a parent of currentId
    * E.g. currentId = 1.4.2.3, endId = 1.4  -> true
    * E.g. currentId = 1.2.5, endId = 1.3    -> false
    */
  const isAncestorParent = (currentId: string, endId: string) => {
    return endId.indexOf(currentId) === 0;
  };
```

packages/iTwinUI-react/src/core/Table/Table.test.tsx
mayank99: `!!` is usually needed to force a value to be boolean (from a different type, e.g. a string). If `checked` is a boolean, it shouldn't be needed

packages/iTwinUI-react/src/core/Table/Table.tsx
mayank99: I don't think this is the right way to do it. Can you find out whether it is actually optional? Might need to look in the documentation/codebase of react-table.

Because if it is not always passed in the JavaScript code and we remove the `?`, then we are lying about it.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: >  Is the `!=` version better since it covers more nully/falsy cases?

That's right

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
gretanausedaite: My thought process would be get all selected rows (from startId to endId) in array and then iterate through them to toggleRowSelected. Not sure if cleaner or better.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,91 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+
+  // Returns true if endId is a parent of currentId
+  // E.g. currentId = 1.4.2.3, endId = 1.4  -> true
+  // E.g. currentId = 1.2.5, endId = 1.3    -> false
+  const isAncestorParent = (currentId: string, endId: string) => {
+    return endId.indexOf(currentId) === 0;
+  };
+
+  // Returns the next row as if it was just a non subrows list
+  const getNextRow = (currentRow: Row<T>, startId: string, endId: string) => {
```
gretanausedaite, line 92: `startId` param is declared but never read in current approach.

gretanausedaite:
Got error when trying to shift select in `Full 2` story 
image

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Ohh, nice to know about JSDocs, thanks

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: Ohh, that's a good concept. But here and on line 1546, I think `checkbox.checked` is a boolean

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Ohh, thanks, `startId` is removed now

rohan-kadkol:
> Got error when trying to shift select in `Full 2` story image

This error is happening when running `resetSelection()` in `selectHandler.ts`. This method goes through all the rows and toggles them to false.
```tsx
const resetSelection = () => {
  instance.rows.forEach((r) => {
    console.log(r);
    r.toggleRowSelected(false);
  });
};
```

But looks like after the row with id=25, `toggleRowSelected()` is not there in the row object.

image

I also got such inconsistent behavior in subrows in some of my trial shift selections. i.e. sometimes the subrows of some of the rows don't have all the attributes that a row should have (or usually has). One of the attributes is `toggleRowSelected()`.

I'm not sure why this is happening. Does anyone know why this could be happening? @mayank99

gretanausedaite:
> I'm not sure why this is happening. 

Found that this happens because virtualization is enabled.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: Thanks, that's similar to as it is now. Currently, we start from `startId` and keep going to the next row/subrow until we reach the `endId`. At each node that we iterate through, we `toggleRowSelected(true)` (if applicable).

rohan-kadkol:
> > I'm not sure why this is happening.
> 
> Found that this happens because virtualization is enabled.

Thanks, I fixed it using what you and Mayank suggested of using `instance.toggleAllRowsSelected(false)`. It works correctly, but I am getting some flickering (not smooth scrolling) when I scroll to a very lower row (e.g. row 70+), especially when I have expanded some sub rows. Is this expected?

[vokoscreenNG-2022-08-29_11-16-31.webm](https://user-images.githubusercontent.com/45748283/187236327-245a1b6e-3f9b-4937-bc23-1f139fca77b1.webm)

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -389,6 +389,88 @@ it('should handle row clicks', async () => {
...
   expect(onSelect).toHaveBeenCalledTimes(3);
   expect(onRowClick).toHaveBeenCalledTimes(3);
+
+  await user.keyboard('[/ControlLeft][ShiftLeft>]'); // Release Control and Press Shift (without releasing it)
+  await user.click(getByText(mockedData()[0].name)); // Perform a click with `shiftKey: true`
+
+  expect(rows[0].classList).toContain('iui-selected');
+  expect(rows[1].classList).toContain('iui-selected');
+  expect(rows[2].classList).toContain('iui-selected');
+  expect(onSelect).toHaveBeenCalledTimes(3 + (1 + 3)); // # times before + (reset + # selected rows from start to end)
```
rohan-kadkol, line 399: After a shift click, `onSelect()` is currently called `1 time to reset all the rows to be unselected + number of selected/toggled rows from start to end (in this test, it is 3)`. Is it alright to call `onSelect()` each time a row is toggled? (i.e. referring to the second part of the above highlighted equation). Or should the second part of the above equation also be `1` (called once after toggling all rows)? I think that would be difficult to achieve so I feel the current way is alright, but just want to make sure.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+  state: TableState<T>,
+  action: ActionType,
+  instance: TableInstance<T>,
+) => {
+  const resetSelection = () => {
+    instance.toggleAllRowsSelected(false);
+  };
+
+  const getGreaterId = (id: string) => {
+    const greaterId = id.split('.');
```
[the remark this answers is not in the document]
rohan-kadkol: Ohh, I'm wondering, can users pass their own id? I thought the React Table package automatically gives the ids to the rows and they are unchangable. Is that true? If it is, then maybe we won't allow users to pass a custom `getGreaterId()` since they can't pass custom ids. Also, to confirm, what is GUID? Is it "greater ID"?

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+    const temp = startId;
+    startId = endId;
+    endId = temp;
+  }
+
+  let currentRow = instance.rowsById[startId];
+
+  while (currentRow != null) {
+    if (currentRow.id === endId) {
+      currentRow.toggleRowSelected(true);
```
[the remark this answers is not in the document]
rohan-kadkol: Thanks, yes, being able to dispatch a list of selected rows in one call sounds way better. However, I'm a little confused on how to implement it. Because currently also we are doing something similar. On a shift click, we are dispatching a table action for shift click that only contains the clicked row (e.g. the end row) in the action.

###### Table.tsx
image

The start row is saved in the state. So the action handler (`onShiftSelectHandler`) can get the start and end rows and toggle them one by one.

###### selectHandler.ts
image

What part of this process can I make more efficient? Maybe the code in the handler (`onShiftSelectHandler`) to toggle multiple rows together? But I'm not sure how to do that.

Also, just to mention, the handler (`onShiftSelectHandler`) doesn't blindly toggle all the rows between the start and the end. If it knows that all the subrows are going to be toggled anyways, it just toggles the row itself and skips toggling the subrows.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
   onSelect?.(selectedData, newState);
 };
 
+/**
+ * Handles selection when clicked on a row while shift is selected.
```
mayank99, line 60:
```suggestion
 * Handles selection when clicked on a row while shift key is pressed.
```

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+ */
+export const onShiftSelectHandler = <T extends Record<string, unknown>>(
+  state: TableState<T>,
+  action: ActionType,
+  instance: TableInstance<T>,
```
mayank99, line 65: did you get a chance to look into whether `instance` should actually be optional or not? if it should be optional in the reducer, then it might make sense to also make it optional here

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+) => {
+  const resetSelection = () => {
+    instance.toggleAllRowsSelected(false);
+  };
+
+  const getGreaterId = (id: string) => {
```
mayank99, line 71: can you add a comment here? it's not obvious what "greter id" means or what this function is doing

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,109 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+      }
+    }
+
+    return nextRow;
+  };
+
+  resetSelection();
+
+  // If no row selected before shift clicking, then assume first row is the last selectedRow (like Windows File Explorer)
+  let startId = state.lastSelectedRow ?? instance.rows[0].id;
```
mayank99, line 136: i think this logic is buggy because `state.lastSelectedRow` is not getting cleared after all rows are deselected.

to see what I mean, try selecting any random row and deselect it. Now try shift clicking any other row. It will behave differently from how it does when the page loads.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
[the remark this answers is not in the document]
mayank99: i think [`selectedFlatRows`](https://react-table-v7.tanstack.com/docs/api/useRowSelect#instance-properties:~:text=will%20be%20false-,selectedFlatRows,-%3A%20Array%3CRow%3E) might help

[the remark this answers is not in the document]
mayank99: i think [`selectedFlatRows`](https://react-table-v7.tanstack.com/docs/api/useRowSelect#instance-properties:~:text=will%20be%20false-,selectedFlatRows,-%3A%20Array%3CRow%3E) might help

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -561,7 +567,15 @@ export const Table = <
         selectRowOnClick &&
         !event.isDefaultPrevented()
       ) {
-        if (!row.isSelected && (selectionMode === 'single' || !event.ctrlKey)) {
+        if (event.shiftKey) {
```
mayank99, line 570: To prevent text selection, there are two approaches. Both of these would either go in the row onclick or keydown and should only be done when `selectionMode` is set to `'multi'`.
- use `getSelection()` to temporarily store the old selection, call `removeAllRanges` on the selection to clear selection, then set the selection to old selection (effectively preserving user's selection state)
- or, apply `user-select: none` on the table when shift selection starts and remove it when it's done

these are all standard platform APIs so you can find more info on MDN

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: if you maintain your own selection state (as mentioned in https://github.com/iTwin/iTwinUI-react/pull/774#discussion_r960020084) then i think you won't need to reset selection

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Ohh, that's true. I renamed `getGreaterId()` to `getNextSiblingId()` and also added a JSDocs comment with some examples.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: I think that behavior is intended. I did it that way since Windows Explorer and other lists like Linux Files Explorer also did it the same way

[vokoscreenNG-2022-09-02_06-31-26.webm](https://user-images.githubusercontent.com/45748283/188121254-47ea828e-9b69-439d-a635-f268227e9bc2.webm)

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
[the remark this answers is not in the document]
rohan-kadkol: > i think [`selectedFlatRows`](https://react-table-v7.tanstack.com/docs/api/useRowSelect#instance-properties:~:text=will%20be%20false-,selectedFlatRows,-%3A%20Array%3CRow%3E) might help

Thanks for the idea, but I'm not sure how you meant I should use that. Because I think it stores all the previously selected rows/subrows. But I'm not sure how it can help avoid toggling all the row/subrows between the start and the end.

I thought that one benefit might be that when resetting the row selections (`resetSelection()`), we can only deselect rows that are not selected in the new selection? E.g. old selection = `1, 2, 3, 4, 5`; new selection = `1, 2, 3`; then only deselect 4 and 5. But I then thought that this might be more inefficient for extremely large tables (1000+ rows) than the idea you suggested of `instance.toggleAllRowsSelected(false);`.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -389,6 +389,88 @@ it('should handle row clicks', async () => {
   expect(rows[2].classList).toContain('iui-selected');
   expect(onSelect).toHaveBeenCalledTimes(3);
   expect(onRowClick).toHaveBeenCalledTimes(3);
+
+  await user.keyboard('[/ControlLeft][ShiftLeft>]'); // Release Control and Press Shift (without releasing it)
+  await user.click(getByText(mockedData()[0].name)); // Perform a click with `shiftKey: true`
+
+  expect(rows[0].classList).toContain('iui-selected');
+  expect(rows[1].classList).toContain('iui-selected');
```
gretanausedaite, line 397: Maybe don't test if all rows are selected. Eg. test if first two rows are selected and last one is not.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -389,6 +389,88 @@ it('should handle row clicks', async () => {
...
+
+  let rows = container.querySelectorAll('.iui-table-body .iui-row');
+  expect(rows.length).toBe(3);
+
+  const user = userEvent.setup();
+  await user.keyboard('[ShiftLeft>]'); // Press Shift (without releasing it)
+  await user.click(getByText(data[1].name)); // [shiftKey: true]
+
+  // By default, when no row is selected before shift click, start selecting from first row to clicked row
+  expect(rows[0].classList).toContain('iui-selected');
```
gretanausedaite, line 422: Look like same test as in lines 393-400. Is there are reason you added both?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -389,6 +389,88 @@ it('should handle row clicks', async () => {
...
+
+  let checkboxes = container.querySelectorAll<HTMLInputElement>(
+    '.iui-table-body .iui-checkbox',
+  );
+  let selectedIndices = [1, 3];
+  Array.from(checkboxes).forEach((checkbox, index) => {
+    expect(!!checkbox.checked).toBe(selectedIndices.includes(index));
+  });
+
+  expect(onSelect).toHaveBeenCalledTimes(3 + (1 + 2)); // # times before + (reset + # selected rows from start to end)
```
gretanausedaite, line 444: This `toHaveBeenCalledTimes` calculations can lead to faulty tests. Maybe there is a better way to test function call. Eg. `toHaveBeenCalledLastWith`? It would test if specific row is selected and not just random row.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -432,6 +434,10 @@ export const Table = <
           );
           break;
         }
+        case shiftRowSelectedAction: {
+          onShiftSelectHandler(newState, action, instance);
```
gretanausedaite, line 438: Don't forget to add user defined onSelect (as in single row select action)

testing.ts
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
```
gretanausedaite, line 1: Are you using this file anywhere? Why did you commit it?

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: the jsdoc examples are still using the old `getGreaterId` name

packages/iTwinUI-react/src/core/Table/Table.tsx
mayank99: Hmm this is causing some flickering when the text selection gets cleared. Not a very nice user experience.

testing.ts
rohan-kadkol: **[DONE]** Haha! Yes, it was not intended. Removed it now

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: Ohh, yes, we could implement some parts our way too. But maybe we could also incorporate some parts from existing essential tools to give a more familiar and intuitive feel. Some reasons I feel the Windows Explorer does it that way:

1. If someone does a shift click after all rows are deselected without `lastSelectedRow` saved, it would be strange to see the same behavior as a normal click (because a normal click toggles the row, and if no `lastSelectedRow` is there, then shift click will also toggle the row). So instead of doing a toggle, we can do a real shift click from the `lastSelectedRow` to the current clicked row. 

2. If the user accidentally clicks outside the table (or similar) and loses their selection, they would have to reselect the first row. This might be a problem if the user went through a lot of effort to find the starting row the first time (e.g. finding the correct image when all images have similar names with random number timestamps 
	```
	  20220902-122553.jpg
	  20220902-123219.jpg
	  20220902-121637.jpg
	  ...
	  ```

	or if the file names are so long that the user has to make the column wider to find the correct name).

	  ```
	  abc<...............................>01.txt
	  abc<...............................>02.txt
	  abc<...............................>03.txt
	  ...
	  ```

	So if the `lastSelectedRow` is saved, the user can assume the start row is still selected and just move to finding the ending row. 

But I do see how saving the `lastSelectedRow` might be confusing to some users too.

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Ohh, right. @bentleyvk implemented a much easier way ([code lines](https://github.com/iTwin/iTwinUI-react/compare/vilius/shift-select-demo#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999R778-R786)) using the `user-select: none` way you suggested. I added that here. It sets `user-select: none` when shift is pressed and removes `user-select` when shift is released. It works without flickering.

Also one thing I noticed: when we scroll out of the selection in a table with virtualization, then the selection is lost (end of video). I'm sure this might be okay since that is how virtualization works, but just noting what I noticed.

[vokoscreenNG-2022-09-02_14-46-33.webm](https://user-images.githubusercontent.com/45748283/188218865-42110b5b-f5f9-46c3-bd68-5b7bf024e575.webm)

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: To be safe, I have temporarily reverted it back to how it was. Since with Vilius's code in `vilius/shift-selection-demo`, `instance?` also worked. But I might be making more slight changes to this code. So I will confirm once again at the end on if everything works with instance being optional.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[DONE]** It now tests that the 0th and 1st rows is selected but the 2nd one is not

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[DONE]** Thanks for idea. After implementing Vilius's idea for the shift select (branch `vilius/shift-select-demo`), `onSelect()` is only called once per total shift select. So not there are no more required calculations for the # of times `onSelect()` is called. The number of times `onSelect()` is called is just one more than the previous one.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: Ohh, right, that's correct. They are the same. But the only difference is that here `lastSelectedRow` is `undefined`. This is because no row was selected (normal click) before the shift click ([old commit line 421](https://github.com/iTwin/iTwinUI-react/pull/774/files/da5ef1854d02894e3c653eaf158ee178b75bd210#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R421)).

So this test here (417-426) is to see that is `lastSelectedRow` is `undefined`, is `lastSelectedRow` set to 0. Whereas in lines 393-400, since a row is selected before the shift click, `lastSelectedRow` is not `undefined`.

Is there a better place you suggest for adding this test? Or maybe it is okay the way it is?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[OUTDATED]**

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: @mayank99 When you get time, can you please let me know if `!!` can be omitted since `checkbox.checked` is already a boolean? Thanks. I know you already said it can be omitted:

> If `checked` is a boolean, it shouldn't be needed

Also, removing `!!` in all places still passes the tests. But since the old code still had `!!`, I just want to confirm if `!!` is needed or not.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: Thanks for the idea. With Vilius's code ideas from the `vilius/shift-select-demo`, we are now maintaining our own `Record/{}` of the selected rows and then setting `selectedRowIds: selectedRowIds` in the new state.

```tsx
const newState = {
  ...state,
  selectedRowIds: selectedRowIds,
};
```

This way works better than the toggle each row version since now `onSelect` is called just once (in the toggle version it was called on each toggle).

But the flickering problem is still unfortunately there. But I think that is not related to this PR. Since this flickering happens also on the hosted storybook.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Thanks for the ideas Vilius and Mayank. As mentioned here ([#774 comment](https://github.com/iTwin/iTwinUI-react/pull/774/files/da5ef1854d02894e3c653eaf158ee178b75bd210#r968990008)), saving our own `Record / {}` and setting the state at the end allowed `onSelect()` to be called only once per shift click, instead of once per toggle.

apps/storybook/src/Table.stories.tsx
```diff
@@ -1438,6 +1438,7 @@ export const Full2: Story<Partial<TableProps>> = (args) => {
 
   const isRowDisabled = useCallback((rowData: TableStoryDataType) => {
     return rowData.product.endsWith('3');
+    // return false;
```
gretanausedaite, line 1441: Commented code can be deleted.

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
```diff
@@ -51,7 +51,7 @@ export type DialogContextProps = {
    * Whether dialog should be positioned relatively to a container or the viewport.
    *
    * Using `'container'` will absolutely position this dialog relative to the closest positioned ancestor.
-   * In other words, you must place the dialog as a child of an element that has `position` set to
```
gretanausedaite, line 54: Could you undo this?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
gretanausedaite: Place is good. Maybe you could add a comment explaining what you are doing? Test name does not specify lastSelectedRow status. "should handle sub-rows shift click selection" could be both either one - defined lastSelectedRow and undefined.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,65 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+    .slice(startIndex, endIndex + 1)
+    .forEach((r) => (selectedRowIds[r.id] = true));
+
+  // 2. Select all children of the last row (endIndex)
+  if (instance != null) {
+    const handleRow = (row: Row<T>) => {
+      selectedRowIds[row.id] = true;
+      row.subRows.forEach((r) => handleRow(r));
+    };
+    handleRow(instance.rowsById[action.id]);
```
gretanausedaite, line 136: What happens when user clicks row 5, then row 2, startIndex > endIndex? I think you should pass `endIndex` here and not action.id.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,65 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+
+  const selectedRowIds: Record<string, boolean> = {};
+
+  // 1. Select all rows between start and end
+  instance?.flatRows
+    .slice(startIndex, endIndex + 1)
+    .forEach((r) => (selectedRowIds[r.id] = true));
+
+  // 2. Select all children of the last row (endIndex)
+  if (instance != null) {
```
gretanausedaite, line 131: When instance is null? Maybe you can move this check to the beggining of this handler.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Ohh, nice catch, thanks

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: I moved this check to the top of the file. Is it alright to add a `console.debug()` on [line 106 of this commit](https://github.com/iTwin/iTwinUI-react/pull/774/commits/cb494c594d2d9e274d88d719c9a62974e03983c1#diff-772f25aaf3cd0aa17d5900a4252050c32c48b2efe18b4f6db0d84df08f96cc0bR106)? Or maybe delete that line?

packages/iTwinUI-react/src/core/Dialog/DialogContext.tsx
rohan-kadkol: **[DONE]** Ohh, I didn't know I made that change. Looks like it was the formatter that did this. Undid it now

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: **[DONE]** I tried to group similar tests together. So I now put the two shift select tests (with `lastSelectedRow`: 1. defined 2. undefined) are moved to the test: `should handle row clicks`. I added two comments `// Shift click test #1` and `#2` to make what we are testing within that test more clear.

I then left my previous code for the sub-row shift selecting test in `should handle sub-rows shift click selection` with slight improvements. Apart from testing if the appropriate checkboxes are checked, we are now also checking if the appropriate indeterminate checkboxes are indeterminate. The code is also more clean now.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
gretanausedaite: You should remove it. It's for internal testing and should not be released in prod version.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[OUTDATED]** Thanks, luckily with the new code, there are no helpers

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Out of curiosity, what is the recommended way in React to log failures in the production version so that we can debug later? Like sometimes when desktop programs or I think even some Android apps crash, they say do you want to send a stack trace to the developer. If we select yes, then I think it sends all the logs about the app crash.

Is there some similar workflow for this in React?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -755,6 +776,14 @@ export const Table = <
             ...style,
           },
         })}
+        onKeyDown={(e) => {
+          if (e.shiftKey) {
```
mayank99, line 780: This will only fire when the table is focused, so it might be better to add it to the document.

Also we only want to disable text selection when the table has multi-selection enabled so it will need an extra condition.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,67 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+  instance.flatRows
+    .slice(startIndex, endIndex + 1)
+    .forEach((r) => (selectedRowIds[r.id] = true));
+
+  // 2. Select all children of the last row (endIndex)
+  const handleRow = (row: Row<T>) => {
+    selectedRowIds[row.id] = true;
+    row.subRows.forEach((r) => handleRow(r));
+  };
+  handleRow(instance.flatRows[endIndex]);
```
mayank99, line 139: Why do you need to handle last row separately?

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,67 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+
+  const newState = {
+    ...state,
+    selectedRowIds: selectedRowIds,
+  };
+
+  // 3.1 Deselect all selected disabled rows and their children
+  // 3.2 Convert all partially selected rows marked with tick mark to horizontal line
+  onSelectHandler(newState, instance, onSelect, isRowDisabled);
```
mayank99, line 148: What are these comments referring to? Because you are just calling the function here.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: That is because we are selecting the rows between the start and the end in the `flatRows` array. But the subRows of the end row comes after the end row in the `flatRows` array. So we need to manually select all the subRows of the end row.

E.g. If user selects 2 to 5, then 5.1 for example will not be selected even though 5 is selected

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: I was just mentioning the reason for calling this function. Because I thought it may not be clear. I just noticed that there is a similar comment under [onSingleSelectHandler's line 86](https://github.com/iTwin/iTwinUI-react/blob/rohan/table-shift-click/packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts#L86).

But I can remove lines 146 and 147 if those comments are not necessary

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -403,6 +405,41 @@ export const Table = <
...
+    }
+
+    // Remove onkeydown event listeners (during destruction)
+    return () => {
+      if (selectionMode === 'multi') {
+        document.removeEventListener('keydown', disableUserSelect);
+        document.removeEventListener('keyup', enableUserSelect);
+      }
+    };
+  }, [selectionMode, disableUserSelect, enableUserSelect]);
```
rohan-kadkol, line 441: Is it okay to merge the on creation logic (429-432) and on destruction logic (435-440) into one useEffect? Or should I separate them into two?

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: I added `keyup` and `keydown` event listeners to the ownerDocument (408-441 of below link). I'm not sure if this is what you meant of if there is a cleaner way to do it. So would be great if you can check how I implemented it [(commits diff)](https://github.com/iTwin/iTwinUI-react/pull/774/files/60657df4a537c23b43f5ff59d4231bcf3b74eda1..82731b8d342cb72acf887a84f7d54a23dfe4ff73).

Additionally, I also added a check to only run the shift select selection logic and to only run the disable selection while Shift is held logic when `selectionMode === 'multi'`. This addition is also in the above link commits diff link (612 of above link).

packages/iTwinUI-react/src/core/Table/Table.tsx
gretanausedaite: It's ok. I think it's a standard practise.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -405,6 +405,41 @@ export const Table = <
...
+        ownerDocument && (ownerDocument.documentElement.style.userSelect = '');
+      }
+    },
+    [ownerDocument],
+  );
+
+  React.useEffect(() => {
+    // Add onkeydown event listeners (during creation)
+    if (selectionMode === 'multi') {
+      document.addEventListener('keydown', disableUserSelect);
```
mayank99, line 438: You should use `ownerDocument` here

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -405,6 +405,41 @@ export const Table = <
...
+      document.addEventListener('keyup', enableUserSelect);
+    }
+
+    // Remove onkeydown event listeners (during destruction)
```
mayank99, line 442: These comments are not necessary because that's just how useEffect works

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -405,6 +405,41 @@ export const Table = <
...
+      if (e.key === 'Shift') {
+        ownerDocument && (ownerDocument.documentElement.style.userSelect = '');
+      }
+    },
+    [ownerDocument],
+  );
+
+  React.useEffect(() => {
+    // Add onkeydown event listeners (during creation)
+    if (selectionMode === 'multi') {
```
mayank99, line 437: nit: you can do an early return to make the code a bit cleaner

```js
if (selectionMode !== 'multi') {
  return;
}
```

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: Would it not be enough to call `handleRow` on every row from start to end?

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Ohh, right, nice idea

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Removed

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Swapped `document` with `ownerDocument`, thanks. But just curious, what is the difference between the two? Also, just to note, `ownerDocument` can be undefined but I don't think `document` can be undefined. So is it okay to add/remove event listeners from `ownerDocument` instead of `document`?

image

image

mayank99: There can be multiple documents associated with a page, e.g. when a popup window is opened (using [`window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open)) and our components are used inside that popup. So we always want to add events to the correct document, and `ownerDocument` helps us do that.

And `document` can also be undefined when running this code _outside_ the browser (e.g. in [SSR](https://www.zachleat.com/web/ssr-overloaded/#component-framework-ssr) or in unit tests). This is why our [style guide](https://github.com/iTwin/iTwinUI-react/blob/main/STYLEGUIDE.md#use-getdocument-getwindow-instead-of-direct-access) recommends guarding document checks.

Those TypeScript types don't always account for code that runs outside the browser.

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: Can you show an example of how the rows are structured? Because I'm not sure I really understand why `2.2.1` is any different from `5.2.1` in terms of handling.

rohan-kadkol: Here is an example tree structure for selecting 2 to 4 (not 5; to keep the example small)

- 1
  - 1.1
  - ...
- **2 [START]**
  - 2.1
  - 2.2
    - 2.2.1
    - 2.2.2
    - 2.2.3
  - 2.3
  	- 2.3.1
  	- 2.3.2
- 3
  - 3.1
  - 3.2
    - 3.2.1
  - 3.3
  - 3.4
- 4
  - 4.1
  - **4.2 [END]**
   - 4.2.1
   - 4.2.2
   - 4.2.3
  - 4.3

Current approach: Treat the nested rows structure (rows has subrows...) as a flat array of just rows with no hierarchy (`flatRows`). Then iterate through this array from the start index to the end index and set each row as checked along the way.

> Why do you need to handle last row separately?

The above mentioned iteration stops at 4.2 since that is the last row (the row we clicked). But this loop does not select 4.2.1 - 4.2.3. This is why we have to handle the last row separately.

> Would it not be enough to call handleRow on every row from start to end?

If we started selection from 2.3 for example, then `handleRow(2.3)` would select itself, 2.3.1, and 2.3.2. The next row in `flatRows` is 2.3.1. Thus, `handleRow(2.3.1)` will select 2.3.1. Thus, 2.3.1 would be selected twice which is inefficient.

mayank99: > The above mentioned iteration stops at 4.2 since that is the last row (the row we clicked). But this loop does not select 4.2.1 - 4.2.3. This is why we have to handle the last row separately.

Ah, this finally clarifies it! Maybe add another line in the comment explaining this.

In general, comments explaining "why" are more useful than ones explaining "what".

mayank99: Currently the comment says _"Select all children of the last row (endIndex) Because the subRows of the lastRow have indices > endIndex and hence won't be selected by step 1"_

This is a bit misleading as we are not selecting "all" children and the reasoning also seems a bit convoluted. Would it be more accurate to say something like "_Handle last row separately because the user might select only a few of the subrows instead of all"_?

rohan-kadkol: Not quite. Actually, lines 136-140 is not because the user may select a few of the subrows instead of all rows. That code block was to indeed to select "all" the children (subRows and their subrows...) of the last row since they will not be selected by the loop in step 1.

But I do agree that is it a little wordy. So I changed it to the below lines. Do you think it is alright? Maybe I can make it shorter? Or a little clearer if it is ambiguous?

```tsx
// 2. Select all children of the last row (endIndex)
// Since lastRow's children come after endIndex + 1 (not selected in step 1)
```

packages/iTwinUI-react/src/types/react-table-config.ts
```diff
@@ -148,6 +148,7 @@ declare module 'react-table' {
       isScrolledToRight?: boolean;
       isScrolledToLeft?: boolean;
     };
+    lastSelectedRow?: string;
```
bentleyvk, line 151: Maybe rename to `lastSelectedRowId`

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,68 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+  if (instance == null) {
+    return state;
+  }
+
+  let startIndex = instance.flatRows.findIndex(
+    (row) => row.id === state.lastSelectedRow,
+  );
+  let endIndex = instance.flatRows.findIndex((row) => row.id === action.id);
+
+  if (startIndex < 0) {
```
bentleyvk, line 114: Nit: you also can write `let startIndex = Math.max(0, instance.flatRows.findIndex(
    (row) => row.id === state.lastSelectedRow,
  ));`

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -88,6 +89,68 @@ export const onSingleSelectHandler = <T extends Record<string, unknown>>(
...
+  // Since lastRow's children come after endIndex + 1 (not selected in step 1)
+  const handleRow = (row: Row<T>) => {
+    selectedRowIds[row.id] = true;
+    row.subRows.forEach((r) => handleRow(r));
+  };
+  handleRow(instance.flatRows[endIndex]);
+
+  const newState = {
+    ...state,
+    selectedRowIds: selectedRowIds,
```
bentleyvk, line 144: You can write 
```ts
const newState = {
    ...state,
    selectedRowIds,
}
```
since key and variable names are the same.

gretanausedaite (approved):
I don't have any additional comments.

packages/iTwinUI-react/src/types/react-table-config.ts
rohan-kadkol: **[DONE]** Renamed. Yes, I think `lastSelectedRowId` is more clearer/accurate

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Ohh, right, thanks

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: **[DONE]** Thanks

mayank99 (approved):

PR #776, opened 2022-08-11
fix(Dialog): Fixed body overflow handling when there is a dialog inside another dialog

### LINKED ISSUE #748: With two modals loaded background is scrollable
Let's say that there are two modals: modal A (parent modal), modal B (modal inside modal A). And when both of them are loaded at the same time, modal A is visible and modal B is not visible, then the background behind modal A is interactive. You can scroll up or down in the page, press some buttons.

#### Expected behavior
When modal is open, background behind the modal should not be interactive

#### Screenshots
There is a modal:
image
Inside the modalContent section there is a different modal:
image
When only main modal is open background behind is scrollable:
image
When both modals are open background behind main modal is not interactive, but main modal is interactive:
image

bentleyvk:
Closes #748  

When parent dialog is closed, its inner dialog `useEffect` would call its clean-up function and set overflow to hidden. 
Now if body already has `overflow: hidden` (possibly from parent dialog), dialog will do nothing with the overflow.

## Checklist


packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -81,16 +81,27 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
...
+    // Prevents document from scrolling when the dialog is open.
     React.useEffect(() => {
       const ownerDocument = dialogRef.current?.ownerDocument;
-      if (!ownerDocument || !preventDocumentScroll) {
+      // If there is no `ownerDocument` or `preventDocumentScroll` is false or
+      // document body originally has `overflow: hidden` (possibly from other/parent dialog), then do nothing.
+      if (
+        !ownerDocument ||
+        !preventDocumentScroll ||
+        originalBodyOverflow.current === 'hidden'
```
mayank99, line 99: Can you elaborate how/why this fixes the bug (*"When only main modal is open, background behind is scrollable"*)? It's not really clear from the description.

bentleyvk: Yeah, I couldn't reproduce this part when having modal inside modal, maybe it is something specific to their use case.

mayank99: then what does your PR fix? i'm confused

bentleyvk: It fixes: when you open parent modal then child modal, close child modal and parent modal, then body won't be scrollable since child modal useEffect cleanup function resets body overflow to hidden.

adamhock (approved):
Looks good to me 🎉

packages/iTwinUI-react/src/core/Dialog/Dialog.tsx
```diff
@@ -65,7 +68,9 @@ export const Dialog = Object.assign(
           trapFocus,
         }}
       >
-        {children}
+        <div style={{ isolation: 'isolate', ...style }} {...rest}>
```
mayank99, line 71: I think this also needs a z-index of around `999`, because the z-indices of the backdrop and dialog are isolated inside this container.

It can be hardcoded for now but should be moved to CSS later.

bentleyvk: When I tested, I had some content near the button with `z-index: 1000` and it didn't bleed trough, so maybe `z-index` on this div is not needed.

mayank99: In my testing I found it's broken even with my suggestion of adding z-index to this div, because it doesn't even have a size currently. I think this div will also need `position: fixed` or `position: absolute` and a size or `inset`.

Quick video demonstration:
https://user-images.githubusercontent.com/9084735/187711177-b2e35f09-322c-43df-a771-424683e6a2a6.mp4

bentleyvk: Removed everything, seems to be working fine.

mayank99 (approved):

PR #777, opened 2022-08-16
fix: Focus is lost while resizing on overflow

### LINKED ISSUE #768: Focus is lost when overflow is used in `ButtonGroup` and `Breadcrumbs`
I have had this issue in `Select` while implementing overflow handling.
If we focus and then resize (in `ButtonGroup`, `Breadcrumbs` stories), focus will be lost.
Need to change code a little to fix that.

[`c7cfb87` (#700)](https://github.com/iTwin/iTwinUI-react/pull/700/commits/c7cfb873f5378ea118564cbc37cc643da1fd1d09)

veekeys:
Closes #768 

Previously focus was being lost whenever we were resizing the window (rerender happened) and that is mostly caused if the DOM structure is changing (then even having a `key` on element does not help). Moved elements around, to fix this issue.

`Breadcrumbs`: Moved inner components outside, because they were getting a new ref and as a result on every rerender new DOM was being created, which led to focus being lost.

`ButtonGroup`: Moved conditions a little, so DOM is kept the same

`MiddleTextTruncation`: Found that `substr` is marked as deprecated

### DIFF
--- a/packages/iTwinUI-react/src/core/Breadcrumbs/Breadcrumbs.tsx
+++ b/packages/iTwinUI-react/src/core/Breadcrumbs/Breadcrumbs.tsx
@@ -62,28 +62,6 @@ export const Breadcrumbs = React.forwardRef(
     const [overflowRef, visibleCount] = useOverflow(items);
     const refs = useMergedRefs(overflowRef, ref);
 
-    const Separator = () => (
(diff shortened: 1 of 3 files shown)


bentleyvk (approved):

elephantcatdog (approved):
Looks good to me

PR #778, opened 2022-08-16
chore: Clean script

### LINKED ISSUE #726: Add clean script
Need to have possibility to fully cleanup the project (node modules, build artifacts, etc.)

veekeys:
Closes #726 

Added `clean` script to clean all `node_modules` and build output from every package.
Was not sure what to do with cypress visual report... Thought it is not breaking anything, so left untouched

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "format": "turbo run format",
     "test": "turbo run test",
     "createComponent": "node scripts/createComponent.js",
-    "push-docker": "yarn workspace storybook push-docker"
+    "push-docker": "yarn workspace storybook push-docker",
+    "clean": "turbo run clean && rimraf node_modules"
   },
   "dependencies": {},
   "devDependencies": {
--- a/packages/iTwinUI-react/package.json
+++ b/packages/iTwinUI-react/package.json
@@ -29,10 +29,11 @@
     "ux"
   ],
   "scripts": {
-    "build": "yarn clean && yarn copy-files && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
-    "build:watch": "yarn clean && yarn copy-files && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
-    "clean": "yarn clean:coverage && rimraf esm && rimraf cjs",
+    "build": "yarn clean:build && yarn copy-files && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
+    "build:watch": "yarn clean:build && yarn copy-files && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
+    "clean:build": "yarn clean:coverage && rimraf esm && rimraf cjs",
     "clean:coverage": "rimraf coverage",
+    "clean": "yarn clean:coverage && yarn clean:build && rimraf node_modules",
     "test": "jest",
     "test:watch": "jest --watch",
     "format": "prettier --config .prettierrc **/*.{tsx,ts,js} --ignore-path .gitignore --write",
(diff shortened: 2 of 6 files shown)


packages/iTwinUI-react/package.json
```diff
@@ -29,10 +29,11 @@
...
   ],
   "scripts": {
-    "build": "yarn clean && yarn copy-files && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
-    "build:watch": "yarn clean && yarn copy-files && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
-    "clean": "yarn clean:coverage && rimraf esm && rimraf cjs",
+    "build": "yarn clean:build && yarn copy-files && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
+    "build:watch": "yarn clean:build && yarn copy-files && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
+    "clean:build": "yarn clean:coverage && rimraf esm && rimraf cjs",
     "clean:coverage": "rimraf coverage",
+    "clean": "yarn clean:coverage && yarn clean:build && rimraf node_modules",
```
mayank99, line 36: Maybe also clean the `.turbo` folder in every package to remove the cache

package.json
```diff
@@ -15,7 +15,8 @@
     "format": "turbo run format",
     "test": "turbo run test",
     "createComponent": "node scripts/createComponent.js",
-    "push-docker": "yarn workspace storybook push-docker"
+    "push-docker": "yarn workspace storybook push-docker",
+    "clean": "turbo run clean && rimraf node_modules"
```
rohan-kadkol, line 19: To confirm, does `turbo` run the command after `run` (in this case `clean`) in all the workspaces/folders as specified in the `workspaces` block on lines 5-8?

image

veekeys: yeah, it goes through all the node projects (folder consisting of `package.json` in specified workspaces - level 1 path) and runs the specified command (`clean`) in each of those.
Run locally to confirm ;)

packages/iTwinUI-react/package.json
veekeys: done

mayank99 (approved):

packages/iTwinUI-react/package.json
```diff
@@ -33,7 +33,7 @@
     "build:watch": "yarn clean:build && yarn copy-files && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
     "clean:build": "yarn clean:coverage && rimraf esm && rimraf cjs",
     "clean:coverage": "rimraf coverage",
-    "clean": "yarn clean:coverage && yarn clean:build && rimraf node_modules",
+    "clean": "rimraf .turbo && yarn clean:coverage && yarn clean:build && rimraf node_modules",
```
rohan-kadkol, line 36: I'm pretty sure it doesn't, but does the order of deleting folders matter? I was thinking what if some cache is added deleting `.turbo` (when running the following command after removing `.turbo`)? E.g. if the commands following the `rimraf .turbo` fail for some reason, turbo might log it in the cache. But I don't think it would happen because you set the `cache` property in `turbo.json`'s `clean` command to false, but I just thought I'd confirm.

veekeys: I guess it would make sense in other scenario, but in here as I needed to disable the cache (otherwise `.turbo` folder gets locked and its impossible to remove it) it does not really matter. And in case it fails, logs in terminal with errors should be available

r100-stack (approved):
Thanks for answering my questions. I think the changes look correct, but I'm not experienced in turbo 😅

PR #779, opened 2022-08-16
fix(Table): Show correct icon on hover for columns with `sortDescFirst` 

### LINKED ISSUE #659: Table: wrong sort icon showed when using `sortDescFirst`
The current conditional logic will show the wrong icons on *hover* when the user overrides sorting order using [`sortDescFirst`](https://react-table.tanstack.com/docs/api/useSortBy#:~:text=will%20be%20disabled-,sortDescFirst,-%3A%20Bool).

Hovering over a DescFirst column should show `SvgArrowDown` because that's the _next_ state (the result of the interaction).

Current behavior:

https://user-images.githubusercontent.com/9084735/166505237-865cab90-1dc3-4182-9491-5e50083cf13b.mov

<sub>_Originally posted in https://github.com/iTwin/iTwinUI-react/issues/645#issuecomment-1116350376_</sub>

elephantcatdog:
image

- Fixed issue by adding logic to check if `sortDescFirst` is true and `isSorted` is false.
- Edited storybook example `Sortable` to have an "ID" column that sorts descending first.
- Tried really hard to have unit or visual tests but is not possible to my knowledge.
  - Unit tests are hard because there's no difference between the "up" and "down" icons except for their paths when looking at the html.
  - Visual tests are hard because cypress doesn't support hovering for css-based visual changes. I tried to use [cypress-real-events](https://github.com/dmtrKovalenko/cypress-real-events), but the `realHover` function does not work with image testing.

Closes #659

## Checklist


apps/storybook/src/Table.stories.tsx
```diff
@@ -314,9 +314,15 @@ export const Sortable: Story<Partial<TableProps>> = (args) => {
         columns: [
           {
             id: 'name',
-            Header: 'Name',
+            Header: 'Name (sort asc first, default)',
```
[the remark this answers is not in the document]
elephantcatdog: That's fine with me. I also spelled out 'ascending' and 'descending' just to make it clearer

apps/storybook/src/Table.stories.tsx
```diff
@@ -341,9 +347,9 @@ export const Sortable: Story<Partial<TableProps>> = (args) => {
...
     () => [
-      { name: 'Name1', description: 'Description1' },
-      { name: 'Name3', description: 'Description3' },
-      { name: 'Name2', description: 'Description2' },
+      { name: 'Name1', description: 'Description1', id: '1' },
+      { name: 'Name3', description: 'Description3', id: '3' },
+      { name: 'Name2', description: 'Description2', id: '2' },
     ],
     [],
   );
```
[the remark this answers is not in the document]
elephantcatdog: Done

apps/storybook/src/Table.stories.tsx
```diff
@@ -362,9 +368,9 @@ export const Sortable: Story<Partial<TableProps>> = (args) => {
...
   data: [
-    { name: 'Name1', description: 'Description1' },
-    { name: 'Name3', description: 'Description3' },
-    { name: 'Name2', description: 'Description2' },
+    { name: 'Name1', description: 'Description1', id: '1' },
+    { name: 'Name3', description: 'Description3', id: '3' },
+    { name: 'Name2', description: 'Description2', id: '2' },
   ],
   isSortable: true,
 };
```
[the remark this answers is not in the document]
elephantcatdog: Done

apps/storybook/src/Table.stories.tsx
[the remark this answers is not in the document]
rohan-kadkol: I actually think maybe `Asc First` and `Desc first` are enough. Because even the video in the linked issue had that.

image

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -813,7 +813,8 @@ export const Table = <
                             )}
                             {showSortButton(column) && (
                               <div className='iui-cell-end-icon'>
-                                {column.isSorted && column.isSortedDesc ? (
+                                {(column.isSorted && column.isSortedDesc) ||
+                                (!column.isSorted && column.sortDescFirst) ? (
```
veekeys, line 817: Do you think we need `isSorted` here from your testing?
Wouldnt it be enough to have `column.isSortedDesc || column.sortDescFirst` ?

apps/storybook/src/Table.stories.tsx
```diff
@@ -314,9 +314,15 @@ export const Sortable: Story<Partial<TableProps>> = (args) => {
         columns: [
           {
             id: 'name',
-            Header: 'Name',
+            Header: 'Name (sorts ascending first)',
             accessor: 'name',
           },
+          {
+            id: 'id',
```
veekeys, line 321: nit, but having `id` after `name` looks uncommon.
id -> name -> description -> actions

But I noticed now in other stories we have even different, where `id` goes after `description` :/
Leave it, I guess...

apps/storybook/src/Table.stories.tsx
```diff
@@ -314,9 +314,15 @@ export const Sortable: Story<Partial<TableProps>> = (args) => {
         columns: [
           {
             id: 'name',
-            Header: 'Name',
+            Header: 'Name (sorts ascending first)',
```
veekeys, line 317: After changing string, you need to update visual tests

veekeys (commented):
Seems to work fine!
For the tests, I think it should be pretty easy to have one. We have an example, where we compare if icon is correct if you check `Table` test file line 1116, unless you have faced some other issues ?

packages/iTwinUI-react/src/core/Table/Table.tsx
elephantcatdog: It is necessary. You can't say `isSorted && (isSortedDesc || sortDescFirst)` because if you're sorting desc first, the column won't be sorted initially when the down icon should appear, and therefore will fail the `isSorted` check. And you can't say `(isSorted && isSortedDesc) || sortDescFirst` because then the down icon will always appear for columns sorted desc first.

apps/storybook/src/Table.stories.tsx
elephantcatdog: Yeah, it seemed weird to have it after description, and I wanted it to be adjacent to the asc first column. But I did change it to your example.

apps/storybook/src/Table.stories.tsx
elephantcatdog: Updated

apps/storybook/src/Table.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: I shortened ascending and descending

elephantcatdog:
> For the tests, I think it should be pretty easy to have one. We have an example, where we compare if icon is correct if you check Table test file line 1116, unless you have faced some other issues ?

Thanks Vykintas, I added two tests for the icons.

apps/storybook/src/Table.stories.tsx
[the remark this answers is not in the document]
rohan-kadkol: Yes, making it to just `Asc` and `Desc` would be great. But about writing `sort`, I think that might be redundant, Because `Asc First` usually means sorted asc first. Also the name of the story is `Sortable` so it would also hint at the same.

But I don't think it might matter that much  😅 Any of those seems okay to me! It was just something small I noticed so I thought I'd mention it

packages/iTwinUI-react/src/core/Table/Table.tsx
veekeys: sorry, you might have misunderstood, but I am asking if you were to remove `isSorted` from both conditions.
So instead of `(column.isSorted && column.isSortedDesc) || (!column.isSorted && column.sortDescFirst)` have just `column.isSortedDesc || column.sortDescFirst`.

I understand why your first given example woudl not work, but I am not sure what you had in mind with the 2nd one.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -587,6 +591,113 @@ it('should not show sort icon if disabled in column level', () => {
...
+  const {
+    container: { firstChild: sortDownIcon },
+  } = render(<SvgSortDown className='iui-icon iui-sort' aria-hidden />);
+  const nameHeader = container.querySelector(
+    '.iui-table-header .iui-cell',
+  ) as HTMLDivElement;
+  expect(nameHeader).toBeTruthy();
+
+  // hover on column header
+  fireEvent.mouseEnter(nameHeader);
```
veekeys, line 623: this hover does nothing actually, because you query DOM and the icon is always there.
We could also check if icon is visible on hover (with `toBeVisible`), but for some reason it did not really work for me as that icon is visible even without hover (test passes if I do not hover and expect it to be visible) :/

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -587,6 +591,113 @@ it('should not show sort icon if disabled in column level', () => {
...
+  expect(container.querySelectorAll('.iui-cell-end-icon > svg')[0]).toEqual(
```
veekeys, line 624: you dont need to query all if you take the first one, just one `querySelector`

packages/iTwinUI-react/src/core/Table/Table.tsx
elephantcatdog: Seems like `column.isSortedDesc || (!column.isSorted && column.sortDescFirst)` works though

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: Removed `mouseEnter`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: Fixed

veekeys (approved):

apps/storybook/src/Table.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: Well, I think I'll leave it as it is now then.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -587,6 +591,111 @@ it('should not show sort icon if disabled in column level', () => {
...
+  const {
+    container: { firstChild: sortUpIcon },
+  } = render(<SvgSortUp className='iui-icon iui-sort' aria-hidden />);
+  const {
+    container: { firstChild: sortDownIcon },
+  } = render(<SvgSortDown className='iui-icon iui-sort' aria-hidden />);
+  const nameHeader = container.querySelector(
+    '.iui-table-header .iui-cell',
+  ) as HTMLDivElement;
+  expect(nameHeader).toBeTruthy();
```
rohan-kadkol, line 620: Don't think this would matter, but this check might be unnecessary here. Since this test is not doing anything with the nameHeader. The other two existing tests that do this `nameHeader` check also use `nameHeader`, which makes the check important.

But this change is optional, just something I noticed

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -587,6 +591,111 @@ it('should not show sort icon if disabled in column level', () => {
...
```
rohan-kadkol, line 673: Same as the previous comment about nameHeader

r100-stack (approved):
Sorry for the delay. Had one observation that I just wanted to mention (no need to act on it).

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: I left this check in because I click on the nameHeader in lines 628, 634, and 640.

PR #787, opened 2022-08-19
feat(MiddleTextTruncation): Add text renderer

### LINKED ISSUE #782: Conditional MiddleTextTruncation tooltip
#### Feature
Requests the ability to display a tooltip only when the text in the MiddleTextTruncation element is truncated.

andriusgv:
Closes #782 

Currently, the user has no way to check whether the text has been truncated or not. Such functionality can be useful, for example, to display a tooltip only for truncated text.

## Checklist


packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
```diff
@@ -59,3 +59,30 @@ it('should leave original text', () => {
   const containerSpan = container.querySelector('span') as HTMLSpanElement;
   expect(containerSpan.textContent).toBe('No trunc');
 });
+
+it('should call onTruncate callback with true', () => {
+  jest.spyOn(UseOverflow, 'useOverflow').mockReturnValue([jest.fn(), 20]);
+  const onTruncate = jest.fn();
+  render(
+    <div>
```
veekeys, line 67: `div` here should not be necessary

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -18,6 +18,10 @@ export type MiddleTextTruncationProps = {
    * @default 6
    */
   endCharsCount?: number;
+  /**
+   * Callback fired when text is truncated.
+   */
+  onTruncate?: (isTruncated: boolean) => void;
```
veekeys, line 24: Just to be sure that this api will be fine and extendable, could you update the existing middle text truncation story in `Select`. It is under our apps/Storybook.
I am also thinking passing children instead of listening to callback might be an option too.

veekeys (commented):
Thank you for the contribution!
Please also accept CLA.

r100-stack (commented):
Looking good! Just two minor nitpick suggestions

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
```diff
@@ -60,29 +60,23 @@ it('should leave original text', () => {
...
-    <div>
-      <MiddleTextTruncation onTruncate={onTruncate} text='No trunc' />
-    </div>,
-  );
-
-  expect(onTruncate).toHaveBeenCalledWith(false);
+  const containerSpan = container.querySelector('span') as HTMLSpanElement;
+  expect(containerSpan.textContent).toBe('This is some …lipsis');
+  expect(
+    containerSpan.querySelector('[data-testid="custom-text"]'),
```
veekeys, line 78: this looks weird, cause its same element.. why dont you just use this selector on line 77 and check its text content.

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -19,33 +19,37 @@ export type MiddleTextTruncationProps = {
...
-
 /**
  * Truncates text with the ellipsis in the middle,
  * leaving defined number of chars at the end.
  * @example
  * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' />
+ * @example
+ * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' endCharsCount={10} />
+ * @example
+ * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' textRenderer={(truncatedText, originalText) => <span title={originalText}>{truncatedText}</span>} />
```
veekeys, line 38: Good that you added examples, but I think we want to recommend memoizing the `textRenderer`, so use `useCallback` in here and in stories too, to avoid not needed rerenders.

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -19,33 +19,37 @@ export type MiddleTextTruncationProps = {
    */
   endCharsCount?: number;
   /**
-   * Callback fired when text is truncated.
+   * Custom renderer for the truncated text.
    */
-  onTruncate?: (isTruncated: boolean) => void;
+  textRenderer?: (
+    truncatedText: string,
+    originalText: string,
```
veekeys, line 26: What is the reason you are passing `originalText`? User would have it as he is passing it to us

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -59,7 +63,7 @@ export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
       ref={ref}
       {...rest}
     >
-      {truncatedText}
+      {textRenderer ? textRenderer(truncatedText, text) : truncatedText}
```
veekeys, line 68: hm... we have this parent `span` element with some styles to make  it work.. wondering if passing some additional children which are going to wrap the truncated text could mess up something.. Not sure yet, but if your examples work, I guess its fine.

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -18,17 +18,26 @@ export type MiddleTextTruncationProps = {
...
  * @example
  * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' />
+ * @example
+ * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' endCharsCount={10} />
+ * @example
+ * <MiddleTextTruncation text='ThisIsMyVeryLongFileName.dgn' textRenderer={(truncatedText, originalText) => <span title={originalText}>{truncatedText}</span>} />
  */
 export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
-  const { text, endCharsCount = 6, style, ...rest } = props;
-
```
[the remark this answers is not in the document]
andriusgv: Ok, minimized for both comments

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
```diff
@@ -59,3 +59,22 @@ it('should leave original text', () => {
...
 });
+
+it('should render custom text', () => {
+  jest.spyOn(UseOverflow, 'useOverflow').mockReturnValue([jest.fn(), 20]);
+  const text = 'This is some very long text to truncate and expect ellipsis';
+  const { container } = render(
+    <MiddleTextTruncation
+      text={text}
+      textRenderer={(truncatedText) => (
+        <span data-testid='custom-text'>{truncatedText}</span>
```
veekeys, line 70: just for fun as your test says `render custom text` try to add some additional chars together with `truncatedText` to check up that content will be rendered.

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
andriusgv: Ok, I removed this check and attribute

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
andriusgv: Added callbacks

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
andriusgv: Added some custom text to the end

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
andriusgv: In general, I think that the user will try to reuse the text rendering functions in multiple places, but after truncation, the text may become heavily modified beyond recognition, so it will lose the reference to the source. Thus, the user won't be able, for example, to show a tooltip with original text.

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
andriusgv: Well, I tested it with the Select component, also tried it on a playground, and seems that it works fine even with nested component

apps/storybook/src/Select.stories.tsx
```diff
@@ -344,12 +344,15 @@ export const TruncateMiddleText: Story<SelectProps<string>> = (args) => {
         <MenuItem>
           <MiddleTextTruncation
             text={option.label}
-            textRenderer={textRenderer}
+            textRenderer={textRenderer(option.label)}
```
veekeys, line 347: That is looking really weird :/
Could have something like `textRenderer={useCallback((truncatedText) => textRenderer(truncatedText, option.label))}` ?
Or ok..  you made your point and return the original text :D I think @rohan-kadkol is also voting to return both truncated and original text..

rohan-kadkol: Yes, thanks! Or maybe something like `textRenderer={textRenderer}`, where `textRenderer` is defined as

```ts
const textRenderer = useCallback(
  (originalText: string, truncatedText: string) => (
    <span title={truncatedText !== originalText ? originalText : undefined}>
      {truncatedText}
    </span>
  ),
  [],
);
```

and `MiddleTextTruncationProps`'s `textRenderer()` has `originalText` and `truncatedText`

```ts
export type MiddleTextTruncationProps = {
  ...

  textRenderer?: (originalText: string, truncatedText: string) => React.ReactNode;
};
```

and then when returning in the component, we pass `text` and `truncatedText` to `textRenderer()`

```tsx
return (
    <span
      ...
    >
      {textRenderer ? textRenderer(text, truncatedText) : truncatedText}
    </span>
  );
```

I feel with this way, users can easily get access to the `originalText` in `textRenderer()` and display what they want accordingly.

@veekeys What do you recommend?

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.test.tsx
```diff
@@ -59,3 +59,21 @@ it('should leave original text', () => {
...
+        <span>{truncatedText} - some additional text</span>
+      )}
+    />,
+  );
+
+  const containerSpan = container.querySelector('span') as HTMLSpanElement;
+  expect(containerSpan.textContent).toBe(
+    'This is some …lipsis - some additional text',
+  );
+});
```
rohan-kadkol, line 79: Maybe adding a short test to verify that the custom React node is indeed rendered. So something like how you did before, but adding the `.toBeTruthy()`

```tsx
expect(
  containerSpan.querySelector('[data-testid="custom-text"]'), // The <span> in customRenderer() has data-testid="custom-text"
).toBeTruthy();
```

rohan-kadkol, line 79: Maybe adding a short test to verify that the custom React node is indeed rendered. So something like how you did before, but adding the `.toBeTruthy()`

```tsx
expect(
  containerSpan.querySelector('[data-testid="custom-text"]'), // The <span> in textRenderer() has data-testid="custom-text"
).toBeTruthy();
```

and `textRenderer` on lines 69-71 to be something like

```tsx
textRenderer={(truncatedText) => (
  <span data-testid='custom-text'>{truncatedText} - some additional text</span>
)}
```

This is because the selector on line 75 is for the parent `<span>` (lines 60-70 of *MiddleTextTruncation.tsx*) and the selector written above is for the `textRenderer()`'s `<span>`

GintV:
Maybe I missed it in tests, but how one should use this to show tooltip (aka title) only if truncation took place?

apps/storybook/src/Select.stories.tsx
veekeys: yeah.. that was the same as it was before done by @andriusgv 😄 
So as seems you both tend to agree to have original text passed, I guess we can do that and be done :)

packages/iTwinUI-react/src/core/utils/components/MiddleTextTruncation.tsx
```diff
@@ -54,7 +68,7 @@ export const MiddleTextTruncation = (props: MiddleTextTruncationProps) => {
       ref={ref}
       {...rest}
     >
-      {truncatedText}
+      {textRenderer ? textRenderer(truncatedText) : truncatedText}
```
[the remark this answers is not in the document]
rohan-kadkol: Yes, that's better!

veekeys (approved):

r100-stack (approved):

PR #791, opened 2022-08-22
fix(Header-v1): Added disabled state for Header breadcrumb items

adamhock:
- Addresses [this comment](https://github.com/iTwin/iTwinUI-react/pull/755#discussion_r950098171) made in #755
- Added `aria-disabled` to `iui-header-breadcrumb-item` to apply the proper styling.


packages/iTwinUI-react/src/core/Header/HeaderButton.tsx
```diff
@@ -113,6 +115,7 @@ export const HeaderButton: HeaderButtonComponent = React.forwardRef(
       <li
         className='iui-header-breadcrumb-item'
         aria-current={isActive ? 'location' : undefined}
+        aria-disabled={disabled ? 'true' : undefined}
```
veekeys, line 118: Can you add disabled button to at least one of our stories? Would be good to have test for it :)

veekeys: And verify that all others work (basic, split and dropdown)

adamhock: I disabled the split button in the Full Header story. These are what each of the following disabled buttons look like (split, dropdown, and basic)

image

veekeys (approved):

r100-stack (approved):
Looks good! Only one thing I wondered about was that there are no unit tests. But maybe they are not needed for this change so it's okay

PR #796, opened 2022-08-22
chore: Updated itwinui-css to dev.1

veekeys:
- Updated `TextArea`
- Fixed some missed `Table` tests selectors
- Removed hardcoded action column menu height
- Updated `SplitButton`
- merged with `main`
- Lots of images changed because of baseline change... Need to carefully review all


mayank99:
@veekeys Did you go through all the images yourself before approving?

I noticed that the Table paginator is missing active state.



Also, he Tree images' height seems to have gone down but this might be an issue with how cypress is capturing the screenshot

gretanausedaite (commented):
Slider end icons are different colour.
image

I think Mayank noticed and documented every issue.

veekeys:
> @veekeys Did you go through all the images yourself before approving?
> 
> I noticed that the Table paginator is missing active state. 
> 
> Same thing in sidenav 
> 
> And the expandable block status icons are missing fill 
> 
> Also, the Tree images' height seems to have gone down but this might be an issue with how cypress is capturing the screenshot 
> 
> Also, the Slider icon labels have become more dark, but I think that's fixed in my variables PR. 

I did it only locally with really bad Cypress html file by scrolling through... It took me a while to actually run all the tests, cause they all were failing..
I will do it through github and image comparer deeper, but your feedback is already really helpful. Thank you

veekeys:
Oh wow...
I have updated `Paginator`, `SideNav`, `Tile` and `ExpandableBlock` which were ALL MISSED either from `Icon` changes or `Button` changes!
This is huge miss!
I will need to verify `Slider` and `Tree` is interesting too...
Apart from these, I have noticed some other changes, which might be worth to raise, I will try to put those in a list tomorrow.

veekeys:
Ok, I checked situation in css, so slider icons are fixed as you mentioned with the css variables PR. 
So here we should be good for now.

rohan-kadkol:
Some test image diffs I found:

* Header.test.ts-Full: The second breadcrumb went from disabled to enabled. But maybe it is supposed to be enabled? Because the current hosted [storybook](https://itwin.github.io/iTwinUI-react/?path=/story/core-header--full) has it enabled.
image
* IconButton.test.ts-Small Active Add: The bottom border seemed to be missing before but was added now. (I don't know if there is a problem with the way the test images are displaying for me. So might just be an issue with my screen)
image
* LabeledTextArea.test.ts-With Custom Icon: Bottom border might be missing (or just a display problem with my image viewer)
image
* SideNavigation.test.ts-Active Item & Basic: The top border of the settings icon might be missing earlier and added now (or just a display problem with my image viewer)
image
* Toasts.test.ts-Negative, Positive, Warning: The red/green/yellow borders might be missing from the toasts (or just a display problem with my image viewer)
image
image
image

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -597,6 +601,111 @@ it('should not show sort icon if disabled in column level', () => {
   ).toBeFalsy();
 });
 
+it('should display correct sort icons for ascending first', async () => {
```
bentleyvk, line 604: Bad merge or what?

gretanausedaite: I think this was added because this branch is updated with main changes (not latests though)

gretanausedaite:
> Some test image diffs I found:

I think we should keep current header test image with all not disabled buttons to keep it as it is currently.
Everything else looked fine to me. Will check again

bentleyvk (approved):

PR #802, opened 2022-08-25
fix(Table): Keyboard Navigation Support

### LINKED ISSUE #488: Table: Can't use keyboard navigation to trigger sort or filter
It is impossible to trigger sort or open filter menu with keyboard.

elephantcatdog:
Fixed sorting, filtering, and anchor navigation issues in the table.

image
Sorting:
 - Added `tabIndex` and `onKeyDown` to the div that contains the sort icon button

image
Filtering
 - Filtering is fixed by no longer making the button `visibility: hidden` when not hovered on. Instead of using `visibility` to control the look of the button, I changed it to use `opacity`. This is done in the CSS repo and more info can be found at that PR: https://github.com/iTwin/iTwinUI/pull/752

image
Anchors
 - I added `tabIndex` and `onKeyDown` props to the anchors found in `Table.stories.tsx`. I also tried fiddling with the table's `tabIndex={-1}`, but that wasn't affecting this issue.
 - One of the stories does not check if the key pressed is "Enter" (lines 1064-1068). I couldn't figure out how to do it and retain the functionality of the actions tab on Storybook.

Added tests for sorting and filtering
 - <b>The test for filtering does not work correctly</b>, even though it passes, because it also passes when it's supposed to fail. I assume that `userEvent.tab()` is a 'better' tab than that made by using the keyboard? Because it's able to correctly tab through the filtering menu even when `visibility: hidden`.

Closes iTwin/iTwinUI-react#488 

## Checklist

- ~[ ] Add component features demo in Storybook (different stories)~ not needed
- ~[ ] Approve test images for new stories~ not needed


elephantcatdog:
There are extraneous changes shown in this PR right now because the main branch hasn't been merged into itwinui-css-v1, and I initially made this branch off of main. If you want to review it, the relevant changes are in Table.tsx, Table.test.tsx, and Table.stories.tsx.

elephantcatdog:
I checked if `Button`s were accessible with the keyboard within a table and they are. The issue with `Anchor`s in a table doesn't seem to be a table issue.

apps/storybook/src/Table.stories.tsx
```diff
@@ -118,7 +118,19 @@ export const Basic: Story<Partial<TableProps>> = (args) => {
             width: 100,
             Cell: (props: CellProps<{ name: string; description: string }>) => {
               const onClick = () => onClickHandler(props);
-              return <Anchor onClick={onClick}>Click me!</Anchor>;
+              return (
+                <Anchor
+                  onClick={onClick}
+                  tabIndex={0}
```
bentleyvk, line 124: Are these needed? Don't native anchor handle focus and keyboard by itself?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3688,3 +3797,60 @@ it('should make column sticky and then non-sticky after dragging sticky column a
     expect(cell).not.toHaveStyle('--iui-table-sticky-left: 400px');
   });
 });
+
+it('should navigate through table sorting with the keyboard', async () => {
+  const onSort = jest.fn();
+  const mockedColumns = [
```
bentleyvk, line 3803: Is it really needed?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3688,3 +3797,60 @@ it('should make column sticky and then non-sticky after dragging sticky column a
...
+  ];
+  renderComponent({
+    columns: mockedColumns,
+    onFilter,
+  });
+
+  await userEvent.tab();
+  await userEvent.keyboard('{Enter}');
+  await userEvent.keyboard('2');
+  await userEvent.tab();
```
bentleyvk, line 3850: Maybe add comments where are you tabbing for better readability.

apps/storybook/src/Table.stories.tsx
elephantcatdog: In the table, anchor isn't able to be navigated to using the keyboard. Buttons are, though, which makes me think that this isn't a table problem, but an anchor problem or a table-anchor interaction problem.

elephantcatdog: Further research shows it's an anchor problem. Anchors inside alerts aren't reachable by keyboard either.

elephantcatdog: I will remove these additions and look into Anchor.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: It's not, I removed it.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: Done

apps/storybook/src/Table.stories.tsx
elephantcatdog: Ah! Thank you!!

elephantcatdog:
I played with the tests, trying to get it to work (fail) for the filter test.

elephantcatdog:
The problems with `Anchor` were further reaching than I thought I should go for this PR, so I made a new issue: https://github.com/iTwin/iTwinUI-react/issues/807

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -813,7 +813,15 @@ export const Table = <
                               />
                             )}
                             {showSortButton(column) && (
-                              <div className='iui-table-cell-end-icon'>
+                              <div
```
bentleyvk, line 816: In CSS it is the whole cell that has `tabIndex`

elephantcatdog: That is true. I feel as though it's confusing to tab onto the filter button for filtering and tab onto the header cell for sorting.

bentleyvk: But this is how even mouse clicks work - you press on the whole cell not the icon.

elephantcatdog: Yep, you're totally right. Do you think the header cell be tab-focusable if there's no sort or filter button? The CSS demo shows the header cell being tab-focusable when there's the filter button but not the sort button, but doesn't have an example of a regular/text header cell with no filter or sort buttons.

elephantcatdog: I followed how the mouse clicks/hovers work for the react table. The header cells change color on hover, so I made it so that all header cells can be tab-focusable.

elephantcatdog:
I removed the test for filtering navigation as I couldn't find a solution to make it fail/pass correctly.

elephantcatdog (commented):
I noticed that the icon colors for sort changed recently from white to black; is this an expected change?
image

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -801,6 +801,12 @@ export const Table = <
                             column.resizeWidth = el.getBoundingClientRect().width;
                           }
                         }}
+                        tabIndex={0}
```
bentleyvk, line 804: I think it should only be focusable if it can be sorted.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3907,3 +3907,15 @@ it.each(['positive', 'warning', 'negative'] as const)(
...
+it('should navigate through table sorting with the keyboard', async () => {
+  const onSort = jest.fn();
+  renderComponent({
+    isSortable: true,
+    onSort,
+  });
+
+  await userEvent.tab(); // tab to sort icon button
+  await userEvent.keyboard('{Enter}');
+  expect(onSort).toHaveBeenCalled();
```
bentleyvk, line 3920: What about filtering?

elephantcatdog: I was having lots of trouble with the filtering test and couldn't make it fail when it was supposed to. Do you have any suggestions? https://teams.microsoft.com/l/message/19:23e76cd5984f40eca958bc496bb00aaa@thread.tacv2/1661891083535?tenantId=067e9632-ea4c-4ed9-9e6d-e294956e284b&groupId=8a0b1bff-5d9d-4a29-9112-7d412098d768&parentMessageId=1661891083535&teamName=iTwinUI%20development&channelName=Issues%20and%20help&createdTime=1661891083535&allowXTenantAccess=false

bentleyvk: Have you tried `fireEvent.focus(el)`?

elephantcatdog: Okay, I tried:
`const filterIconButton = container.querySelector(
    '.iui-table-filter-button',
  ) as HTMLElement;`
  `fireEvent.focus(filterIconButton);`
and
  `fireEvent.blur(filterIconButton);`
and
  `filterIconButton.focus();`
and
  `const table = container.querySelector('.iui-table') as HTMLElement;`
  `fireEvent.keyDown(table, { key: 'Tab' });`

Only `blur` passes-ish, but it passes every time and it gives a console error:
Screenshot 2022-09-09 082108
I'm in the process of looking the error up but haven't found anything yet.

elephantcatdog: Readded filter test

packages/iTwinUI-react/src/core/Table/Table.tsx
elephantcatdog: I did it this way because it followed how the mouse clicks/hovers work for the react table. The header cells change color on hover, so I made it so that all header cells can be tab-focusable.

I changed it to be focusable only if sortable though.

elephantcatdog: @adamhock What do you think?

bentleyvk (approved):

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: I agree with Vilius. I think that unless there's an action you can take on the header cells (such as sorting), then it should not be focusable.

adamhock (approved):
LGTM. Nice job 🎉

PR #809, opened 2022-08-31
fix(Dialog): Only focus dialog when `setFocus` is `true`

### LINKED ISSUE #784: Dialog always hijacks focus
Even when `trapFocus` is set to false, using `Dialog.Main` with `isOpen={true}` causes it to focus the dialog.

This is not ideal because sometimes you don't want the focus to move. e.g. for a cookie dialog, it might not necessarily be the best move to autofocus the dialog.

The problem is likely this line:

https://github.com/iTwin/iTwinUI-react/blob/2de1c9a49ee08a7afbf58f0576f12a522cf5b7f3/packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx#L73

bentleyvk:
Closes #784

## Checklist


packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -79,7 +84,7 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
         ref?.contains(document.activeElement) &&
           previousFocusedElement.current?.focus();
       };
-    }, [isOpen]);
+    }, [isOpen, setFocus]);
```
mayank99, line 87: Won't this unnecessarily trigger this entire effect when `setFocus` changes from `false` to `true`? We only want the effect to be triggered when `isOpen` changes, so I'd say put `setFocus` in a ref.

bentleyvk: And what if someone wants to set focus after dialog was opened?

mayank99: so if `setFocus` is set to true before the dialog is opened, then it should still set focus after the dialog opens

bentleyvk: No, I mean if dialog was opened and some other action happened that makes already opened dialog to be focused/
But still `setFocus` is boolean so it won't change on every rerender so I don't see a problem here.

mayank99: Why would anyone want to focus the dialog using the prop after it's already opened? It will also break the `previousFocusedElement` logic, so there's an even stronger reason to prevent this kind of behavior. Plus it will trigger this useEffect when setFocus changes even when the dialog is closed.

If the user really has a valid use case for this, then they should call `dialogRef.current.focus()`.

adamhock (commented):
I do agree with Mayank that `setFocus` should be in its own ref

Outside of that, I have no additional comments

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
elephantcatdog: I don't think it matters if having `setFocus` doesn't break anything. But if it's breaking something else and it can be achieved in another way then I'd say don't include `setFocus`. But generally I think it's good to enable users to have more options even if we don't see the point. I've talked to people that want to use/change our stuff in a way we didn't consider and sometimes that makes using our stuff more difficult than it needs to be.

bentleyvk: Changed to ref although I completely disagree with this...

mayank99:
Looks like tests are failing after latest merge.

mayank99 (approved):

adamhock (approved):
Looks good to me 🎉

PR #811, opened 2022-08-31
fix(VirtualScroll): Show all children upon toggling display/visibility

### LINKED ISSUE #718: Virtualized table doesn't render all elements
When table is created but not shown, only 20 rows are rendered afterwards. Scrolling the table doesn't make any more rows appear. Only occurs when virtualization is enabled.

#### Expected behavior
Table should render all rows in viewport.

#### Reproduction
https://codesandbox.io/s/itwinui-react-minimal-example-forked-zxyp3w?file=/src/App.tsx

##### Steps to reproduce
1. Press 'Show' button.
2. Scroll
3. See that no more rows are rendered

#### Additional information
Looks like in VirtualScroll.js `childHeight` is initially set to 0 and doesn't update. It depends on `visibleChildren`, which is calculated using `visibleNodeCount`. The problem is, if `childHeight == 0`, `visibleNodeCount` doesn't update either. Would call this a circular dependency?

r100-stack:
Closes #718 

## Checklist

- *[N/A]* Add meaningful unit tests for your component (verify that all lines are covered)
- *[N/A]* Add component features demo in Storybook (different stories)
- *[N/A]* Approve test images for new stories

##### Old

[vokoscreenNG-2022-09-02_15-43-48.webm](https://user-images.githubusercontent.com/45748283/188225908-5e63647f-d731-423b-b334-58248fdfedbc.webm)

##### New

[vokoscreenNG-2022-09-02_15-42-23.webm](https://user-images.githubusercontent.com/45748283/188225927-af93e9fc-5936-49cc-b666-5a81deaef93a.webm)


packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -215,8 +184,7 @@ export const useVirtualization = (props: VirtualScrollProps) => {
     return arr;
   }, [itemsLength, itemRenderer, bufferSize, startNode, visibleNodeCount]);
 
-  // Get child height when children available
-  React.useLayoutEffect(() => {
+  const updateChildHeight = React.useCallback(() => {
```
rohan-kadkol, line 187: **[COMMENT]** The main change I made was to abstract the setting of `childHeight` (old lines 220-242). Then `updateChildHeight()` can be called in `React.useLayoutEffect()` and in `onResize()`. That way when the resize happens (`display` goes from `display: none` to `display: block`), the `childHeight` is updated which allows all us to visit all the items.

apps/playground/src/main.css
```diff
@@ -5,4 +5,4 @@
 
 html {
   color-scheme: dark light;
-}
\ No newline at end of file
+}
```
rohan-kadkol, line 8: When I saved this file, it automatically added a new line at the end. Is this added new line okay?

gretanausedaite: It would be nice if you removed this empty line. To not include changes that are not actually changing anything.

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -242,6 +210,45 @@ export const useVirtualization = (props: VirtualScrollProps) => {
     };
   }, [visibleChildren.length]);
 
+  const onResize = React.useCallback(
```
adamhock, line 213: Minor, but could you move this variable back up to line 175 where it was previously? It makes it easier to review when the changes are side by side

rohan-kadkol: Ohh, I had tried doing that before, but I left it the current way because `onResize` now depends on `updateChildHeight`. So `onResize` now needs to be below `updateChildHeight`.

The lines after `onResize` 226-250 have to be in the order it is in since the next one is dependent on the previous one. However, I noticed that `getScrollableContainer` could be placed on line 175 to avoid an unnecessary change. Thus, I made that small [change](https://github.com/iTwin/iTwinUI-react/pull/811/commits/4e9b6f9eb11ae675960f0d00f5201a9d90eb7c1b)

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -242,6 +214,41 @@ export const useVirtualization = (props: VirtualScrollProps) => {
...
```
gretanausedaite, line 217: How do you know that `display: 'none'` to `display:'block'` is resize event?

adamhock:
[188225927-af93e9fc-5936-49cc-b666-5a81deaef93a.webm](https://user-images.githubusercontent.com/78564221/189424125-19bcc6f7-6d7f-48c4-8073-5a17f9a36f07.webm)

I don't like how the virtual scroll bar resizes a few frames after the table is open, thus causing the scroll bar to be bigger. I'm not sure if there is a solution for this, but outside of that, the PR looks good to me

rohan-kadkol:
> 188225927-af93e9fc-5936-49cc-b666-5a81deaef93a.webm 
> I don't like how the virtual scroll bar resizes a few frames after the table is open, thus causing the scroll bar to be bigger. I'm not sure if there is a solution for this, but outside of that, the PR looks good to me

Yes, that's true. The user sees the resetting of the number of items for a short moment after pressing the show button. I'm not sure at the moment why it is happening. It may be because the `onResize` callback is called after the first frame is rendered, instead of before. So the `childHeight` is updated after the first frame is rendered. If so, I don't know how can that be avoided.

Does anyone know why this could be happening? @gretanausedaite @mayank99

packages/iTwinUI-react/src/core/utils/components/VirtualScroll.tsx
```diff
@@ -242,6 +214,41 @@ export const useVirtualization = (props: VirtualScrollProps) => {
...
+    resizeRef(scrollableParent);
+  }, [resizeRef]);
+
+  // Stop watching resize, when virtual scroll is unmounted
+  React.useLayoutEffect(() => {
+    return () => resizeObserver?.disconnect();
+  }, [resizeObserver]);
+
+  // Get child height when children available
+  React.useEffect(() => updateChildHeight(), [updateChildHeight]);
```
gretanausedaite, line 250: previously updateChildHeight was layoutEffect. I think you could use the same here.

rohan-kadkol: **[DONE]** I changed `useEffect()` to `useLayoutEffect()`. Just curious, what is the difference between the two? I tried looking online but am still not totally sure of the difference

gretanausedaite: Found really good explanation here: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

rohan-kadkol: Very helpful, thanks! The main difference was the synchronous vs asynchronous part. Mentioning that makes everything clearer :smile:

gretanausedaite:
> 
> Does anyone know why this could be happening? @gretanausedaite @mayank99

I did not find a way to fix this. Maybe @mayank99 or @bentleyvk can take a look?

bentleyvk:
> > Does anyone know why this could be happening? @gretanausedaite @mayank99
> 
> I did not find a way to fix this. Maybe @mayank99 or @bentleyvk can take a look?

I have no idea. All effects are `useLayoutEffect` so I have no other suggestion. I would say it is not nice but minor.

gretanausedaite (approved):
Looks good to me 🌵

adamhock (approved):

PR #823, opened 2022-09-12
feat(Dialog): Allows dialog to be resizable

### LINKED ISSUE #614: Resizable and Movable Modal Dialog

bentleyvk:
Closes #614 

dialog-resize

## Checklist


packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+            height = elementRef.current.style.height = `${newHeight}px`;
+            translateY = initialTranslateY - (initialPointerY - clientY);
+          }
+          width = elementRef.current.style.width = `${
+            initialWidth - (initialPointerX - clientX)
+          }px`;
+          elementRef.current.style.transform = `translate(${translateX}px, ${translateY}px)`;
+          break;
+        }
+        case 'right': {
```
bentleyvk, line 127: For right, bottom-right and bottom no need to check min-width and min-height since CSS handles it.

packages/iTwinUI-react/src/core/utils/components/Resizer.test.tsx
```diff
@@ -0,0 +1,368 @@
...
+ * |      |               |       |
+ * |      |    element    |       |
+ * |      +---------------+       |
+ * |                  (200;200)   |
+ * |                              |
+ * |         container            |
+ * +------------------------------+
+ *            w=300px         (300;300)
+ */
+
```
bentleyvk, line 76: Thought of using `it.each` but it looked that it will take almost same amount of code and might not be that readable.

packages/iTwinUI-react/src/core/utils/hooks/useDragAndDrop.tsx
```diff
@@ -68,9 +51,9 @@ export const useDragAndDrop = (
       left,
     } = elementRef.current?.getBoundingClientRect();
 
-    let newTranslateX = translateX.current;
-    let newTranslateY = translateY.current;
+    let [newTranslateX, newTranslateY] = getTranslateValues(elementRef.current);
```
bentleyvk, line 54: Getting `translate` value from the element and not from the `translateX.current` because when resizing, it triggers resize hook inside `useDragAndDrop` and it will use old translate values that breaks the resizing.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * Ref of the container element in order to avoid resizing past container boundaries.
+   * If not passed, viewport will be used.
+   */
+  containerRef?: React.RefObject<HTMLElement>;
+  /**
+   * Callback that is being called on resize end.
+   * Useful to preserve state if element is being closed.
+   */
+  setStyle?: (style: React.CSSProperties) => void;
```
mayank99, line 22: would it make more sense to call this `onResize`? in case we want to pass more things when resize ends

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+      getComputedStyle(elementRef.current).minWidth.replace('px', ''),
+    );
+    const minHeight = Number(
+      getComputedStyle(elementRef.current).minHeight.replace('px', ''),
+    );
+
+    const resizer = event.currentTarget.dataset.iuiResizer;
+
+    const originalUserSelect =
+      elementRef.current.ownerDocument.body.style.userSelect;
```
mayank99, line 70: looks weird performing side effects on another element. maybe this should be handled in dialog with inversion of control?

but minor

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+    const onResizePointerMove = (event: PointerEvent) => {
+      if (!elementRef.current) {
+        return;
+      }
+
+      const containerRect = containerRef?.current?.getBoundingClientRect();
+      const clientX = getBoundedValue(
+        event.clientX,
+        containerRect?.left ?? 0,
+        containerRect?.right ?? getWindow()?.innerWidth ?? 0,
```
mayank99, line 82: I think `document.documentElement.clientWidth` would be better than `window.innerWidth` here because `window.innerWidth` considers the "visual viewport" which can be affected things like zoom

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+          elementRef.current.style.transform = `translate(${translateX}px, ${translateY}px)`;
+          break;
+        }
+        case 'left': {
+          const newWidth = initialWidth + (initialPointerX - clientX);
+          if (newWidth < minWidth) {
+            break;
+          }
+          width = elementRef.current.style.width = `${newWidth}px`;
+          translateX = initialTranslateX - (initialPointerX - clientX);
```
mayank99, line 166: maybe save the `initialPointerX - clientX` part in a variable for readability and reuse

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+          elementRef.current.style.transform = `translate(${translateX}px, ${translateY}px)`;
```
mayank99, line 167: oh this one is also a huge side effect, i think this should be handled in the dialog for sure. Do we even need to set it here if we are returning it in `setStyle`?

mayank99:
looks like there are a few merge conflicts

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -144,10 +170,19 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
         tabIndex={-1}
         style={{
           transform,
+          overflow: 'unset',
```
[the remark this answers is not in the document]
mayank99: Can you leave a comment on https://github.com/iTwin/iTwinUI/issues/759?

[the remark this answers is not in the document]
bentleyvk: Done

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+        }
+      },
+      { once: true },
+    );
+  };
+
+  return (
+    <>
+      <div
+        data-iui-resizer='top-left'
```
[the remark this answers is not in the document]
bentleyvk: > but now I'm thinking we _should_ have styling for these like a colored border when hovering edges.

I guess question for @FlyersPh9

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+        onPointerDown={onResizePointerDown}
+        style={{
+          position: 'absolute',
+          top: -4,
+          left: -4,
+          width: 8,
+          height: 8,
```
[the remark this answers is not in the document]
bentleyvk: idk, copied from iTwin.js, seemed something similar to the Windows file explorer.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: I would expect `onResize` to be called on every `onMouseMove` call, maybe then it could be `onResizeEnd`

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: I was making this component that it could be reused so this `userSelect` part is kind of also needed for other components that is going to use it in order to prevent unwanted selections.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: I would disagree with this one cause the purpose of this component is to handle resizing and so it does that, also calling `setStyle` on each pointer move would be causing a lot of rerenders.

packages/iTwinUI-react/src/core/utils/functions/styles.tsx
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+export const getTranslateValues = (element: HTMLElement | null | undefined) => {
```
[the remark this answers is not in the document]
bentleyvk: Added

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+   * Callback that is being called on resize end.
+   * Useful to preserve state if element is being closed.
+   */
+  setStyle?: (style: React.CSSProperties) => void;
+};
+
+/**
+ * Component that allows to resize parent element.
+ * Parent must have `position: relative`.
+ * @example
```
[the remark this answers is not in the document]
bentleyvk: Made

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: Changed

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: Moved to a variable

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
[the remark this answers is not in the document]
bentleyvk: ok, windows has it bigger so increased the size

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
```diff
@@ -0,0 +1,298 @@
...
+      height: initialHeight,
+    } = elementRef.current.getBoundingClientRect();
+
+    let width = `${initialWidth}px`;
+    let height = `${initialHeight}px`;
+    let translateX = initialTranslateX;
+    let translateY = initialTranslateY;
+
+    const minWidth = Number(
+      getComputedStyle(elementRef.current).minWidth.replace('px', ''),
```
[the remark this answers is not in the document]
bentleyvk: Changed to `parseFloat`

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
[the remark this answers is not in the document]
FlyersPh9: Can we try releasing without coloring the edges and see if we get any feedback on users not knowing it can be resized?  I'm not used to seeing dialogs where the edges / corners light up to show that they can be resized.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
mayank99: `onResizeEnd` works 👍

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
mayank99: But this relies on knowing that the thing that is being resized _always_ needs a transform applied. Rerenders are not always necessary because you can call this same line `style.transform = ...` inside Dialog without putting it in state.

And again, why are we returning it in setStyle currently if it will cause extra rerenders?

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
```diff
@@ -143,6 +154,33 @@ export const DialogMain = React.forwardRef<HTMLDivElement, DialogMainProps>(
       [isDraggable, onPointerDown],
     );
 
+    // Prevents dialog from moving when window is being resized
+    React.useLayoutEffect(() => {
```
mayank99, line 158: useLayoutEffect causes warnings when doing SSR. https://reactjs.org/link/uselayouteffect-ssr

You can conditionally call useEffect on the first render (server) and useLayoutEffect on every subsequent render (client).

I introduced a new util in #825 but that's in the v1-update branch (v2?).
https://github.com/iTwin/iTwinUI-react/blob/a30870382e312be40678e386ea28160ac54a9e6d/packages/iTwinUI-react/src/core/utils/hooks/useIsomorphicLayoutEffect.tsx#L12-L13

Currently in the main branch every component already calls `useLayoutEffect` (via `useTheme`) so maybe it's fine for this PR but I feel like we should replace all remaining instances of useLayoutEffect in the v1-update branch.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: `Resizer` code relies on `transform` cause its value is taken on resize start, so `Resizer` only works with `transform`. `setStyle` is needed to preserve resize state when `Dialog` is being closed by setting `isOpen={false}` and reopening again, the same as `useDragAndDrop`.

packages/iTwinUI-react/src/core/utils/components/Resizer.tsx
bentleyvk: Renamed

packages/iTwinUI-react/src/core/Dialog/DialogMain.tsx
bentleyvk: Maybe we can leave it for v1 update with your newly introduced hook? Or we can even do it after v1-update branch is merged.

mayank99 (approved):

PR #825, opened 2022-09-12
feat: Allow scoped styles using `ThemeProvider`

### LINKED ISSUE #813: Scoped styles
#### Feature
Related to: https://github.com/iTwin/viewer/issues/201

I would like to be able to render an iModel in a div on a page and be contained within that div. However, the iModel adds classes to the body instead of adding them to the container div where the iModel is rendered. I think this may also fix another issue I have noticed: the popovers are rendered at the bottom of the page:

image

mayank99:
This PR adds the ability to use scoped styles by wrapping subtrees in `ThemeProvider`. 🎉

Closes #813

### Changes to `ThemeProvider`

1.  Renders a `<div>` with the `iui-root` class and data attributes (`data-iui-theme` and `data-iui-contrast`). This requires the user to wrap their app tree in `ThemeProvider`.
2. Uses the React Context API to make theme info available to descendants.
3. If no children are specified, then it will default to the old behavior where it simply calls `useTheme`.

### Changes to `useTheme`

1. Uses the React Context API to detect if theme was provided somewhere up the tree. If yes, then does nothing.
2. If no theme is provided, then it falls back to old behavior where the theme is applied dynamically to the root element.
3. Now uses an isomorphic layout effect to prevent SSR warnings.

### Tests

Added new unit tests for when children are passed, and kept old tests (with slight changes) for the fallback behavior.

Unsure if we need stories.

For a quick way to test a few different cases, replace `apps/playground/src/App.tsx` with the following code:

<details>
  <summary> App.tsx </summary>

```tsx
import styled from '@emotion/styled';
import {
  Button,
  LabeledInput,
  Surface,
  ThemeProvider,
  Alert,

const App = () => {
  return (
    <>
      <Wrapper>
        <Content theme='Default (global useTheme switcher)' />

        <ThemeProvider>
          <Content theme='Default (with extra ThemeProvider wrapper)' />
        </ThemeProvider>

        <ThemeProvider theme='os'>
          <Content theme='OS' />
        </ThemeProvider>

        <ThemeProvider theme='light'>
          <Content theme='Light' />
        </ThemeProvider>

        <ThemeProvider theme='dark'>
          <Content theme='Dark' />
        </ThemeProvider>

        <ThemeProvider theme='light' themeOptions={{ highContrast: true }}>
          <Content theme='Light (high contrast)' />
        </ThemeProvider>

        <ThemeProvider theme='dark' themeOptions={{ highContrast: true }}>
          <Content theme='Dark (high contrast)' />
        </ThemeProvider>
      </Wrapper>
    </>
  );
};

const Content = ({ theme = 'Default' }) => {
  return (
    <ContentWrapper elevation={2}>
      <ContentLabel>{theme}</ContentLabel>
      <Alert clickableText='Learn more'>This is an alert</Alert>
      <Button>Hello world</Button>
      <LabeledInput label='Label' placeholder='Placeholder...' />
      <Button styleType='high-visibility'>Goodbye cruel world</Button>
    </ContentWrapper>
  );
};

const ContentWrapper = styled(Surface)`
  padding: var(--iui-size-m);
  display: grid;
  gap: var(--iui-size-s);
  max-width: max(40vw, 300px);
`;

const ContentLabel = styled.div`
  font-style: italic;
  position: relative;

  &::before {
    content: '';
    bottom: 0;
    position: absolute;
    width: 100%;
    border-top: 1px solid;
  }
`;

const Wrapper = styled.div`
  display: grid;
  gap: 1rem;
`;

export default App;
```
</details>


README.md
```diff
@@ -62,29 +60,13 @@ import { ThemeProvider } from '@itwin/itwinui-react';
...
-    <ThemeProvider theme='dark' />
-    // Your code goes here.
+    <ThemeProvider theme='dark'>
+      {/* Your components go here. */}
+    </ThemeProvider>
   </>
 );
 ```
 
-- The `useTheme` hook also provides the same functionality as `ThemeProvider`.
```
mayank99, line 71: I've removed `useTheme` from the README because our users should wrap their app in `ThemeProvider` whenever possible.

packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.tsx
```diff
@@ -3,31 +3,77 @@
...
+  const prefersDark = useMediaQuery('(prefers-color-scheme: dark)');
+  const prefersHighContrast = useMediaQuery('(prefers-contrast: more)');
+
+  const shouldApplyDark = theme === 'dark' || (theme === 'os' && prefersDark);
+  const shouldApplyHC = themeOptions?.highContrast ?? prefersHighContrast;
+
+  // only provide context if wrapped around children
+  return hasChildren ? (
+    <ThemeContext.Provider value={{ theme, themeOptions }}>
+      <Element
```
mayank99, line 47: I made this polymorphic because a `<div>` might not always be desirable.

The user will now be able to do this for better SSR support:
```jsx
<ThemeProvider as='body'>
  <App />
</ThemeProvider>
```

gretanausedaite (commented):
I have no comments. Looks good.

packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -3,186 +3,265 @@
...
+    window.matchMedia = jest.fn().mockReturnValueOnce({ matches: true });
+
+    render(<ThemeProvider theme='os' />);
     expectDarkTheme();
+  });
 
-    changeOSTheme('light');
-    expectLightTheme();
+  // TODO: fix or remove this test
+  describe.skip('media query', () => {
```
mayank99, line 144: I cannot get these two tests to pass so I'm skipping them. I'm considering removing them completely.

Also, for reviewing changes in this file, I suggest "hiding whitespace changes" as most of the old tests are unchanged.

gretanausedaite (approved):
Code looks good. Skipped tests could be fixed or removed (or add backlog item to fix them). Now it just takes code lines without purpose.

adamhock (approved):
I agree with Greta's comment about the skipped unit test. Other than that, this look good. Nice work 🎉

mayank99:
Tests have been fixed 🎉

PR #827, opened 2022-09-14
fix(ButtonGroup): Handle overflow when available space is smaller than one element

### LINKED ISSUE #793: ButtonGroup: Overflow is broken when size is less than 1 element
If resizing to less than 1 element it is getting broken and never recovers.
In our overflow story I added left margin of 300px and get this:
oveflow

Need to check if its specific to `ButtonGroup` or our `useOverflow` and if that is the same case with other components we use `useOverflow` (i.e. `breadcrumbs`)

elephantcatdog:
image 
image

Fixed `ButtonGroup` overflow breaking when the container was too small. This was specific to `ButtonGroup` and was occurring when `splice` was called with `-1` when `visibleCount = 1`.

Closes #793 

## Checklist

- [ ] ~Add component features demo in Storybook (different stories)~
- [ ] ~Approve test images for new stories~


packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
```diff
@@ -79,6 +79,47 @@ it.each(['start', 'end'] as const)(
...
 
+it.each(['start', 'end'] as const)(
+  'should handle overflow when available space is smaller than one element (%s)',
+  (overflowPlacement) => {
+    const scrollWidthSpy = jest
+      .spyOn(HTMLElement.prototype, 'scrollWidth', 'get')
+      .mockReturnValue(200);
+    const offsetWidthSpy = jest
+      .spyOn(HTMLElement.prototype, 'offsetWidth', 'get')
+      .mockReturnValue(50);
```
rohan-kadkol, line 90: When trying to replicate the issue, I found that the `scrollWidth` and `offsetWidth` were `50` and `0` respectively.

image

I did this by adding ``console.log(`offsetWidth: ${containerRef.current.offsetWidth}`,`scrollWidth: ${containerRef.current.scrollWidth}`);`` below [line 72 of useOverflow.tsx](https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/utils/hooks/useOverflow.tsx#L72).

Maybe I misunderstood where these `scrollWidth` and `offsetWidth` values are being used. Are they being used in `useOverflow` for the `containerRef` (`iui-button-group`)? Or are these values being mocked for something else?

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
```diff
@@ -79,6 +79,47 @@ it.each(['start', 'end'] as const)(
...
+          {buttons}
+        </ButtonGroup>
+      );
+    };
+    const { container } = render(<OverflowGroup />);
+
+    expect(container.querySelector('.iui-button-group')).toBeTruthy();
+
+    const buttons = container.querySelectorAll('.iui-button');
+    expect(buttons).toHaveLength(1);
```
rohan-kadkol, line 116: Maybe we should also test a more complex case as mentioned in the issue? Something about the overflow not restoring after coming down to just one button (the `overflowButton`).

So maybe after this test, you could change the mock to simulate three buttons being displayed. Then you can verify if 3 buttons are displayed or 10 (9 (because `.slice(0, -1)` gives all elements except the last one) + 1 (overflow button)).

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
```diff
@@ -79,6 +79,47 @@ it.each(['start', 'end'] as const)(
...
+  (overflowPlacement) => {
+    const scrollWidthSpy = jest
+      .spyOn(HTMLElement.prototype, 'scrollWidth', 'get')
+      .mockReturnValue(200);
+    const offsetWidthSpy = jest
+      .spyOn(HTMLElement.prototype, 'offsetWidth', 'get')
+      .mockReturnValue(50);
+
+    const OverflowGroup = () => {
+      const buttons = [...Array(3)].map((_, index) => (
```
rohan-kadkol, line 93: Maybe we can directly use Array(3) instead of using the spread operator? The tests passed even after making this change.

elephantcatdog: So I copied this from the overflow test that was already there and didn't think about why it was that way. I did some research and testing though.

This stackoverflow question has some good explanations for the difference: https://stackoverflow.com/questions/66077917/creating-array-with-spread-operator
I particularly like the examples given in the top comment.

I also put some `console.logs` in the code to see the difference between using the spread operator vs not using it:
image
image
The results show that the spread operator is necessary for `.map` to work properly.

Additionally, the tests passed because only one button was created without the spread operator, and my test was only looking for one button.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: > So maybe after this test, you could change the mock to simulate three buttons being displayed. Then you can verify if 3 buttons are displayed or 10 (9 (because .slice(0, -1) gives all elements except the last one) + 1 (overflow button)).

I just realized, I was thinking there are 10 buttons since there were 10 on the storybook's Overflow example. So in your case, I would mean: 

*"So maybe after this test, you could change the mock to simulate **2 buttons** being displayed. Then you can verify if **2 buttons** are displayed or **3** (**2** (because .slice(0, -1) gives all elements except the last one) + 1 (overflow button))."*

elephantcatdog: I don't think it's necessary because the button group only doesn't restore itself after breaking, and my test checks that it doesn't break. Thoughts?

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: Ohh, I meant like `Array(3).map(...` and not `[Array(3)].map(...`. I think that might work because at least from what I know, `[...Array(3)]` just creates a duplicate (new object in memory with the same contents) of `Array(3)`. 

Here is an example
image

elephantcatdog: Hm, I tested it with just Array(3) and console.log shows a blank space instead of an array of objects. Also using `Array(3)` instead of `[...Array(3)]` in the existing test causes failures. So this doesn't work.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
elephantcatdog: I've been playing with it and I don't think it works at the same location in `useOverflow`, or `offsetWidth = 150` and `scrollWidth = 150` would show 3 button elements.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: Ohh, sorry, I don't think I properly saw the StackOverflow link you sent. After trying it out myself, l realize that `[...Array(3)]` is correct. `Array(3)` is not iterable, so we need `...[Array(3)]` since we need to iterate through it. I learned something new! :smile:

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: That makes sense. Also, I now think that adding it might make the test more complex than needed. Also, if we add it, future contributors might be confused about why we are testing such a specific case. So I too now feel that there's no need of adding it

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: Ohh, I see. I'm not too sure. Maybe I misundertood where those mocks for `scrollWidth` and `offsetWidth` are being used. So maybe they are used in some other place than the place where I put the comment. If that is the case, maybe your mock values are correct. @bentleyvk What do you think?

bentleyvk: `scrollWidth` and `offsetWidth` are used in `useOverflow` hook. `offsetWidth` is width of available space and `scrollWidth` is width of the space that is needed for all visible buttons to fit. So if `scrollWidth` is `200px` that means that each button's width is `200 / 3 = 66px` and `66 > 50` so no buttons fit here. Not sure if I answered your question.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -110,7 +110,7 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
             )}
 
           {visibleCount < items.length
-            ? items.slice(0, visibleCount - 1)
+            ? items.slice(0, visibleCount < 1 ? 0 : visibleCount - 1)
```
bentleyvk, line 113: No need for a change but just to note that you can also just write `Math.max(0, visibielCount - 1)`

bentleyvk (approved):

r100-stack (approved):

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
rohan-kadkol: Ohh, I see. I'm slightly confused. But wouldn't all the buttons be of width of `50px` (Since the `offsetWidth` for even the button children is `50px`)? I'm slightly confused why would the buttons change its width according to the available space (`scrollWidth of parent container / number of buttons`).

Also, I'm a bit confused that if no buttons would fit here, then why is the test on [line 123 of this PR](https://github.com/iTwin/iTwinUI-react/pull/827/files#diff-bda7aa7aeee0828d62bfdcba442079e287ab78c173fc5037015a8d3e6913cec5R123) expecting 1 button?

I might be totally wrong, so I guess I'll ask you these questions later to better understand how mocks work, in order to not hold up this PR.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.tsx
elephantcatdog: I was looking for that!!

elephantcatdog: I think this will be more clear

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
elephantcatdog: Me too!

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.test.tsx
elephantcatdog: Yeah, I'm still having trouble with this too.

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.tsx
elephantcatdog: Updated it, thanks!

packages/iTwinUI-react/src/core/ButtonGroup/ButtonGroup.tsx
```diff
@@ -110,7 +110,7 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
             )}
 
           {visibleCount < items.length
-            ? items.slice(0, visibleCount - 1)
+            ? items.slice(0, Math.max(0, visibleCount - 1))
```
rohan-kadkol, line 113: Yes, I think that is what Vilius meant. I also tested it and it worked.

elephantcatdog: Thanks!

PR #829, opened 2022-09-14
feat(NotificationMarker): Add `NotificationMarker` component

### LINKED ISSUE #348: Button: notification marker
Add notification marker to the button as we have CSS for it https://github.com/iTwin/iTwinUI/pull/293
image

r100-stack:
Closes #348

* Adds `NotificationMarker` component.
* `NotificationMarker` adds a small notification circle to the top-right of the passed `children` prop. This can be applied to pretty much anything but is mostly intended for icons within default / borderless buttons.
* Prerequisite: PR #895 

#### 1. Primary Intended Use-case

[vokoscreenNG-2022-09-20_10-21-41.webm](https://user-images.githubusercontent.com/45748283/191283495-36c80f60-42c0-4533-8f4c-17f2ac1a3349.webm)

```html
<IconButton styleType='borderless'>
  <NotificationMarker><SvgNotification /></NotificationMarker>
</IconButton>
```

#### 2. Example with non-icon children

[vokoscreenNG-2022-09-20_10-22-45.webm](https://user-images.githubusercontent.com/45748283/191283744-2e937987-4550-4cd6-8e65-0e7bfa0d1acd.webm)

```html
<NotificationMarker type='positive' pulsing={true}>Live</NotificationMarker>
<NotificationMarker type='negative' pulsing={true}>Rec</NotificationMarker>
<NotificationMarker type='primary' pulsing={true}>1 new message</NotificationMarker>
```

#### 3. Using `enabled` prop

[vokoscreenNG-2022-09-20_10-28-24.webm](https://user-images.githubusercontent.com/45748283/191285303-062aabfe-ee93-4791-9ad4-a3bee618ae39.webm)

```jsx
let [demoNotificationCount, ...] = useState(0);
...
<NotificationMarker enabled={demoNotificationCount > 0}>
  <SvgNotification />
</NotificationMarker>
```

## Checklist


packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+    className,
+    children,
+    type = 'primary',
+    urgent = false,
+    active = true,
+    ...rest
+  } = props;
+  useTheme();
+  return (
+    <div
```
mayank99, line 64: divs inside buttons is invalid markup. Does it still work with `<span>`?

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+      className={cx(
+        active
+          ? { [`iui-notification-${type}`]: true, 'iui-urgent': urgent }
+          : {},
```
mayank99, line 68:
```suggestion
          { [`iui-notification-${type}`]: active, 'iui-urgent': active && urgent }
```

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+   *   <SvgNotification />
+   * </NotificationMarker>
+   */
+  active?: boolean;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * A small notification circle to the top-right of the passed children prop. This can be applied to pretty much anything but mostly intended for icons within default / borderless buttons.
+ * @example
+ * // Primary Intended Use-case
```
mayank99, line 46: No need for this `//` comment

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+   * ...
+   * <NotificationMarker active={newMessagesCount > 0}>
+   *   <SvgNotification />
+   * </NotificationMarker>
+   */
+  active?: boolean;
+} & Omit<CommonProps, 'title'>;
+
+/**
+ * A small notification circle to the top-right of the passed children prop. This can be applied to pretty much anything but mostly intended for icons within default / borderless buttons.
```
mayank99, line 44: Line wrapping would help here

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+  urgent: true,
+};
+
+CustomChild.argTypes = {
+  type: {
+    control: false,
+  },
+};
+
+export const ProgrammaticDisplay: Story<
```
mayank99, line 81: I don't think this story is needed. Everything is "programmatic" if you think about it.

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Type of notification
+   *
+   * 'primary' = blue,
+   * 'positive' = green,
+   * 'warning' = orange,
+   * 'negative' = red,
+   * @default 'primary'
+   */
+  type?: 'primary' | 'positive' | 'warning' | 'negative';
```
mayank99, line 24: I think this prop should be called `status`

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+   * Set this programmatically to false when you just want to render the passed children without the notification
+   * @default true
+   * @example
+   * let [newMessagesCount, ...] = useState(0);
+   * ...
+   * <NotificationMarker active={newMessagesCount > 0}>
+   *   <SvgNotification />
+   * </NotificationMarker>
+   */
+  active?: boolean;
```
mayank99, line 40: `active` can be confused with selected state. Maybe `enabled` would be clearer

packages/iTwinUI-react/src/core/index.ts
```diff
@@ -311,3 +311,6 @@ export {
   MiddleTextTruncation,
 } from './utils';
 export type { ThemeType, MiddleTextTruncationProps } from './utils';
+
+export { NotificationMarker } from './NotificationMarker';
+export type { NotificationMarkerProps } from './NotificationMarker';
```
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Ohh, I forgot it, thanks

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+  children: <SvgNotification />,
+};
+
+Basic.argTypes = {
+  children: {
+    control: false,
+  },
+};
+
+export const CustomChild: Story<NotificationMarkerProps> = (args) => {
```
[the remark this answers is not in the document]
rohan-kadkol: I was trying to have a story to show that we can pass anything as a child instead of just SVGs like `<SvgNotification />`.

Maybe `CustomChild` is not an accurate name. What name do you recommend for such a story?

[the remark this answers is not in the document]
rohan-kadkol: I was trying to have a story to show that we can pass anything as a child instead of just SVGs like `<SvgNotification />`.

Maybe `CustomChild` is not an accurate name. What name do you recommend for such a story?

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Ohh, I think I got confused with the `type` from `Alert.tsx`. Changed it to `status` now

image

apps/storybook/src/NotificationMarker.stories.tsx
rohan-kadkol: I removed this `ProgrammaticDisplay` story. Do you think there should be documentation, a story, or something similar to show how to use the `enabled` (currently called `active`) prop? Because without mentioning it, people might do something like:

```jsx
{numNotifications > 0 ? <NotificationMarker>child</NotificationMarker> : child}
```

instead of:

```jsx
{<NotificationMarker enabled={numNotifications > 0}>child</NotificationMarker>}
```

Using the first way would lead to an extra dom `<div>` when numNotifications > 0. This might cause CSS styling problems (users' direct descendent styles with `>`) and also maybe longer untidy code

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Ohh, right, renamed it to `enabled`

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Split line 44 into two lines

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
mayank99: dont forget the jsdoc

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: To confirm, did you mean to remove line 46? Or just remove the "`//`" from line 46?

I thought you meant the former, so I removed line 46. But just confirming if that's what you meant.

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: To confirm, are you referring to what is mentioned in this [StackOverflow answer](https://stackoverflow.com/a/12982293/11547064)? This concept is new to me. I have a few questions:

* What is "markup" exactly?
* What is the meaning of "valid" markup?
* Why `<div>` inside `<button>` is invalid markup?
* How to know what is valid markup and what isn't? Do we always have to put it in a [validator tool](https://validator.w3.org/#validate_by_input+with_options) like the one mentioned in the answer? Or is there a website/source with rules for valid markup?

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Nice idea, thanks

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Oops, thanks, changed now

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,78 @@
...
+   * 'warning' = orange,
+   * 'negative' = red,
+   * @default 'primary'
+   */
+  type?: 'primary' | 'positive' | 'warning' | 'negative';
+  /**
+   * Set this to true for important notifications
+   * @default false
+   */
+  urgent?: boolean;
```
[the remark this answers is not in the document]
rohan-kadkol: Changed `urgent` to `pulsing`. I added a warning to the JSDocs too.

To confirm is there a better way to add the warning? Maybe something like an `@abc` tag that we can use for a warning, or use something similar?

image

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+  demoNotificationCount: 0,
+  urgent: true,
+  type: 'primary',
+};
+
+ProgrammaticDisplay.argTypes = {
+  active: {
+    control: false,
+  },
+};
```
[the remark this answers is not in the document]
rohan-kadkol: I added uses of `<NotificationMarker>` in `Header`'s Full story ([commit](https://github.com/iTwin/iTwinUI-react/pull/829/commits/3b029f0c1df338e65334384f13d79c8c7168d3b7)) and `SideNavigations`'s Basic story ([commit](https://github.com/iTwin/iTwinUI-react/pull/829/commits/bee281d467f8697e832763e203652837999daa8e)).

I did not add any props to modify the `NotificationMarker`s since it might be confusing whether `status` and `pulsing` are parameters of `Header`/`SideNavigation` or `NotificationMarker`.

To confirm, is the way I have added it alright?

apps/storybook/src/Header.stories.tsx
```diff
@@ -165,7 +166,9 @@ export const Full: Story<HeaderProps> = (args) => {
           onClick={() => action('Clicked on the notification bell')()}
           styleType='borderless'
         >
-          <SvgNotification />
+          <NotificationMarker status='positive' pulsing={true}>
+            <SvgNotification />
+          </NotificationMarker>
```
elephantcatdog, line 171: The test image that goes along with this story should be updated.

apps/storybook/src/NotificationMarker.stories.tsx
elephantcatdog: You have a good example on the property itself. Not sure how else to disseminate that info with the current documentation setup we have.

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: Maybe you could just put one text example into Basic? And replace the Custom example with a Pulsing example. Because in both Basic and Custom examples, the notification marker itself is basic.

apps/storybook/src/SideNavigation.stories.tsx
```diff
@@ -50,18 +51,39 @@ export const Basic: Story<SideNavigationProps> = (args) => {
...
       secondaryItems={[
-        <SidenavButton startIcon={<SvgSettings />} key={3}>
+        <SidenavButton
+          startIcon={
+            <NotificationMarker status='warning'>
+              <SvgSettings />
+            </NotificationMarker>
+          }
+          key={3}
+        >
```
elephantcatdog, line 86: The test image that goes along with this story needs to be updated.

elephantcatdog (commented):
Don't forget that the programmatic screenshot needs to be removed.

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: Don't forget to update the story test images for Header and SideNavigation.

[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Thanks, approved Header and SideNavigation images

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Ohh, nice idea, thanks. I changed `CustomChild` story to `Pulsing`.

`Basic` now shows one icon button and one text. `Pulsing` shows four texts (for each status) with pulsing notifications

apps/storybook/src/Header.stories.tsx
rohan-kadkol: **[DONE]**

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import {
+  IconButton,
+  NotificationMarker,
+  NotificationMarkerProps,
+} from '@itwin/itwinui-react';
+import SvgNotification from '@itwin/itwinui-icons-react/cjs/icons/Notification';
+
+export default {
```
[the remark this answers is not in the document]
rohan-kadkol: **[DONE]** Nice idea, thanks

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
mayank99: markup is just html. "valid" means it's not allowed and will not be properly announced by assistive technologies (e.g. screen readers). and yes you need to validate it through a tool if you're not sure.
https://validator.w3.org/docs/help.html#validation_basics

there is an html spec but you're not expected to read the whole thing all at once 😄 https://html.spec.whatwg.org/

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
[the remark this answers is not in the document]
mayank99: There is nothing in jsdoc, it just comes down to better wording.

apps/storybook/src/NotificationMarker.stories.tsx
mayank99: I don't think it's necessarily a big deal if users do a ternary to toggle the notification-marker, so i think it's fine if we don't even show an example tbh

rohan-kadkol:
> Are you planning to do the CSS changes (e.g. `mask`) before using it in React?

As of now, I am working on the masking approach on the side. So I think the masking change would come after this PR. 

After making the masking change in CSS, we might need to change the class names and dom structure slightly in react, which I don't think would be a big PR.

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
rohan-kadkol: **UPDATE**: I now created another story called `Status`. So now, `Basic` is same as before, `Pulsing` shows one icon and one bold text (similar to Basic), and `Status` shows all four statuses for icons and texts at once

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    </>
+  );
+};
+
+Pulsing.args = {
+  status: 'negative',
+  pulsing: true,
+};
+
+export const Status: Story<NotificationMarkerProps> = (args) => {
```
rohan-kadkol, line 89: Is this story showing too much information? Is there a better way? Or this is alright?

Basically I intended to display all four statuses for text and icons at once, similar to how we show all things at once for most components on the CSS repo's Github demo website.

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+  return (
+    <>
+      <IconButton styleType='borderless'>
+        <NotificationMarker status='primary' {...args}>
+          <SvgNotification />
+        </NotificationMarker>
+      </IconButton>
+
+      <NotificationMarker status='primary' {...args}>
+        <Text style={{ fontWeight: '700' }}>URGENT</Text>
```
rohan-kadkol, line 78: How to add bold text using our `<Text>` component? Is this the way?

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: Hm 🤔 I just thought about how there could be a problem with visual testing the pulse if one time it's 'on' and another time it's 'off'. Have you noticed a problem with that if you run that test a few times?

apps/storybook/src/NotificationMarker.stories.tsx
elephantcatdog: I like the story for the statuses. I think you don't need to show both text and icon. You could do something like this: 
image
That tells users also what the name of the color is.

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
rohan-kadkol: I thought of that but then realized that the CSS repo also does a similar form of image testing. The [header](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/backstop/results/bitmaps_reference/iTwinUI_dark_header_Type_All_-_narrow_screen_0_demo-all_0_.png) and [side-navigation](https://github.com/iTwin/iTwinUI/blob/main/packages/itwinui-css/backstop/results/bitmaps_reference/iTwinUI_dark_side-navigation_State_collapsed_hover_and_focus_0_demo-side-navigation_0_desktop.png) tests also have a NotificationMarker in the screenshot. Since it worked there with Backstop, I thought it might work here too with Cypress.

Additionally, I ran the tests a couple of times and they always passed. Also, there are multiple attempts for each test (around 4), so I think they will pass even if they fail in one of the attempts

@mayank99 What do you think?

apps/storybook/src/NotificationMarker.stories.tsx
rohan-kadkol: Thanks for the idea. I added the status names as the dummy names for the text notification examples for the `Status` story. 

However, I kept the icons examples in that story too since I felt it was worth showing both icons and texts. The user can then decide whether they want to use an icon, text, or something else.

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]** Changed div to span and it seems to do the same effect.

apps/storybook/src/NotificationMarker.stories.tsx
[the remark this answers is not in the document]
elephantcatdog: Good info, thanks!

mayank99:
> As of now, I am working on the masking approach on the side. So I think the masking change would come after this PR.
> 
> After making the masking change in CSS, we might need to change the class names and dom structure slightly in react, which I don't think would be a big PR.

If there is not a lot to do in the CSS PR, I'd suggest doing that first. There is no rush to release NotificationMarker so we can wait a bit longer.

apps/storybook/src/NotificationMarker.test.ts
```diff
@@ -0,0 +1,16 @@
...
+ *--------------------------------------------------------------------------------------------*/
+describe('NotificationMarker', () => {
+  const storyPath = 'Core/NotificationMarker';
+  const tests = ['Basic', 'Pulsing', 'Status'];
+
+  tests.forEach((testName) => {
+    it(testName, function () {
+      const id = Cypress.storyId(storyPath, testName);
+      cy.visit('iframe', { qs: { id } });
+      cy.compareSnapshot(testName);
```
rohan-kadkol, line 13: I could not fix the NotificationMarker test images. In the `NotificationMarker -> Status` story for example, the notification marker circle appears on the top right when viewing in Storybook.

image

But the test images show the circle on the top center for the `IconButton`.

image



I tried using @mayank99 's suggestion to use `cy.wait()` but that didn't work. Does anyone have any ideas on why this is happening and how to fix this?

mayank99: After spending some time on this, I was able to narrow down the issue to `IconButton` but couldn't really figure out the exact reason it gets misaligned like that. Could be a bug in Chrome 94 (which is what Cypress 9.6.0 comes with).

I was able to work around it by removing `IconButton` and instead adding the fill/size attributes on the svg (which will not be necessary after https://github.com/iTwin/iTwinUI/issues/851 and/or https://github.com/iTwin/iTwinUI-icons/issues/55). I've pasted all my local changes below, feel free to copy/change as you see fit. Hopefully this unblocks you for now.

<details>

<summary>NotificationMarker.stories.tsx</summary>

```tsx
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
import { Story, Meta } from '@storybook/react';
import React from 'react';
import {
  Text,
  NotificationMarker,
  NotificationMarkerProps,
import { SvgNotification } from '@itwin/itwinui-icons-react';

export default {
  component: NotificationMarker,
  args: {
    enabled: true,
    pulsing: false,
  },
  argTypes: {
    className: { control: { disable: true } },
    style: { control: { disable: true } },
    status: {
      control: 'radio',
      options: ['primary', 'positive', 'warning', 'negative'],
    },
  },
  title: 'Core/NotificationMarker',
  decorators: [
    (Story) => (
      <div
        style={{
          display: 'grid',
          placeItems: 'start',
          gap: '20px',
        }}
      >
        <Story />
      </div>
    ),
  ],
} as Meta<NotificationMarkerProps>;

export const Basic: Story<NotificationMarkerProps> = (args) => {
  return (
    <>
      <NotificationMarker {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>
      <NotificationMarker {...args}>Normal</NotificationMarker>
    </>
  );
};

Basic.args = {
  status: 'primary',
  children: <SvgNotification />,
};

Basic.argTypes = {
  children: {
    control: false,
  },
};

export const Pulsing: Story<NotificationMarkerProps> = (args) => {
  return (
    <>
      <NotificationMarker status='primary' {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>

      <NotificationMarker status='primary' {...args}>
        <Text style={{ fontWeight: '700' }}>URGENT</Text>
      </NotificationMarker>
    </>
  );
};

Pulsing.args = {
  status: 'negative',
  pulsing: true,
};

export const Status: Story<NotificationMarkerProps> = (args) => {
  return (
    <>
      <NotificationMarker status='primary' {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>
      <NotificationMarker status='positive' {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>
      <NotificationMarker status='warning' {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>
      <NotificationMarker status='negative' {...args}>
        <SvgNotification width={16} height={16} fill='currentColor' />
      </NotificationMarker>
      <NotificationMarker status='primary' {...args}>
        Primary
      </NotificationMarker>
      <NotificationMarker status='positive' {...args}>
        Positive
      </NotificationMarker>
      <NotificationMarker status='warning' {...args}>
        Warning
      </NotificationMarker>
      <NotificationMarker status='negative' {...args}>
        Negative
      </NotificationMarker>
    </>
  );
};

Status.argTypes = {
  status: {
    control: false,
  },
};
```
</details>

rohan-kadkol: Thanks. I ended up adding a story called `NotificationMarker -> Basic` that directly uses an SVG (e.g. `<SvgNotification ... />`).

The other three stories in `NotificationMarker` use `IconButton`/`Button` and so the visual tests still give the above-mentioned bug. Thus, these three stories are left out of the visual tests.

The visual test for `Button -> WithNotification` is excluded from visual tests for the same reason.

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,85 @@
...
+   * @default true
+   * @example
+   * let [newMessagesCount, ...] = useState(0);
+   * ...
+   * <NotificationMarker enabled={newMessagesCount > 0}>
+   *   <SvgNotification />
+   * </NotificationMarker>
+   */
+  enabled?: boolean;
+} & Omit<CommonProps, 'title'>;
```
mayank99, line 44: this should take all props 
```suggestion
} & React.ComponentProps<'span'>;
```

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * A small notification circle to the top-right of the passed children prop.
+ * This can be applied to almost anything but mostly intended for icons within default / borderless buttons.
+ * @example
+ * <IconButton styleType='borderless'>
+ *   <NotificationMarker><SvgNotification /></NotificationMarker>
+ * </IconButton>
+ * @example
+ * <NotificationMarker status='positive' pulsing={true}>Live</NotificationMarker>
+ */
+export const NotificationMarker = (props: NotificationMarkerProps) => {
```
mayank99, line 56: forward ref might help here if the user wants to access the dom element

```suggestion
export const NotificationMarker = React.forwardRef(props: NotificationMarkerProps, ref: React.Ref<HTMLSpanElement>) => {
```

(the suggestion will need formatting and closing brackets)

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,85 @@
...
+    ? {
+        'data-iui-variant': status,
+        'data-iui-urgent': pulsing,
+      }
+    : {};
+
+  return (
+    <span
+      className={cx({ 'iui-notification-marker': enabled }, className)}
+      {...notificationAttributes}
```
mayank99, line 77: why not apply the data attributes directly?
```suggestion
  return (
    <span
      className={cx({ 'iui-notification-marker': enabled }, className)}
      data-iui-variant={enabled && status}
      data-iui-urgent={enabled && pulsing)
```

apps/storybook/src/NotificationMarker.test.ts
rohan-kadkol: **[DONE]** Added comments in `NotificationMarker.test.ts` and `Button.test.ts` explaining why certain tests are left out

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+Basic.args = {
+  status: 'primary',
+};
+
+export const WithButtons: Story<NotificationMarkerProps> = (args) => {
+  return (
+    <>
+      <IconButton styleType='borderless' onClick={action('clicked')}>
```
mayank99, line 65: no need for click action imo. we want to show the notification marker, so the less code in the surrounding elements, the better

rohan-kadkol: Removed

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: I wanted to not include the data attributes when the `NotificationMarker` is disabled. To keep the same functionality and also reduce the lines of code, I set the data attributes to `null` inline when `enabled = false`

```tsx
data-iui-variant={enabled ? status : null}
data-iui-urgent={enabled ? pulsing : null}
```

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: Added `React.forwardRef`

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: **[DONE]**

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * A small notification circle to the top-right of the passed children prop.
+ * This can be applied to almost anything but mostly intended for icons within default / borderless buttons.
+ * @example
+ * <IconButton styleType='borderless'>
+ *   <NotificationMarker><SvgNotification /></NotificationMarker>
+ * </IconButton>
+ * @example
+ * <NotificationMarker status='positive' pulsing={true}>Live</NotificationMarker>
+ */
+// export const NotificationMarker = (props: NotificationMarkerProps) => {
```
mayank99, line 56: leftover

mayank99 (approved):

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
```diff
@@ -0,0 +1,84 @@
...
+      enabled = true,
+      ...rest
+    } = props;
+    useTheme();
+
+    return (
+      <span
+        ref={ref}
+        className={cx({ 'iui-notification-marker': enabled }, className)}
+        // Exclude data attributes when enabled = false
```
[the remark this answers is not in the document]
rohan-kadkol: Removed

packages/iTwinUI-react/src/core/NotificationMarker/NotificationMarker.tsx
rohan-kadkol: Thanks, removed

apps/storybook/src/Button.stories.tsx
```diff
@@ -87,6 +88,69 @@ WithIcon.args = {
...
+      </Button>
+    </div>
+  );
+};
+
+WithNotification.args = {
+  children: 'Inbox',
+};
+
+WithIcon.argTypes = {
```
elephantcatdog, line 150: I think this should be "WithNotification.argTypes"?

rohan-kadkol: **[DONE]**

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,146 @@
...
+        Positive
+      </NotificationMarker>
+      <NotificationMarker status='warning' {...args}>
+        Warning
+      </NotificationMarker>
+      <NotificationMarker status='negative' {...args}>
+        Negative
+      </NotificationMarker>
+    </>
+  );
```
elephantcatdog, line 139: Maybe add an example for 'white' status in here? It'd probably be helpful visually to go along with explanation 
image

apps/storybook/src/NotificationMarker.stories.tsx
```diff
@@ -0,0 +1,146 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import {
+  Text,
+  NotificationMarker,
+  NotificationMarkerProps,
```
elephantcatdog, line 10: Want to confirm that this issue is just on my end: 
image

apps/storybook/src/NotificationMarker.stories.tsx
elephantcatdog: In `ComboBox.stories.tsx` the fontWeight is set to 'bold', but it looks like it's the same thing (ie 'bold' == '700'), so I think it's fine as is. Doesn't seem to be another way to do it.
image

apps/storybook/src/NotificationMarker.stories.tsx
elephantcatdog: I found an example in `Button.stories.tsx` that has 'white' status. Even though it's an example there, I think it'd still be relevant here if you want to just copy it over or do something similar.

rohan-kadkol: I added examples for `styleType = white` in `NotificationMarker -> Status` story. I also changed the grid layout to make it more intuitive.

apps/storybook/src/NotificationMarker.stories.tsx
rohan-kadkol: This is not happening to me, but it might be alright. I too get similar errors sometimes but it occasionally goes away after restarting VSCode or re-running `yarn install`. Mayank also suggested to try running the `TypeScript: Restart TS Server` VSCode command to see if it helps, since Storybook is usually unreliable.

But for now, I think this error can be ignored.

apps/storybook/src/NotificationMarker.stories.tsx
elephantcatdog: Looks good; I like it!

elephantcatdog (approved):

Issue #794, opened 2022-08-22
MultiSelect: Screen reader does not read the selected value


veekeys:
https://github.com/iTwin/iTwinUI-react/pull/700#discussion_r951662931

PR #830, opened 2022-09-16
feat(Combobox): Add multi-select option

### LINKED ISSUE #650: Support multi select for combobox
#### Feature
Be able to select multiple values from a combobox and have them appear as tags which can be removed individually.

#### Examples
image
image

gretanausedaite:
Closes: iTwin/iTwinUI-react#650 

Added multiselect option to ComboBox.
image

New props:
- `multiple` - sets ComboBox to multiselect;
- ComboboxMultipleTypeProps - to manage types for multiple and single select;

Added 'multiselect' reducer action to handle array selection.

New components:
- `ComboBoxMultipleContainer`
- `SelectTagContainer`

Changed ComboBoxMenuItem role to `option` for better a11y (screen-readers).


packages/iTwinUI-react/src/core/ComboBox/ComboBoxInputContainer.tsx
```diff
@@ -30,7 +30,9 @@ export const ComboBoxInputContainer = (props: ComboBoxInputContainerProps) => {
       {...rest}
       id={id}
     >
-      <div className='iui-input-with-icon'>{children}</div>
+      <div className='iui-input-with-icon' style={{ position: 'relative' }}>
```
gretanausedaite, line 33: Add modifier to `input-with-icon` to change position to relative.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -363,18 +515,37 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
           <ComboBoxInputContainer disabled={inputProps?.disabled} {...rest}>
-            <ComboBoxInput
-              value={inputValue}
-              {...inputProps}
-              onChange={handleOnInput}
-            />
+            <>
+              <ComboBoxInput
+                style={{
+                  paddingLeft: tagContainerWidth + 18,
```
gretanausedaite, line 529: Add dynamic padding change and dynamic height change (for custom rendering if users don't handle overflow)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -394,4 +565,56 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+            right: 'unset',
+          }}
+        >
+          {selectedItemsRenderer(selectedItems)}
+        </div>
+      )}
+      {selectedItems && !selectedItemsRenderer && (
+        <SelectTagContainer
+          style={{
+            maxWidth: `70%`,
```
gretanausedaite, line 609: Add maxWidth and right: unset to tag container

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -394,4 +565,56 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  }, [selectedItems, tagRenderer]);
+
+  return (
+    <>
+      {selectedItems && selectedItemsRenderer && (
+        <div
+          className='iui-select-tag-container'
+          ref={mRef}
+          style={{
+            maxWidth: `70%`,
```
gretanausedaite, line 596: For custom implementation change container height.

gretanausedaite (commented):
@mayank99 @bentleyvk 
I would like to get some comments on multi-select combobox implementation. I wanted to add those changes in React first to see if it actually works. It's not finished and has some issues. Just wanted to know if I'm on the right track.
I've added some comments in places I am planning to make css changes.

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
```diff
@@ -5,11 +5,19 @@
...
-  [type, value]: [ComboBoxAction] | [ComboBoxAction, number | undefined],
+  state: {
+    isOpen: boolean;
+    selectedIndex: number | number[];
+    focusedIndex: number;
+  },
+  [type, value, action]:
+    | [ComboBoxAction]
+    | [ComboBoxAction, number | undefined]
+    | [ComboBoxAction, number | undefined, SelectedAction | undefined],
```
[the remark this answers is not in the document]
mayank99: i think doing this might also avoid the need to cast to `number[]` below

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
```diff
@@ -25,6 +33,30 @@ export const comboBoxReducer = (
...
+      }
+      if (action === 'added') {
+        console.log('added', value);
+        console.log([...(state.selectedIndex as number[]), value]);
+        return {
+          ...state,
+          focusedIndex: value ?? state.focusedIndex,
+          selectedIndex:
+            value !== null && value !== undefined
+              ? [...(state.selectedIndex as number[]), value].sort()
```
mayank99, line 55: the default `sort` function (without an argument) has quite unexpected results. Are you sure this is what you want?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,6 +32,54 @@ import { ComboBoxInput } from './ComboBoxInput';
...
+const isSingleOnChange = <T,>(
+  onChange:
+    | (((value: T) => void) | undefined)
+    | (((value: T, event: SelectValueChangeEvent) => void) | undefined),
+  multiple: boolean,
+): onChange is ((value: T) => void) | undefined => {
+  return !multiple;
+};
+
+export type ComboboxMultipleTypeProps<T> =
```
mayank99, line 55: this doesn't need to be exported

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -48,7 +98,7 @@ export type ComboBoxProps<T> = {
   /**
    * Callback fired when selected value changes.
    */
-  onChange?: (value: T) => void;
+  // onChange?: (value: T) => void;
```
mayank99, line 101: Remove commented out props?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,18 +233,45 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     });
   }
 
+  const getSelectedIndexes = () => {
+    const indexArray: number[] = [];
+    isMultipleEnabled(valueProp, multiple) &&
```
mayank99, line 238: why not just an `if` condition? 😄

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,18 +233,45 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
   const [{ isOpen, selectedIndex, focusedIndex }, dispatch] = React.useReducer(
     comboBoxReducer,
     {
       isOpen: false,
-      selectedIndex: valueProp
+      selectedIndex: isMultipleEnabled(valueProp, multiple)
+        ? valueProp
+          ? getSelectedIndexes()
+          : []
+        : valueProp
```
mayank99, line 256: Would be good to explain/simplify what is going on here. I think we could avoid a nested ternary

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,6 +32,54 @@ import { ComboBoxInput } from './ComboBoxInput';
 import { ComboBoxInputContainer } from './ComboBoxInputContainer';
 import { ComboBoxMenu } from './ComboBoxMenu';
 import { ComboBoxMenuItem } from './ComboBoxMenuItem';
+import SelectTagContainer from '../Select/SelectTagContainer';
+import SelectTag from '../Select/SelectTag';
+
+const isMultipleEnabled = <T,>(
+  variable: (T | undefined) | (T[] | undefined),
```
mayank99, line 39: what is "variable"?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -394,4 +565,56 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
 };
 
+type MultipleComboboxProps<T> = {
+  selectedItems?: SelectOption<T>[];
+  selectedItemsRenderer?: (options: SelectOption<T>[]) => JSX.Element;
+  tagRenderer: (item: SelectOption<T>) => JSX.Element;
+  mRef: React.ForwardedRef<HTMLDivElement>;
+};
+
+const MultipleSelectedContainer = <T,>({
```
mayank99, line 575: this should go in a separate file. I think that would also help clean up the code in main ComboBox component, and more things would be handled through Context

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -394,4 +565,56 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  selectedItems,
+  selectedItemsRenderer,
+  tagRenderer,
+  mRef,
```
mayank99, line 579: what is "mRef"?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,6 +32,54 @@ import { ComboBoxInput } from './ComboBoxInput';
...
+      /**
+       * Enable multiple selection.
+       * @default false
+       */
+      multiple?: false;
+      /**
+       * Custom renderer for the selected item in select.
+       * If `multiple` is enabled, it will give array of options to render.
+       */
+      selectedItemRenderer?: (option: SelectOption<T>) => JSX.Element;
```
mayank99, line 66: I don't think we should allow custom renderer for initial release tbh. Makes it way more complex in every way

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -363,18 +515,37 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+                style={{
+                  paddingLeft: tagContainerWidth + 18,
+                  minHeight: tagContainerHeight,
+                }}
+                value={inputValue}
+                {...inputProps}
+                onChange={handleOnInput}
+              />
+              {isMultipleEnabled(selectedIndex, multiple) && (
+                <MultipleSelectedContainer
```
mayank99, line 537: Have you researched the accessibility concerns that come with this markup? The multi-select implementation was already inaccessible (https://github.com/iTwin/iTwinUI-react/issues/794) and combobox, being more complex, requires even more thought to make accessible.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,18 +233,45 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
     },
   );
 
+  const selectedOptions = React.useMemo(() => {
+    if (selectedIndex == null) {
+      return undefined;
+    }
+    const item = isMultipleEnabled(selectedIndex, multiple)
+      ? selectedIndex != undefined && selectedIndex.length > 0
+        ? selectedIndex.map((index) => optionsRef.current[index])
```
mayank99, line 269: Again hard to understand what is going on here with the nested ternaries. Could you explain/simplify?

apps/storybook/src/ComboBox.stories.tsx
```diff
@@ -27,7 +27,8 @@ export default {
   },
   decorators: [
     (Story) => (
-      <div style={{ maxWidth: 'clamp(300px, 50%, 100%)', minHeight: 380 }}>
+      <div>
+        {/* <div style={{ maxWidth: 'clamp(300px, 50%, 100%)', minHeight: 380 }}> */}
```
mayank99, line 31: why commented out?

mayank99 (commented):
From initial look, I'm finding it hard to follow the flow. It might help to explain the logic through text or video recording.

But I have a feeling we might have better luck moving the reducer/provider into the main combobox component so that we can handle more things through `dispatch`ing events, rather than relying on `useEffect`s to perform the updates based on reducer state.

That is my main feedback, but I left a few minor comments as well. I will probably have more useful feedback once I better understand the code.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: should we create a css file to collect all the necessary changes? it's getting hard to keep track when everything is inline style objects scattered across different components

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: should we create a css file to collect all the necessary changes? it's getting hard to keep track when everything is inline style objects scattered across different components

apps/storybook/src/ComboBox.stories.tsx
```diff
@@ -575,3 +576,36 @@ Virtualized.args = {
 Virtualized.argTypes = {
   value: { control: { disable: true } },
 };
+
```
bentleyvk, line 579: After deselecting all items, I get this error:
image

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
bentleyvk: Second ternary is not needed because if `valueProp` is undefined, `getSelectedIndexes` will return empty array.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -276,13 +359,49 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
       return;
     }
-    const currentValue = optionsRef.current[selectedIndex]?.value;
+    if (isMultipleEnabled(selectedIndex, multiple)) {
+      const currentOptions = selectedOptions as SelectOption<T>[];
+      if (
+        !isSingleOnChange(onChangeProp.current, multiple) &&
+        currentOptions.length > 0
+      ) {
+        currentOptions.map((option) =>
```
bentleyvk, line 368: This part is not nice, since it will trigger `onChange` for every already selected value, no matter that only one item was selected.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -301,8 +420,17 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
       return customItem ? (
         React.cloneElement(customItem, {
           onClick: (e: unknown) => {
```
bentleyvk, line 422: This part could be moved to a variable and reused.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -394,4 +565,56 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+    }
+
+    return selectedItems.map((item) => tagRenderer(item));
+  }, [selectedItems, tagRenderer]);
+
+  return (
+    <>
+      {selectedItems && selectedItemsRenderer && (
+        <div
+          className='iui-select-tag-container'
```
bentleyvk, line 593: Why here you used div with tag container class and bellow you used SelectTagContainer component?

packages/iTwinUI-react/src/core/utils/hooks/useContainerHeight.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * @private
+ * @param watchResizes If false, ResizeObserver will not be connected and only the initial height will be returned
+ * @returns [ref to attach to the element, stateful height of the element]
+ *
+ * @example
+ * const [ref, height] = useContainerHeight();
+ * ... // do something with height
+ * return <div ref={ref}>...</div>;
+ */
+export const useContainerHeight = <T extends HTMLElement>(
```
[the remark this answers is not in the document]
gretanausedaite: I don't think we can rename the hook. It's exported and if developers might be using it.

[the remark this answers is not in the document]
mayank99: It's marked as private so I think it's fine to do it

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Added onClickHandler

apps/storybook/src/ComboBox.stories.tsx
gretanausedaite: I think I fixed that. I was returning undefined instead of empty array.

apps/storybook/src/ComboBox.stories.tsx
gretanausedaite: Testing purposes. Undid this change

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Added some comments

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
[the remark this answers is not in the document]
gretanausedaite: Removed type and now checking if selected item already exist in array - remove, if not - add.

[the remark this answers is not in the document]
gretanausedaite: Not sure about casting `number[]`. I think I might need that.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: I wanted to make things more complex 😃 
Added if condition.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed second ternary

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: It is value we are testing. Eg. valueProp or selectedIndex can be both T and T[]. So this way we can check types of these variables. (copied from multiselect)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Did not found a way to add forwardRef with <T,> type. Found that I can add custom ref props, so added mRef.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Should we give users back single changed value or full array? 
`onChange => ((value: T, event: SelectValueChangeEvent) => void)` (multiselect does this)
or
`onChange => ((values: T[]) => void)`

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: moved everything under select tag container

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: removed export for type

packages/iTwinUI-react/src/core/Select/SelectTagContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const SelectTagContainer = React.forwardRef(
+  (props: SelectTagContainerProps, ref: React.RefObject<HTMLDivElement>) => {
+    const { tags, className, ...rest } = props;
+
+    useTheme();
+    const [containerRef, visibleCount] = useOverflow(tags);
+    const refs = useMergedRefs(ref, containerRef);
+
+    return (
+      <span className='iui-content'>
```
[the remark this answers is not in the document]
gretanausedaite: Probably yes, removed and moved back to select

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
bentleyvk: In `Select` in calls `onChange` only one time with the changed item, whether it was added or removed. And now when you select or deselect an item, `onChange` is being called on each selected item, not just the newest one.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,18 +221,46 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  };
+
   // Reducer where all the component-wide state is stored
   const [{ isOpen, selectedIndex, focusedIndex }, dispatch] = React.useReducer(
     comboBoxReducer,
     {
       isOpen: false,
-      selectedIndex: valueProp
+      // if multiple enabled -> get selected indexes, if not -> is valueProp defined -> yes - get index of selected value -> no - set index as -1
+      selectedIndex: isMultipleEnabled(valueProp, multiple)
```
bentleyvk, line 243: Maybe instead of this ternary with comment, we can move everything to `getSelectedIndexes` that would have normal if else.

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMultipleContainer.tsx
```diff
@@ -0,0 +1,48 @@
...
+    return (
+      <>
+        {selectedItems && (
+          <SelectTagContainer
+            style={{
+              maxWidth: `70%`,
+              right: 'unset',
+            }}
+            ref={ref}
+            tags={selectedItemsElements}
```
mayank99, line 40: Do you really need this special selectedItemsElements and tagRenderer? I think you could combine both of those here like:

```js
<SelectTagContainer
  tags={selectedItems.map(({ label }) => <SelectTag key={label} label={label} />)}
>
```

Together with assigning a default value of `[]` to the `selectedItems` prop

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
```diff
@@ -63,6 +35,15 @@ export const ComboBoxStateContext = React.createContext<
...
-  ((x: [ComboBoxAction] | [ComboBoxAction, number]) => void) | undefined
+  | ((
+      x:
+        | { type: 'multiselect'; value: number }
+        | { type: 'multi-override'; value: number[] }
+        | { type: 'open'; value: undefined }
+        | { type: 'close'; value: undefined }
+        | { type: 'select'; value: number }
+        | { type: 'focus'; value: number | undefined },
+    ) => void)
```
mayank99, line 46: It might not make sense to have these types in this file anymore if the reducer's types are defined in ComboBox.tsx. I wonder if it's possible to make a generic that gets inferred by the `value` we pass to the provider.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -363,18 +503,36 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
             id,
             minWidth,
             isOpen,
-            focusedIndex,
+            focusedIndex: focusedIndex,
```
mayank99, line 506: this can be reverted

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      });
+    }
+    return indexArray;
+  }, [multiple, options, valueProp]);
+
   // Reducer where all the component-wide state is stored
+  const comboBoxReducer = (
+    state: {
+      isOpen: boolean;
+      selectedIndex: number | number[];
```
mayank99, line 226: `selectedIndex` implies that it's a number, not an array. Maybe rename to `selected`.

Although, it might also be better to have separate properties `selectedIndex` and `selectedIndices` (both optional) and use them depending on whether `multiple` is false/true.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+          onChangeProp.current?.(optionsRef.current[value]?.value);
+        }
+        return {
+          ...state,
+          selectedIndex: value ?? state.selectedIndex,
+          focusedIndex: value ?? state.focusedIndex,
+        };
+      }
+      case 'multi-override': {
+        return { ...state, selectedIndex: value };
```
mayank99, line 258: why do you need `multi-override` separately from `multiselect`?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      }
+      case 'multiselect': {
+        const actionIndex = (state.selectedIndex as number[]).findIndex(
```
mayank99, line 261: I don't think we should be type casting to `as number[]` everywhere without actually checking the type, because then we are *lying* to both typescript and ourselves.

A better strategy would be to check using `Array.isArray` and early return/throw if that's false.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+          (item) => item === value,
+        );
+        if (actionIndex >= 0) {
+          onChangeProp.current?.(optionsRef.current[value]?.value, 'removed');
+          return {
+            ...state,
+            focusedIndex: value ?? state.focusedIndex,
```
mayank99, line 268: Looks like this part is not working. It is resetting the focused index, both with keyboard and mouse selection.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      selectedIndex: number | number[];
+      focusedIndex: number;
+    },
+    {
+      type,
+      value,
+    }:
+      | { type: 'multiselect'; value: number }
+      | { type: 'multi-override'; value: number[] }
+      | { type: 'open'; value: undefined }
```
mayank99, line 235: If you remove `value` from these types, you won't need to pass `undefined` to the dispatch.

After that you won't be able to destructure `{ type, value }` above, so you would need to keep the non-destructured version (call it `action` for example) and then access `type` and `value` from inside it (e.g. `type` will become `action.type`).

apps/storybook/src/ComboBox.stories.tsx
```diff
@@ -575,3 +577,30 @@ Virtualized.args = {
...
+              ? prev.filter((item) => item !== value)
+              : [...prev, value],
+          );
+        }}
+        multiple={true}
+        value={selectedOptions}
+        {...args}
+      />
+      <Button onClick={() => setSelectedOptions(['AL', 'US'])}>
+        change selected
```
mayank99, line 602: what's the purpose of this button?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,25 +30,54 @@ import { ComboBoxInput } from './ComboBoxInput';
...
    * Controlled value of ComboBox.
+   * If `multiple` is enabled, it is an array of values.
+   */
+  value?: T | T[];
+  /**
+   * Callback fired when selected value changes.
+   */
+  onChange?:
+    | ((value: T) => void)
+    | ((value: T, event: SelectValueChangeEvent) => void);
```
mayank99, line 68: so we don't have stricter conditional typing anymore? what problem were you running into?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -263,26 +379,42 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      dispatch({ type: 'multi-override', value: indexes });
+    } else {
+      dispatch({
+        type: 'select',
+        value: options.findIndex((option) => option.value === valueProp),
+      });
     }
-    const currentValue = optionsRef.current[selectedIndex]?.value;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [valueProp, options]);
```
mayank99, line 394: why did you disable eslint here? putting `multiple` in this deps array makes sense to me

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -263,26 +379,42 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-    ]);
-  }, [options, valueProp]);
-
-  // Call user-defined onChange when the value actually changes
-  React.useEffect(() => {
-    // Prevent user-defined onChange to be called on mount
-    if (!mounted.current) {
-      mounted.current = true;
-      return;
+    if (isMultipleEnabled(valueProp, multiple) && !!valueProp) {
```
mayank99, line 382: I think this useEffect might need an early return if the `valueProp` is the same as what is currently in the state. We were doing something similar previously.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +208,99 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      }
+      case 'multi-override': {
+        return { ...state, selectedIndex: value };
+      }
+      case 'multiselect': {
+        const actionIndex = (state.selectedIndex as number[]).findIndex(
+          (item) => item === value,
+        );
+        if (actionIndex >= 0) {
+          onChangeProp.current?.(optionsRef.current[value]?.value, 'removed');
```
mayank99, line 265: Did you run into any problem by putting onChange here?

apps/storybook/src/ComboBox.stories.tsx
gretanausedaite: To test if combobox works when user changes value. Will be removed for final product.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: To not have `number | number[]` type in reducer action.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed eslint line

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Agree. It will be interesting to add check if two arrays are the same. Do we care about order of items in array?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: I was getting error in story that `type multiple: true is not compatible with false | undefined` (or something). I could add conditional typing back to try fix that issue some other way.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Renamed to selected. Not sure about having two properties.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -353,6 +491,8 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     [emptyStateMessage],
   );
 
+  const [tagContainerWidthRef, tagContainerWidth] = useContainerWidth();
```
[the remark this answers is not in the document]
gretanausedaite: This is used in Combobox input component inline styles. Should I move tagContainerWidth into context?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -363,18 +503,36 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-              value={inputValue}
-              {...inputProps}
-              onChange={handleOnInput}
-            />
+            <>
+              <ComboBoxInput
+                style={
+                  isMultipleEnabled(selectedIndex, multiple)
+                    ? {
+                        paddingLeft: tagContainerWidth + 18,
```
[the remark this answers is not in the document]
gretanausedaite: Should I move tagContainerWidth into context?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Focusing with selected as array is difficult. Which value we should focus? Last selected?

elephantcatdog (commented):
Is there a way to delete a selection without unselecting it in the dropdown list? I think it's common to be able to backspace/delete or have little [x] buttons on the tags to facilitate that.

Examples: [Vue](https://vue-multiselect.js.org/#:~:text=.%0A%20%20%20%20%20%20%7B%7B%20value%20%20%7D%7D-,Multiple%20select,-To%20allow%20multiple), [Mantine](https://mantine.dev/core/multi-select/)

gretanausedaite:
> Is there a way to delete a selection without unselecting it in the dropdown list? I think it's common to be able to backspace/delete or have little [x] buttons on the tags to facilitate that.
> 
> Examples: [Vue](https://vue-multiselect.js.org/#:~:text=.%0A%20%20%20%20%20%20%7B%7B%20value%20%20%7D%7D-,Multiple%20select,-To%20allow%20multiple), [Mantine](https://mantine.dev/core/multi-select/)

Current approach is same as in [Multiple Select](https://github.com/iTwin/iTwinUI-react/pull/700#pullrequestreview-1080855368). The `x` was removed because of a11y issues. 
We could add that `delete` button action, `backspace` can be tricky because we have input and it could cause problems if users accidentally delete more than they wanted.

gretanausedaite:
voiceOver on combobox multiselect: 

https://user-images.githubusercontent.com/83585998/194859910-c1663947-627b-43b4-9490-b1b4970ccbbe.mp4

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -242,47 +287,81 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      } else {
+        onChangeProp.current?.(
+          optionsRef.current[__originalIndex]?.value,
+          isMenuItemSelected(__originalIndex) ? 'removed' : 'added',
+        );
+      }
+    },
+    [isMenuItemSelected, multiple, onChangeProp, optionsRef],
+  );
+  console.log(focusedIndex);
```
elephantcatdog, line 352: console.log should be removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,25 +31,54 @@ import { ComboBoxInput } from './ComboBoxInput';
...
+
+// Type guard for multiple did not work
+const isSingleOnChange = <T,>(
+  onChange:
+    | (((value: T) => void) | undefined)
+    | (((value: T, event: SelectValueChangeEvent) => void) | undefined),
+  multiple: boolean,
+): onChange is ((value: T) => void) | undefined => {
+  return !multiple;
+};
```
elephantcatdog, line 51: Can you please explain why you have these functions `isMultipleEnabled` and `isSingleOnChange`? I see the comment that type guard didn't work for `isSingleOnChange`, but I don't understand why you can't use `if (multiple) {...}` and are using `if (isMultipleEnabled(value, multiple) {...}` and `if (isSingleOnChange(onChange, multiple) {...}`.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -242,47 +287,81 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-        dispatch(['focus', -1]);
+        dispatch({ type: 'focus', value: -1 });
       }
       inputProps?.onChange?.(event);
     },
     [filterFunction, focusedIndex, inputProps, optionsRef],
   );
 
-  // When the value prop changes, update the selectedIndex
+  // When the value prop changes, update the selected
```
elephantcatdog, line 310: This would be a bit more clear if you changed the comment to "When the value prop changes, update selected" or "When the value prop changes, update the selected index/indices"

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -196,21 +236,26 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     // When the dropdown opens
     if (isOpen) {
       inputRef.current?.focus(); // Focus the input
-      setFilteredOptions(optionsRef.current); // Reset the filtered list
-      dispatch(['focus']);
+      // Reset the filtered list (does not reset when multiple enabled)
+      !multiple && setFilteredOptions(optionsRef.current);
```
elephantcatdog, line 240: Just want to confirm that `isSingleOnChange` or `!isMultipleEnabled` is not needed here

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -196,21 +236,26 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+      if (!isMultipleEnabled(selected, multiple)) {
+        dispatch({ type: 'focus' });
+      }
```
elephantcatdog, line 243: Seems like you could do
`if (!isMultipleEnabled(selected, multiple)) {`
        `dispatch({ type: 'focus' });`
      `} else { setFilteredOptions(optionsRef.current) };`
instead of needing the first line `!multiple && setFilteredOptions(optionsRef.current);`?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Updated comment

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Removed console.log

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: in function
```ts
const isMultipleEnabled = <T,>(
  variable: (T | undefined) | (T[] | undefined),
  multiple: boolean,
): variable is T[] | undefined => {
  return multiple;
};
```
middle part `variable is T[] | undefined` "checks" if variable provided is specific type. 
Eg. `selected` can be number or number[]. If I want to do something only when `multiple` is enabled, I need `selected` to be `number[]`.
So code `if (isMultipleEnabled(selected, multiple)) { doSomethingWith(selected) }` checks if multiple is enabled and also checks if selected type is what I need. I don't need casts `selected as number[]`.

Not sure if make sense. If not, feel free to ping me.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: setFilteredOptions does not use values that changes types with multiple enabled. Therefore, isMultipleEnabled is not required.

gretanausedaite: Edit: changed code based on comment on line 243.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Both actions are performed when multiple is NOT enabled. Moved both under one if condition.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
elephantcatdog: Ahhh, that makes sense. Thanks for explaining!

elephantcatdog (approved):
Looks good to me!

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
mayank99: yeah that's one option.

Alternatively, put `tagContainerRef` in context, and move `useContainerWidth` inside ComboBoxInput because that's the only place it is used

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -196,21 +236,26 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
-      dispatch(['focus']);
-      // Reset the input value
-      setInputValue(
-        selectedIndex != undefined && selectedIndex >= 0
-          ? optionsRef.current[selectedIndex]?.label
-          : '',
-      );
+      dispatch({ type: 'focus' });
+      // Reset the input value if not multiple
+      if (!isMultipleEnabled(selected, multiple)) {
```
mayank99, line 250: Similarly, this can use `Array.isArray`

```suggestion
if (!multiple && !Array.isArray(selected)) {
```

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -30,25 +31,54 @@ import { ComboBoxInput } from './ComboBoxInput';
...
    * Controlled value of ComboBox.
+   * If `multiple` is enabled, it is an array of values.
+   */
+  value?: T | T[];
+  /**
+   * Callback fired when selected value changes.
    */
-  value?: T;
+  onChange?:
+    | ((value: T) => void)
```
mayank99, line 68: What if get rid of this overload and just make `event` optional?

e.g. `onChange?: (value: T, event?: SelectValueChangeEvent) => void`


That is effectively what the type is currently anyway (for the user), so simplifying this would just make it easier for us to use inside the component and we could get rid of the `isSingleOnChange` type guard.

mayank99 (commented):
A few minor comments about the type guards as I think they can be removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: I think order shouldn't matter, but the user should not change the order anyway, so it's also fine if we execute this when the order is different.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
mayank99: Last selected makes sense to me

apps/storybook/src/ComboBox.stories.tsx
```diff
@@ -575,3 +575,26 @@ Virtualized.args = {
...
+  return (
+    <ComboBox
+      options={options}
+      inputProps={{ placeholder: 'Select a country' }}
+      onChange={(value: string, event: string) => {
+        action(value + ' ' + event ?? '')();
+        setSelectedOptions((prev) =>
+          event === 'removed'
+            ? prev.filter((item) => item !== value)
+            : [...prev, value],
```
mayank99, line 592: I just realized it's not very nice that the user has to do all this logic in `onChange`. Should we handle it for them and provide the whole array as an argument?

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
```diff
@@ -5,56 +5,117 @@
...
 export const comboBoxReducer = (
-  state: { isOpen: boolean; selectedIndex: number; focusedIndex: number },
-  [type, value]: [ComboBoxAction] | [ComboBoxAction, number | undefined],
+  state: {
+    isOpen: boolean;
+    selected: number | number[];
+    focusedIndex: number;
+  },
+  action:
+    | { type: 'multiselect'; value: number }
```
mayank99, line 28: might want to save this type because it's reused down below in line 127. Could create `ComboBoxAction` similar to before with this new value

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -128,28 +136,41 @@ export const ComboBoxInput = React.forwardRef(
...
             if (isOpen) {
-              dispatch(['select', focusedIndexRef.current]);
-              dispatch(['close']);
+              if (multiple) {
+                if (focusedIndexRef.current > -1) {
+                  dispatch({
+                    type: 'multiselect',
+                    value: focusedIndexRef.current,
+                  });
+                  onChangeHandler && onChangeHandler(focusedIndexRef.current);
```
mayank99, line 153: nit: why not use `?.`?

i.e. `onChangeHandler?.(focusedIndexRef.current)`

same thing below in line 159

mayank99 (commented):
one potentially important question about API (`onChange`)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +209,25 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     });
   }
 
+  // Get indices of selected elements in options array when we have selected values.
+  const getSelectedIndexes = React.useCallback(() => {
+    const indexArray: number[] = [];
+    if (isMultipleEnabled(valueProp, multiple) && !!valueProp) {
```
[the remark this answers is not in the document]
gretanausedaite: Added

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -180,14 +209,25 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
...
+  const [{ isOpen, selected, focusedIndex }, dispatch] = React.useReducer(
     comboBoxReducer,
     {
       isOpen: false,
-      selectedIndex: valueProp
-        ? optionsRef.current.findIndex((option) => option.value === valueProp)
-        : -1,
+      selected: isMultipleEnabled(valueProp, multiple)
+        ? getSelectedIndexes()
+        : optionsRef.current.findIndex((option) => option.value === valueProp),
```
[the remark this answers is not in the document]
gretanausedaite: Moved to function. Why rename it?

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -196,21 +236,26 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     // When the dropdown opens
     if (isOpen) {
       inputRef.current?.focus(); // Focus the input
-      setFilteredOptions(optionsRef.current); // Reset the filtered list
-      dispatch(['focus']);
+      // Reset the filtered list (does not reset when multiple enabled)
+      if (!isMultipleEnabled(selected, multiple)) {
```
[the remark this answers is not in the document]
gretanausedaite: removed

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Added

packages/iTwinUI-react/src/core/ComboBox/helpers.ts
gretanausedaite: Added ComboBoxAction

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
gretanausedaite: Added this magic.

mayank99:
Visual tests are failing. I assume they will work after #962. But unit tests are also failing.

mayank99:
@gretanausedaite Can you leave a comment with the current status of this PR? I see you updated tests but I was going through my old reviews and found unresolved comments like [this one](https://github.com/iTwin/iTwinUI-react/pull/830#discussion_r996195590). I plan to review again later today.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: Some time ago we discussed adding accessibility fixes in other PR.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
gretanausedaite: This is internal function, leaving name as is, will rename later (if needed)

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
gretanausedaite: I added both type guards back. It nice to have some types enforcements in code.

gretanausedaite:
It's better to review full thing from scratch. I added a lot of stuff (then later removed that stuff) since last time you reviewed. Main things about multiselect: 
- Keeping type guards - it's cleaner way to ensure selected is array, multiple is true, etc.
- Multiselect on change return { array, event?: type, value}
- reducer has simple multiselect action that sets array of selected items
- onClickHandler works with button clicks (no more useEffects for selectedIndex). It is passed to keyboard navigation in input.
- onChangeHandler handles user provided onChange.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -181,14 +221,28 @@ export const ComboBox = <T,>(props: ComboBoxProps<T>) => {
     });
   }
 
+  // Get indices of selected elements in options array when we have selected values.
+  const getSelectedIndexes = React.useCallback(() => {
+    if (isMultipleEnabled(valueProp, multiple)) {
+      const indexArray: number[] = [];
+      valueProp &&
+        valueProp.forEach((value) => {
```
mayank99, line 229: nit
```suggestion
      valueProp?.forEach((value) => {
```

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -155,37 +178,56 @@ export const ComboBoxInput = React.forwardRef(
...
+          spellCheck={false}
+          autoCapitalize='none'
+          autoCorrect='off'
+          style={
+            multiple
+              ? {
+                  paddingLeft: tagContainerWidth + 18,
+                }
+              : {}
+          }
```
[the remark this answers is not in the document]
mayank99: also we might want to move this paddingLeft to a css variable in the future. we should not be setting regular css properties as inline styles usually

packages/iTwinUI-react/src/core/Select/SelectTagContainer.tsx
```diff
@@ -0,0 +1,44 @@
...
+import cx from 'classnames';
+import { useTheme, CommonProps, useOverflow, useMergedRefs } from '../utils';
+import SelectTag from './SelectTag';
+
+export type SelectTagContainerProps = {
+  /**
+   * Select tags.
+   */
+  tags: React.ReactNode[];
+} & Omit<CommonProps, 'title'>;
```
mayank99, line 15: This should use `Omit<React.ComponentPropsWithoutRef<'div'>, 'children'>`

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMultipleContainer.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const ComboBoxMultipleContainer = React.forwardRef(
+  (
+    { selectedItems = [] }: ComboBoxMultipleContainerProps,
+    ref: React.RefObject<HTMLDivElement>,
+  ) => {
+    return <SelectTagContainer ref={ref} tags={selectedItems} />;
+  },
+);
```
mayank99, line 19: Readability and flexibility of this component can be improved.
```suggestion
type ComboBoxMultipleContainerProps = {
  selectedItems?: JSX.Element[];
} & Omit<React.ComponentPropsWithoutRef<'div'>, 'children'>;

export const ComboBoxMultipleContainer = React.forwardRef(
  (props: ComboBoxMultipleContainerProps, ref: React.Ref<HTMLDivElement>) => {
    const { selectedItems = [], ...rest } = props;
    return <SelectTagContainer ref={ref} tags={selectedItems} {...rest} />;
  },
);
```

yarn.lock
```diff
@@ -1967,7 +1967,6 @@
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/@itwin/itwinui-variables/-/itwinui-variables-1.0.0.tgz#e108778511efbfda6197eed1b3b6a1e3fd19402c"
   integrity sha512-f78S401k1romzL6jgJuNZTe8P0/PCtrKUDVZwMsIJHLlgwyMKGR7nqDWQBV03tY2/TF/qMLt9S96Tej4f1Xrtg==
-
```
mayank99, line 1970: Undo this

mayank99 (commented):
Few minor comments and should be good to go

packages/iTwinUI-react/src/core/Select/SelectTagContainer.tsx
gretanausedaite: Changed

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
```diff
@@ -125,28 +140,36 @@ export const ComboBoxInput = React.forwardRef(
...
+                  dispatch({ type: 'close' });
+                }
+              } else {
+                onClickHandler?.(focusedIndexRef.current);
+                dispatch({ type: 'close' });
+              }
             } else {
-              dispatch(['open']);
+              dispatch({ type: 'open' });
             }
```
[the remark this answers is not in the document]
gretanausedaite: We want to keep menu open when multiselect is enabled and user select an item. 
If multiselect and item is selected - select item;
If multiselect and no item is selected - close;

apps/storybook/src/ComboBox.stories.tsx
```diff
@@ -575,3 +575,27 @@ Virtualized.args = {
 Virtualized.argTypes = {
   value: { control: { disable: true } },
 };
+
+export const MultipleSelect: Story<Partial<ComboBoxProps<string>>> = (args) => {
+  const options = React.useMemo(() => countriesList, []);
+  const [selectedOptions, setSelectedOptions] = React.useState<
+    string[] | undefined
```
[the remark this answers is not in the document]
gretanausedaite: Not needed. Removed.
I added it here to test "clearing" value. When user want to programmatically set value to undefined to clear any selection.

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
```diff
@@ -32,24 +33,64 @@ import { ComboBoxInputContainer } from './ComboBoxInputContainer';
 import { ComboBoxMenu } from './ComboBoxMenu';
 import { ComboBoxMenuItem } from './ComboBoxMenuItem';
 
+// Type guard for enabling multiple
+const isMultipleEnabled = <T,>(
+  variable: (T | undefined) | (T[] | undefined),
+  multiple: boolean,
+): variable is T[] | undefined => {
+  return multiple;
```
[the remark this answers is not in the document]
gretanausedaite: We lose `undefined` when using Array.isArray.

yarn.lock
gretanausedaite: undid

packages/iTwinUI-react/src/core/ComboBox/ComboBoxInput.tsx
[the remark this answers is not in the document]
mayank99: Ah that makes sense. Could you add a brief comment saying this is to "keep menu open when multiselect is enabled and user selects and item"?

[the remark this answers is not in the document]
gretanausedaite: Added comment

packages/iTwinUI-react/src/core/ComboBox/ComboBox.tsx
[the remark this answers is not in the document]
gretanausedaite: Updated code to check for array and undefined.

mayank99 (approved):
LGTM

elephantcatdog (commented):
I found two things when I was messing around with it.

The red marks in the below image show where `cursor: pointer` for `iui-input-with-icon` is being applied. When I click here when the combobox is closed nothing happens, but when it's open clicking there closes the combobox.

I think it would be better without `cursor: pointer` there, but I don't think it's a big deal if it requires a lot of effort. Also, on the far left the cursor is an I-beam cursor as if it's part of the text box, which I suppose it is? It acts like it is (ie, when you click there the cursor is moved to the text box and the combobox opens).
image

Also, I was initially able to make the combobox flicker a lot when (I think) the box was just smaller than the text/items inside, but I tried to take a video of it and haven't been able to reproduce it since. That's not super helpful, but I'll update if I can reliably reproduce it. Maybe it was just something wrong with my browser.

gretanausedaite:
> Also, I was initially able to make the combobox flicker a lot when (I think) the box was just smaller than the text/items inside, but I tried to take a video of it and haven't been able to reproduce it since. That's not super helpful, but I'll update if I can reliably reproduce it. Maybe it was just something wrong with my browser.

I also noticed this issue. I believe it's unavoidable. Even if I check for smaller container there will be a breaking point in which combobox is going to flicker.

gretanausedaite:
> I think it would be better without `cursor: pointer` there, but I don't think it's a big deal if it requires a lot of effort. Also, on the far left the cursor is an I-beam cursor as if it's part of the text box, which I suppose it is? It acts like it is (ie, when you click there the cursor is moved to the text box and the combobox opens).

To remove `cursor: pointer` is css issue (or maybe not even the issue and expected behaviour).
The "correct" Tag should have x icon to remove it from the list, should be added some time later.

elephantcatdog (approved):

PR #835, opened 2022-09-21
feat(Table): Column resize modes

### LINKED ISSUE #824: Expand table on column resize
We could have some flag that toggles resize mode like react-grid https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/column-resizing/#resizing-modes

### Discussed in https://github.com/iTwin/iTwinUI-react/discussions/822

<div type='discussions-op-text'>

<sup>Originally posted by **HaveSpacesuit** September  9, 2022</sup>
It is difficult to expand columns to a wide width. Currently it looks like when a column is resized, the next column loses or gains width based on the first column resize. The overall table width stays the same.

Column resize static

It would be nice to expand/shrink the entire table width when a column is resized. If I make column A bigger, I likely do not want to make column B smaller.

This should work in situations when the table has horizontal scroll and does not have horizontal scroll - adding or removing the scroll as needed.

Here is a sandbox showing that it is not easy to expand a column past the point where the next column is at minimum width. There are examples with and without horizontal scroll. (I usually have the last column width set to undefined in order to allow the table to fill horizontal space if the columns do not all fit.)

https://codesandbox.io/s/itwinui-react-minimal-example-forked-7wizuj?file=/src/columns.ts

Column width scroll
</div>

bentleyvk:
Closes #824 

Added new prop `resizeMode` and it can have two values `next-column` (default one) and `current-column` (new one). Better names suggestions are welcome.

ztable-resize-current

## Checklist


packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -280,12 +276,20 @@ const reducer = <T extends Record<string, unknown>>(
...
-      !isNewColumnWidthsValid(newNextColumnWidths, instance?.flatHeaders)
+      !isNewColumnWidthsValid(newColumnWidths, instance) ||
+      !isNewColumnWidthsValid(newNextColumnWidths, instance)
     ) {
       return newState;
     }
 
+    // Setting `width` here because it might take several rerenders until actual column width is set.
+    // Also setting after the actual resize happened.
+    instance?.flatHeaders.forEach((h) => {
```
bentleyvk, line 287: Moved here from pointer down handler because it was causing a bug when nothing was resized because e.g. you try to reduce size of a column that already is in its minimum width.
table-bug-resize

packages/iTwinUI-react/src/types/react-table-config.ts
```diff
@@ -100,6 +100,15 @@ declare module 'react-table' {
...
     data: D[];
+    /**
+     * Column's resize mode.
+     *  - `next-column` - when resizing it affects current and the next column,
+     *   e.g. when increasing width of current column, next column's width will decrease.
+     *  - `current-column` - when resizing it affects only the current column,
+     *   e.g. when increasing width of the current column, next column's width remains the same.
+     * @default 'next-column'
+     */
+    resizeMode?: 'next-column' | 'current-column';
```
bentleyvk, line 111: Added new prop here instead of `Table.tsx` so I can avoid typescript errors and still have `resizeMode` in table instance.

packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -348,6 +352,19 @@ const isNewColumnWidthsValid = <T extends Record<string, unknown>>(
...
 
+  if (instance && instance.resizeMode === 'current-column') {
+    let newTableWidth = 0;
+    for (const header of instance.flatHeaders) {
+      newTableWidth += columnWidths[header.id]
+        ? columnWidths[header.id]
+        : getHeaderWidth(header);
+    }
+    // `tableWidth` is whole number therefore we need to round the `newTableWidth`
+    if (Math.round(newTableWidth) < instance.tableWidth) {
```
gretanausedaite, line 363: Why we are not allowing to decrese table width?

packages/iTwinUI-react/src/types/react-table-config.ts
gretanausedaite: `resizeMode` already exists in `<Image />`. Maybe `columnResizeMode`?
Or maybe reuse some resizeMode names and have `resizeMode?: 'contain' | 'strech'` ? With contain table keeps its width, with strech - it changes.

packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
bentleyvk: I don't like the idea that table will be smaller because that can cause some layout shifts or they can even break without devs even thinking about that. Also I checked other libs and although they allow to reduce column size but table size remains the same but their table columns are a bit different, they don't have strict borders.
image

gretanausedaite: I agree that it could break some layouts.
But users can reduce column width to get table to original size if other column is expanded. So not being able to decrease size of column in expand mode might confuse users and they could think it's a bug. 
Maybe we could consult UX team about expected behaviour or just talk within a team.

gretanausedaite (commented):
I found issues with last column resizing in expand mode. It's not smooth and it gets `maximum update depth exceeded` error after some attempts to resize.
image

(00:16 shift happens on hover)
https://user-images.githubusercontent.com/83585998/191941136-2de2f1b2-349c-4e51-9089-8a24381506e7.mov

packages/iTwinUI-react/src/types/react-table-config.ts
bentleyvk: Renamed

bentleyvk:
> I found issues with last column resizing in expand mode. It's not smooth and it gets `maximum update depth exceeded` error after some attempts to resize. image
> 
> (00:16 shift happens on hover) https://user-images.githubusercontent.com/83585998/191941136-2de2f1b2-349c-4e51-9089-8a24381506e7.mov

This issue is reproducable in deployed storybook too, you only need to resize very fast and release left mouse button while doing so. I will fix this with another PR cause it might take some time and users are complaining about our current resize mode.

packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -268,24 +264,45 @@ const reducer = <T extends Record<string, unknown>>(
...
+        ? getColumnWidths(nextHeaderIdWidths, -deltaX / nextColumnWidth)
+        : {};
 
     if (
-      !isNewColumnWidthsValid(newColumnWidths, instance?.flatHeaders) ||
-      !isNewColumnWidthsValid(newNextColumnWidths, instance?.flatHeaders)
+      !isNewColumnWidthsValid(newColumnWidths, instance.flatHeaders) ||
+      !isNewColumnWidthsValid(newNextColumnWidths, instance.flatHeaders) ||
+      !isNewTableWidthValid(
+        { ...newColumnWidths, ...newNextColumnWidths },
```
gretanausedaite, line 291: I think you are expanding next column. Maybe this could be changed to next avaible for resize? Because right now in storybook when I try to make 'Description' smaller - nothing happens. I think 'End date' should be expanded.

bentleyvk: 'End Date` has max width so I can't expand past it. And I don't really want to write some more code to make things even more complex to check whether it reached max width and if so, expand another column. At least until we get another issue regarding this.

gretanausedaite (commented):
Resizing columns looks good except for last column.
If we don't want to solve everything now, we should at least create issue with known bugs/issues to solve them later.
And for now I think it would be important to fix this problem when expanding last column and expand handle goes out of view - it is not being released. In video I'm expanding last column and then I want to scroll, but when I move my mouse, last column is resized and it gives this crazy table height changes.
Also, maybe we could keep cursor style during whole resize process, not only when hovered on header?

gretanausedaite (commented):
Resizing columns looks good except for last column.
If we don't want to solve everything now, we should at least create issue with known bugs/issues to solve them later.
And for now I think it would be important to fix this problem when expanding last column and expand handle goes out of view - it is not being released. In video I'm expanding last column and then I want to scroll, but when I move my mouse, last column is resized and it gives this crazy table height changes.

https://user-images.githubusercontent.com/83585998/192736927-6f02a5d1-7912-40ad-a86e-216702ba7805.mp4


Also, maybe we could keep cursor style during whole resize process, not only when hovered on header?

bentleyvk:
> Resizing columns looks good except for last column. If we don't want to solve everything now, we should at least create issue with known bugs/issues to solve them later. And for now I think it would be important to fix this problem when expanding last column and expand handle goes out of view - it is not being released. In video I'm expanding last column and then I want to scroll, but when I move my mouse, last column is resized and it gives this crazy table height changes.
> 
> Also, maybe we could keep cursor style during whole resize process, not only when hovered on header?

Stopping to resize when mouse leaves the screen

gretanausedaite (approved):
Code looks good 🌵

PR #836, opened 2022-09-21
feat(Table): Row selection count for paginator

### LINKED ISSUE #834: Table: Row selection count for Paginator
Display the number of rows selected in a paginated table

image

adamhock:
There is now a counter in the bottom left section the paginated tables that shows how many rows have been selected if multi-selection mode is enabled.

Closes #834


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -599,6 +603,10 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  const numSelectedRows =
+    selectionMode === 'single'
+      ? 0
```
rohan-kadkol, line 608: What if one row is selected when `selectionMode` is single? Then `numSelectedRows` should be 1?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -599,6 +603,10 @@ export const Table = <
...
+      : rows.filter((row) => row.isSelected).length;
```
rohan-kadkol, line 609: Are we also counting subrows in this `numSelectedRows` count?

If so, then you might need to use `flatRows`. It is an ordered array of all rows (even subrows).

packages/iTwinUI-react/src/core/Table/TablePaginator.tsx
```diff
@@ -282,7 +283,15 @@ export const TablePaginator = (props: TablePaginatorProps) => {
       ref={paginatorResizeRef}
       {...rest}
     >
-      <div className='iui-left' />
+      <div className='iui-left'>
+        {totalSelectedRowsCount > 0 && (
+          <span className='iui-table-paginator-selection-count'>
+            {`${totalSelectedRowsCount} ${
+              totalSelectedRowsCount > 1 ? 'rows' : 'row'
```
rohan-kadkol, line 290:
```suggestion
              totalSelectedRowsCount === 1 ? 'row' : 'rows'
```
nit: Maybe a more safer boolean could be the one above?

If in the future we want to show even this count when 0 rows are selected, we don't need to change this boolean.

PR #837, opened 2022-09-21
feat(Table): Row selection count for paginator

### LINKED ISSUE #834: Table: Row selection count for Paginator

adamhock:
PR continued from #836 

There is now a counter in the bottom left section the paginated tables that shows how many rows have been selected if multi-selection mode is enabled.

Closes https://github.com/iTwin/iTwinUI-react/issues/834


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -659,6 +663,10 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  const numSelectedRows =
+    selectionMode === 'single'
+      ? 0
```
adamhock, line 668: comment from @rohan-kadkol 

> What if one row is selected when `selectionMode` is single? Then `numSelectedRows` should be 1?

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -659,6 +663,10 @@ export const Table = <
...
+      : rows.filter((row) => row.isSelected).length;
```
adamhock, line 669: comment from @rohan-kadkol 

> Are we also counting subrows in this `numSelectedRows` count?
> 
> If so, then you might need to use `flatRows`. It is an ordered array of all rows (even subrows).

packages/iTwinUI-react/src/core/Table/TablePaginator.tsx
```diff
@@ -281,7 +282,15 @@ export const TablePaginator = (props: TablePaginatorProps) => {
       ref={paginatorResizeRef}
       {...rest}
     >
-      <div className='iui-left' />
+      <div className='iui-left'>
+        {totalSelectedRowsCount > 0 && (
+          <span className='iui-table-paginator-selection-count'>
+            {`${totalSelectedRowsCount} ${
+              totalSelectedRowsCount > 1 ? 'rows' : 'row'
```
adamhock, line 289: comment from @rohan-kadkol 

```suggestion
              totalSelectedRowsCount === 1 ? 'row' : 'rows'
```
 > nit: Maybe a more safer boolean could be the one above?

> If in the future we want to show even this count when 0 rows are selected, we don't need to change this boolean.

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: This was an issue for me. For one, if there is single selection mode, I do not think we should show how many rows are selected since the max amount of rows that can be selected in single selection mode is 1. I set this to 0 so that the checked rows counter isn't displayed in this case. 

With that being said, it does make the variable name misleading :/

mayank99:
Can you reassign the reviewers to be the same as the old PR?

Also I have a question: what does the `iui-table-paginator-selection-count` do? We don't have any styles for it afaict.

adamhock:
> Also I have a question: what does the `iui-table-paginator-selection-count` do? We don't have any styles for it afaict.

This class name has been removed

packages/iTwinUI-react/src/core/Table/TablePaginator.tsx
adamhock: This change has been made

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -2499,6 +2499,33 @@ it('should change page size', async () => {
...
+    data: mockedData(100),
+    pageSize: 10,
+    paginatorRenderer: (props) => <TablePaginator {...props} />,
+    isSelectable: true,
+  });
+
+  const rowCheckboxes = container.querySelectorAll(
+    '.iui-table-body .iui-row .iui-checkbox',
+  );
+
```
bentleyvk, line 2513: Nit picking but here are too many new lines 😀 Try grouping related code with new lines, like you click something and then expect some result.

packages/iTwinUI-react/src/core/Table/Table.tsx
bentleyvk: I don't think there is a best way here. For example, it could be beneficial to show that there is already one row selected when you filtered table and that selected row is not visible or when you go to the next page, but also you should remember that you selected just one row.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -659,6 +663,10 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  const numSelectedRows =
```
bentleyvk, line 666: Wrap this with `React.useMemo` or move under `paginatorRendererProps` because right not it is going trough all rows on each rerender.

bentleyvk: I don't maybe it's just for me but `selectedRowsCount` sounds better.

packages/iTwinUI-react/src/core/Table/Table.tsx
bentleyvk: This should also fix a bug where you apply filter and it no longer shows how many rows are selected if they are not visible.

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: Ohh, right, that is a good point. Also if we don't want to display it when `selectionMode = single`, can we maybe pass the selectionMode to the TablePaginator so that it can only display the row count when `selectionMode === single && totalSelectedRowsCount > 0` instead of just the second condition? (on line 286 of `TablePaginator.tsx`)

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: On that note, can we maybe make it to `totalSelectedRowsCount` to be the same name as the property on [line 97](https://github.com/iTwin/iTwinUI-react/pull/837/files#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999R97)? Or maybe make changed line 97 to `selectedRowsCount` so both have the same names?

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol, line 669: I'm not sure because I'm not the most confident with using hooks, but do we usually wrap such things with `useMemo` or `useCallback`, etc.? Not sure

```suggestion
  const numSelectedRows = React.useMemo(
    () =>
      selectionMode === 'single'
        ? 0
        : rows.filter((row) => row.isSelected).length,
    [rows, selectionMode],
  );
```

bentleyvk: Yes, this should be wrapped with `useMemo` because it is a variable that changes only when `rows` or `selectionMode` changes. We use `useCallback` for function calls and in most cases we do that in order to prevent component that is getting that function from rerendering unnecessary.

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: I think it's unnecessary to pass the `selectionMode` prop into `TablePaginatorRendererProps`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
adamhock: The new lines have been removed

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: I have moved this count under `paginatorRendererProps` and removed the `numSelectedRows` variable; however, now the row selection count is not updating in the unit test which is causing it to fail. I'm still looking into this

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: I have moved this count under `paginatorRendererProps` and removed the `numSelectedRows` variable

adamhock: This has been changed to use `flatRows` here, but now the row selection count includes all rows and their subrows.

Imagine a table with Row 1 which has subrows Row 1a and Row 1b. If Row 1 is selected, the row count shows as 3 rows selected (Row 1, Row 1a, and Row 1b). If all of a row's subrows are selected, that row is also counted as selected. Is this an issue?

rohan-kadkol: I think that is how it works. If one row is selected, all subrows, and subrows' subrows (excl. disabled) have to be selected

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -2499,6 +2499,31 @@ it('should change page size', async () => {
...
+  fireEvent.click(rowCheckboxes[0]);
+  expect(container.querySelector('.iui-left span')?.textContent).toBe(
+    '1 row selected',
+  );
+
+  fireEvent.click(rowCheckboxes[1]);
+  expect(container.querySelector('.iui-left span')?.textContent).toBe(
+    '2 rows selected',
+  );
+});
```
rohan-kadkol, line 2525: Maybe you could add a test case to test flatRows. So by selecting one row, ensure that the subrow count is also included in the paginator count.

Optionally, you could also test that the disbaled subrows count is not added to the total count.

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: What if we have a prop to allow users to not display the selected count? That way `totalSelectedRowCount` can store the real selectedCount for both `single` (0-1) and `multiple` (0+) selectionModes. Then if the user doesn't want the selected count, they could pass `false`.

rohan-kadkol: I think then let's go ahead with having that 0 there. But it would help to add a comment saying why `totalSelectedRowsCount` can be 0 even when one row is selected. Else, someone in the future might get confused that this is a bug.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: I think this test would be useful. Like testing what happens if you select a row with subrows. The row selection count should also include the subrow count.

adamhock: The test has now been updated and uses a table with subrows

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: I saw your comment on [L94-96](https://github.com/iTwin/iTwinUI-react/pull/837/files#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999R94-R96). But better to also write a comment on [L669-670](https://github.com/iTwin/iTwinUI-react/pull/837/files#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999R669-R670). Probably something like:

```tsx
totalSelectedRowsCount:
        selectionMode === 'single' ? 0 : instance.selectedFlatRows.length, // 0 when selectionMode = 'single' since totalSelectedRowCount is for multi-selection mode only
```

adamhock: This comment has been added

r100-stack (approved):

bentleyvk (approved):

packages/iTwinUI-react/src/core/Table/TablePaginator.tsx
```diff
@@ -282,7 +283,15 @@ export const TablePaginator = (props: TablePaginatorProps) => {
       ref={paginatorResizeRef}
       {...rest}
     >
-      <div className='iui-left' />
+      <div className='iui-left'>
+        {totalSelectedRowsCount > 0 && (
+          <span>
+            {`${totalSelectedRowsCount} ${
+              totalSelectedRowsCount === 1 ? 'row' : 'rows'
+            } selected`}
```
mayank99, line 291: this needs to be localized

elephantcatdog: This is going to be fixed in an upcoming PR

PR #845, opened 2022-09-29
fix: Add exports field

### LINKED ISSUE #727: Importing itwinui-react components in vitest environment results in an error
Importing a component from this library in vitest environment results in an error. More exactly:
```
SyntaxError: Unexpected token '{'
 ❯ wrapSafe ../../../../../../../internal/modules/cjs/loader.js:1001:16
 ❯ Module._compile ../../../../../../../internal/modules/cjs/loader.js:1049:27
 ❯ Object.Module._extensions..js ../../../../../../../internal/modules/cjs/loader.js:1114:10
 ❯ Module.load ../../../../../../../internal/modules/cjs/loader.js:950:32
 ❯ Function.Module._load ../../../../../../../internal/modules/cjs/loader.js:790:12
 ❯ Module.require ../../../../../../../internal/modules/cjs/loader.js:974:19
 ❯ require ../../../../../../../internal/modules/cjs/helpers.js:101:18
 ❯ Object.<anonymous> ../../common/temp/node_modules/.pnpm/@itwin/itwinui-react@1.40.1_react-dom@16.14.0+react@16.14.0/node_modules/@itwin/itwinui-react/cjs/core/utils/hooks/useTheme.
 ❯ Module._compile ../../../../../../../internal/modules/cjs/loader.js:1085:14
 ❯ Object.Module._extensions..js ../../../../../../../internal/modules/cjs/loader.js:1114:10
```

#### Expected behavior
Running tests through vitest should be able to import itwinui-react components.

#### Reproduction
https://stackblitz.com/edit/github-w9fjy7-cacfzq?file=package.json,src%2Fitwinui%2Fimport-test.test.tsx

Commenting out the `<Button>Itwinui button</Button>` line in the linked file will cause the test to pass.

##### Steps to reproduce
1. Setup a react project with vitest as the test runner
1. Import a `itwinui-react` component from a vitest test
1. Notice the error

#### Additional information
I tried digging into possible problems that could be causing this, and inspecting the line where this failed in `useTheme`, it's failing on an import of a css file. It seems like this is a known issue on the vitest github page: https://github.com/vitest-dev/vitest/issues/496

And according to the discussion there, it seems like just adding the newly supported `exports` field in to the package.json file of library resolves this problem. 

To check this, I tried modifying the `package.json` file in my node_modules folder, and indeed, afterwards importing and rendering of the components in vitest then work.

The working package.json looked as so: 
```
{
  "name": "@itwin/itwinui-react",
  "version": "1.40.1",
  "author": "Bentley Systems",
  "license": "MIT",
  "main": "cjs/index.js",
  "module": "esm/index.js",
  "typings": "cjs/index.d.ts",
  "exports": {
    "import": "./esm/index.js"
  },
  ...
```

veekeys:
Added `exports` field to package json, which should not produce any breaking changes and also make vitest work (havent tested astro).

This just sets the correct entry point for esm import and cjs require. (allows only imports from path `@itwin/itwinui-react`
```
".": {
      "import": "./esm/index.js",
      "require": "./cjs/index.js"
},
```

This allows any deep level import path from both cjs and esm folders. If you want to limit that to certain scope (i.e. utils only), this can be done with v2 only as otherwise we risk delivering a breaking change.
```
"./cjs/": "./cjs/",
"./esm/": "./esm/"
```
Docs I read
https://nodejs.org/api/packages.html#exports
https://webpack.js.org/guides/package-exports/

Closes #727


mayank99:
Storybook is currently failing:

```
ERR! Error: [vite]: Rollup failed to resolve import "@itwin/itwinui-react/esm/core/utils" from "src/Badge.stories.tsx".
```

Do you need a `*` too? e.g. `./esm/*`

Also, we want to make a v2 release anyway, so if you can get it to work with a certain combination of exports, then I think that's totally fine to do that without worrying about breaking user imports. We could even have a special key for utils.
```
"./utils": {
  "import": "./esm/core/utils/index.js",
  "require": "./cjs/core/utils/index.js"
},
```

veekeys:
> Storybook is currently failing:
> 
> ```
> ERR! Error: [vite]: Rollup failed to resolve import "@itwin/itwinui-react/esm/core/utils" from "src/Badge.stories.tsx".
> ```
> 
> Do you need a `*` too? e.g. `./esm/*`
> 
> Also, we want to make a v2 release anyway, so if you can get it to work with a certain combination of exports, then I think that's totally fine to do that without worrying about breaking user imports. We could even have a special key for utils.
> 
> ```
> "./utils": {
>   "import": "./esm/core/utils/index.js",
>   "require": "./cjs/core/utils/index.js"
> },
> ```

yea, that didnt help....
What is more interesting, when I added this explicit resolution `"./esm/core/utils": "./esm/core/utils/index.js"`, it started to throw this 
Screenshot 2022-09-30 at 17 01 52

packages/iTwinUI-react/package.json
```diff
@@ -6,6 +6,14 @@
   "main": "cjs/index.js",
```
rohan-kadkol, line 6: If we are using `exports`, maybe the `main` is overridden and not needed?

veekeys: Well, I thought its needed for lower version of node and potentially for backwards compatibility if some bundlers fail with it
https://nodejs.org/docs/latest-v16.x/api/packages.html#package-entry-points

but not sure.. we could risk getting rid of it ? But again..safety reasons if `exports` is not supported..

veekeys:
Made it pass, but I have no idea, why this weird export of `./` is needed (it shows warning as invalid...).
Blame typescript as I think it still has experimental support for exports (tried updating ts to the newest, didnt help)... or rollup has some weird behaviour on top of that..

gretanausedaite (approved):
I have no comments on code.

apps/storybook/src/Tabs.stories.tsx
```diff
@@ -11,12 +11,12 @@ import {
...
   VerticalTabsProps,
   Tab,
+  TabsProps,
 } from '@itwin/itwinui-react';
-import { Tabs, TabsProps } from '@itwin/itwinui-react/esm/core/Tabs/Tabs';
 
 export default {
   title: 'Core/Tabs',
-  component: Tabs,
+  component: HorizontalTabs,
```
rohan-kadkol, line 19: Just curious, do we keep this `Tabs` since this story shows both Horizontal and Vertical tabs? Also to confirm, what does this `component` prop actually do?

r100-stack (approved):

apps/storybook/src/Text.stories.tsx
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import { Meta, Story } from '@storybook/react';
 import React from 'react';
-import { Text, TextProps } from '@itwin/itwinui-react/';
+import { Text, TextProps } from '@itwin/itwinui-react';
```
mayank99, line 7: So if the user imports with `/`, what error would they get?

mayank99 (commented):
Can you change the target branch to `v2` since this will potentially break user imports?

apps/storybook/src/Text.stories.tsx
veekeys: Something like this...
Screenshot 2022-10-05 at 16 08 42

But i am sure, that the first time i got the error, it had `./` export was missing, not the `.`. Thats how I figured to global search import with ending slash...
Basically, user would always get export path is missing, if they are importing from path we have not defined in our package.json

mayank99:
Check out https://github.com/iTwin/iTwinUI-icons/commit/5ba48903694154004ebded4def8fe5f944b48222 for how I'm handling this in https://github.com/iTwin/iTwinUI-react/pull/845 in accordance with new TS features.

Not sure if we want to do all of that here yet, because proper ESM support would mean we need to change all imports/exports to reference full file names (e.g. `utils/index.js`, `Button.js` etc instead of `utils`, `Button`).

But at the very least, I would like these two entries in `exports` so that the user has a nicer way of explicitly choosing which folder to import from.
```json
    "./esm": "./esm/index.js",
    "./cjs": "./cjs/index.js",
```

mayank99 (approved):
This LGTM now. I would say lets make a dev release after merging and test in vitest (or ask the user who opened this to test it).

PR #848, opened 2022-09-30
fix(Table): Fixed when fast resizing caused Table to throw an error

bentleyvk:
table-bug-resize-rerenders

When you resize table very fast and release mouse button, it will throw an error that too many setState were called.
After debugging I found out that for some reason `columnResizeDone` was being called 20+ times although it was dispatched only once. I tried our `useColumnResize` hook inside one of the react-table examples and everything was working properly. So I tried dropping parts of the table until I came down to removing useState setter calls from refs. Another issue was also that when you release mouse in the middle of column, it would trigger a sort, it was also causing the same problem, so changed back sorting to `onMouseDown`.

## Checklist

### DIFF
--- a/packages/iTwinUI-react/src/core/DropdownMenu/DropdownMenu.test.tsx
+++ b/packages/iTwinUI-react/src/core/DropdownMenu/DropdownMenu.test.tsx
@@ -19,7 +19,10 @@ function assertBaseElement(menu: HTMLUListElement, role = 'menu') {
   });
 }
 
-function renderComponent(props?: Partial<DropdownMenuProps>) {
+function renderComponent(
+  props?: Partial<DropdownMenuProps>,
+  container?: HTMLElement,
+) {
   return render(
     <DropdownMenu
       menuItems={(close) => [
@@ -37,19 +40,23 @@ function renderComponent(props?: Partial<DropdownMenuProps>) {
     >
       <Button>Click here</Button>
     </DropdownMenu>,
+    { container },
   );
 }
 
 it('should show menu only after click', () => {
-  renderComponent();
+  const mockDocument = document.implementation.createHTMLDocument();
+  const div = mockDocument.createElement('div');
+  const mockContainer = mockDocument.body.appendChild(div);
+  renderComponent(undefined, mockContainer);
 
-  let menu = document.querySelector('.iui-menu') as HTMLUListElement;
+  let menu = mockDocument.querySelector('.iui-menu') as HTMLUListElement;
   expect(menu).toBeFalsy();
 
-  const button = screen.getByText('Click here');
-  fireEvent.click(button);
+  const button = mockDocument.querySelector('.iui-button') as HTMLButtonElement; // screen.getByText('Click here');
+  button.click(); // fireEvent.click(button);
 
-  menu = document.querySelector('.iui-menu') as HTMLUListElement;
+  menu = mockDocument.querySelector('.iui-menu') as HTMLUListElement;
   assertBaseElement(menu);
 });
 
--- a/packages/iTwinUI-react/src/core/DropdownMenu/DropdownMenu.test.tsx
+++ b/packages/iTwinUI-react/src/core/DropdownMenu/DropdownMenu.test.tsx
@@ -19,10 +19,7 @@ function assertBaseElement(menu: HTMLUListElement, role = 'menu') {
   });
 }
 
-function renderComponent(
-  props?: Partial<DropdownMenuProps>,
-  container?: HTMLElement,
-) {
+function renderComponent(props?: Partial<DropdownMenuProps>) {
   return render(
     <DropdownMenu
       menuItems={(close) => [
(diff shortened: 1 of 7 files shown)


packages/iTwinUI-react/src/core/DropdownMenu/DropdownMenu.test.tsx
```diff
@@ -19,7 +19,10 @@ function assertBaseElement(menu: HTMLUListElement, role = 'menu') {
   });
 }
 
-function renderComponent(props?: Partial<DropdownMenuProps>) {
+function renderComponent(
```
bentleyvk, line 22: I don't understand why test `should render filter dropdown in the correct document` in Table fails. So I tried to do the same in other tests and it fails here too. Anyone has any suggestions?

packages/iTwinUI-react/src/core/Table/filters/FilterToggle.tsx
```diff
@@ -8,12 +8,11 @@ import React from 'react';
 import cx from 'classnames';
 import { HeaderGroup } from 'react-table';
 import '@itwin/itwinui-css/css/table.css';
-import { useTheme, Popover, getDocument, StylingProps } from '../../utils';
+import { useTheme, Popover, StylingProps } from '../../utils';
 import { IconButton } from '../../Buttons';
 
 export type FilterToggleProps<T extends Record<string, unknown>> = {
   column: HeaderGroup<T>;
-  ownerDocument?: Document;
```
bentleyvk, line 16: By default Popover appends content to `ownerDocument.body`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -1654,7 +1654,9 @@ it('should render filter dropdown in the correct document', async () => {
     '.iui-filter-button',
   ) as HTMLElement;
   expect(filterToggle).toBeTruthy();
-  act(() => filterToggle.click());
+  act(() => {
+    filterToggle.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
```
bentleyvk, line 1658: Change is needed because it now has `onMouseDown` handler instead of `onClick`. I can't use `fireEvent` or `userEvent` because they are throwing this error here: `It looks like the window object is not available for the provided node`

packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -186,14 +184,12 @@ const defaultGetResizerProps = (ownerDocument: Document | undefined) => (
   return [
     props,
     {
-      onClick: (e: React.MouseEvent) => {
-        // Prevents from triggering sort
-        e.stopPropagation();
-      },
       onMouseDown: (e: React.MouseEvent) => {
```
mayank99, line 187: can we use `onPointerDown` to combine `mousedown` and `touchstart`?

packages/iTwinUI-react/src/core/Table/filters/FilterToggle.tsx
```diff
@@ -53,13 +52,12 @@ export const FilterToggle = <T extends Record<string, unknown>>(
...
           visible={isVisible}
           onClickOutside={close}
-          appendTo={ownerDocument?.body}
         >
           <IconButton
             styleType='borderless'
             isActive={isVisible || isColumnFiltered}
             className={cx('iui-filter-button', className)}
-            onClick={(e) => {
+            onMouseDown={(e) => {
```
mayank99, line 60: This will break for keyboard users because `click` automatically handles keyboard events too. Also it's a bit unexpected because click handlers usually activate when the mouse is released. 

Same comment in other places like BaseFilter

I think we should try to prevent the sort handler from being called when resizing.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -414,36 +414,32 @@ export const Table = <
...
-      }
-    },
-    [ownerDocument],
-  );
+  const disableUserSelect = React.useCallback((e: KeyboardEvent) => {
+    if (e.key === 'Shift') {
+      ownerDocument.current &&
+        (ownerDocument.current.documentElement.style.userSelect = 'none');
+    }
+  }, []);
```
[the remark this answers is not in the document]
bentleyvk: When using `useState`, its value can be changed with setter function, therefore it has to be in the deps list to prevent calling function in `useCallback` with an old value. When using `useRef`, value never changes as it is an object whose reference is always the same, only `value.current` changes. Therefore it is not needed in deps list since reference never changes and `value.current` inside `useCallback` function will take new values whenever it is called.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
bentleyvk: Reverted.

packages/iTwinUI-react/src/core/Table/hooks/useResizeColumns.tsx
```diff
@@ -113,14 +111,18 @@ const defaultGetResizerProps = (ownerDocument: Document | undefined) => (
...
+          ownerDocument.current?.removeEventListener(
             'mousemove',
             handlersAndEvents.mouse.moveHandler,
           );
-          ownerDocument.removeEventListener(
+          ownerDocument.current?.removeEventListener(
             'mouseup',
             handlersAndEvents.mouse.upHandler,
           );
+          ownerDocument.current?.removeEventListener(
```
bentleyvk, line 122: I saw that it was triggering on resize end every time mouse leaves the screen. Now it only triggers that once.

mayank99 (approved):

PR #850, opened 2022-09-30
docs: Add recommended node version to CONTRIBUTING.md

r100-stack:
Added recommendation to use node version 16 to avoid build issues

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ Want to contribute by creating a pull request? Great! [Fork iTwinUI-react](https
 
 ## How to setup
 
-To clone and build iTwinUI-react, you'll need [Git](https://git-scm.com), [Yarn 1](https://yarnpkg.com/getting-started/install) and [Node 16](https://nodejs.org/en/download/current/) installed on your computer.
+To clone and build iTwinUI-react, you'll need [Git](https://git-scm.com), [Node 16](https://nodejs.org/en/download/releases/), and [Yarn 1](https://classic.yarnpkg.com/en/docs/install) installed on your computer.
 
 1. [Create a local clone](https://docs.github.com/en/get-started/quickstart/fork-a-repo#cloning-your-forked-repository) of your forked repository. You can do this from the command line or using the Github Desktop app.
 2. Go to the directory where you cloned iTwinUI-react. e.g. `cd iTwinUI-react`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -37,6 +37,8 @@ e.g. to build all workspaces together, run the following command from the root:
 yarn build
 ```
 
+**NOTE:** To avoid build issues, we recommend using [Node](https://nodejs.org/en/download/current/) *v16.x (16+)* and not *v17.x (17+)*. You can install and use [nvm](https://github.com/nvm-sh/nvm#intro) to change your node version.
```
mayank99, line 40: This is the wrong place for this note. This section is talking about the monorepo workflow and has nothing to do with node.

I think you can just add "Node 16" in line 15 where it mentions git and yarn.

And there is no need to mention `nvm` (in fact, it is better to use [`fnm`](https://github.com/Schniz/fnm)) as it's cross platform). It's upto the contributor how they want to manage their versions.

Lastly, "16+" implies 17 and 18, so you should just say "16".

rohan-kadkol: **[DONE]** Thanks, I just mentioned [Node 16](https://nodejs.org/en/download/current/) on line 15.

I had placed it earlier on line 40 since with v17, I think there were no issues with `yarn install`. But it started causing issues when running `yarn build` (not sure). So I placed it near the mention of `yarn build`

mayank99 (approved):

CONTRIBUTING.md
```diff
@@ -12,7 +12,7 @@ Want to contribute by creating a pull request? Great! [Fork iTwinUI-react](https
 
 ## How to setup
 
-To clone and build iTwinUI-react, you'll need [Git](https://git-scm.com) and [Yarn 1](https://yarnpkg.com/getting-started/install) installed on your computer.
+To clone and build iTwinUI-react, you'll need [Git](https://git-scm.com), [Yarn 1](https://yarnpkg.com/getting-started/install) and [Node 16](https://nodejs.org/en/download/current/) installed on your computer.
```
mayank99, line 15: Unrelated but I noticed the yarn 1 link is pointing to the wrong place. Can you update it to https://classic.yarnpkg.com/en/docs/install?

Also I think node should be mentioned before yarn because it is a prereq

rohan-kadkol: Ohh, good points, thanks. I changed the yarn link and put node before yarn.

I also noticed that the old node link was for the current version (18+). So I replaced the old link (https://nodejs.org/en/download/current/) with the link to the LTS version (16.17.1) (https://nodejs.org/en/download/). Is this alright?

mayank99: Looks good

adamhock (approved):

Issue #843, opened 2022-09-28
Remove deprecated things for iTwinUI-React v2


FlyersPh9:
- [ ] Alert
- [ ] Checkbox
- [ ] Tabs

PR #852, opened 2022-09-30
feat(Tabs)!: Remove deprecated `HorizontalTab`

mayank99:
Removed the already deprecated `HorizontalTab` and `HorizontalTabProps` from both the source and the barrel.

Added ! to the title to trigger a major release.

Closes part of #843


gretanausedaite (approved):
Code looks good.
Shouldn't we trigger major release with `v2 -> main` merge?

mayank99:
> Shouldn't we trigger major release with `v2 -> main` merge?

So previously I was targeting `main` branch, that's why I added `!`. But even with `v2` branch, I think we might want to rebase the commits onto main so this will show up in the final commit message too.

adamhock (approved):

PR #861, opened 2022-10-06
fix(Table): Select all works in indeterminate state

### LINKED ISSUE #731: Table: Unselect all filtered rows
The select all button works to first select all rows, then unselect all rows in a table.

When a filter is applied, the select all button only selects all rows, but there is no way to unselect all rows.

Animation

[Here is a test table](https://codesandbox.io/s/itwinui-react-minimal-example-forked-qwvlr8?file=/src/App.tsx) with selection and filter enabled.

adamhock:
- Clicking the "Select all" checkbox when it is in the indeterminate state now unselects all rows.

- Filtered rows can now be deselected with the "Select all" checkbox.

Closes #731

### DIFF
--- a/packages/iTwinUI-react/src/core/Table/Table.test.tsx
+++ b/packages/iTwinUI-react/src/core/Table/Table.test.tsx
@@ -352,10 +352,10 @@ it('should handle checkbox clicks', async () => {
   expect(onSelect).toHaveBeenCalledWith([mockedData()[1]], expect.any(Object));
 
   await userEvent.click(checkboxCells[0]);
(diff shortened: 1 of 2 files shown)


gretanausedaite (commented):
This change should go to `main` as it solves a bug not related to v1 or v2.

r100-stack (commented):
Code LGTM :) but just a few comments:
* As Greta said here and as Mayank suggested before, I think this PR should point to main
* The word "works" in the PR title was a little confusing to me. Did you maybe mean "rows" or something similar?

adamhock:
A new PR #864 has been made which merges these changes into main. Closing

PR #865, opened 2022-10-07
fix: Add `tippy.js` as a direct dependency

### LINKED ISSUE #856: Display the ComboBox dropdown (from a Dialog) inside AppUI is throwing an exception
We can run the following [sandbox ](https://codesandbox.io/s/itwinui-react-minimal-example-forked-0phivr?file=/src/App.tsx:557-1010) and the ComboBox dropdown works as expected.

But if we copy the identical dialog to ninezone (Design Review), and try to display the ComboBox dropdown, we get the following error:

image

image

We are using:
 "@itwin/itwinui-react": "1.46.0"

MindaugasButkus:
tippyjs version 6.3.1 had some bugs in it that are fixed in version 6.3.7. One of which was clicking on a Dropdown would cause an error because "props.onShow is not a function". Upgrading to 6.3.7 resolves the issue.

Closes # <no issue was created>

## Checklist

- [ ] Add meaningful unit tests for your component (verify that all lines are covered) - Code was not modified
- [ ] Add component features demo in Storybook (different stories) - No new components were added
- [ ] Approve test images for new stories - No new stories were added
- [ ] Add screenshots of the key elements of the component - No components were modified

### DIFF
--- a/packages/iTwinUI-react/package.json
+++ b/packages/iTwinUI-react/package.json
@@ -47,11 +47,12 @@
     "@itwin/itwinui-css": "^0.63.0",
     "@itwin/itwinui-icons-react": "^1.10.1",
     "@itwin/itwinui-illustrations-react": "^1.3.1",
-    "@tippyjs/react": "^4.2.5",
+    "@tippyjs/react": "^4.2.6",
     "@types/react-table": "^7.0.18",
     "classnames": "^2.2.6",
     "react-table": "^7.1.0",
-    "react-transition-group": "^4.4.2"
+    "react-transition-group": "^4.4.2",
+    "tippy.js": "^6.3.7"
   },
   "devDependencies": {
     "@babel/core": "^7.12.10",
@@ -79,7 +80,6 @@
     "markdown-to-jsx": "6.11.4",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
-    "tippy.js": "^6.3.1",
     "ts-jest": "^28.0.2",
     "ts-loader": "^9.2.8",
     "ts-node": "^8.0.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1828,9 +1828,9 @@
     source-map "^0.7.3"
 
 "@popperjs/core@^2.9.0":
-  version "2.11.5"
-  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.11.5.tgz#db5a11bf66bdab39569719555b0f76e138d7bd64"
-  integrity sha512-9X2obfABZuDVLCgPK9aX0a/x4jaOEweTTWE2+9sr0Qqqevj2Uv5XorvusThmc9XGYpS9yI+fhh8RTafBtGposw==
+  version "2.11.6"
+  resolved "https://registry.yarnpkg.com/@popperjs/core/-/core-2.11.6.tgz#cee20bd55e68a1720bdab363ecf0c821ded4cd45"
+  integrity sha512-50/17A98tWUfQ176raKiOGXuYpLyyVMkxxG6oylzL3BPOlA6ADGdK7EYunSa4I064xerltq9TGXs8HmOk5E+vw==
 
 "@rollup/pluginutils@^4.2.1":
   version "4.2.1"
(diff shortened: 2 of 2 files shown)


mayank99 (commented):
Hi @MindaugasButkus, thanks for the PR. Can I ask why you need this change to be made in iTwinUI-react?

We only have a dependency on `@tippyjs/react@^4.2.5`. The `^` means that 4.2.6 is allowed. And it depends on `tippy.js@^6.3.1`, which again means that 6.3.7 is compatible. So if you run `npm install`, it should automatically install those versions as they are the latest. If your lockfile already had the old versions, then you can use [`npm update`](https://docs.npmjs.com/cli/v8/commands/npm-update) on these two packages.

MindaugasButkus:
> Hi @MindaugasButkus, thanks for the PR. Can I ask why you need this change to be made in iTwinUI-react?
> 
> We only have a dependency on `@tippyjs/react@^4.2.5`. The `^` means that 4.2.6 is allowed. And it depends on `tippy.js@^6.3.1`, which again means that 6.3.7 is compatible. So if you run `npm install`, it should automatically install those versions as they are the latest. If your lockfile already had the old versions, then you can use [`npm update`](https://docs.npmjs.com/cli/v8/commands/npm-update) on these two packages.

The problem was that the consumer of this package was using yarn, not npm, and there is no easy way to upgrade indirect dependencies in yarn.

Now, with this change there should be a version bump in this package, so it is a direct dependency update. With this direct dependency update comes the requirement for newer versions of tippy, which I added to `dependencies` list.

mayank99:
> Hi @mayank99. The problem was that the consumer of this package was using yarn, not npm, and there is no easy way to upgrade indirect dependencies in yarn.

There is a [`yarn upgrade`](https://classic.yarnpkg.com/lang/en/docs/cli/upgrade/) command too

> Now, with this change there should be a version bump in this package, so it is a direct dependency update. With this direct dependency update comes the requirement for newer versions of tippy, which I added to `dependencies` list.

But we don't really have a direct runtime dependency on tippy.js, we only use types from there. Theoretically, if `@tippyjs/react` was to remove tippy.js as a dependency in a future minor version, our package should still work.

Am I missing something? @bentleyvk

mayank99 (approved):
If it helps people, let's go with it I guess. I don't see much of a downside, as `tippy.js` will be included one way or another.

adamhock (approved):

packages/iTwinUI-react/package.json
```diff
@@ -47,11 +47,12 @@
     "@itwin/itwinui-css": "^0.63.0",
     "@itwin/itwinui-icons-react": "^1.10.1",
     "@itwin/itwinui-illustrations-react": "^1.3.1",
-    "@tippyjs/react": "^4.2.5",
+    "@tippyjs/react": "^4.2.6",
```
rohan-kadkol, line 50: Just curious, is there a reason (e.g. fixes something) for making this upgrade to tippyjs/react? Or would just upgrading tippy.js (not react) from "6.3.1 -> 6.3.7" fix the problem you were mentioning?

r100-stack (commented):
Just reporting what I saw. I reconfirmed that tippyjs react was using version "^6.3.1", and that `yarn upgrade` does not upgrade indirect dependencies. But I wasn't sure how to confirm if tippyjs's version "6.3.7" fixes the "props.onShow is not a function" you are mentioning.

I also tried looking for uses of tippy.js in our repo, there were very few. Since it is not a major upgrade of the tippy.js (x.x.1 -> x.x.7) and since all our tests pass, I think this update might be alright.

But probably better to wait for a review/approval from Greta or Adam.

packages/iTwinUI-react/package.json
MindaugasButkus: @rohan-kadkol there is some kind of fix for dynamic singleton headless content: https://github.com/atomiks/tippyjs-react/commit/d4c915de38b246d87abfad576887063d0a09d463 . It is unrelated, but I see no reason keeping the outdated version. It's a related package, so why not upgrade?

Regarding tippy.js package, I think the issue was fixed in version 6.3.3 with this commit: https://github.com/atomiks/tippyjs/commit/22481d7655b6eed7884ce84daef52471d526d1dc

MindaugasButkus:
Could anyone tell me what the "Project / Add to project (pull_request)" check does and why did it fail? Can the PR be completed without it passing? If no, how can I rerun it?

gretanausedaite:
> Could anyone tell me what the "Project / Add to project (pull_request)" check does and why did it fail? Can the PR be completed without it passing? If no, how can I rerun it?

This action is for our internal task management and is not required, PR can be merged without this.

> Also, I do not have the required permissions to merge this PR. @gretanausedaite @adamhock Could you merge it?

Sure, doing it now.

Issue #868, opened 2022-10-10
Add `Tile` status and loading state


gretanausedaite:
Add status and loading state to tiles.

PR #872, opened 2022-10-11
feat(Tile): Add status and loading state

gretanausedaite:
#868 

Add: 
- status?: 'positive' | 'warning' | 'negative';
- isLoading?: boolean;


packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -73,6 +73,10 @@ export type TileProps = {
    * Dropdown menu containing `MenuItem`s.
    */
   moreOptions?: React.ReactNode[];
+  /**
+   * Status of the tile.
+   */
+  status?: 'positive' | 'warning' | 'negative';
```
gretanausedaite, line 79: Do we need to add status and loading icons as in css repo?
image

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -150,6 +161,8 @@ export const Tile = (props: TileProps) => {
           'iui-new': isNew,
           'iui-selected': isSelected,
           'iui-actionable': isActionable,
+          [`iui-${status}`]: !!status,
```
gretanausedaite, line 164: I don't think we are adding disabled state to the tiles. Should we add that here?

apps/storybook/src/Tile.stories.tsx
```diff
@@ -285,3 +285,75 @@ Folder.args = {
...
+      moreOptions={moreOptions}
+      status={status}
+      {...rest}
+    />
+  );
+};
+Status.argTypes = {
+  ...Basic.argTypes,
+  thumbnail: { control: { disable: true } },
+};
```
rohan-kadkol, line 314: Can make `status` into a radio button list

```suggestion
Status.argTypes = {
  ...Basic.argTypes,
  thumbnail: { control: { disable: true } },
  status: { control: 'radio', options: ['positive', 'warning', 'negative'] },
};
```

packages/iTwinUI-react/src/core/Tile/Tile.tsx
rohan-kadkol: Yes, I feel so. It gives a 1-1 mapping between this and the CSS repo. Also, it looks better. Also, other aspects of the tile seem to be almost identical to the css examples, so would be better that this is also based off of the css examples.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
rohan-kadkol: If we are adding disabled state to the tiles, we can add it as a prop separate from status. Similar to [L53](https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/Checkbox/Checkbox.tsx#L53) & [L71](https://github.com/iTwin/iTwinUI-react/blob/main/packages/iTwinUI-react/src/core/Checkbox/Checkbox.tsx#L71) in `Checkbox.tsx`

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -73,6 +73,10 @@ export type TileProps = {
    * Dropdown menu containing `MenuItem`s.
    */
   moreOptions?: React.ReactNode[];
+  /**
+   * Status of the tile.
+   */
+  status?: 'positive' | 'warning' | 'negative';
   /**
```
rohan-kadkol, line 80: Just curious, does the order of defining props matter? Because status is on L79 but loading is on L107.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
rohan-kadkol, line 79: Does this PR also aim to implement the status and loading for Folder tiles?

image

rohan-kadkol: Ohh, looks like there already is a variant prop that we can edit in storybook. So maybe we can also make `variant` into a radio button list like this [comment](https://github.com/iTwin/iTwinUI-react/pull/872/files#r1004481871). Since variant is used in the `Folder` story too, might be better to update the global `argTypes` on [L28-32](https://github.com/iTwin/iTwinUI-react/blob/main/apps/storybook/src/Tile.stories.tsx#L28-L32) instead of the local one.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Order should not matter. I just added new props to these lines based on relevance. To me it makes sense to have `status` near other props, like options, thumbnail, etc., and to have `isLoading` near `isActionable`.

apps/storybook/src/Tile.stories.tsx
gretanausedaite: Added

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -73,6 +74,14 @@ export type TileProps = {
...
    */
   moreOptions?: React.ReactNode[];
+  /**
+   * Status of the tile.
+   */
+  status?: 'positive' | 'warning' | 'negative';
+  /**
+   * Custom icon. Will override status and loading icon if specified.
+   */
+  titleIcon?: JSX.Element;
```
gretanausedaite, line 84: Icon prop name could be changed (if needed).

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -141,6 +158,31 @@ export const Tile = (props: TileProps) => {
   const showMenu = React.useCallback(() => setIsMenuVisible(true), []);
   const hideMenu = React.useCallback(() => setIsMenuVisible(false), []);
 
+  const renderTitleIcon = () => {
+    if (!!titleIcon) {
+      return React.cloneElement(titleIcon, {
```
gretanausedaite, line 163: I tried adding
```ts
<span className='iui-tile-status-icon' aria-hidden>
   {titleIcon}
</span>
```

But styling does not work perfectly when titleIcon is ProgressRadial (which is one of the main items to add there).

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Added icons.

gretanausedaite: Added argType to variant in Status, Loading and Custom Loading Icon stories. It made sense to have it only there.

apps/storybook/src/Tile.stories.tsx
```diff
@@ -285,3 +286,118 @@ Folder.args = {
...
+    />
+  );
+};
+Status.argTypes = {
+  ...Basic.argTypes,
+  thumbnail: { control: { disable: true } },
+  status: { control: 'radio', options: ['positive', 'warning', 'negative'] },
+  variant: { control: 'radio', options: ['default', 'folder'] },
+};
+Status.args = {
```
rohan-kadkol, line 318: I noticed that the `isNew` prop is not available to toggle. Might be good to add it to the global `args` on L34

packages/iTwinUI-react/src/core/Tile/Tile.tsx
rohan-kadkol: Actually, looks like `variant` is in the global `args` on [L34](https://github.com/iTwin/iTwinUI-react/blob/main/apps/storybook/src/Tile.stories.tsx#L34). So would be better to add the radio button lines you added like [L316](https://github.com/iTwin/iTwinUI-react/pull/872/files#diff-198e29f0f82eaa85da799271dd3ea518d23c0672cdbabc74f6f41c24e73c5b77R316) to the global `argTypes` on [L28-32](https://github.com/iTwin/iTwinUI-react/blob/main/apps/storybook/src/Tile.stories.tsx#L28-L32) instead of the three local stories

packages/iTwinUI-react/src/core/Tile/Tile.tsx
rohan-kadkol: Seems like an appropriate name.

Some other ideas could be `leadingIcon`, or `startIcon`. This is because they can then be paired with `trailingIcon` or `endIcon` if we ever want to add an icon to the right of the title in the future.

mayank99: i'm not sure about `titleIcon` because we don't even have `title` (it's called `name`).

i'm also not sure if we should add another prop for this. this is another opportunity to create a generic "text with icon" component that i proposed here: https://github.com/iTwin/iTwinUI/issues/586#issuecomment-1098008414

gretanausedaite: So skip adding custom icon for now? Or add this new component to CSS V1 - React v2?

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: We don't have disabled state examples for tile with buttons. When aria-disabled is added - main buttons and thumbnail buttons are still actionable.
Do we leave for our users to add disabled state to these buttons? Do we clone elements and add it ourselves?
@mayank99 @FlyersPh9 

image

packages/iTwinUI-react/src/core/Tile/Tile.tsx
mayank99: I'd say skip it for now and create an issue

packages/iTwinUI-react/src/core/Tile/Tile.tsx
mayank99: The user is in charge of specifying the button so I think they should disable it manually. Maybe we can provide an example and mention it in jsdoc.

gretanausedaite: Disabled button looks a bit strange to me:
image

Especially because thumbnails background is different kind of gray.
image

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Removed prop, added ProgressRadial for loading.

apps/storybook/src/Tile.stories.tsx
gretanausedaite: Updated Tiles args with different boolean controls. Disabled some of them in individual stories.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -141,6 +160,29 @@ export const Tile = (props: TileProps) => {
   const showMenu = React.useCallback(() => setIsMenuVisible(true), []);
   const hideMenu = React.useCallback(() => setIsMenuVisible(false), []);
 
+  const renderTitleIcon = () => {
```
mayank99, line 163: This can be moved into a private component at the bottom of the file, and then we can just use it in the JSX markup like this: `<TitleIcon status={status} isSelected={isSelected} isLoading={isLoading} isNew={isNew} />`.

The benefit is that react will not re-create this function on every render.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -141,6 +160,29 @@ export const Tile = (props: TileProps) => {
...
+      );
+    }
+    if (isSelected) {
+      return <SvgCheckmark className='iui-tile-status-icon' aria-hidden />;
+    }
+    if (isNew) {
+      return <SvgNew className='iui-tile-status-icon' aria-hidden />;
+    }
+    if (!!status) {
+      return StatusIconMap[status]({ className: `iui-tile-status-icon` });
```
mayank99, line 180: nit: This can be "JSX-ify'd" if you assign it to a variable (see example in Alert). Makes it easier to work with.

https://github.com/iTwin/iTwinUI-react/blob/7913c3d0b5418d95988d1c74527e422839f542f1/packages/iTwinUI-react/src/core/Alert/Alert.tsx#L71

https://github.com/iTwin/iTwinUI-react/blob/7913c3d0b5418d95988d1c74527e422839f542f1/packages/iTwinUI-react/src/core/Alert/Alert.tsx#L84

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -96,6 +101,17 @@ export type TileProps = {
...
    */
   isActionable?: boolean;
+  /**
+   * Display a loading state.
+   * @default false
+   */
+  isLoading?: boolean;
+  /**
+   * Flag whether the tile is disabled.
+   * Users should change buttons and icon buttons to disabled ones when this flag is set.
```
mayank99, line 111:
```suggestion
 *
 * Note: This only affects the tile. You need to manually disable
 * the buttons and other interactive elements inside the tile.
 *
```

packages/iTwinUI-react/src/core/Tile/Tile.tsx
```diff
@@ -150,9 +192,12 @@ export const Tile = (props: TileProps) => {
           'iui-new': isNew,
           'iui-selected': isSelected,
           'iui-actionable': isActionable,
+          [`iui-${status}`]: !!status,
+          'iui-loading': isLoading,
         },
         className,
       )}
+      aria-disabled={isDisabled}
       tabIndex={isActionable ? 0 : undefined}
```
mayank99, line 201: Probably should prevent focusing the tile when it's disabled. It's not a good user experience, but it will be consistent with our components.
```suggestion
tabIndex={isActionable && !isDisabled ? 0 : undefined}
```

packages/iTwinUI-react/src/core/Tile/Tile.tsx
mayank99: yeah i'm not a fan of the disabled styling

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Using this method.

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Leaving as is. It's css issue.

mayank99 (approved):

packages/iTwinUI-react/src/core/Tile/Tile.tsx
gretanausedaite: Added new issue https://github.com/iTwin/iTwinUI-react/issues/910

r100-stack (approved):

PR #878, opened 2022-10-13
fix(Table): Select all rows for current filter

### LINKED ISSUE #731: Table: Unselect all filtered rows

adamhock:
Fixes the issue mentioned in [this](https://github.com/iTwin/iTwinUI-react/issues/731#issuecomment-1273687956
) comment regarding selecting all visible rows when a filter is applied

Closes #731


elephantcatdog (approved):
Looks good to me!

mayank99 (approved):

PR #888, opened 2022-10-19
feat(Table): Ctrl + Shift click keeps previous selection & also shift clicks

r100-stack:
**[CONTINUES]** PR #886 (closed as it was pointing to `main` and not `v2`). Pointing this to `v2` to avoid delaying `v2` due to merge conflicts with `main`.

---

Closes #880 

* Shift click keeps previous selection when holding Ctrl when Shift clicking

**NOTE**: The related issue #879 about `lastSelectedRowId` not updating on checkbox click, will come in a separate PR, hopefully from the external contributor. After that PR, this Ctrl + Shift functionality can be completely realized.

#### Steps

* Click `1.1`
* Shift click `1.2.2`
* Ctrl click `2.1` and `2.3`
* Ctrl + Shift click `1.2.4`

#### New

* Previous selection of `2.1` and `2.3` is preserved

[vokoscreenNG-2022-10-18_16-33-25.webm](https://user-images.githubusercontent.com/45748283/196538489-57f3cef7-52a9-4e5f-860a-61a334ebd274.webm)

#### Old

* Previous selection of `2.1` and `2.3` is lost

[vokoscreenNG-2022-10-18_16-41-20.webm](https://user-images.githubusercontent.com/45748283/196540551-67576324-b5a0-4eb1-b173-18e2e92e78fa.webm)

## Checklist

- [x] Add meaningful unit tests for your component (verify that all lines are covered)
- [x] Verify that all existing tests pass
- ~[ ]~ Add component features demo in Storybook (different stories)
- ~[ ]~ Approve test images for new stories
- [x] Add screenshots of the key elements of the component

PR #889, opened 2022-10-19
fix(Table): Ctrl and checkbox clicks update start row of shift selection

### LINKED ISSUE #879: Shift selection ignores any ctrl/checkbox/toggle selections in multi selection mode table
When using multi selection mode table, clicking more than one item (using ctrl or checkbox), and then using shift selection, it ignores any other selected items and takes the first selected item as a starting point for the shift range.
image

#### Expected behavior
Shift should take the last selected item as a starting point for range selection

##### Steps to reproduce
1. Select any row
2. Select a second row using the checkbox or ctrl
3. Use shift to select any row

##### Additional information
Related issue #880

r100-stack:
Closes #879

* Now both Checkbox clicks, and Ctrl clicks update `lastSelectedRowId`, which is the variable storing the starting row of the shift selection.

#### Steps

* Click `1.2.2`
* Checkbox click `1.2.3`
* Shift click `1.4` **(Observation 1)**
* Ctrl click `2.3`
* Shift click `2.1` **(Observation 2)**

#### Old

* **(Observation 1)**: Shift click starts from `1.2.2` (Not `1.2.3`)
* **(Observation 2)**: Shift click starts from `1.2.2` (Not `2.3`)

[vokoscreenNG-2022-10-19_13-31-49.webm](https://user-images.githubusercontent.com/45748283/196763656-44604031-939b-4408-8103-9d8208159bdf.webm)

#### New

* **(Observation 1)**: Shift click starts from `1.2.3`
* **(Observation 2)**: Shift click starts from `2.3`

[vokoscreenNG-2022-10-19_13-33-52.webm](https://user-images.githubusercontent.com/45748283/196764087-2db3decc-6c64-4cb5-bca8-af517e1a7986.webm)

## Checklist


packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,20 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
   onSelect?.(selectedData, newState);
 };
 
+export const onToggleHandler = <T extends Record<string, unknown>>(
```
elephantcatdog, line 59: I don't know a lot about this - why did you make a new function instead of updating `onSelectHandler` to include the `action` property? I see that other functions are using it, but maybe they could be updated too, or it could be optional?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -361,62 +361,100 @@ it('should handle checkbox clicks', async () => {
...
   expect(rows[0].classList).toContain('iui-selected');
   expect(rows[1].classList).toContain('iui-selected');
-  expect(rows[2].classList).not.toContain('iui-selected');
+  expect(rows[2].classList).toContain('iui-selected');
+  expect(rows[3].classList).not.toContain('iui-selected');
   expect(onSelect).toHaveBeenCalledTimes(1);
   expect(onRowClick).toHaveBeenCalledTimes(1);
 
-  await userEvent.click(getByText(mockedData()[1].name)); // Deselect
+  await userEvent.click(getByText(data[1].name)); // Deselect; lastSelectedRowId = undefined -> 1
```
elephantcatdog, line 390: Should lastSelectedRowId be 1 if it's deselected? Previously this comment was on "Reselect"

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -361,62 +361,100 @@ it('should handle checkbox clicks', async () => {
...
+
   expect(rows[0].classList).toContain('iui-selected');
   expect(rows[1].classList).not.toContain('iui-selected');
-  expect(rows[2].classList).not.toContain('iui-selected');
+  expect(rows[2].classList).toContain('iui-selected');
```
elephantcatdog, line 394: If shift is still pressed here and you click on row 1, why is row 2 still selected? When I try this sequence in storybook I get different results - row 0 and 1 are selected and row 2 and 3 aren't.

rohan-kadkol: Ohh! That's true. Looks like `await user.keyboard('[ShiftLeft>]');` ([L380](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R380)) only holds shift for one click? But the [documentation](https://testing-library.com/docs/user-event/keyboard/#:~:text=Keys%20can%20be%20kept%20pressed%20by%20adding%20a%20%3E%20to%20the%20end%20of%20the%20descriptor.) says `>` should keep the key held continuously.

Or maybe I misunderstood. Maybe the docs means that if `>` is provided without a number at the end (e.g. `[ShiftLeft>5]`), then the key is only held for 1 click.

Do you know which is this true?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: Earlier, only selecting updating `lastSelectedRowId`. Now even deselecting does that. So I included this to test this special case.

E.g. After deselecting `1.2.2` and then shift clicking on `1.1`, selection starts from `1.2.2` and not `1.2.4`

[vokoscreenNG-2022-10-21_14-43-58.webm](https://user-images.githubusercontent.com/45748283/197267645-30046366-14d8-451d-863b-c76890a25f33.webm)

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
rohan-kadkol: We want to update `lastSelectedRowId` only on checkbox click and row click, but not on shift click.

Checkbox click calls onSelectHandler() ([L495](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999L495)). So we could update lastSelectedRowId in onSelectHandler(). But the problem is that onShiftSelectHandler() also calls onSelectHandler() ([L101](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-772f25aaf3cd0aa17d5900a4252050c32c48b2efe18b4f6db0d84df08f96cc0bR101)). So we can't update checkbox click in onSelectHandler() since then lastSelectedRowId also updates on shift click.

Thus, checkbox click now calls `onToggleHander()` and updates lastSelectedRowId there. Shift select does not update lastSelectedRowId since onShiftSelectHandler() still calls `onSelectHandler()` and not `onToggleHandler()`

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: I think the second thing. If you want a button held down for multiple keypresses then you need to add a number to signify how many.

rohan-kadkol: **[DONE]** Ohh, I see, thanks. To be safe, I explicitly released the shift before deselecting the row ([L390](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R390)).

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: This feels weird to me. If I deselect something, why would I want to then select it?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: Also you changed this test from 'shift click 2 shift click 1' to 'shift click 2 click 1'

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: That's true. I tried to base it off of other examples. When doing the same in Windows Explorer, it does the same.

Example
* Click `packages`
* Ctrl click `CODE OF CONDUCT`
* Ctrl click `packages`
* Shift click `.vscode`

[Video.webm](https://user-images.githubusercontent.com/45748283/198062006-41dba5f7-2f1b-4c24-904c-6016867e8025.webm)

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: > `{ }` should be used with 'Shift' instead of 'ShiftLeft' like `{Shift>}`

Yes, as per the documentation `{...}` is for [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) and `[...]` is for [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). I saw that other places in the repo were using the `{...}` version, so I changed all `[...]` examples to use that version too, for uniformity.

> Maybe # 1 is why the shift key wasn't being held forever?

I'm not sure, because like you mentioned, both versions don't hold shift continuously for me either.

> Also you changed this test from 'shiftclick 2 shiftclick 1' to 'shiftclick 2 click 1'

I think both the old and current versions are actually doing 'shiftclick 2 click 1'. The old version doesn't have the explicit release shift key like on [L390](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R390), but since `>` on [L378](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9L378) was not holding shift for multiple click like discussed above, that click was also actually a non-shift click.

Or maybe I misunderstood what you meant?

elephantcatdog: > Yes, as per the documentation `{...}` is for [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) and `[...]` is for [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code). I saw that other places in the repo were using the `{...}` version, so I changed all `[...]` examples to use that version too, for uniformity.

Ah, I see

> I'm not sure, because like you mentioned, both versions don't hold shift continuously for me either.

Well that's unfortunate.

> I think both the old and current versions are actually doing 'shiftclick 2 click 1'. The old version doesn't have the explicit release shift key like on [L390](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R390), but since `>` on [L378](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9L378) was not holding shift for multiple click like discussed above, that click was also actually a non-shift click.

> Or maybe I misunderstood what you meant?

It looked like you initially meant for the test to be 'shiftclick 2 shiftclick 1'. If you meant for it to be 'shiftclick 2 click 1' then that's fine. But if you meant for it to be 'sc2 sc1' then you should change it to run like that.

rohan-kadkol: Ohh, thanks for pointing that out. I guess I was just testing `sc c` and not `sc sc` in both cases. So I think it's alright to leave it as it.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
elephantcatdog: Ah, okay

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -361,62 +361,102 @@ it('should handle checkbox clicks', async () => {
...
+  // (Here startIndex = 1, endIndex = 0)
+  await user.keyboard('{Shift>}'); // Hold Shift
+  await user.click(getByText(data[0].name));
+
+  expect(rows[0].classList).toContain('iui-selected');
   expect(rows[1].classList).toContain('iui-selected');
   expect(rows[2].classList).not.toContain('iui-selected');
+  expect(rows[3].classList).not.toContain('iui-selected');
   expect(onSelect).toHaveBeenCalledTimes(3);
   expect(onRowClick).toHaveBeenCalledTimes(3);
```
elephantcatdog, line 412: I'm performing this test on the table story, but it's not doing the same thing as in the tests. So from the beginning of this test I shiftclick #2 -> click #1 -> shiftclick #0 -> controlclick #3 -> shiftclick #4 -> click checkbox#1 -> shiftclick #3. The third action, shiftclick #0 (after deselecting #1), does not select rows #0 & #1, just #0.


https://user-images.githubusercontent.com/37936169/199308008-c9e5bd68-0c3d-49c5-b43b-64a83bf8774c.mp4

rohan-kadkol: Not sure why that's happening. I was able to make it work correctly on my end.

[Video4.webm](https://user-images.githubusercontent.com/45748283/199328257-6c6f6970-e067-4f88-a215-53523238475e.webm)

To confirm, are you testing the `rohan/table-shift-click-checkbox-bug-1` branch? Because there is another branch with a similar name (`rohan/table-ctrl-shift-click-1`)

elephantcatdog: Hm, maybe I need to update it

elephantcatdog: That was it. Everything seems fine now!

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -361,62 +361,102 @@ it('should handle checkbox clicks', async () => {
...
   expect(rows[2].classList).not.toContain('iui-selected');
+  expect(rows[3].classList).toContain('iui-selected');
+  expect(rows[4].classList).toContain('iui-selected');
   expect(onSelect).toHaveBeenCalledTimes(5);
   expect(onRowClick).toHaveBeenCalledTimes(5);
+
+  const checkboxes = container.querySelectorAll<HTMLInputElement>(
+    '.iui-table-body .iui-checkbox',
+  );
+  expect(checkboxes.length).toBe(5);
```
elephantcatdog, line 437: Maybe move this checkbox check to be before or after the series of shift/control/regular clicking? It seems kinda random to be here to me

elephantcatdog (approved):
I had one more comment, but it's a small and unecessary change. Everything looks good!

GintV:
What's the progress on this one? When we could expect this to get released? Will this be included only in 2.x?

cc @ainepi

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
```diff
@@ -56,6 +56,20 @@ export const onSelectHandler = <T extends Record<string, unknown>>(
...
+  action: ActionType,
+  instance?: TableInstance<T>,
+  onSelect?: (
+    selectedData: T[] | undefined,
+    tableState?: TableState<T>,
+  ) => void,
+  isRowDisabled?: (rowData: T) => boolean,
+) => {
+  onSelectHandler(newState, instance, onSelect, isRowDisabled);
+  newState.lastSelectedRowId = action.id;
```
mayank99, line 70: This line is very cryptic so a comment would be useful here.

Also we might want to document somewhere what is the expected flow for the various permutations of multi-selection. Because I would never have guessed that clicking on a checkbox should update "start of shift selection".

mayank99 (approved):

packages/iTwinUI-react/src/core/Table/actionHandlers/selectHandler.ts
mayank99: > Not sure, but is this what you meant?
 
That's not super helpful, I think it should explain *why* this line is needed or what value it adds, rather than simply what it does.

> Would the documentation website be the right place for this? Or are you suggesting somewhere in the code/JSDocs?

Documentation website might work yeah, but I expect the Table page to get very long, so I'm not sure if that's the only place we should mention it.

rohan-kadkol: > That's not super helpful, I think it should explain why this line is needed or what value it adds, rather than simply what it does.

Added comment on [L71](https://github.com/iTwin/iTwinUI-react/pull/889/files#diff-772f25aaf3cd0aa17d5900a4252050c32c48b2efe18b4f6db0d84df08f96cc0bR71-R72)

rohan-kadkol: > Documentation website might work yeah, but I expect the Table page to get very long, so I'm not sure if that's the only place we should mention it.

I tried adding comments to help explain the expected workflow of ctrl/shift selection.

* In this PR, I added some JSDocs for some of the methods in `selectHandler.ts`. [Commits](https://github.com/iTwin/iTwinUI-react/pull/889/files/a52301ad0889ebd3794ee73c6d1d1242de587607..0eaab948add2cd1535f078c1b2dd5d5e270515d7)
* In the related PR [#888](https://github.com/iTwin/iTwinUI-react/pull/888), I added a small comment talking about previous selection handling depending on if it is a shift click or a ctrl + shift click. [Commit](https://github.com/iTwin/iTwinUI-react/pull/888/commits/f37c20902961eb76118aa148f7a0d21f467688c2)

I will also add it in some place in the documentation website soon.

packages/iTwinUI-react/src/core/Table/Table.test.tsx
rohan-kadkol: I had kept it here just to confirm that there are indeed five checkboxes since there are five rows.

But the number of checkboxes would not change after shift/control/regular clicks. So, I think doing this check just once in the current line location is good.

elephantcatdog: Ah, I don't know that I noticed that 'checkboxes' was used down below \/. This is fine

PR #892, opened 2022-10-21
fix(Modal): Apply z-index on dialog wrapper

mayank99:
In itwinui-react 1.48.0, the `Modal` component uses `iui-dialog-wrapper`, which creates a stacking context and prevents children's z-index from being applied correctly. (regression)

So I've added a z-index of 999 to the wrapper, which results in the correct behavior:

I was going to make this change in css and release another 0.X version but that would require more PRs and result in merge conflicts so I've just added a small style tag to the portal container (can be removed later). I used the `:where` selector to keep the specificity to a minimum, because some of our users want to override the z-index.

### DIFF
--- a/packages/iTwinUI-react/src/core/Modal/Modal.test.tsx
+++ b/packages/iTwinUI-react/src/core/Modal/Modal.test.tsx
@@ -200,12 +200,12 @@ it('should work with portal container properly', () => {
 
   let container = document.querySelector('body > #test-id') as HTMLElement;
   expect(container).toBeTruthy();
-  expect(container.children.length).toBe(1);
+  expect(container.children.length).toBe(2);
 
   renderComponent({ modalRootId: 'test-id' });
   container = document.querySelector('body > #test-id') as HTMLElement;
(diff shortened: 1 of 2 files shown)


gretanausedaite (commented):
I think you need to update tests. Now modal root has minimum 2 children. 
image

gretanausedaite (approved):
Should be good for now.
Are you considering adding "proper" css in v1.0?

r100-stack (approved):
A thought came to maybe test near [L203](https://github.com/iTwin/iTwinUI-react/pull/892/files#diff-92d941ca9af6ee9515887b5d23918e2e2bb23267a8e48189ee84d14e2c1ebadcR203) that the `<style>` tag exists. But since this `<style>` tag is a temporary addition, I don't think it would be needed. LGTM

mayank99:
> Should be good for now. Are you considering adding "proper" css in v1.0?

Yeah, we can do that. This PR is just to get the fix out for people who are still on css 0.X and react 1.x.

PR #901, opened 2022-10-28
fix: Append popover elements to scoped root

### LINKED ISSUE #898: v2: Portaled elements don't use scoped theme by default
Follow up to #813 which was added in #825. I was testing out the v2 branch and noticed that when a `ComboBox` is used inside a `ThemeProvider` deep inside the app tree, the dropdown menu (popover) doesn't inherit the theme and instead uses the default theme.

<details>
<summary>Code to reproduce</summary>
   
```tsx
export default () => {
  return (
    <ThemeProvider theme='dark'>
      <ComboBox options={options} />
    </ThemeProvider>
  );
}
const options = [
  { label: 'Apple', value: 'apple' },
  { label: 'Banana', value: 'banana' },
  { label: 'Lychee', value: 'lychee' },
];
```
</details>

I see two ways of fixing this:
1. append the popover to the nearest scoped root by default. OR
2. keep appending the popover to body, and create a duplicate scoped root around it.

mayank99:
Fixes #898 for popovers. Modals and toasts will still respect the *topmost* ThemeProvider but not any scoped roots.

#### Main changes

- **ThemeProvider**: Now saves `rootRef`.
- **Popover**: Access `rootRef` from theme context and uses it for `appendTo`.
- **getContainer**: Now tries to add container to `.iui-root` if one exists.
- **Toast**: Removed `useTheme` because it unnecessarily runs a layout effect outside the main tree, adding redundant classes to `<body>`.


mayank99:
I have decided to leave out modals/toasts from this PR because I think those are less important from a scoping perspective (they always appear in the same place, regardless of what part of the app is scoped).

In a future PR I plan to introduce two new hooks:
- `useRoot` which will return the nearest `ThemeProvider` element (and can then be passed into Modal or Toast).
- `useToaster` which will automatically put toasts in a scoped root.

gretanausedaite (approved):
LGTM 🌵

Issue #849, opened 2022-09-30
v2 release


mayank99:
- [ ] https://github.com/iTwin/iTwinUI-react/issues/153
- [x] https://github.com/iTwin/iTwinUI-react/issues/727
- [ ] #912
- [x] https://github.com/iTwin/iTwinUI-react/issues/843
- [ ] https://github.com/iTwin/iTwinUI-react/issues/898
- [ ] Rename components that have changed in css v1 (deprecate old ones and point them to new)
	- [x] #902
	- [x] #905
- [ ] Deprecate (but not remove) the following components: #858
	- [x] #908
	- [x] #913

Issue #367, opened 2021-10-11
add window control options to header for frameless-window electron applications


MichaelBelousov:
#### Feature

add items to the right of the `userIcon` spot in the `Header` component,
particularly for window controls in frameless-window electron applications that use their header as the frame.

Example:
![MicrosoftTeams-image (3)](https://user-images.githubusercontent.com/16315257/136790289-93a69162-2caa-47c0-ae13-832bce79c919.png)

So the header would need some new prop like `windowControls`, similar in type to the children prop which get placed in the center currently.

It may also be of important to make sure that the header has good drag-ability in this mode. See the documentation here:
https://www.electronjs.org/docs/api/frameless-window#draggable-region
I have tested adding the following style to my header which seemed to do the trick:

```scss
.myHeaderClass {
  -webkit-app-region: drag;
  > * {
    -webkit-app-region: none;
  }
}
```

MichaelBelousov:
I'll try that out, can a note be added about that for other people that try to use the `userIcon` prop but want it moved?

mayank99:
`userIcon` has been deprecated in [v2](https://github.com/iTwin/iTwinUI-react/releases/tag/v2.0.0)

PR #902, opened 2022-10-31
feat(Avatar, AvatarGroup): Deprecate `UserIcon`/`UserIconGroup`, replace with `Avatar`/`AvatarGroup`

r100-stack:
> - Rename components that have changed in css v1
> 	- e.g. Avatar (**Rohan**), Stepper
> 	- Deprecate old names while pointing them to new ones

_From https://github.com/iTwin/iTwinUI-react/issues/849_

---

* Deprecated `UserIcon` and `UserIconGroup`
* Added `Avatar` and `AvatarGroup` (replacements)

## Checklist


packages/iTwinUI-react/src/core/UserIcon/UserIcon.tsx
```diff
@@ -52,6 +64,8 @@ export const defaultStatusTitles: StatusTitles = {
 };
 
 /**
+ * @deprecated Since v2, this has been renamed to `Avatar`
```
mayank99, line 67: This is not the way to do it, because now the component code is duplicated for no reason.

Here's how I would do it:
0. Rename all instances of `UserIcon` and `UserIconGroup` to `Avatar` and `AvatarGroup`.
1. Delete every trace of `UserIcon` and `UserIconGroup`. This includes test files and test images.
2. In `Avatar`, add `export const UserIcon = Avatar`. In `AvatarGroup`, add `export const AvatarGroup = UserIconGroup`. Same thing with props. And then add `@deperacted` JSDoc above these.
3. In the barrel file, export all four components and their props.

rohan-kadkol: **[DONE]** Good idea, thanks.

To confirm, is the old JSDocs needed for `UserIcon` and `UserIconGroup`? Or is the `@deprecation` tag enough? ([L128](https://github.com/iTwin/iTwinUI-react/pull/902/files#diff-316029e888ce93252e0623c6960c4df4d157b8aa6403db243a65a030da39345aR128) and [L144](https://github.com/iTwin/iTwinUI-react/pull/902/files#diff-b55ac7cab532d4e79274a4b578b5348be2369f627f66d51da118b7a3ef3ec2c9R144)) I felt the `@deprecation` tag was enough since users should be encouraged to not use our deprecated components/props.

packages/iTwinUI-react/src/core/Header/Header.tsx
```diff
@@ -68,6 +71,20 @@ export type HeaderProps = {
...
+   * Header
+   * (expects `Avatar`, can be wrapped in `IconButton` and `DropdownMenu` if needed)
+   * @example
+   * <DropdownMenu menuItems={...}>
+   *   <IconButton styleType='borderless'>
+   *     <Avatar ... />
+   *   </IconButton>
+   * </DropdownMenu>
+   */
+  avatar?: React.ReactNode;
```
mayank99, line 87: 1. I don't think changes to the `Header` API should be part of this PR
2. I don't think we need a new `avatar` prop. We can just deprecate `userIcon` and suggest using `actions` instead. See #367

packages/iTwinUI-react/src/core/UserIcon/UserIcon.tsx
mayank99: deprecation should be enough

packages/iTwinUI-react/src/core/Header/Header.tsx
rohan-kadkol: **[DONE]** Removed `avatar`.

I think `userIcon` might have been there to enforce the `UserIcon` to always be to the left of `menuItems`. Without it, it can be anywhere in the `actions`.

But even deprecating it makes sense. Since we have a `Header` stories to show where we recommend to place `Avatar` in `actions`. It is also alright since, like you said, there was nothing special being done specifically to the `userIcon`.

mayank99: Just to be clear: you need to bring back `userIcon`. Deprecating does not mean removing it (you have removed it from the header code).

rohan-kadkol: Ohh, so deprecate means it _should work even if users use it_, but we do not recommend it as we won't maintain it and it also might be removed in the future?

So deprecate doesn't mean that it won't work anymore so you _must_ change to what we recommend?

packages/iTwinUI-react/src/core/AvatarGroup/AvatarGroup.tsx
```diff
@@ -24,55 +24,61 @@ export type UserIconGroupProps = {
...
   const {
     children,
     animated = false,
     stacked = true,
-    maxIcons = 5,
-    iconSize = 'small',
-    countIconProps,
+    maxAvatars = 5,
+    avatarSize = 'small',
+    countAvatarProps,
```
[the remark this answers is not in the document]
elephantcatdog: Why does `AvatarGroup` need to be backwards compatible with `UserIcon` when we still have `UserIconGroup`?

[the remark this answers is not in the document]
rohan-kadkol: > Why does `AvatarGroup` need to be backwards compatible with `UserIcon` when we still have `UserIconGroup`?

I think Mayank explained it here: https://github.com/iTwin/iTwinUI-react/pull/902#discussion_r1010713129. I guess we want to have as less breaking changes as possible to encourage users to upgrade to v2. So all users have to do is rename `UserIconGroup` -> `AvatarGroup` and `UserIcon` -> `Avatar` and everything should work.

But I think, we would want to make this breaking change of new prop names at some point. And according to [semver](https://semver.org/#summary), `MAJOR` version is for breaking changes like this and `MINOR` for backward compatible changes.

image

So we should include these changes in v2? Or maybe I misunderstood how semver works?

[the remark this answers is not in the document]
rohan-kadkol: > These prop names need to remain the same as before to keep backwards compatibility with `UserIcon`.

I reverted those changes for now, like you suggested. I can re-add it if we decide to do that.

packages/iTwinUI-react/src/core/Header/Header.tsx
rohan-kadkol: > Just to be clear: you need to bring back `userIcon`.

Brought `userIcon` back to `Header`. Also added some JSDocs to `actions` to recommend users to add `Avatar` to the end of the `actions` array ([L43-46](https://github.com/iTwin/iTwinUI-react/pull/902/commits/4120bb19e67d3b48175032a715d4bdfd064bbf42#diff-c792e3ac6e0ca0e4361d7efffdd239f466a22831e373546c4e9fd52c786e37ebR43-R46)).

packages/iTwinUI-react/src/core/AvatarGroup/AvatarGroup.tsx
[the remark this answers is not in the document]
mayank99: When I opened the original issue, I laid out the plan as: deprecate in v2, remove in v3.

A deprecation doesn't need to happen in a major version, it's just convenient to do it now because the components were renamed in css v1.

mayank99 (approved):

adamhock (approved):

PR #905, opened 2022-11-01
feat(Wizard): Deprecate `Wizard` to use `Stepper`/`WorkflowDiagram`

r100-stack:
> - Rename components that have changed in css v1
> 	- e.g. Avatar (**Rohan** PR [#902](https://github.com/iTwin/iTwinUI-react/pull/902)), Stepper (**Rohan**)
> 	- Deprecate old names while pointing them to new ones

_From https://github.com/iTwin/iTwinUI-react/issues/849_

---

* Deprecate `Wizard` but not delete it
  * Did not delete it since this is not a simple rename like `UserIcon` -> `Avatar` (PR [#902](https://github.com/iTwin/iTwinUI-react/pull/902)).
  * Instead, we are now deprecating a ternary operator type component wrapper that returns `Stepper` when Wizard.type = `default` | `long`, and `WorkflowDiagram` when Wizard.type = `workflow`
  * Thus, deleting Wizard's code and setting `Wizard = Stepper` is not possible
* Add stories and visual tests for `Stepper` and `WorkflowDiagram`

## Checklist


gretanausedaite (approved):
No comments, looks good!

apps/storybook/src/WorkflowDiagram.test.ts
```diff
@@ -0,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+describe('WorkflowDiagram', () => {
```
elephantcatdog, line 5: Is WorkflowDiagram a completely separate component now?

packages/iTwinUI-react/src/core/Wizard/Stepper.tsx
```diff
@@ -32,7 +32,7 @@ export type StepperProps = {
    */
   steps: StepProperties[];
   /**
-   *  The type of Wizard to display.
+   *  The type of Stepper to display.
```
elephantcatdog, line 35: You didn't change the name of the parent file from 'Wizard' to 'Stepper'.

elephantcatdog (commented):
It would be easier to tell if the tests have changed if they're 'renamed' vs deleted and created again.

apps/storybook/src/WorkflowDiagram.test.ts
rohan-kadkol: `WorkflowDiagram` actually already existed. But `Wizard` was a wrapper around `WorkflowDiagram`.

So people earlier had to do `<Wizard type:'workflow''/>` to get `WorkflowDiagram`. Now they can directly use `WorkflowDiagram`

You can see [L59-70](https://github.com/iTwin/iTwinUI-react/pull/905/files#diff-31a31bcd2c19607b8a9dfdfdd6da3ac2990b7dfd28d59ec85314055793880fa8R59-R70) to get a better idea.

packages/iTwinUI-react/src/core/Wizard/Stepper.tsx
rohan-kadkol: Yes, I just deprecated Wizard but left `Wizard.tsx` as it was because of the following:

image

Also, Stepper actually already existed, but it was just being wrapped with Wizard ([L59-70](https://github.com/iTwin/iTwinUI-react/pull/905/files#diff-31a31bcd2c19607b8a9dfdfdd6da3ac2990b7dfd28d59ec85314055793880fa8R59-R70)).

Or maybe I misunderstood what you meant?

rohan-kadkol: Just to mention, I changed the highlighted L35 because Wizard is now deprecated. So I thought the JSDocs in `Stepper.tsx` should also not use the word "Wizard".

But `Stepper.tsx` is a file that already existed and I left it unchanged apart from L35.

elephantcatdog: I meant 
image

apps/storybook/src/WorkflowDiagram.test.ts
elephantcatdog: I guess my question is... do you know why we moved WorkflowDiagram out from under Wizard (now Stepper)? 
Why don't we do `<Stepper type="workflow"/>`?

packages/iTwinUI-react/src/core/Wizard/Stepper.tsx
rohan-kadkol: Ah, nice catch! Renamed `src/core/Wizard` to `src/core/Stepper`

apps/storybook/src/WorkflowDiagram.test.ts
rohan-kadkol: I thought since we are deprecating `Wizard`, I could expose `Stepper` and `WorkflowDiagram`. But maybe I could somehow combine `Wizard` into `Stepper` to make it just like a rename change `Wizard` -> `Stepper` on the users end (since that might be helpful for users to easily update to v2)

Will confirm which approach to take.

elephantcatdog: I saw the discussion, makes sense to me. Just wanted to check.

elephantcatdog (commented):
Did you run the visual tests and then clean them after you updated the images?

rohan-kadkol:
> It would be easier to tell if the tests have changed if they're 'renamed' vs deleted and created again.

Good idea. Changed the test image git changes to "rename" type changes instead of "delete and add" changes. I then re-ran `yarn test` to catch any image changes.

> Did you run the visual tests and then clean them after you updated the images?

Yes, I ran `yarn test`. The old `Wizard` test images are removed (renamed to `Stepper`/`WorkflowDiagram`)

elephantcatdog:
> Good idea. Changed the test image git changes to "rename" type changes instead of "delete and add" changes. I then re-ran yarn test to catch any image changes.

Thank you, this is great!

> Yes, I ran yarn test. The old Wizard test images are removed (renamed to Stepper/WorkflowDiagram)

Hm, I was pretty sure that you have to run `yarn clean`?
image

I'm mostly asking because I don't see a removed file for `Stepper.test.ts-WorkflowDiagram.png`.

elephantcatdog (approved):

PR #911, opened 2022-11-03
feat(Table): First level header now respected

### LINKED ISSUE #153: Table: columns first level Header is not respected.
In order to get headers to be displayed in the table, we must actually create a root group (which is never displayed)
It is a bit unexpected behavior.

#### Environment
- __Package version(s)__: 1.7.2
- __Operating System__: Windows 10
- __Browser name and version__: Chrome, 91.0.4472.114

#### Code Sandbox
Link to a minimal repro: https://codesandbox.io/s/shy-wind-1sp5q?file=/src/App.js
This pits react-table example against itwinUi table, click on the Toggle Header Level to add/remove the grouping, the same data is fed to their example and iui table, notice that, the groupings are off.

#### Steps to reproduce
Please see the repo, compare the columns definition between grouped and not grouped and the behavior.

#### Actual behavior
In the sandbox, I get either "Ungrouped" headers, or no headers.

#### Expected behavior
In the sandbox, I expect a grouping level, and an "Ungrouped" header.

#### Additional information
There is a `.splice(1)` in the Table header code that seems to cause this issue, not sure why it is there.

adamhock:
`Table` components no longer have to specify a `Header`. The following basic `Table` structure has changed

Before: 
image

After:
image

Tables that do specify a `Header` will have that table header displayed on top of the column headers like so
image

Fixes #153 

TODO

- [x] Fix unit tests
- [x] Verify visual tests
- [ ] Test group header scenarios


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -824,23 +818,29 @@ export const Table = <
...
-        >
-          <div className='iui-table-header'>
-            {headerGroups.slice(1).map((headerGroup: HeaderGroup<T>) => {
-              const headerGroupProps = headerGroup.getHeaderGroupProps({
-                className: 'iui-table-row',
-              });
-              return (
-                <div {...headerGroupProps} key={headerGroupProps.key}>
+        {headerGroups.map((headerGroup: HeaderGroup<T>) => {
+          headerGroup.headers = headerGroup.headers.filter(
```
adamhock, line 822: This filter removes unnecessary headers that are being placed when the `Header` is specified for tables with multi-select or expandable rows. This shouldn't be the permanent solution because users who put their header ids as either of these strings will not have their header shown

TableRowHeader

elephantcatdog: Maybe make a comment about needing a better solution here so we won't forget, or make another issue?

gretanausedaite (commented):
Have you tested grouped headers scenario described in this issue?

elephantcatdog (commented):
I hate to ask this because we're on a time crunch, but I can't tell that this works to fix the problem with stacked/grouped headers. Can you make a storybook example and maybe a unit test?

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -1562,24 +1450,19 @@ it('should disable global filter column with disableGlobalFilter', async () => {
 });
 
 it('should not global filter with manualGlobalFilter', async () => {
-  const mockedColumns = [
+  const mockedColumns: Column<TestDataType>[] = [
```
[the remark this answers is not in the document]
adamhock: Some of the unit tests weren't building and others were. It seems very inconsistent and adding this seemed to work for the ones that weren't building. The suggestion came from @bentleyvk

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: Comment has been added

adamhock:
> I hate to ask this because we're on a time crunch, but I can't tell that this works to fix the problem with stacked/grouped headers. Can you make a storybook example and maybe a unit test?

A Table Header story has been added (if there are any other name suggestions for this story, feel free to chime in!).

Unit tests have also been added

gretanausedaite:
> I haven't yet. I'll update PR with TODO
I added double headers to story to check that: 
image
So I think that should be good (maybe we need to think about styling)

gretanausedaite:
> > A Table Header story has been added (if there are any other name suggestions for this story, feel free to chime in!).
> 
> Stacked Header? Grouped Header? Detailed Header? Custom Header?

'Grouped header' or 'header groups' sounds good to me.

bentleyvk:
You can take inspiration for the story from the `react-table` examples: https://react-table-v7.tanstack.com/docs/examples/grouping

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -828,23 +822,30 @@ export const Table = <
...
+            (header) =>
+              !header.id.includes('iui-table-checkbox-selector_placeholder') &&
+              !header.id.includes('iui-table-expander_placeholder'),
+          );
+          const headerGroupProps = headerGroup.getHeaderGroupProps({
+            className: 'iui-table-row',
+          });
+          return (
+            <div
+              className='iui-table-header-wrapper'
```
bentleyvk, line 837: This might not work with grouped headers and horzontal scroll or sticky columns. I think wrapper must go above all headers

packages/iTwinUI-react/src/core/Typography/Text/Text.tsx
```diff
@@ -37,9 +37,8 @@ type TextOwnProps = {
   isSkeleton?: boolean;
 };
 
-export type TextProps<
-  T extends React.ElementType = 'div'
-> = PolymorphicComponentProps<T, TextOwnProps>;
+export type TextProps<T extends React.ElementType = 'div'> =
```
bentleyvk, line 40: Huh?

adamhock: A lot of the recent PRs have been having formatting changes such as this one

adamhock:
I ended up removing the Grouped Header story and test since the feature isn't entirely flushed out and the goal for this PR is just to fix the first level header bug

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -947,10 +948,11 @@ export const Table = <
...
                 </div>
-              );
-            })}
-          </div>
-        </div>
+              </div>
+            </div>
+          );
+        })}
+
```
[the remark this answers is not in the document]
adamhock: removed

mayank99:
@adamhock What's the status of this PR? Is your PR tackling both parts of issue? If not, what is your plan to move forward with it?

Also if you're not able to revert the large number of file changes in the diff, I would suggest opening a new PR.

adamhock:
> @adamhock What's the status of this PR? Is your PR tackling both parts of issue? If not, what is your plan to move forward with it?
> 
> Also if you're not able to revert the large number of file changes in the diff, I would suggest opening a new PR.

My PR is tackling both parts of the issue. First level `Header` is no longer required, and when users put a first level `Header` in their Tables, a table header is shown instead of being ignored.

There's an open comment about Grouped Headers not working for Horizontal Scrolling and Sticky Columns, but I'm not sure if that should just be fixed in another PR so that this one can get merged to fix the major bug.

As for the image changes, I'm stuck on how to revert them, so I will just open a new PR

adamhock:
Closing this PR and reopening in #935

Issue #60, opened 2021-05-13
Margin Prop for Typography Components


CadenAustin:
#### Feature
At the moment to strip the bottom margin from typography you have to override it using css, I think it would be nice to have a prop for typography components that determines if margin is applied to the text.

#### Examples
A Boolean prop like:
`<Header useMargin={true/false} />` 
or maybe a prop to set your own margin value:
`<Header marginSize={value} />`

mayank99:
Not sure I understand the benefit of a margin prop.

You can already do this:
```jsx
<Headline style={{ margin: 0 }} />
```

PR #914, opened 2022-11-07
feat(Typography): Deprecate Headline, Leading, Small, Subheading, Title, Body

### LINKED ISSUE #913: v2: Deprecate Headline, Leading, Small, Subheading, Title, Body (all superseded by `Text`)

gretanausedaite:
Solves #913


apps/storybook/src/Text.stories.tsx
```diff
@@ -28,8 +32,18 @@ const TextStory: Story<TextProps> = (args) => {
...
+export const Small: Story<TextProps> = TextStory.bind({});
+Small.args = { variant: 'small' };
+export const Leading: Story<TextProps> = TextStory.bind({});
+Leading.args = { variant: 'leading' };
+export const Subheading: Story<TextProps> = TextStory.bind({});
+Subheading.args = { variant: 'subheading' };
+export const Title: Story<TextProps> = TextStory.bind({});
+Title.args = { variant: 'title' };
+export const Headline: Story<TextProps> = TextStory.bind({});
+Headline.args = { variant: 'headline' };
```
mayank99, line 46: These examples need to also carry over the `as` prop to be the same as the element that was previously being rendered by these individual components. e.g. `<Headline>` becomes `<Text variant='headline' as ='h1'>`.

Also storybook is showing errors for these.

mayank99: `<Small>` should use `as: 'small'`

and this is the error in visual tests:

gretanausedaite: Updated. Should `leading` be h4 or h3? `Leading` component is h3 but jsdoc in `Text` says h4

mayank99: Oh wow, now I don't know if the jsdoc is wrong or if the component is wrong. See, this is why I think it was a good idea to deprecate these old components! They abstract the heading levels to the point where it will hurt accessibility.

I would say leave the component as `h3`. This will make it so that we don't break heading levels for existing users.

I'm also thinking we should emphasize that the user should decide what to put in the `as` prop depending on the context of their app, rather than mindlessly replacing `<Leading>` with `<Text variant='leading' as='h4'>`.

gretanausedaite: Changed h4 to h3.

> I'm also thinking we should emphasize that the user should decide what to put in the `as` prop depending on the context of their app, rather than mindlessly replacing `<Leading>` with `<Text variant='leading' as='h4'>`.

Do you have in mind jsdoc or maybe doc website? I think jsdoc might be too limiting when explaining users how to use polymorphic components.

mayank99: > And what about margin, which was always there with Typography components, but was not with `Text` ? (sorry if it is added now)

I think spacing should also be controlled by the user. Having margins built into the components usually hurts more than helps. We've had complaints in the past about this margin: https://github.com/iTwin/iTwinUI-react/issues/60

> Changed h4 to h3.

I think we could keep the story as h4, since it is a new addition that does not need backwards compatibility.

> Do you have in mind jsdoc or maybe doc website? I think jsdoc might be too limiting when explaining users how to use polymorphic components.

Maybe we can add one sentence in the jsdoc and link to a resource about heading levels? Docs site is not published yet, and we want this deprecation to be released as part of v2.

veekeys: Right, but thinkin from the breaking changes and how everyone relied on this one, it will be huge.
Plus, having our margins set, we can guarantee consistency between apps. 
All the heading elements having margin makes sense to me (you always need some kind of a spacing between that and content).
Plus, if we still follow the native elements and keeping ours as close as those, native `h` elements do have margins which we do reset because of our css. So I think if we just deliver `Text` element and actually reset the native elements margin, we do more bad than good.
IMO
@FlyersPh9 anything on this one from the UI standards perspective? Or we have fully removed all spacings from our typography elements in css too?

mayank99: > Right, but thinkin from the breaking changes and how everyone relied on this one, it will be huge.

That's why we are not removing the old components completely. They can continue using the deprecated components.

> Plus, having our margins set, we can guarantee consistency between apps.
All the heading elements having margin makes sense to me (you always need some kind of a spacing between that and content).

We don't know where exactly each heading will be used. The same heading can be used in two completely different contexts where they need different spacing.

And even if they need the same spacing, the dev might want to apply that spacing through flex/grid `gap` rather than margin.

---

Here's an example where `Subheading` is used without spacing:


Here's an example where `Title` is used without spacing:

FlyersPh9: Mayank captured 2 good examples of when we'd not want to include that spacing.  The spacing is useful for when building pages like a blog post or documentation page, but the spacing gets in the way when trying to build parts of a component.

rohan-kadkol: Might not be a good idea, but what if we provide some type of prop for toggling whether to use default text margins or not in the `ThemeProvider`? Just a thought, not sure if it is good, maybe overkill.

That way devs can use the default margins (we provide consistency across iTwinUI websites) or override the behavior in the ThemeProvider just _once_ to handle _all_ the text spacing themselves (we also provide easy customizability).

> we are not removing the old components completely. They can continue using the deprecated components.

Also, that way users don't need to use deprecated Text components to use our default text spacing values. Because we may remove the deprecated Text components in v3.

packages/iTwinUI-react/src/core/Typography/Text/Text.tsx
```diff
@@ -48,10 +47,12 @@ type TextComponent = PolymorphicForwardRefComponent<'div', TextOwnProps>;
  * @example
  * <Text>Some normal paragraph text rendered as a div</Text>
  * @example
- * <Text variant='title' as='h3'>I'm some h2 text rendered as an h3 element!</Text>
- * @example
  * <Text isMuted>Some muted text.</Text>
  * <Text isSkeleton>Skeleton text</Text>
+ *
+ * Users should decide which element to render based on the context of their app. Link to heading levels docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements
```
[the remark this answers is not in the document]
mayank99: can you move this above the examples and change the link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#accessibility_concerns ?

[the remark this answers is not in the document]
gretanausedaite: Updated

mayank99 (approved):

adamhock (approved):
Aside from my one comment, the code looks good to me 🎉

packages/iTwinUI-react/src/core/Typography/Small/Small.tsx
```diff
@@ -16,10 +16,9 @@ export type SmallProps = {
 } & React.HTMLAttributes<HTMLParagraphElement>;
 
 /**
- * Small text, renders a paragraph element
```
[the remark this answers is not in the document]
gretanausedaite: Variants are explained in Text.tsx line 16.

PR #919, opened 2022-11-09
fix(Header): Change `iui-slim` to `data-iui-size: slim`

gretanausedaite:
- Table image: fix from previous merge
- Add Slim header story
- Change `iui-slim` to `data-iui-size: slim`


mayank99 (approved):
Is Table image change expected?

apps/storybook/src/Header.stories.tsx
```diff
@@ -329,3 +345,7 @@ export const CenterContent: Story<HeaderProps> = (args) => {
     </Header>
   );
 };
+
+CenterContent.args = {
+  isSlim: false,
```
rohan-kadkol, line 350: Since `isSlim` is set to `false` in the parent args on [L51](https://github.com/iTwin/iTwinUI-react/blob/b2f08e9742a3494475d9afae587826e6cf0d97a2/apps/storybook/src/Header.stories.tsx#L51), I think we can leave out setting it to `false` in `<Story>.args` for all stories stories except `Slim`. Only for the `Slim` story, we need to set it to `true`.

gretanausedaite:
> Is Table image change expected?

Image tests are failing after adding Initial state table story.

elephantcatdog (approved):
Looks good to me

r100-stack (approved):

PR #923, opened 2022-11-09
feat(Alert): Use `data-iui-status`

adamhock:
- `class="iui-alert-(status)` -> `class="iui-alert" data-iui-status="(status)"` where `(status)` can be `informational`, `positive`, `negative`, or `warning`.
- `class="iui-alert iui-sticky"` -> `class="iui-alert" data-iui-sticky="true"`

These are the changes in React as a result of [this](https://github.com/iTwin/iTwinUI/pull/803) Alert PR in CSS

This PR will come out of draft once the CSS PR is merged


packages/iTwinUI-react/src/core/Alert/Alert.test.tsx
```diff
@@ -63,24 +61,27 @@ it('renders sticky alert correctly', () => {
...
 });
 
-(['informational', 'positive', 'warning', 'negative'] as Array<
-  'informational' | 'positive' | 'warning' | 'negative'
->).forEach((type) => {
+(
+  ['informational', 'positive', 'warning', 'negative'] as Array<
+    'informational' | 'positive' | 'warning' | 'negative'
+  >
+).forEach((type) => {
```
rohan-kadkol, line 74: optional: Can revert this formatting change. Or maybe better to keep it since it is a small PR, so better to do formatting changes along with this PR.

r100-stack (approved):

packages/iTwinUI-react/src/core/Alert/Alert.test.tsx
mayank99: Every new PR seems to have formatting issues. I think this could potentially be because the `website` uses a different version of `prettier` which is taking preference. Feel free to look into it if you have time

mayank99 (approved):

mayank99:
@adamhock I don't get it. Why did you merge this before updating to the latest css that I specifically released for your PR? The visual tests are now failing in CI: https://github.com/iTwin/iTwinUI-react/actions/runs/3440659055/jobs/5739369985

PR #935, opened 2022-11-11
fix(Table): First level header is no longer required

### LINKED ISSUE #153: Table: columns first level Header is not respected.

adamhock:
Continuation from #911 

Fixes #153

### DIFF
--- a/apps/storybook/src/Table.stories.tsx
+++ b/apps/storybook/src/Table.stories.tsx
@@ -109,29 +109,24 @@ export const Basic: Story<Partial<TableProps>> = (args) => {
   const columns = useMemo(
     () => [
       {
-        Header: 'Table',
-        columns: [
-          {
-            id: 'name',
-            Header: 'Name',
-            accessor: 'name',
-          },
-          {
-            id: 'description',
-            Header: 'Description',
-            accessor: 'description',
-            maxWidth: 200,
-          },
-          {
-            id: 'click-me',
-            Header: 'Click',
-            width: 100,
-            Cell: (props: CellProps<{ name: string; description: string }>) => {
-              const onClick = () => onClickHandler(props);
-              return <Anchor onClick={onClick}>Click me!</Anchor>;
-            },
-          },
-        ],
+        id: 'name',
+        Header: 'Name',
+        accessor: 'name',
+      },
+      {
+        id: 'description',
+        Header: 'Description',
+        accessor: 'description',
+        maxWidth: 200,
+      },
+      {
+        id: 'click-me',
+        Header: 'Click',
+        width: 100,
+        Cell: (props: CellProps<{ name: string; description: string }>) => {
+          const onClick = () => onClickHandler(props);
+          return <Anchor onClick={onClick}>Click me!</Anchor>;
+        },
       },
     ],
     [],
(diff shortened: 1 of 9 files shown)


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -828,23 +822,30 @@ export const Table = <
...
+            (header) =>
+              !header.id.includes('iui-table-checkbox-selector_placeholder') &&
+              !header.id.includes('iui-table-expander_placeholder'),
+          );
+          const headerGroupProps = headerGroup.getHeaderGroupProps({
+            className: 'iui-table-row',
+          });
+          return (
+            <div
+              className='iui-table-header-wrapper'
```
mayank99, line 837: Shouldn't this `header-wrapper` wrap all groups?

mayank99 (commented):
PR title should be renamed if this is not adding full support for grouped headers.

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: This is what Horiztonal Scroll looks like when the `Header` is specified now
image

After wrapping all of the header groups in `iui-table-header-wrapper`, the result looks like this
TableRowHeaderHorizontalScroll

packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -3932,21 +3655,16 @@ it('should add expander column manually', () => {
...
+      Header: 'Name',
+      accessor: 'name',
+    },
+    {
+      id: 'description',
+      Header: 'Description',
+      accessor: 'description',
     },
   ];
   const { container } = renderComponent({
```
elephantcatdog, line 3670: Lines 3698 and 3732 need to be updated to remove the `Header: 'Table'` thing

packages/iTwinUI-react/src/core/Table/Table.tsx
mayank99: i think this is a css bug. the wrapper needs flex-direction: column probably

packages/iTwinUI-react/src/core/Table/Table.test.tsx
adamhock: Removed

apps/storybook/src/Table.stories.tsx
```diff
@@ -3130,80 +2983,73 @@ export const ColumnManager: Story<Partial<TableProps>> = (args) => {
...
+            <>{props.row.original.startDate.toLocaleDateString('en-US')}</>
+          );
+        },
+      },
+      {
+        id: 'endDate',
+        Header: 'End date',
+        accessor: 'endDate',
+        Cell: (props: CellProps<TableStoryDataType>) => {
+          return <>{props.row.original.endDate.toLocaleDateString('en-US')}</>;
```
elephantcatdog, line 3022: This return statement no longer has () around it. Is that okay?

adamhock: I believe this is some `prettier` print issue

elephantcatdog: ah ok

elephantcatdog:
Seems ok to me. I approve whenever Mayank does

packages/iTwinUI-react/src/core/Table/Table.tsx
elephantcatdog: @adamhock Does adding `flex-direction` fix the problem with `header-wrapper` wrapping all groups?

bentleyvk: I don't think we are supporting columns grouping so there is no need to fix this.

adamhock: Adding `flex-direction` does solve this problem, but there's another problem that occurs after this change is made.

TableRowHeaderHorizontalScroll

mayank99: > I don't think we are supporting columns grouping so there is no need to fix this.

Ok, so lets merge this as-is. But we need to clear in our messaging that we do not support grouped headers yet.

mayank99 (approved):

PR #937, opened 2022-11-14
chore: Storybook side navigation style update

gretanausedaite:
Updated iTwinUI-css to 1.0.0-dev.16
Updated iTwinUI-variables to 1.0.0-dev.1
Updated selected item style in Storybook

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
   <picture>
     <source
       media='(prefers-color-scheme: dark)'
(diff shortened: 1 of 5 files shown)


mayank99:
Nothing changed here? Then I guess there is no need to merge it. We can just update to stable 1.0 when it's released.

Main thing I really wanted to test was whether export maps from css work in react, and it looks like they do.

mayank99 (approved):
Changes look good. PR title should reflect storybook change, which is bigger than the package update change.

Issue #948, opened 2022-11-21
Update codesandboxes with the latest changes


gretanausedaite:
Update code sandbox examples with react v2
<img width="1544" alt="image" src="https://user-images.githubusercontent.com/83585998/203056117-7ab62b64-0b4d-4c48-bfc9-e89cee0b2950.png">

mayank99:
I think @veekeys owns the codesandbox we currently use. Thinking from a maintainability perspective, we should probably have the code for the sandbox be _inside_ our repo (kinda like how we want to add examples in #929). Codesandbox and Stackblitz both support using github as a source.

Currently if you try to use the vite playground as a source, it fails because it relies on private deps (`configs`) in the monorepo. We shouldn't really rely on that tsconfig because it's meant for our package, not for playgrounds or examples (which should be standalone imo).
- https://stackblitz.com/github/iTwin/iTwinUI-react/tree/main/apps/playground
- https://codesandbox.io/s/github/iTwin/iTwinUI-react/tree/main/apps/playground

Once we do that, we would just need to update the links in our README and issue templates.

veekeys:
As for now, I fixed the compilation issue.
Check up if need to change something else.

veekeys:
Updated.
I guess you can close this and update other issue with support of playground complication in cloud.

PR #951, opened 2022-11-29
chore: Add nextjs and astro playgrounds

### LINKED ISSUE #929: (maintenance) Add examples to monorepo
Not required for v2 release but it would be nice to have more than just a vite playground in the monorepo. At minimum we should have one playground for Nextjs and one for Astro. 

This will allow us to test things like export maps and CSS imports, as well as have an example to show our users, especially because setup in SSR environments is not self-explanatory. Nextjs requires using `next-global-css` and Astro requires adding `@itwin/itwinui-css` to `vite.ssr.noExternal`. Also `ThemeProvider` goes in very specific places to make SSR work so it would be nice to show.

mayank99:
This PR adds examples for nextjs and astro to showcase the ability to SSR our ThemeProvider, as well as allow easily testing things. The nextjs example uses [`next-transpile-modules`](https://github.com/martpie/next-transpile-modules) to resolve css imports within components (this is a limitation in nextjs). The astro example doesn't hydrate the `ThemeProvider` (no [`client:*`](https://docs.astro.build/en/core-concepts/framework-components/#hydrating-interactive-components)) because hydrating it results in an extra `<body>` tag being added. I think this is a bug in astro.

`yarn build` will build these playgrounds (as well as the existing vite one) so we can be sure that our package will continue to work with these frameworks in every future release. Currently these examples only contain a `Button` and `ThemeProvider`. We could add more things in the future.

Other interesting notes:
- Moved vite-playground into `playgrounds/` folder.
- Removed tsconfig from `configs`. Every playground should be self-contained so that it can be opened in stackblitz/codesandbox through url. See https://github.com/iTwin/iTwinUI-react/issues/948#issuecomment-1329478969 for more context.
- Removed copyright linter from all playgrounds. It's not required when it's not part of the published package, and it adds noise when users open our examples in stackblitz.
- Moved astro prettier config into `configs` so that it can be shared by `website` and `astro-playground`.
- Playgrounds will *not* open by default when running `yarn dev`. Auto-opening every dev server is an annoying experience, so they will now need to be manually opened. Storybook and website will still auto-open. The urls/ports will be as follows (incrementing from 1700 which i picked randomly), assuming they are available:
  - docs website:  `http://localhost:1700`
  - vite playground: `http://localhost:1701`
  - next playground: `http://localhost:1702`
  - astro playground: `http://localhost:1703`
  - storybook: `http://localhost:6006` (storybook default)


examples/next/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "next-playground",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev -p 1702",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
```
gretanausedaite, line 9: It would be nice to have prod build preview in next too.

gretanausedaite (commented):
Looks good as first examples. Of course we should add more to them later.
Could you add some info in readme how to run these? It's kinda obvious but it's nice to have some info for users or new team members who is not familiar with this environment.

examples/next/package.json
mayank99: that's what `start` does

gretanausedaite: Did not know that. Thanks!

mayank99:
> Could you add some info in readme how to run these? It's kinda obvious but it's nice to have some info for users or new team members who is not familiar with this environment.

Updated the contributing guide in [`1210885`](https://github.com/iTwin/iTwinUI-react/pull/951/commits/12108851894ab274f88dae48f8a796c08584162b). It mentions the dev server and ports. I didn't add any READMEs to individual workspaces because i thought that would kinda become hidden and too many files to maintain.

gretanausedaite (approved):

playgrounds/astro/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "type": "module",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "dev": "astro dev",
+    "start": "astro dev",
+    "build": "astro build",
+    "preview": "astro preview",
+    "astro": "astro",
+    "clean": "rimraf .turbo && rimraf node_modules && rimraf dist"
```
rohan-kadkol, line 12: I'm not sure when is `.turbo` generated, but after running `yarn dev`, `.turbo` is not generated in `/playgrounds/astro`. So just confirming if this is needed.

Also, is `dist` some type of output folder from turbo?

playgrounds/astro/package.json
```diff
@@ -0,0 +1,27 @@
...
+  },
+  "dependencies": {
+    "@astrojs/react": "^1.2.2",
+    "@itwin/itwinui-react": "*",
+    "astro": "~1.6.3",
```
rohan-kadkol, line 17: Any specific reason for `~`? Would `^` work?

playgrounds/astro/src/pages/_layout.astro
```diff
@@ -0,0 +1,25 @@
...
+  <head>
+    <meta charset='utf-8' />
+    <meta name='viewport' content='width=device-width' />
+    <title>Astro</title>
+  </head>
+  <ThemeProvider as='body' theme='dark'>
+    <slot />
+  </ThemeProvider>
+
+  <style is:global>
```
rohan-kadkol, line 15: What is the `is:global` for? Is it an astro specific thing?

playgrounds/astro/src/env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="astro/client" />
```
rohan-kadkol, line 1: I'm new to astro. What is the need of this file/line? Should this be `.gitignore`d, similar to [L6](https://github.com/iTwin/iTwinUI-react/pull/951/files#diff-33c4e506f435461adae7f2fd3139ab1c45d9ce24daff6307f4e132b32a785f74R6) in the next playground?

playgrounds/vite/src/main.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <>
+      <Global
+        styles={css`
+          * {
+            box-sizing: border-box;
+            margin: 0;
+          }
+        `}
+      />
```
rohan-kadkol, line 24: I think this is an addition to the old `itwinui-playground`. Just curious, what does this do and why was this needed?

apps/website/package.json
```diff
@@ -23,7 +24,6 @@
     "astro": "~1.6.3",
     "nanostores": "^0.7.0",
     "postcss-preset-env": "^7.8.2",
-    "prettier": "^2.8.0",
```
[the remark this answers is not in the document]
mayank99: prettier version will be managed by `configs`

apps/website/package.json
```diff
@@ -32,5 +32,6 @@
     "rehype-slug": "^5.0.1",
     "sass": "^1.55.0",
     "snarkdown": "2.0.0"
-  }
+  },
+  "prettier": "configs/prettier-astro-config.js"
```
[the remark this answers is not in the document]
mayank99: `configs` is a private workspace for putting our config files so we can share them in other workspaces. and yes `"prettier"` needs a file. the file must be a valid prettier configuration

packages/configs/package.json
```diff
@@ -19,7 +19,8 @@
     "eslint-plugin-react": "^7.29.0",
     "eslint-plugin-react-hooks": "^4.2.0",
     "fast-glob": "^3.2.5",
-    "prettier": "^2.8.0"
+    "prettier": "^2.8.0",
+    "prettier-plugin-astro": "^0.7.0"
```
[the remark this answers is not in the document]
mayank99: prettier-plugin-astro is a plugin for prettier, not a replacement

playgrounds/next/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "next-playground",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
mayank99: it doesn't matter, these versions don't do anything and will never be changed, the playgrounds are private / not published

playgrounds/astro/package.json
mayank99: `dist` is the output from astro. `.turbo` is where turbo puts any cached stuff and i think logs.

playgrounds/astro/package.json
mayank99: i want to keep it `~` because minor version bumps can often have soft breaking changes which might make our build fail

playgrounds/astro/src/pages/_layout.astro
mayank99: yes https://docs.astro.build/en/reference/directives-reference/#isglobal

playgrounds/astro/src/env.d.ts
mayank99: it should not be gitignored. this adds types for astro and vite. for example when you type `Astro.` in a `.astro` file, you will get auto complete

playgrounds/vite/src/main.tsx
mayank99: this is a replacement for the `main.css` file which i deleted. i'm trying to keep this repo free of css files.

r100-stack (approved):

Issue #915, opened 2022-11-08
Table: global filter should trigger `onFilter` callback


bentleyvk:
When using global filter it should trigger `onFilter` callback in order to get filtered data like with column filters.

PR #958, opened 2022-11-30
feat(Table): `onFilter` gets filtered rows

### LINKED ISSUE #816: Table: need a way to get filtered rows
We could add additional parameter to `onFilter` so users could get filtered rows from there.

adamhock:
- `onFilter` now has an additional parameter `filteredData` which users can get their filtered rows from.
- `onFilter` is now called in `Table.tsx` instead of `filterHandler.ts` because there's no way to get filtered rows in the state reducer. Because of this, `onFilterHandler` now only returns the filters.
- `previousFilter` and `currentFilter` are used to track when the filters are changed so `onFilter` is only called when this occurs. These are checked in a `.useEffect` after the `stateReducer` runs all the handling.

Closes iTwin/iTwinUI-react#816


packages/iTwinUI-react/src/core/Table/Table.test.tsx
```diff
@@ -865,6 +865,7 @@ it('should filter table', async () => {
   expect(onFilter).toHaveBeenCalledWith(
     [{ fieldType: 'text', filterType: 'text', id: 'name', value: '2' }],
     expect.objectContaining({ filters: [{ id: 'name', value: '2' }] }),
+    expect.any(Object),
```
gretanausedaite, line 868: Does not seem like a valid test. At least check returned rows count to see if it match.

apps/storybook/src/Table.stories.tsx
```diff
@@ -498,11 +498,22 @@ export const Filters: Story<Partial<TableProps>> = (args) => {
   );
 
   const onFilter = React.useCallback(
-    (filters: TableFilterValue<TableStoryDataType>[], state: TableState) => {
+    (
+      filters: TableFilterValue<TableStoryDataType>[],
+      state: TableState,
+      rows: Row<{ name: string; description: string }>[],
```
gretanausedaite, line 504: Maybe use same name `filteredData` for prop as in Table?

adamhock: Changed

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -459,7 +465,18 @@ export const Table = <
           onSort?.(newState);
           break;
         case TableActions.setFilter:
-          onFilterHandler(newState, action, previousState, instance, onFilter);
+          // Check to see if filters have changed
+          const previousFilter = previousState.filters.find(
```
gretanausedaite, line 469: Maybe move to function to keep this part of code minimal.

packages/iTwinUI-react/src/core/Table/actionHandlers/filterHandler.ts
```diff
@@ -10,21 +10,15 @@ export const onFilterHandler = <T extends Record<string, unknown>>(
   action: ActionType,
```
gretanausedaite, line 10: I believe you can remove not used props.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -654,6 +670,15 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  React.useEffect(() => {
+    if (previousFilter.current === currentFilter.current) {
+      return;
+    } else {
```
gretanausedaite, line 676: I think you could remove `else` statement and keep next two line as just code.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -654,6 +670,15 @@ export const Table = <
...
+      previousFilter.current = currentFilter.current;
+      onFilter?.(currentFilter.current, state, instance?.filteredRows);
+    }
+  }, [state, instance?.filteredRows, onFilter]);
```
gretanausedaite, line 680: I don't think `instance` can be undefined. Why using `instance?.<....>`?

gretanausedaite (commented):
Have you tested with subrows?

packages/iTwinUI-react/src/core/Table/Table.tsx
elephantcatdog: Or reverse it? Like

> if (previousFilter.current !== currentFilter.current) {
      previousFilter.current = currentFilter.current;
      onFilter?.(currentFilter.current, state, instance?.filteredRows);
    }

elephantcatdog (commented):
No other comments

apps/storybook/src/Table.stories.tsx
```diff
@@ -498,11 +498,22 @@ export const Filters: Story<Partial<TableProps>> = (args) => {
...
-    (filters: TableFilterValue<TableStoryDataType>[], state: TableState) => {
+    (
+      filters: TableFilterValue<TableStoryDataType>[],
+      state: TableState,
+      filteredData: Row<{ name: string; description: string }>[],
+    ) => {
+      // rowInfo is used due to JSON errors when displaying row data
+      let rowInfo = '[';
+      filteredData.forEach((row) => {
+        rowInfo += `{\"name\":\"${row.original.name}\",\"description\":\"${row.original.description}\"},`;
```
bentleyvk, line 509: Or just `JSON.strigify(row.original)`.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -654,6 +670,15 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  React.useEffect(() => {
+    if (previousFilter.current === currentFilter.current) {
```
bentleyvk, line 674: Need a test that `onFilter` is not triggered when the same filter is applied.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -459,7 +465,18 @@ export const Table = <
           onSort?.(newState);
           break;
         case TableActions.setFilter:
-          onFilterHandler(newState, action, previousState, instance, onFilter);
+          // Check to see if filters have changed
```
bentleyvk, line 468: I think the same is needed for global filters but maybe with a separate PR because they are not even triggering `onFilter`

adamhock: There is an issue filed for that here #915 that will be handled in a separate PR

apps/storybook/src/Table.stories.tsx
adamhock: Fixed

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: This has been fixed. I was using it inside of `filterHandler.ts` but forgot to remove the check whenever code was moved to `Table.tsx`

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: This has been fixed

packages/iTwinUI-react/src/core/Table/actionHandlers/filterHandler.ts
adamhock: These props are now being used since the filter checking is done in `filterHandler.ts`

adamhock:
> Have you tested with subrows?

subrows have been tested!

packages/iTwinUI-react/src/core/Table/Table.tsx
adamhock: This has been moved!

gretanausedaite (approved):
Looks good to me.

elephantcatdog (approved):

PR #960, opened 2022-12-01
website: Add all components page

gretanausedaite:
- Add all components page to website
- Add link in homepage to components page
- Remove Typography - deprecated in v2

image


gretanausedaite:
To do:
- [ ] Update cards to be more accessible

apps/website/src/components/ComponentIcon.astro
```diff
@@ -0,0 +1,102 @@
...
+if (icon) {
+  const codePath = new URL(`./src/icons/component-icons/${icon}.astro`, baseUrl);
+  code = await fs.promises.readFile(codePath, 'utf8');
+}
+
+// const componentName = icon.split('.')[0];
+// const capitalize = (s: string) => s.charAt(0).toUpperCase() + s.slice(1);
+---
+
+<div class='thumbnail' set:html={code} />
```
mayank99, line 28: I think you might not need to do all this extra work because [`Astro.glob`](https://docs.astro.build/en/guides/imports/#astroglob) can do it for you. Only catch is that it does not support dynamic strings so you need glob everything and then filter inside it.

```suggestion
const allThumbnails = await Astro.glob('../icons/component-icons/*.astro');
const thumbnail = allThumbnails.find(({ file }) => file.endsWith(`${icon}.astro`));
---

<thumbnail.default class='thumbnail' />
```

apps/website/src/pages/docs/_layout.astro
```diff
@@ -17,6 +17,10 @@ const currentPage = new URL(Astro.request.url).pathname;
 </Global>
 
 <style lang='scss'>
+  :root {
+    --_content-width: min(100%, 70ch);
```
mayank99, line 21: If this is set on root and changed in other files, maybe it should not be prefixed by `_`.

There is also this convention of creating two layers of variables such that it can be controlled both from the outside and the inside. Might make sense here. See https://lea.verou.me/2021/10/custom-properties-with-defaults/#:~:text=pseudo%2Dprivate%20custom%20properties

apps/website/src/components/ComponentIcon.astro
```diff
@@ -0,0 +1,102 @@
...
+
+<style lang='scss' is:global>
+  .thumbnail svg {
```
mayank99, line 31: If these styles are global, we need to come up with a way to scope them. I can think of two approaches to this:
1. move most of the styles into the thumbnails and scope them there. then expose a few custom properties to control them from outside.
2. create a scoped root class name, and then use `:global` to target the classes inside the svg. this is similar to what i'm doing with the stuff inside content-wrapper in `_layout.astro`.

apps/website/src/components/LeftSidebarContent.astro
```diff
@@ -88,11 +90,35 @@ sections.set(
...
+
+    @media (max-width: 50em) {
+      color: var(--color-text);
+      font-size: var(--type-3);
+      padding-inline: var(--space-6);
+      margin-bottom: var(--space-3);
+    }
+  }
+
+  .nav-group-subtitle {
```
mayank99, line 116: the styling is a bit confusing now and it's not obvious what is clickable vs what isn't.

apps/website/src/components/ComponentIcon.astro
```diff
@@ -0,0 +1,102 @@
...
+    height: calc(var(--space-6) * 7);
+    width: 100%;
+  }
+  .main-color {
+    fill: #fff;
```
mayank99, line 36: contrast is pretty bad in a few places

apps/website/src/pages/docs/components.mdx
```diff
@@ -0,0 +1,11 @@
...
+title: Components
+description:
+layout: ./_layout.astro
+propsPath:
+group: main
+---
+
+import ComponentsPage from '~/components/ComponentsPage.astro';
+
+<ComponentsPage />
```
mayank99, line 11: i think there is kinda no need to create this mdx file. most of the frontmatter properties are not even being used or are turned into exceptions.

you can name it `components.astro` and move all the code inside here. only thing you would need to add the wrapping layout yourself.

```astro
---
import Layout from './_layout.astro';
---

<Layout content={{ title: 'Components' }}>
  // ...
</Layout>
```

apps/website/src/pages/index.astro
```diff
@@ -71,6 +72,13 @@ const currentPage = new URL(Astro.request.url).pathname;
           <h3>ComboBox</h3>
           <p>Allows typing a value to filter and choose from the options in dropdown list.</p>
         </div>
+        <div class='demo-wrapper'>
+          <div class='demo-box demo-box-muted'>
+            <a class='discover-more' href='/docs/components'
+              >Discover more components <MoreIcon />
```
mayank99, line 78: Arrow makes more sense compared to carets.

gretanausedaite: Added arrow

apps/website/src/pages/docs/components.mdx
gretanausedaite: Moved to astro file

apps/website/src/icons/MoreIcon.astro
```diff
@@ -0,0 +1,4 @@
+<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16' width='16' height='16' fill='#d6d9db'
```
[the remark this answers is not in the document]
gretanausedaite: Added astro.props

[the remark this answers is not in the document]
mayank99: it's missing in RightArrowIcon

elephantcatdog (commented):
I get an error when I try to go to `docs/components`: 
image

The console has this error: 
image

apps/website/src/icons/MoreIcon.astro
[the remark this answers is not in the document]
gretanausedaite: fixed

apps/website/src/components/ComponentIcon.astro
gretanausedaite: Changed color a bit

gretanausedaite:
> I get an error when I try to go to `docs/components`

Should be fixed now.

apps/website/src/components/LeftSidebarContent.astro
gretanausedaite: Should be good now. Used --color-line-2 for muted look. 
image

apps/website/src/pages/docs/_layout.astro
mayank99: yes

gretanausedaite: Added

mayank99: so now you can move it back down, it's no longer required in `:root` in this file

in `components.astro` it can be set anywhere so `:root` is fine there

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,81 @@
+---
+import React from 'react';
+import ComponentIcon from './ComponentIcon.astro';
+
+export type Props = {
+  url: string;
+  title: React.ReactNode;
+  description?: React.ReactNode;
```
mayank99, line 8: this is an astro file, react node is not relevant here. did you mean to use `string` ?

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,81 @@
...
+  image?: string;
+  className?: string;
+};
```
mayank99, line 11: better to forward all props (which includes `class`).
```suggestion
} & astroHTML.JSX.LiHTMLAttributes;
```

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,81 @@
...
+
+const { url, title, description, image, className } = Astro.props;
```
mayank99, line 13: after forwarding all props, we need to assign `class` to `className` because `class` is a reserved keyword. also grab `...rest`
```suggestion
const { url, title, description, image, class: className, ...rest } = Astro.props;
```

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,81 @@
...
+---
+
+<li class={`card ${className}`}>
```
mayank99, line 16: astro has [`class:list`](https://docs.astro.build/en/reference/directives-reference/#classlist) which makes this cleaner. also apply rest props

```suggestion
<li class:list={['card', className]} {...rest}>
```

apps/website/src/components/ComponentIcon.astro
```diff
@@ -0,0 +1,91 @@
+---
+import { renderJSX } from 'astro/dist/runtime/server';
+import fs from 'node:fs';
+
```
mayank99, line 4: these are not used
```suggestion
```

apps/website/src/pages/docs/components.astro
```diff
@@ -0,0 +1,54 @@
...
+  ><main>
+    <ul class='card-list'>
+      {
+        componentPages.map((page) => (
+          <Card
+            title={page.title}
+            description={page.description}
+            url={page.url}
+            image={page.thumbnail}
+            className='card-background'
```
mayank99, line 26:
```suggestion
            class='card-background'
```

apps/website/src/components/Card.astro
gretanausedaite: I wanted to have more generic component where title could be string or string with icon. But that's probably not going to happen. Need to add other component to home page demo cards.

mayank99: [slots](https://docs.astro.build/en/core-concepts/astro-components/#slots) is the way to do that in astro, but string works for now

apps/website/src/components/LeftSidebarContent.astro
mayank99: alignment looks better now but `--color-line-2` has bad contrast because it's not really meant to be used as for text. might want to add a new muted color that meets 4.5 contrast score

apps/website/src/icons/component-icons/Alert.astro
```diff
@@ -0,0 +1,20 @@
+<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100' {...Astro.props}>
+  <g id='b'>
```
mayank99, line 2: most of these svgs have single letter ids.

`id` must be globally unique across the whole page, so this is invalid html. one option is to run all the svgs through [`svgo`](https://github.com/svg/svgo), or you could remove the ids manually for now.

apps/website/src/pages/index.astro
```diff
@@ -311,10 +321,18 @@ const currentPage = new URL(Astro.request.url).pathname;
     }
     .library-item {
       padding: var(--space-5);
-      text-align: start;
-      border: 1px solid var(--color-text);
       border-radius: var(--border-radius-1);
+      box-shadow: 0 0 1px var(--color-highlight-1), 0 0 10px var(--color-highlight-2);
```
mayank99, line 325: Are these changes intentional? Previously it only had box-shadow on hover/focus and had better visibility. It looks too dim/blue now.

| Before | After |
| --- | --- |
|  |  |

apps/website/src/pages/index.astro
```diff
@@ -353,15 +371,49 @@ const currentPage = new URL(Astro.request.url).pathname;
...
+          text-decoration: none;
+        }
+
+        &::after {
+          content: '';
+          position: absolute;
+          top: 0;
+          bottom: 0;
+          left: 0;
+          right: 0;
```
mayank99, line 412:
```suggestion
          inset: 0;
```

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,80 @@
...
+      text-decoration: none;
+    }
+
+    &::after {
+      content: '';
+      position: absolute;
+      top: 0;
+      bottom: 0;
+      left: 0;
+      right: 0;
```
mayank99, line 62:
```suggestion
      inset: 0;
```

apps/website/src/components/Card.astro
```diff
@@ -0,0 +1,80 @@
+---
+import React from 'react';
```
mayank99, line 2:
```suggestion
```

apps/website/src/pages/docs/components.astro
```diff
@@ -0,0 +1,54 @@
+---
+import Layout from './_layout.astro';
+import Card from '../../components/Card.astro';
```
mayank99, line 3:
```suggestion
import Card from '~/components/Card.astro';
```

apps/website/src/components/LeftSidebarContent.astro
gretanausedaite: image

Ratio 4:63

apps/website/src/pages/index.astro
gretanausedaite: I wanted it to match copy to clipboard button but I undid all the changes because of better visibility.

apps/website/src/icons/component-icons/Alert.astro
gretanausedaite: removed IDs

apps/website/src/components/ComponentIcon.astro
gretanausedaite: Should be good now.

mayank99 (approved):

PR #961, opened 2022-12-01
fix(ModalButtonBar, ModalContent): Wrap `DialogButtonBar` and `DialogContent` instead of directly assigning

mayank99:
I was testing some build improvements to storybook when I started getting this error:
```
DialogButtonBar is not defined
```

Investigating it, I found that docgen was trying to use `DialogButtonBar`. But `DialogButtonBar` and `DialogContent` are not exported (they get exposed as subcomponents of `Dialog`, like `Dialog.ButtonBar`, etc) so it could not find it.

To fix this, instead of directly assigning `DialogButtonBar` to `ModalButtonBar` and `DialogContent` to `ModalContent`, I am now creating a wrapper component over it. I would imagine this would potentially fix some similar issues on the user's side too.


adamhock (approved):

gretanausedaite (commented):
We are assigning Avatar to UserIcon. Could this have similar problem?

mayank99:
> We are assigning Avatar to UserIcon. Could this have similar problem?

UserIcon is exported so it should be fine. And I didn't get any other errors in the storybook change I was testing.

gretanausedaite (approved):

PR #963, opened 2022-12-01
fix(useTheme): Exit early if theme is set on body

### LINKED ISSUE #944: ThemeProvider: using `as='body'` inside nextjs `_document` adds duplicate attributes
Open https://stackblitz.com/edit/nextjs-ujsdru?file=pages%2F_document.js and inspect the preview to find duplicate `data-iui-` attributes (both on `<html>` and `<body>`). 

The ones on `<html>` are likely added by `useTheme` (which should not happen because it checks for `ThemeContext` first). It's possible that the context is not being set when `ThemeProvider` is used in `_document.js`.

---

Contents of `_document.js`:
```jsx
import { ThemeProvider } from '@itwin/itwinui-react';
import { Html, Head, Main, NextScript } from 'next/document';

export default function Document() {
  return (
    <Html>
      <Head />
      <ThemeProvider theme="dark" as="body">
        <Main />
        <NextScript />
      </ThemeProvider>
    </Html>
  );
}

```

mayank99:
Fixes #944

Our current recommendation is to use `<ThemeProvider as='body'>` which sounds nice in theory, but hydrating a react component rendered as a `<body>` seems to be a bit difficult in practice. Next doesn't properly set context. Astro renders a duplicate a `<body>` element, which can be worked around by not hydrating it, but then again context will not be set.

When context is not set, this line will return undefined:
https://github.com/iTwin/iTwinUI-react/blob/1b4e5c51ec6c29e8d2c4b28e6f54ebfba26e3b52/packages/iTwinUI-react/src/core/utils/hooks/useTheme.ts#L55

So instead of relying entirely on context check, I'm now explicitly checking if `<body>` has `data-iui-theme` set, and exiting from the useEffect if it does. This avoids running the code which erroneously adds the theme on `<html>` (mentioned in #944).

As a bonus, this approach will also allow users to bypass using our `ThemeProvider` and manually set the data attributes in whatever they prefer.


gretanausedaite (approved):

packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.test.tsx
```diff
@@ -293,4 +293,16 @@ describe('Fallback (without children)', () => {
     expect(document.documentElement.dataset.iuiContrast).toBeUndefined();
     expect(document.body.classList).not.toContain('iui-root');
   });
+
+  it('should not modify root if <body> has theme', () => {
+    render(
+      <body data-iui-theme='dark'>
+        <ThemeProvider theme='light' />
+      </body>,
+      { container: document.documentElement },
```
rohan-kadkol, line 302: A little confused why is this needed? What does this line do?

packages/iTwinUI-react/src/core/utils/hooks/useTheme.ts
```diff
@@ -56,8 +56,8 @@ export const useTheme = (
   const ownerDocument = themeOptions?.ownerDocument ?? getDocument();
```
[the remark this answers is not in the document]
mayank99: it's deprecated for our users. we still need to use it internally and keep supporting it

packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.test.tsx
mayank99: normally, testing-library will add the element inside a div and use that as a container. this line is telling it to use `<html>` as the container since we want `<body>` to be a direct child of it.

when confused, you can always look up the api docs: https://testing-library.com/docs/react-testing-library/api/#container

mayank99: > ... and this is wrong since there are two `<body>` and also since `<body>` is a child of a `<div>`?

Yeah, we only want one `<body>` and it should be in the correct place.
 
> So with this line, is the test dom exactly what we provide? So something like this?

It will be like this (includes the container, and ThemeProvider doesn't render anything):
```html
<html>
  <body data-iui-theme='dark'>
  </body>
</html>
```

> Also just curious, is there a way to `console.log` the dom tree in a easily readable form and not like a JS object type?

You can use `debug` which is returned by `render`.
```js
    const { debug } = render(...);
    debug();
```

r100-stack (approved):

PR #965, opened 2022-12-01
docs: Add "edit page on github" link

mayank99:
Added github edit url based on what page is currently shown. Looks like this:

This is the first step towards accepting docs contributions from non-technical folks (https://github.com/iTwin/iTwinUI/issues/885).

Also (unrelated): renamed `Typography` to `typography` for sorting order. Currently it appears at the top of the list in https://github.com/iTwin/iTwinUI-react/tree/main/apps/website/src/pages/docs.


apps/website/src/components/EditPageLink.astro
```diff
@@ -0,0 +1,33 @@
+---
+import EditIcon from '~/icons/EditIcon.astro';
+
+type Props = { page: string };
+const { page } = Astro.props;
+
+const githubEditUrl = `https://github.com/iTwin/iTwinUI-react/edit/main/apps/website/src/pages/${page}.mdx`;
```
gretanausedaite, line 7: Future problem: I think it's possible to add this link to config to not have hardcoded values. When we deploy this page to cloud, we could just update config and not full app if we ever change docs location.

gretanausedaite: Actually same with all links.

apps/website/src/components/EditPageLink.astro
```diff
@@ -0,0 +1,33 @@
...
+---
+
+<a href={githubEditUrl}>
```
gretanausedaite, line 10: Open in new tab.

gretanausedaite (commented):
Seems like layout is broken. 
image

apps/website/src/components/EditPageLink.astro
mayank99: I'm not sure I understand how the place we *deploy* affects this link. The github link will stay the same regardless of whether it is deployed to a custom domain or one from azure or netlify

apps/website/src/components/EditPageLink.astro
mayank99: Open in new tab is generally considered an accessibility pattern. Any reason you want it here?

mayank99:
> Seems like layout is broken. 

Fixed 🎉

elephantcatdog (commented):
When I try to visit other pages in the documentation side of things (except the 'Variables' page), I get errors like this: 
image


Also when the 'Edit this page on GitHub' link is clicked on the 'Getting Started' page in the docs, the link leads here: https://github.com/iTwin/iTwinUI-react/edit/main/apps/website/src/pages//docs.mdx
which isn't a valid link.

apps/website/src/components/EditPageLink.astro
gretanausedaite: It would be more convenient (for me). You don't have to navigate back after changing text in github and creating PR.
I guess it's ok if it's considered an anti-pattern.

apps/website/src/components/EditPageLink.astro
gretanausedaite: True. I just thought it would be better practise to have customisable settings in config file. But probably we are not going to change that often or in runtime (eg. other apps have different URLs for dev, qa and prod environments).

gretanausedaite (approved):

PR #969, opened 2022-12-05
chore: Use `swc` for dev server

mayank99:
This is the second of my three proposed monorepo improvements (first one being #951, third one being storybook improvements).

**Previously**: Every `dev` script was bottlenecked by `yarn build` in itwinui-react, which takes ~10 seconds because `tsc` is slow.

**Now**: `yarn dev` no longer requires `yarn build` in itwinui-react (it has been removed in `turbo.json`). Running `yarn dev` does two things:
1. start [`swc`](https://swc.rs/) in watch mode to transpile TS source to JS.
    - the initial build is super fast, taking a fraction of a second instead of 10. the watch mode is also reliable and fast.
2. starts `tsc` compiler in watch mode to generate types _only_.
    - the initial build takes a couple seconds but is non-critical. the watch mode is fast enough.
 
This means most of our dev servers start instantly and are usable without waiting for `tsc`. Storybook still takes over a minute to start (because it uses webpack) and it is the only noticeably slow part of our DX. But the hot refresh is still instant.


gretanausedaite (commented):
`yarn dev` crashes if code has type errors.

mayank99:
> `yarn dev` crashes if code has type errors.

I think that is the unfortunate side effect of running `tsc` without `--watch`. To avoid this problem I have removed `yarn build:types`.

packages/iTwinUI-react/esbuild.mjs
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import esbuild from 'esbuild';
+import fg from 'fast-glob';
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+const argsList = process.argv.length > 2 ? process.argv.slice(2) : [];
```
rohan-kadkol, line 10: nit: I think slice returns an empty list when the start index is higher than the last index in the list. So maybe the check can be removed.

```suggestion
const argsList = process.argv.slice(2);
```

packages/iTwinUI-react/esbuild.mjs
```diff
@@ -0,0 +1,29 @@
...
+const isCjs = argsList.includes('--cjs');
+const shouldWatch = argsList.includes('--watch');
+
+const entryPoints = fg.sync(['src/**/*.{ts,tsx}'], {
+  ignore: ['src/**/*.test.{ts,tsx}'],
+  cwd: dirname(fileURLToPath(import.meta.url)),
```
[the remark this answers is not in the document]
mayank99: `__dirname` does not work in ESM

packages/iTwinUI-react/package.json
```diff
@@ -47,16 +47,16 @@
   ],
   "scripts": {
     "build": "yarn clean:build && tsc -p tsconfig.cjs.json && tsc -p tsconfig.esm.json",
-    "build:watch": "yarn clean:build && concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
-    "clean:build": "yarn clean:coverage && rimraf esm && rimraf cjs",
+    "build:watch": "concurrently \"tsc -p tsconfig.cjs.json --watch\" \"tsc -p tsconfig.esm.json --watch\"",
+    "clean:build": "rimraf esm && rimraf cjs",
```
rohan-kadkol, line 51: Just curious, why are we not running `yarn clean:coverage` anymore? Also, what is stored in `packages/iTwinUI-react/coverage`? Because I could not find this folder locally

mayank99: `coverage/` is produced when running `yarn test --coverage`. i removed it because it has nothing to do with build

r100-stack (approved):

veekeys:
Have you compared swc to esbuild? Seems at least swc core has stable v1 version.
It also uses chokidar for watching files and I think then it would recompile only the file changed rather than everything, which might be useful in this case. Of course, the only issues is with type files...

mayank99:
> Have you compared swc to esbuild? Seems at least swc core has stable v1 version. It also uses chokidar for watching files and I think then it would recompile only the file changed rather than everything, which might be useful in this case. 

I did try manually with `chokidar` but it was a lot of code and still not working perfectly.

Just tried with `swc` and it seems to work great! Thanks for the suggestion.

https://user-images.githubusercontent.com/9084735/206235178-043c6881-c9c7-4eb9-be50-59fc175e08e9.mov

> Of course, the only issue is with type files...

I will use `tsc` to generate types only. I think that should be fine.

veekeys:
> > Have you compared swc to esbuild? Seems at least swc core has stable v1 version. It also uses chokidar for watching files and I think then it would recompile only the file changed rather than everything, which might be useful in this case.
> 
> I did try manually with `chokidar` but it was a lot of code and still not working perfectly.
> 
> Just tried with `swc` and it seems to work great! Thanks for the suggestion.
> 
>  Untitled.mov 
> > Of course, the only issue is with type files...
> 
> I will use `tsc` to generate types only. I think that should be fine.

Nice!!
I am jealous now, cause I could not figure out why Vite app is refreshing the page even though it sees that only single file from the dependency has changed (writes in console hmr, not page reload like in other cases).

packages/iTwinUI-react/package.json
```diff
@@ -47,16 +47,19 @@
...
     "clean": "rimraf .turbo && yarn clean:coverage && yarn clean:build && rimraf node_modules",
     "test": "jest",
     "test:watch": "jest --watch",
     "format": "prettier --config .prettierrc **/*.{tsx,ts,js} --ignore-path .gitignore --write",
     "lint": "eslint \"**/*.{js,ts,tsx}\" --max-warnings=0",
     "lint:fix": "yarn lint --fix && node ../configs/copyrightLinter.js --fix \"*/**/*.{js,ts,tsx}\"",
-    "dev": "yarn build:watch",
+    "dev": "yarn clean:build && concurrently \"yarn dev:swc\" \"yarn dev:swc:cjs\" \"yarn dev:types\"",
+    "dev:swc": "swc src -d esm --watch",
+    "dev:swc:cjs": "swc src -d cjs --watch -C module.type=commonjs",
```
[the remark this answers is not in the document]
mayank99: what would be the benefit? we never run `dev:*` scripts directly. the separation is just for readability

r100-stack (approved):

gretanausedaite (approved):

PR #971, opened 2022-12-05
chore: Fix eslint blocking commits when changing .js files

### LINKED ISSUE #970: eslint prevents commits when committing changes in ignored files

When committing changes of files that are ignored by eslint like the below three `.js` files, it gives the below error and prevents a commit.

image

#### Possible solution

As discussed with Mayank, one recommendation Mayank gave was to not run eslint for `jsx, js` files.

So something similar to replacing: 

https://github.com/iTwin/iTwinUI-react/blob/5be110de87ce7c719d27c137f882200dc9cc01d7/apps/storybook/package.json#L46-L50

with

```
    "*.{tsx,ts,jsx,js}": [
      "prettier --write",
      "node ../configs/copyrightLinter.js --fix"
    ],
    "*.{tsx,ts}": [
      "eslint --max-warnings=0 --fix"
    ]
```

#### Expected behavior
Should not block commits when eslint ignored files are changed.

##### Steps to reproduce
To reproduce this, we can edit eslint ignored files, like any of the three files mentioned in the screenshot above.

#### Additional information
I encountered this when merging the latest `main` branch (`v2.0.1`) with the old `rohan/table-ctrl-shift-click-1` branch of my PR #888

mayank99:
No longer running `eslint` on `js` files in pre-commit hook.

Fixes #970.

### DIFF
--- a/packages/iTwinUI-react/package.json
+++ b/packages/iTwinUI-react/package.json
@@ -108,8 +108,10 @@
   "lint-staged": {
     "*.{tsx,ts,jsx,js}": [
       "prettier --write",
-      "eslint --max-warnings=0 --fix",
       "node ../configs/copyrightLinter.js --fix"
+    ],
+    "*.{tsx,ts}": [
+      "eslint --max-warnings=0 --fix"
     ]
   },
   "prettier": "configs/prettier-config",
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -45,8 +45,10 @@
   "lint-staged": {
     "*.{tsx,ts,jsx,js}": [
       "prettier --write",
-      "eslint --max-warnings=0 --fix",
       "node ../../packages/configs/copyrightLinter.js --fix"
+    ],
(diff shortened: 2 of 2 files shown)


packages/iTwinUI-react/package.json
```diff
@@ -108,8 +108,10 @@
   "lint-staged": {
     "*.{tsx,ts,jsx,js}": [
       "prettier --write",
-      "eslint --max-warnings=0 --fix",
       "node ../configs/copyrightLinter.js --fix"
+    ],
+    "*.{tsx,ts}": [
```
gretanausedaite, line 113: Why separating eslint setting solves the issue?

mayank99: The issue happens because we are ignoring some js files in our eslint config. When we pass files that we have ignored back to eslint, it fails instead of just ignoring it. So excluding all js files is a way of avoiding the error by avoiding running eslint on those files.

gretanausedaite (approved):

PR #973, opened 2022-12-06
fix(ThemeProvider): Exclude `ownerDocument` from `themeOptions` if children passed

mayank99:
Instead of relying on `UseThemeProps`, `ThemeProvider` now defines its own props and uses conditional typing to exclude `ownerDocument` from `themeOptions` if children is passed, and keep `ownerDocument` if children not passed (for backwards compatibility). This change is intended to prevent accidentally passing `ownerDocument`. Would be nice to release it as a patch of 2.0 rather than in 2.1 so that users have a smaller window to make this mistake.

Also, since we are defining a new type, I used the opportunity to write a better jsdoc for the `theme` prop, warning about SSR behavior.

### DIFF
--- a/packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.tsx
+++ b/packages/iTwinUI-react/src/core/ThemeProvider/ThemeProvider.tsx
@@ -10,13 +10,31 @@ import type {
   PolymorphicForwardRefComponent,
   ThemeOptions,
   ThemeType,
-  UseThemeProps,
 } from '../utils';
 import '@itwin/itwinui-css/css/global.css';
 import '@itwin/itwinui-variables/index.css';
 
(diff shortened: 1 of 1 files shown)


gretanausedaite (commented):
It would be nice to have tests for these scenarios to make sure we don't break it in the future.

gretanausedaite (approved):

mayank99:
> It would be nice to have tests for these scenarios to make sure we don't break it in the future.

This is all typescript stuff and there is no change in runtime, so I'm not really sure how to test it tbh.

gretanausedaite:
> This is all typescript stuff and there is no change in runtime, so I'm not really sure how to test it tbh.

Add examples with different usage might be a start [Another PR problems]

adamhock (approved):

PR #983, opened 2022-12-09
fix(Table): Reset `columnOrder` when `columns` changes

### LINKED ISSUE #900: Table: Column order is not respected when external columns change
When changing columns after dragging a column, the new columns are in an unexpected order.

To reproduce:
* Create a table with `enableColumnReordering`
* Drag a column to a different position
* Update the columns passed to the table
  * Must be done in code by changing columns object, rather than the built-in [column manager](https://itwin.github.io/iTwinUI-react/?path=/story/core-table--column-manager)
* The updated columns do not follow the same order as the columns passed to the table

https://codesandbox.io/s/itwinui-react-minimal-example-forked-3cxy87?file=/src/App.tsx
Column order changes

r100-stack:
Closes #900

When new `columns` array contents are passed, the old/stale `columnOrder` is still used. This could cause unintended column ordering as described in #900. 

To fix this, this PR resets the `columnOrder` to `[]` whenever the `columns` array contents change.

## Checklist


packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -655,6 +655,12 @@ export const Table = <
     setPageSize(pageSize);
   }, [pageSize, setPageSize]);
 
+  // Reset the column order whenever new columns are passed
+  // This is to avoid the old columnOrder from affecting the new columns' columnOrder
+  React.useEffect(() => {
```
gretanausedaite, line 660: To me this seems like a drastic action to reset order anytime new columns are passed.
I'm trying to think of a use case when user want to keep old order but nothing comes to my mind right now.

One idea might be to add prop for toggling reset/no-reset. Default would be reset.

But I'm not sure about that.

rohan-kadkol: > To me this seems like a drastic action to reset order anytime new columns are passed.

Yes, I too am a bit hesitant to doing this reset. Because if users do not memoize the columns they pass (using `React.memo` or similar), then the columns object could be re-created on each render.
e.g.
```tsx
const columns = () => [ ... ];

<Table
  columns={columns()}
  ...
/>
```

In this case, reordering columns may not work as the order is always reset.

---

> I'm trying to think of a use case when user want to keep old order but nothing comes to my mind right now.

I tried running the existing tests. All passed except one that was re-rendering with a new columns object (like the above example) (near [L3321-L3350](https://github.com/iTwin/iTwinUI-react/pull/983/files#diff-c4aa401f671f853692462c6d1ee006969f80e3d02863b8a1a8919154aa3f67d9R3321-R3350)). I edited that test to check that the column order is not `resultingColumns` but instead `columns` itself since the order is expected to reset.

---

> One idea might be to add prop for toggling reset/no-reset. Default would be reset.

Maybe we name it something like `autoResetColumnOrder`.

```tsx
/**
* If this is:
* - true, then the `columnOrder` is reset whenever the contents of the passed `columns` changes.
* - false, then `columnOrder` does not reset. This is useful when wanting to manually maintain columnOrder
* @default = true
*/
autoResetColumnOrder?: boolean
```

Maybe we also provide an optional `columnOrder` prop if they want to programmatically change the column order? It would make more sense to provide the above prop if users can actually handle the `columnOrder` themselves.

One other easier option could be to internally check if the content of `columns` have changed and only then do the reset. That way even the above example works.

What do you think?

mayank99: We need to be careful with adding props. Too many props are not helpful for the user. In this case, I think we can reset based on some if condition rather than on every render.

gretanausedaite: @rohan-kadkol Check if columns actually changed in this useEffect.

rohan-kadkol: > One other easier option could be to internally check if the content of columns have changed and only then do the reset. That way even the above example with un-memomized columns works.

> @rohan-kadkol Check if columns actually changed in this useEffect.

**[DONE]** I now added a check to reset the `columnOrder` only when the "contents" of `columns` change ([L687](https://github.com/iTwin/iTwinUI-react/pull/983/files#diff-9edd44b431060c27e0260d286ee188bdf901225b1ee84ff6540f9d796dd0f999R687)). That way, even if a new `columns` object with the same contents is passed on each render, the `columnOrder` is not reset.

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -676,6 +676,20 @@ export const Table = <
     }
   }, [state, instance.filteredRows, onFilter]);
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
gretanausedaite, line 679: Remove eslint-disable and change `Columns<any>` to `Columns<T>`.

gretanausedaite (commented):
Have you tested user example with this new code?
[Add user provided example to one of our test apps and see if this update helps with their issue]

rohan-kadkol:
> Have you tested user example with this new code? [Add user provided example to one of our test apps and see if this update helps with their issue]

Yes, I tested it by adding the code from the sandbox to the Vite Playground. I added it in the commits near https://github.com/iTwin/iTwinUI-react/pull/983/commits/a86264e3ce19956012c4f6b550b425a405a98324 and removed it after testing that it worked.

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]**

gretanausedaite (approved):
🌵

packages/iTwinUI-react/src/core/Table/Table.tsx
```diff
@@ -676,6 +676,19 @@ export const Table = <
     }
   }, [state, instance.filteredRows, onFilter]);
 
+  const lastPassedColumns: React.MutableRefObject<Column<T>[] | null> =
```
adamhock, line 679: `const lastPassedColumns = React.useRef([] as Column<T>[]);`

This looks a bit cleaner in my opinion but it's a nitpick

adamhock (approved):
🎉

packages/iTwinUI-react/src/core/Table/Table.tsx
rohan-kadkol: **[DONE]** Thanks, done

PR #840, opened 2022-09-26
chore: Replaced `max-height` inline style with `.iui-scroll` for few Menu derivatives

r100-stack:
* Implements [v1 (issue iTwin/iTwinUI-react#831)](https://github.com/iTwin/iTwinUI-react/issues/831) CSS change: [feat(Menu): Add CSS variable for max-height on scrollable menu](https://github.com/iTwin/iTwinUI/pull/756/commits/804c819792ec382f83a357cd855f407a6e1f5c4d)
* Removed hardcoded react `max-height` of 315px from `VirtualizedComboBoxMenu`, `ComboBoxMenu`, `Select` and instead used `.iui-scroll`

## Checklist

- ~[ ]~ Add meaningful unit tests for your component (verify that all lines are covered)
- [x] Verify that all existing tests pass
- ~[ ]~ Add component features demo in Storybook (different stories)
- ~[ ]~ Approve test images for new stories
- [x] Add screenshots of the key elements of the component

## TODOs

- [x] Pass CI [#3293](https://github.com/iTwin/iTwinUI-react/actions/runs/3146038506)
- [x] Tests pass locally
  - The `Select -> TruncateMiddleText` test is failing, even before these changes. Thus ignoring this failure
- ~[ ]~ Fix the bug of subMenu of a subMenu not displaying after adding `.iui-scroll` to the `.iui-menu`
  - **[OUTDATED]** *after removing `.iui-scroll` from Menu. Would be handled in [#853](https://github.com/iTwin/iTwinUI/issues/899) ([comment iTwin/iTwinUI-react#840](https://github.com/iTwin/iTwinUI-react/pull/840#discussion_r985834228))*
  - (`DropdownMenu -> Subrows` story): [Actual (after PR)](https://user-images.githubusercontent.com/45748283/192341362-2b662620-170a-40d3-9c6c-2ff218aa4739.webm) | [Expected (before PR)](https://user-images.githubusercontent.com/45748283/192341616-50e4192c-6c62-4490-bec6-6ea3dbdcafb9.webm)

#### `Table -> ColumnManager` story: `DropdownMenu` `max-height` = `--iui-menu-max-height` (`315px`)

image

PR #986, opened 2022-12-09
fix(ComboBox): Move max-height to outer element to fix virtual scroll

### LINKED ISSUE #984: ComboBox: broken layout when using virtualization
In v2, combobox with `enableVirtualization` looks broken. The menu takes up all the available height and doesn't scroll properly.

https://itwin.github.io/iTwinUI-react/?path=/story/core-combobox--virtualized
https://itwin.github.io/iTwinUI-react/iframe.html?args=&id=core-combobox--virtualized&viewMode=story

mayank99:
Virtualization broke after #840 because `maxHeight: 315` was removed from `Surface` (presumably because the max-height is specified in the css through `iui-menu.iui-scroll`). But that doesn't work, because for virtual scroll to work, the max-height must be on the outermost element.

This PR unsets the `Menu` max-height by removing `iui-scroll`, and manually adds the same max-height to the outermost element (`Surface`). Also adds a visual test for it to help catch any future regressions.

Unrelated: `elevation={1}` was removed from `Surface` because that's the default in css now.


packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -56,31 +56,27 @@ const VirtualizedComboBoxMenu = React.forwardRef(
...
-      >
+      <Surface style={surfaceStyles} {...rest}>
         <div {...outerProps}>
           <Menu
             id={`${id}-list`}
             setFocus={false}
             role='listbox'
             ref={mergeRefs(menuRef, innerProps.ref, forwardedRef)}
-            className={cx('iui-scroll', className)}
+            className={className}
```
rohan-kadkol, line 79: Just curious, would still using the `.iui-scroll` for `overlay-y` but just overriding the `max-height` remove the need for `isOverflowOverlaySupported`?

https://github.com/iTwin/iTwinUI/blob/9912acf9a0a526ccf34d1089ef03722ed26d5c62/packages/itwinui-css/src/menu/menu.scss#L30-L34

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -56,31 +56,27 @@ const VirtualizedComboBoxMenu = React.forwardRef(
...
+      getWindow()?.CSS?.supports?.('overflow: overlay');
 
-    const styles = React.useMemo(
-      () => ({
-        minWidth,
-        maxWidth: `min(${minWidth * 2}px, 90vw)`,
-      }),
-      [minWidth],
-    );
+    const surfaceStyles = {
```
veekeys, line 62: you think component is too light to memoize this object?

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
```diff
@@ -56,31 +56,27 @@ const VirtualizedComboBoxMenu = React.forwardRef(
       scrollToIndex: focusedVisibleIndex,
     });
 
-    const overflowY = getWindow()?.CSS?.supports?.('overflow-x: overlay')
-      ? { overflowY: 'overlay' }
-      : { overflowY: 'auto' };
+    const isOverflowOverlaySupported =
```
veekeys, line 59: to same comment, this could be moved outside the component code too

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: this object itself is not really expensive to create, so i dont think there's a need to memoize

packages/iTwinUI-react/src/core/ComboBox/ComboBoxMenu.tsx
mayank99: done

gretanausedaite (approved):

PR #987, opened 2022-12-09
chore: Safer visual tests for stories with `<img>`

r100-stack:
## Issue

I was on the iTwinUI-react GitHub page and saw that the CI is failing.
image

I think it is from the latest merged PR [#974](https://github.com/iTwin/iTwinUI-react/pull/974) in which the Avatar test is failing in this [CI action](https://github.com/iTwin/iTwinUI-react/actions/runs/3660131263).
image

## Fix

Avatar's WithImage story is an Avatar with an  with an src that needs to download the image.
image

The baseline image so far was actually an empty circle:
image

> #### OLD METHOD
>
> Adding a `cy.wait(2000)` actually downloads the image and displays/renders it. So now the new test image generated is:
> image

#### NEW METHOD

Instead of `cy.wait(2000)`, I am now using `hide()` to hide the `` tags in all stories with ``. I got this idea from [Tile.test.ts L25-31](https://github.com/iTwin/iTwinUI-react/blob/13cff7fab356e1015e2e7c361f15c6a7c8fe4d6b/apps/storybook/src/Tile.test.ts#L25-L31). I used more specific selectors whenever possible (e.g. `img.iui-icon`) and otherwise used a normal `img` selector.

I also applied this fix for all stories with ``.

This is a quick PR with this small fix so that CI does not fail. Hopefully we can merge it soon. Later if there is a better fix, we can do that.

## Checklist


elephantcatdog (commented):
I think that the test for combobox loading state is unecessary if they end up just looking loaded

rohan-kadkol:
> I think that the test for combobox loading state is unecessary if they end up just looking loaded

I think it would be safer to ignore the  there too. Because if for some reason the image doesn't load later, the test would fail.

Actually, the test failure mentioned in the PR description might have been because the baseline image was without loading the `` but the test image on GitHub loading with the image. Thus, I feel it is safer to ignore all ``s

elephantcatdog:
> I think it would be safer to ignore the  there too. Because if for some reason the  doesn't load later, the test would fail.
> 
> Actually, the test failure mentioned in the PR description might have been because the baseline image was "without" loading the  but the test image on GitHub was "with" loading with the .
> 
> Thus, I feel it is safer to ignore all s with urls as their src.

I meant more like - don't add the test if it doesn't add value.

elephantcatdog:
Other than that, it looks good.

elephantcatdog (approved):

elephantcatdog:
I approved it; don't know if we should wait for an approval from someone else

apps/storybook/src/Avatar.test.ts
```diff
@@ -16,6 +16,12 @@ describe('Avatar', () => {
     it(testName, function () {
       const id = Cypress.storyId(storyPath, testName);
       cy.visit('iframe', { qs: { id } });
+
+      // Hide images if present to avoid uncertainties in testing
+      if (testName == 'With Image') {
```
adamhock, line 21: Perhaps use `===` here

apps/storybook/src/ComboBox.test.ts
```diff
@@ -13,6 +13,7 @@ describe('ComboBox', () => {
     'With Message',
     'With Status',
     'Multiple Select',
+    'Disabled',
```
mayank99, line 16: i'm surprised (and glad) this works now. initially we excluded it because it was taking too long to load (it was freezing up and timing out). i believe that is also the reason we excluded `Virtualized` (fixed in #986).

apps/storybook/src/ComboBox.test.ts
```diff
@@ -27,6 +28,12 @@ describe('ComboBox', () => {
           items[3].click();
         });
       }
+
+      // Hide images if present to avoid uncertainties in testing
+      if (testName === 'Disabled') {
+        cy.get('img.iui-icon').hide();
```
mayank99, line 34: it should also work if we don't target the class here. will be easier to maintain if we change the class in the future.
```suggestion
        cy.get('img').hide();
```

mayank99 (approved):

apps/storybook/src/ComboBox.test.ts
rohan-kadkol: Ohh, I see. To be sure, I ran this test a couple of times and it passed alright. So including this test.

apps/storybook/src/ComboBox.test.ts
rohan-kadkol: **[DONE]** I had actually tried to make the selectors specific since I noticed `Tile.test.ts` [L30](https://github.com/iTwin/iTwinUI-react/blob/main/apps/storybook/src/Tile.test.ts#L30) also selected by class name.

But I guess that was because there was no `` tag there. The image was in the `<div>` tag itself and so needed the class name selector.

PR #917, opened 2022-11-08
feat!: Remove dependency on itwinui-icons

mayank99:
Removed `@itwin/itwinui-icons-react` and moved the icons that are used by our components into `src/utils/icons`.

First step towards #912. In a separate PR, I will change illustrations to use dynamic imports.

PR #991, opened 2022-12-12
chore: Update STYLEGUIDE.md with import icons for `itwinui-react`

### LINKED ISSUE #925: Update style guide with latest way to import icons in components (`src/core`)
> > To confirm, do we need to update the style guide regarding importing icons in components (`src/core`)?
> 
> Yeah we should.

_Originally posted by @mayank99 in https://github.com/iTwin/iTwinUI-react/issues/917#issuecomment-1310375639_

r100-stack:
Closes #925 

* Added an importing icons section in `CONTRIBUTING.md` with the process of adding/using new icons in `packages/iTwinUI-react`.
* Removed importing icons styling recommendations in `STYLEGUIDE.md` since it is partially included in `CONTRIBUTING.md` ([#991 (comment)](https://github.com/iTwin/iTwinUI-react/pull/991#discussion_r1047752267))

<details>
<summary>Old PR description before https://github.com/iTwin/iTwinUI-react/pull/991#discussion_r1047682881</summary>

* Updated `STYLEGUIDE.md` to recommend importing icons from `src/core/utils` in `packages/itwinui-react` (new recommendation) and import from cjs in other modules (existing recommendation).
* Added styling recommendations for imports for illustrations to import illustrations similar to icons.
* Corrected incorrectly imported icons and illustrations, according to the style guide.

</details>

## Checklist


packages/iTwinUI-react/src/core/NonIdealState/ErrorPage.tsx
```diff
@@ -6,17 +6,15 @@ import React from 'react';
...
-  Svg403,
-  Svg404,
-  Svg500,
-  Svg502,
-  Svg503,
-  SvgError,
-  SvgRedirect,
-  SvgTimedOut,
-} from '@itwin/itwinui-illustrations-react';
+import Svg401 from '@itwin/itwinui-illustrations-react/cjs/illustrations/401';
```
mayank99, line 9: what's the reason for this change?

there is really _no_ need to import from internal cjs paths anymore, whether it's illustrations in this component or icons in stories/playgrounds.

rohan-kadkol: For the icons, I was making these changes to match the [STYLEGUIDE.md](https://github.com/iTwin/iTwinUI-react/blob/main/STYLEGUIDE.md#import-individual-icons-directly-from-the-cjs-module) recommendation.

image

For the illustrations, there was no style guide recommendation. So I thought of adding a recommendation to import illustrations similar to icons. I mentioned this in the PR desc.

But if there is no benefit for importing icons and illustrations from cjs, then I can undo these changes and remove those recommendations from the style guide. So, contributors can import icons and illustrations in whatever way they prefer.

mayank99: yes, the task is to update the styleguide which has become outdated after #912. for more context, you can see the history in #917 and #924.

mayank99: Looks like i didn't clearly document the thought process in those PRs so i'll explain:

In v1, we were relying on `@itwin/itwinui-icons-react`. This is a huge package with tons of icons, so treeshaking is very important to reduce client bundle size. But treeshaking was not working correctly with module imports in certain environments (i think webpack4), so we ended up using internal imports from the `cjs/` folder. The idea is that if you only import the files you are using, instead of the barrel, then you don't need to rely on bundlers for treeshaking. (Also, this only needed to be done in `@itwin/itwinui-react` which gets delivered to the user. Anything else like storybook, tests, etc are completely irrelevant to this conversation and can use module imports with no problems.)

But using these internal cjs imports breaks ESM environments. So in #917, we removed the dependency completely and are now just inlining the icons that we actually need, instead of bringing in the whole package.

Similarly, for ESM compatibility, we don't want to rely on internal imports for illustrations. But `ErrorPage` actually imports all (or most) of the svgs in `@itwin/itwinui-illustrations-react` so it would _not_ get treeshaked anyway. This is why, in #924, we switched to module imports, deprecated the `ErrorPage` component, and introduced a new `NonIdealState` component which does not import all of these illustrations. The idea is that if the user doesn't import `ErrorPage`, then the whole component will be treeshaked. And then if they bring in `NonIdealState`, they can import just the one illustration instead of all.

Both of these changes, combined with https://github.com/iTwin/iTwinUI-icons/pull/49, result in lower bundle sizes for everyone and better ESM compatibility.

As for what this means for the styleguide:
1. There is no need to ever import from `cjs/` anywhere in the repo. Those points should be removed.
2. Illustrations are fine to not mention. Eventually we can remove `ErrorPage` and therefore remove the dependency.
3. We could mention the new way of importing icons, which involves:
  - Manually add the svg component to `utils/icons` and `utils/index.ts`.
  - Import it from `utils`

I think point 3 might not belong in styleguide though.

CONTRIBUTING.md
```diff
@@ -156,6 +156,27 @@ apps/storybook/src/**Alert.test.tsx**
 
 - Contains the Cypress test for the component.
 
+### Importing icons
+
+#### In `packages/iTwinUI-react`:
+
+* Manually add the `<svg>` component to `utils/icons` and `utils/index.ts`
```
mayank99, line 163:
```suggestion
* Manually add the `<svg>` component to `utils/icons` and `utils/icons/index.ts`
```

CONTRIBUTING.md
```diff
@@ -156,6 +156,27 @@ apps/storybook/src/**Alert.test.tsx**
...
+* Import it from `utils`
+
+e.g.
+```tsx
+import { SvgClose, SvgInfoCircular} from '../utils';
```
mayank99, line 168:
```suggestion
import { SvgClose, SvgInfoCircular } from '../utils';
```

CONTRIBUTING.md
```diff
@@ -156,6 +156,27 @@ apps/storybook/src/**Alert.test.tsx**
...
+
+#### In other modules:
+
+* Import it from `@itwin/itwinui-icons-react`
+
+e.g.
+```tsx
+import { SvgChevronRightDouble, SvgFolder } from '@itwin/itwinui-icons-react';
+```
+
```
mayank99, line 179: I don't think this is necessary. It doesn't matter what we do in other workspaces.

STYLEGUIDE.md
```diff
@@ -146,27 +146,37 @@ export const Alert = ({
   ...
 ```
 
-### Import individual icons directly from the cjs module
+### Import icons
```
mayank99, line 149: I think this whole section can be removed now that we have it in the contributing guide.

CONTRIBUTING.md
rohan-kadkol: **[DONE]** Sounds good, removed

STYLEGUIDE.md
rohan-kadkol: Removing it sounds good. But before removing it, to confirm, would it be good to mention importing icons from `utils` in `CONTRIBUTING.md` but recommend users not to import individually from `utils/icons` in `STYLEGUIDE.md` ([L149-158](https://github.com/iTwin/iTwinUI-react/pull/991/files#diff-18b19378fcb4038a2ce14f3ef8dd8ffa0bacd8f4e795cef51d08f3178d7e946cR149-R158))?

mayank99: no need. if we see anyone doing it like that in the future, we can reconsider

rohan-kadkol: **[DONE]** Sounds good, removed

mayank99 (approved):

elephantcatdog (approved):

